PR #7, opened 2020-06-02
Add landing page with list of releases

mfojtik:
Defining list of supported releases in code is not great, we might think of having some YAML file config and an observer that will restart sippy when the config change (that will list all supported releases).

### DIFF
--- a/main.go
+++ b/main.go
@@ -646,9 +646,7 @@ func (s *Server) printHtmlReport(w http.ResponseWriter, req *http.Request) {
 
 	release := req.URL.Query().Get("release")
 	if _, ok := s.analyzers[release]; !ok {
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -27,6 +27,14 @@ import (
 var (
 	dashboardTemplate = "redhat-openshift-ocp-release-%s-%s"
 	TagStripRegex     = regexp.MustCompile(`\[Skipped:.*?\]|\[Suite:.*\]`)
+
+	releases = []string{
+		"4.2",
+		"4.3",
+		"4.4",
+		"4.5",
+		"4.6",
```
bparees, line 36: why not get this from the args?

bparees:
thanks @mfojtik !

PR #8, opened 2020-06-02
Add metal-ipi as a potential platform

stbenjam:
Added to testgrids just now: https://github.com/kubernetes/test-infra/pull/17806

I think it makes sense to show metal and metal-ipi separately, since they're very different.

### DIFF
--- a/pkg/html/html.go
+++ b/pkg/html/html.go
@@ -104,16 +104,15 @@ Data current as of: %s
 
 	// 1 encoded job name
 	// 2 test name
-	// 3 job name
-	// 4 release
-	// 5 encoded test name
+	// 3 job name regex
+	// 4 encoded test name
 	testGroupTemplate = `
 		<tr class="collapse %s">
 			<td colspan=2>
 			%s
 			<p>
-			<a target="_blank" href="https://search.svc.ci.openshift.org/?maxAge=168h&context=1&type=junit&maxMatches=5&maxBytes=20971520&groupBy=job&name=%[3]s.*%[4]s&search=%[5]s">Job Search</a>
-			<a style="padding-left: 20px" target="_blank" href="https://search.svc.ci.openshift.org/?maxAge=168h&context=1&type=bug&maxMatches=5&maxBytes=20971520&groupBy=job&search=%[5]s">Bug Search</a>
+			<a target="_blank" href="https://search.svc.ci.openshift.org/?maxAge=168h&context=1&type=junit&maxMatches=5&maxBytes=20971520&groupBy=job&name=%[3]s&search=%[4]s">Job Search</a>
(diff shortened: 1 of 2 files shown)


pkg/util/utils.go
```diff
@@ -260,7 +261,9 @@ func FindPlatform(name string) []string {
 	if openstackRegex.MatchString(name) {
 		platforms = append(platforms, "openstack")
 	}
-	if metalRegex.MatchString(name) {
+	if metalIPIRegex.MatchString(name) {
+		platforms = append(platforms, "metal-ipi")
+	} else if metalRegex.MatchString(name) {
```
stbenjam, line 266: Does the fall through make sense? So a metal job is selected as ipi or not. I could make the metal regex more robust if this is too unclear

bparees: yeah i'd rather see you make the existing metal regex more robust(mutually exclusive with the new one), lest this get reordered at some point in the future and screw it up.

stbenjam: Hm, problem is golang doesn't do negative lookaheads in the regexp libarary, I'll have to play around with it.

There's also a bunch of problems that need fixing... metal-ipi is the first multi-stage job we've turned release informing on for, the job names are reversed because we use variants in openshift/release to create the periodic, so it ends up being named `ocp-4.5-e2e-metal-ipi`, but the job search wants the release as a suffix.

Also it's trying to query `operator.Run multi-stage test e2e-metal-ipi`, but you get no results in the search dashboard unless you remove `operator.`

bparees: if there's no easy way to disambiguate the non-ipi jobs, we can just do it as you have it, just please add a comment on both the regex and the if/else making it very clear what the behavior is and why, so someone doesn't refactor it unknowingly.

regarding the other issues, let's chat on slack

stbenjam: Updated, and sent you a message about the other issues.

pkg/html/html.go
```diff
@@ -292,7 +291,18 @@ func summaryJobsByPlatform(report, reportPrev util.TestReport, endDay, jobTestCo
 
 			encodedTestName := url.QueryEscape(regexp.QuoteMeta(test.Name))
 
-			rows = rows + fmt.Sprintf(testGroupTemplate, strings.ReplaceAll(v.Platform, ".", ""), test.Name, v.Platform, report.Release, encodedTestName,
+			// Multi-stage jobs that use variants will have the OCP release
+			// before the name of the job, while the older template based
+			// tests will have it after.
```
bparees, line 296: can we do a single regex which is `release.platform | platform.release` and do away w/ the special case for metal-ipi?

stbenjam: Oh yea, makes sense. Updated!

PR #20, opened 2020-07-02
Truncate Summary and Send Version to Bugzilla

coreydaley:
- Truncate the summary to 255 characters when opening a Bugzilla
 - Send the version when opening a Bugzilla

### DIFF
--- a/pkg/html/html.go
+++ b/pkg/html/html.go
@@ -78,7 +78,7 @@ Data current as of: %s
 }
 </style>
 
-<h1 class=text-center>CI Release Health Summary</h1>
+<h1 class=text-center>CI Release {{ .Release }} Health Summary</h1>
 
 <p class="small mb-3 text-nowrap">
 	Jump to: <a href="#SummaryAcrossAllJobs">Summary Across All Jobs</a> | <a href="#FailureGroupings">Failure Groupings</a> | 
@@ -95,7 +95,7 @@ Data current as of: %s
 
 {{ summaryJobsByPlatform .Current .Prev .EndDay .JobTestCount }}
 
-{{ summaryTopFailingTests .Current.TopFailingTestsWithoutBug .Current.TopFailingTestsWithBug .Prev.All .EndDay }}
+{{ summaryTopFailingTests .Current.TopFailingTestsWithoutBug .Current.TopFailingTestsWithBug .Prev.All .EndDay .Release }}
 
 {{ summaryJobPassRatesByJobName .Current .Prev .EndDay .JobTestCount }}
 
@@ -345,7 +345,7 @@ func getPrevTest(test string, testResults []util.TestResult) *util.TestResult {
 	return nil
 }
 
-func summaryTopFailingTests(topFailingTestsWithoutBug, topFailingTestsWithBug []*util.TestResult, resultPrev map[string]util.SortedAggregateTestResult, endDay int) string {
+func summaryTopFailingTests(topFailingTestsWithoutBug, topFailingTestsWithBug []*util.TestResult, resultPrev map[string]util.SortedAggregateTestResult, endDay int, release string) string {
 	allPrev := resultPrev["all"]
 
 	// test name | bug | pass rate | higher/lower | pass rate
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -725,7 +725,7 @@ func (s *Server) printHtmlReport(w http.ResponseWriter, req *http.Request) {
 		html.WriteLandingPage(w, s.options.Releases)
 		return
 	}
-	html.PrintHtmlReport(w, req, s.analyzers[release].Report, s.analyzers[release+"-prev"].Report, s.options.EndDay, 15)
+	html.PrintHtmlReport(w, req, s.analyzers[release].Report, s.analyzers[release+"-prev"].Report, s.options.EndDay, 15, release)
```
bparees, line 728: analyzers[release].Report already has a Release field in it, you shouldn't need to pass this arg in.

pkg/html/html.go
```diff
@@ -385,7 +385,7 @@ func summaryTopFailingTests(topFailingTestsWithoutBug, topFailingTestsWithBug []
...
-			bug = fmt.Sprintf("<a target=\"_blank\" href=https://bugzilla.redhat.com/enter_bug.cgi?classification=Red%%20Hat&product=OpenShift%%20Container%%20Platform&cf_internal_whiteboard=buildcop&short_desc=%[1]s&cf_environment=%[1]s&comment=test:%%0A%[1]s%%20%%0A%%0Ais%%20failing%%20frequently%%20in%%20CI,%%20see%%20search%%20results:%%0A%s%%0A%%0A%s>Open a bug</a>", url.QueryEscape(test.Name), url.QueryEscape(searchUrl), url.QueryEscape(exampleJob))
+			bug = fmt.Sprintf("<a target=\"_blank\" href=https://bugzilla.redhat.com/enter_bug.cgi?classification=Red%%20Hat&product=OpenShift%%20Container%%20Platform&cf_internal_whiteboard=buildcop&short_desc=%[1]s&cf_environment=%[1]s&comment=test:%%0A%[1]s%%20%%0A%%0Ais%%20failing%%20frequently%%20in%%20CI,%%20see%%20search%%20results:%%0A%s%%0A%%0A%s&version=%[1]s>Open a bug</a>", url.QueryEscape(test.Name[:255]), url.QueryEscape(searchUrl), url.QueryEscape(exampleJob), release)
```
bparees, line 388: this is truncating the test name everywhere in the bz template, you should only be truncating it in the summary field.

main.go
coreydaley: fixed in current push

pkg/html/html.go
coreydaley: fixed in current push

bparees:
@coreydaley were you able to test this locally?

coreydaley:
Yes, and I fixed an issue with the short description truncating just now.

pkg/html/html.go
```diff
@@ -385,7 +385,11 @@ func summaryTopFailingTests(topFailingTestsWithoutBug, topFailingTestsWithBug []
...
 			exampleJob := "FIXME: Replace this paragraph with a particular job URI from the search results to ground discussion.  A given test may fail for several reasons, and this bug should be scoped to one of those reasons.  Ideally you'd pick a job showing the most-common reason, but since that's hard to determine, you may also chose to pick a job at random.  Release-gating jobs (release-openshift-...) should be preferred over presubmits (pull-ci-...) because they are closer to the released product and less likely to have in-flight code changes that complicate analysis."
-			bug = fmt.Sprintf("<a target=\"_blank\" href=https://bugzilla.redhat.com/enter_bug.cgi?classification=Red%%20Hat&product=OpenShift%%20Container%%20Platform&cf_internal_whiteboard=buildcop&short_desc=%[1]s&cf_environment=%[1]s&comment=test:%%0A%[1]s%%20%%0A%%0Ais%%20failing%%20frequently%%20in%%20CI,%%20see%%20search%%20results:%%0A%s%%0A%%0A%s>Open a bug</a>", url.QueryEscape(test.Name), url.QueryEscape(searchUrl), url.QueryEscape(exampleJob))
+			short_desc := url.QueryEscape(test.Name)
+			if len(short_desc) > 255 {
+				short_desc = short_desc[:255]
+			}
```
bparees, line 391: doesn't this risk truncating an escaped character and screwing everything up?

coreydaley: fixed

bparees: so the escape characters don't count towards the limit?  (makes sense i guess that BZ only sees the unescaped value)

PR #29, opened 2020-07-10
JSON api endpoint

allenmqcymp:


pkg/jsonAPI/jsonapi.go
```diff
@@ -0,0 +1,514 @@
...
+
+		platformTests := report.ByPlatform[v.Platform]
+		for _, test := range platformTests.TestResults {
+			if util.IgnoreTestRegex.MatchString(test.Name) {
+				continue
+			}
+
+			encodedTestName := url.QueryEscape(regexp.QuoteMeta(test.Name))
+			// NOTE: not really sure what this represents
+			// jobQuery := fmt.Sprintf("%s.*%s|%s.*%s", report.Release, v.Platform, v.Platform, report.Release)
```
bparees, line 224: it's part of the ci-search query that filters the query down to just job names that match the job name we're listing test failures for.  

your `testLink` below should include this value as the `name` query parameter.  see:
https://github.com/openshift/sippy/blob/master/pkg/html/html.go#L310
https://github.com/openshift/sippy/blob/master/pkg/html/html.go#L322-L329

which is populating:
https://github.com/openshift/sippy/blob/master/pkg/html/html.go#L128

It allows someone to see a search result that only shows "failures of this specific test, when run in jobs that match this specific platform"

i think your "testLink" url should include it in the same way since this is a "by platform" set of results you're building.

pkg/jsonAPI/jsonapi.go
```diff
@@ -0,0 +1,514 @@
...
+						Percentage:          v.PassPercentage,
+						ProjectedPercentage: v.PassPercentageWithKnownFailures,
+						Runs:                v.Successes + v.Failures,
+					},
+				},
+			}
+		}
+
+		// FIXME: not sure if I need this
+		count := jobTestCount
```
bparees, line 400: you don't, if you're trying to return all the tests that failed for the job, not a limited set.

you can remove it just like you removed it from the ByPlatform results above.

It does mean you're going to potentially return a *lot* of data (every test that failed across every job, multiplied by 2 since you're also doing it in the byplatform results).

So you may want to look at how big a json this is producing and evaluate whether returning all the tests is really viable or useful.

bparees: (if you don't remove it, you should also considering adding additionalMatches back in so the api can inform the caller of how many results were skipped)

pkg/jsonAPI/jsonapi.go
```diff
@@ -0,0 +1,514 @@
...
+			Job:          fg.Job,
+			Url:          fg.Url,
+			TestFailures: fg.TestFailures,
+		})
+	}
+	return failureGroups
+}
+
+func jsonTestImpactingBugs(testImpactingBugs []util.Bug) []util.Bug {
+	return testImpactingBugs
```
bparees, line 479: this is just returning its own arg?  that doesn't seem useful :)

pkg/jsonAPI/jsonapi.go
```diff
@@ -0,0 +1,514 @@
...
+			"jobRunsWithFailureGroups":  jsonFailureGroupList(data.Current),
+			"testImpactingBugs":         jsonTestImpactingBugs(data.Current.BugsByFailureCount),
+		},
+	}
+
+	enc := json.NewEncoder(w)
+	err := enc.Encode(jsonObject)
+	if err != nil {
+		// klog.Errorf("unable to render json %v", err)
+		fmt.Printf("unable to render json %v", err)
```
bparees, line 511: use klog.

pkg/util/utils.go
```diff
@@ -118,10 +118,10 @@ type BugList map[string]BugResult
 type BugResult map[string][]Bug
 
 type Bug struct {
-	Summary      string `json:"name"`
-	ID           string
-	Url          string
-	FailureCount int32
+	Summary      string `json:"summary,omitempty"`
```
bparees, line 121: this json field name value needs to stay as is, it has to match the json struct being returned by ci-search's json api:
https://search.ci.openshift.org/search?search=storage&maxAge=48h&context=-1&type=bug

(this struct is a bit overloaded, it's used both to deserialize the results coming from ci-search, as well as to storage bug metadata for reporting, but it saves mapping back and forth between two structs that otherwise serve basically the same purpose)

pkg/jsonAPI/jsonapi.go
```diff
@@ -0,0 +1,514 @@
+package jsonAPI
```
bparees, line 1: just call it `package json`.  you can rename it on import to de-conflict it with the golang json package.

or maybe just call it `api` for that matter.  This is/will be the programmatic api for sippy.

pkg/jsonAPI/jsonapi.go
```diff
@@ -0,0 +1,514 @@
...
+	"net/http"
+	"net/url"
+	"regexp"
+
+	"github.com/openshift/sippy/pkg/html"
+	"github.com/openshift/sippy/pkg/util"
+)
+
+// NOTE: these functions are mirrored in html.go
+// Copied over here as a quick fix
```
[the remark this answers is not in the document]
bparees: (and make both the json and html endpoints use them, so when i say move them i mean move them out of html)

main.go
```diff
@@ -735,6 +737,15 @@ func (s *Server) printHtmlReport(w http.ResponseWriter, req *http.Request) {
 	html.PrintHtmlReport(w, req, s.analyzers[release].Report, s.analyzers[release+"-prev"].Report, s.options.EndDay, 15)
 }
 
+func (s *Server) printJSONReport(w http.ResponseWriter, req *http.Request) {
+	release := req.URL.Query().Get("release")
+	if _, ok := s.analyzers[release]; !ok {
+		fmt.Println("must specify a release")
```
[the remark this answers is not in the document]
allenmqcymp: yeah, definitely. That was just placeholder for something I wasn't sure what to do about

pkg/jsonAPI/jsonapi.go
```diff
@@ -0,0 +1,514 @@
...
+		countPrev += group.TestFailures
+	}
+	if len(failureGroups) != 0 {
+		median = failureGroups[len(failureGroups)/2].TestFailures
+		avg = count / len(failureGroups)
+	}
+	if len(failureGroupsPrev) != 0 {
+		medianPrev = failureGroupsPrev[len(failureGroupsPrev)/2].TestFailures
+		avgPrev = count / len(failureGroupsPrev)
+	}
```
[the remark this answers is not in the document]
allenmqcymp: I agree. Do you think utils is a good place to put it?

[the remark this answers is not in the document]
bparees: there's no place better at the moment :)

feel free to create new files within the utils package though, if it helps you organize things.

pkg/jsonAPI/jsonapi.go
allenmqcymp: renamed to api

pkg/jsonAPI/jsonapi.go
allenmqcymp: yeah. I just created a function because I wanted a function to correspond to every section in the dashboard. I guess I can just refer to it directly.

pkg/jsonAPI/jsonapi.go
allenmqcymp: for now, I think I'm going to remove it, because I think it I could add a feature to only request specific portions so that less data is returned

allenmqcymp: I agree that there is too much data at the moment though

allenmqcymp:
Hi Ben, I've resolved most of the comments you left. I do agree that there is too much data outputted at the moment, and would like your advice on how to potentially restructure the API -- perhaps adding URL arguments to specify section to include/omit?

bparees:
> Hi Ben, I've resolved most of the comments you left. I do agree that there is too much data outputted at the moment, and would like your advice on how to potentially restructure the API -- perhaps adding URL arguments to specify section to include/omit?

i'd consider starting by just leaving that data out entirely.  Talk with Jessica+Nick(and maybe Tim Moran) but i think the goal here was primarily about having the pass/fail data for the platforms+tests on the whole, not necessarily the details about which specific tests were failing on which specific platforms/jobs.

Whille I do think there's useful analysis one might be able to do w/ the full set of data, i'd say let's not add it until we have a clear need/use for it.

But having it be "opt-in" based on a parameter is totally reasonable too, given you've already implemented the capability.

Maybe make it an arg like "failing tests to show per job/platform" and show the top N tests based on the arg value.  Where the default is zero, "-1" is all of them, and anything else is "however many they asked for"

allenmqcymp:
> > Hi Ben, I've resolved most of the comments you left. I do agree that there is too much data outputted at the moment, and would like your advice on how to potentially restructure the API -- perhaps adding URL arguments to specify section to include/omit?
> 
> i'd consider starting by just leaving that data out entirely. Talk with Jessica+Nick(and maybe Tim Moran) but i think the goal here was primarily about having the pass/fail data for the platforms+tests on the whole, not necessarily the details about which specific tests were failing on which specific platforms/jobs.
> 
> Whille I do think there's useful analysis one might be able to do w/ the full set of data, i'd say let's not add it until we have a clear need/use for it.
> 
> But having it be "opt-in" based on a parameter is totally reasonable too, given you've already implemented the capability.
> 
> Maybe make it an arg like "failing tests to show per job/platform" and show the top N tests based on the arg value. Where the default is zero, "-1" is all of them, and anything else is "however many they asked for"

Sounds good. I think I will opt for making it take optional arguments, since I've already implemented the capability to return everything.

allenmqcymp:
Hi Ben @bparees, I've deleted the unnecessary associated bug lists since they aren't that useful for data analysis, at least right now. Let me know if there are any more issues before merging.

main.go
```diff
@@ -735,6 +736,23 @@ func (s *Server) printHtmlReport(w http.ResponseWriter, req *http.Request) {
...
 }
 
+func (s *Server) printJSONReport(w http.ResponseWriter, req *http.Request) {
+	release := req.URL.Query().Get("release")
+	w.Header().Set("Content-Type", "application/json; charset=utf-8")
+	if _, ok := s.analyzers[release]; !ok {
+		// return a 404 error along with the list of available releases in the detail section
+		errMsg := map[string]interface{}{
+			"code":   "404",
+			"detail": fmt.Sprintf("list of releases: %v", s.options.Releases),
```
bparees, line 746:
```suggestion
			"detail": fmt.Sprintf("No valid release specified, valid releases are: %v", s.options.Releases),
```

main.go
```diff
@@ -735,6 +736,23 @@ func (s *Server) printHtmlReport(w http.ResponseWriter, req *http.Request) {
...
+		}
+		errMsgBytes, _ := json.Marshal(errMsg)
+		w.WriteHeader(404)
```
bparees, line 749:
```suggestion
		w.WriteHeader(http.StatusNotFound)
```

pkg/api/api.go
```diff
@@ -0,0 +1,410 @@
...
+)
+
+// PassRate describes statistics on a pass rate
+type PassRate struct {
+	Percentage          float64 `json:"percentage"`
+	ProjectedPercentage float64 `json:"projectedPercentage,omitempty"`
+	Runs                int     `json:"runs"`
+}
+
+// SummaryAcrossAllJobs describes the category failuregroups
```
bparees, line 22: looks like a cut+paste issue on the doc

pkg/api/api.go
```diff
@@ -0,0 +1,410 @@
...
+// passRate may include projected pass rate (float64), percentage (float64), and number of runs (int)
+type FailingTest struct {
+	Name     string     `json:"name"`
+	Url      string     `json:"url"`
+	PassRate PassRate   `json:"passRate"`
+	Bugs     []util.Bug `json:"bugs"`
+}
+
+// FailingTestBug describes a single instance of failed test with bug or failed test without bug
+// differs from failingtest in that it includes pass rates for previous days and latest days
```
bparees, line 63: can't you just use FailingTestBug everywhere since it's a superset of FailingTest?

pkg/api/api.go
```diff
@@ -0,0 +1,410 @@
...
+type FailingTestBug struct {
+	Name      string              `json:"name"`
+	Url       string              `json:"url"`
+	PassRates map[string]PassRate `json:"passRates"`
+	Bugs      []util.Bug          `json:"bugs,omitempty"`
+}
+
+// JobSummaryPlatformdescribes a single platform and its associated jobs, their pass rates, and failing tests
```
bparees, line 71:
```suggestion
// JobSummaryPlatform describes a single platform and its associated jobs, their pass rates, and failing tests
```

pkg/api/api.go
```diff
@@ -0,0 +1,410 @@
...
+	latestDays := fmt.Sprintf("latest%dDays", endDay)
+	prevDays := "prev7Days"
+	var topFailingTests []FailingTestBug
+
+	allPrev := resultPrev["all"]
+
+	for _, test := range topFailingTestsWithBug {
+		encodedTestName := url.QueryEscape(regexp.QuoteMeta(test.Name))
+
+		testLink := fmt.Sprintf("https://search.svc.ci.openshift.org/?maxAge=168h&context=1&type=bug%%2Bjunit&name=&maxMatches=5&maxBytes=20971520&groupBy=job&search=%s", encodedTestName)
```
bparees, line 192: make the format string a constant and reuse it in html.go so we don't have two copies of the value potentially getting out of sync

pkg/api/api.go
```diff
@@ -0,0 +1,410 @@
...
+	prevDays := "prev7Days"
+
+	allPrev := resultPrev["all"]
+
+	var topFailingTests []FailingTestBug
+
+	for _, test := range topFailingTestsWithoutBug {
+		encodedTestName := url.QueryEscape(regexp.QuoteMeta(test.Name))
+
+		testLink := fmt.Sprintf("https://search.svc.ci.openshift.org/?maxAge=168h&context=1&type=bug%%2Bjunit&name=&maxMatches=5&maxBytes=20971520&groupBy=job&search=%s", encodedTestName)
```
bparees, line 247: same here, make it a shared constant

pkg/api/api.go
```diff
@@ -0,0 +1,410 @@
...
+		return nil
+	}
+
+	for i := len(all) - 1; i > len(all)-10; i-- {
+		test := all[i]
+		encodedTestName := url.QueryEscape(regexp.QuoteMeta(test.Name))
+		canaryFailures = append(canaryFailures,
+			CanaryTestFailInstance{
+				Name: test.Name,
+				Url:  fmt.Sprintf("https://search.svc.ci.openshift.org/?maxAge=168h&context=1&type=bug%%2Bjunit&name=&maxMatches=5&maxBytes=20971520&groupBy=job&search=%s", encodedTestName),
```
bparees, line 355: shared constant

pkg/api/api.go
allenmqcymp: yeah, my bad.

pkg/api/api.go
allenmqcymp: yeah I realized failingTest is actually wholly redundant since I'm not returning bug data, so I deleted it.

bparees:
@allenmqcymp squash it down to a few logical commits and i'll merge.  you can update html.go to use the constant you've added as a follow up (I realize it'll be a bit complicated because of how that search url is already embedded in existing strings)

PR #31, opened 2020-07-17
introduce json api endpoint for retrieving data

allenmqcymp:
Created a new api branch and copied all the code from jsonapi.


pkg/api/api.go
```diff
@@ -0,0 +1,400 @@
...
+	}
+	return &failureGroupStruct
+}
+
+func summaryJobsByPlatform(report, reportPrev util.TestReport, endDay, jobTestCount int) []JobSummaryPlatform {
+	jobsByPlatform := util.SummarizeJobsByPlatform(report)
+	jobsByPlatformPrev := util.SummarizeJobsByPlatform(reportPrev)
+
+	latestDays := fmt.Sprintf("latest%dDays", endDay)
+	prevDays := "prev7Days"
```
bparees, line 126: just make these `latest` and `prev` so it still make sense if this gets used with custom start/end days later. and makes it easier for a consumer to process the keys.

pkg/api/api.go
```diff
@@ -0,0 +1,400 @@
...
+		jobSummariesByPlatform = append(jobSummariesByPlatform, jobSummaryPlatform)
+	}
+	return jobSummariesByPlatform
+}
+
+// top failing tests with a bug
+func summaryTopFailingTestsWithBug(topFailingTestsWithBug []*util.TestResult, resultPrev map[string]util.SortedAggregateTestResult, endDay int) []FailingTestBug {
+
+	latestDays := fmt.Sprintf("latest%dDays", endDay)
+	prevDays := "prev7Days"
```
bparees, line 173: same here

pkg/api/api.go
```diff
@@ -0,0 +1,400 @@
...
+
+	return topFailingTests
+
+}
+
+// top failing tests without a bug
+func summaryTopFailingTestsWithoutBug(topFailingTestsWithoutBug []*util.TestResult, resultPrev map[string]util.SortedAggregateTestResult, endDay int) []FailingTestBug {
+
+	latestDays := fmt.Sprintf("latest%dDays", endDay)
+	prevDays := "prev7Days"
```
bparees, line 227: and here

pkg/api/api.go
```diff
@@ -0,0 +1,400 @@
...
+		}
+
+		topFailingTests = append(topFailingTests, failedTestWithoutBug)
+	}
+	return topFailingTests
+}
+
+func summaryJobPassRatesByJobName(report, reportPrev util.TestReport, endDay, jobTestCount int) []PassRatesByJobName {
+	latestDays := fmt.Sprintf("latest%dDays", endDay)
+	prevDays := "prev7Days"
```
bparees, line 278: and here

PR #36, opened 2020-08-03
add functionality to get all releases via json?release=all, changed output to include version, refactoring.

allenmqcymp:


main.go
```diff
@@ -767,7 +767,21 @@ func (s *Server) printHtmlReport(w http.ResponseWriter, req *http.Request) {
...
+		for _, r := range s.options.Releases {
+			if _, ok := s.analyzers[r]; ok {
+				releaseReports[r] = []util.TestReport{s.analyzers[r].Report, s.analyzers[r+"-prev"].Report}
+				// NOTE: not sure if safe to assume that prev always exists given r exists
+			} else {
+				klog.Errorf("unable to load test report for release version %s", r)
+				continue
+			}
+		}
+		api.PrintAllJSONReport(w, req, releaseReports, s.options.EndDay, 15)
```
bparees, line 783: add a return here and you don't need the else below.

bparees: (guiding philosophy here is "earliest possible return")

pkg/api/api.go
```diff
@@ -350,36 +350,58 @@ func failureGroupList(report util.TestReport) []FailureGroup {
...
-func PrintJSONReport(w http.ResponseWriter, req *http.Request, report, prevReport util.TestReport, endDay, jobTestCount int) {
-
+func formatJSONReport(report, prevReport util.TestReport, endDay, jobTestCount int) map[string]interface{} {
 	data := html.TestReports{
 		Current:      report,
 		Prev:         prevReport,
 		EndDay:       endDay,
 		JobTestCount: jobTestCount,
 		Release:      report.Release}
+	// NOTE: should I use report.Release or pass in release from main.go
```
bparees, line 360: since you have to pass the report object in anyway, this seems reasonable rather than adding a redundant arg

main.go
```diff
@@ -777,8 +791,9 @@ func (s *Server) printJSONReport(w http.ResponseWriter, req *http.Request) {
 		w.WriteHeader(http.StatusNotFound)
 		w.Write(errMsgBytes)
 		return
+	} else {
+		api.PrintJSONReport(w, req, s.analyzers[release].Report, s.analyzers[release+"-prev"].Report, s.options.EndDay, 15)
```
bparees, line 795: Seems like you should be able to have a single PrintJSONReport function whether you are reporting a single release, or multiple.  Printing a single release report is just a special case of printing multiple and i'd expect the output to look the same (other than only having a single release entry)

PR #45, opened 2020-08-17
add ovn as a platform

deads2k:
I suspect that many ovn failures are specific to ovn.  Having a platform checks the general feature and classifies the top failures so we can quickly compare.


bparees:
@deads2k this should roll out automatically once it merges.

PR #49, opened 2020-08-18
add ocp and origin platforms

deads2k:
The pass rates are very different and the top failing test cases are different, so we have to attack them differently.


pkg/util/utils.go
```diff
@@ -379,7 +387,7 @@ func FindPlatform(name string) []string {
 	if ovnRegex.MatchString(name) {
 		platforms = append(platforms, "ovn")
 	}
-	if ovnRegex.MatchString(name) {
+	if fipsRegex.MatchString(name) {
```
bparees, line 390: whoops

PR #51, opened 2020-08-20
suppress single failure, one off, long tail failures from the top failing test lists

deads2k:
Our goal is primary failers.  We're now at a point where low frequency tests can fail a single time at random and they clog the top failures list.  It's usually not helpful to open bugs on these because (at least in the ones I've seen at the top), they are either systemic and we have bugs open on the platform (ovn or fips) or they are just one-offs for the week and don't really matter.

This doesn't impact the data, only the presentation.

### DIFF
--- a/pkg/html/html.go
+++ b/pkg/html/html.go
@@ -396,6 +396,12 @@ func summaryTopFailingTests(topFailingTestsWithoutBug, topFailingTestsWithBug []
 	`
 
 	for _, test := range topFailingTestsWithoutBug {
+		// if we only have one failure, don't show it on the glass.  Keep it in the actual data so we can choose how to handle it,
+		// but don't bother creating the noise in the UI for a one-off/long tail.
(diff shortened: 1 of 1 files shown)


pkg/html/html.go
```diff
@@ -396,6 +396,12 @@ func summaryTopFailingTests(topFailingTestsWithoutBug, topFailingTestsWithBug []
 	`
 
 	for _, test := range topFailingTestsWithoutBug {
+		// if we only have one failure, don't show it on the glass.  Keep it in the actual data so we can choose how to handle it,
+		// but don't bother creating the noise in the UI for a one-off/long tail.
+		if (test.Failures + test.Flakes) == 1 {
+			continue
+		}
```
bparees, line 403: is the minTestRuns not already filtering this out?

if you only have 1 failure and you have at least 10 runs, you're passing at 90% and i wouldn't expect you to be a top failure anyway.

PR #52, opened 2020-08-20
remove some of the e2e test steps from excluded tests list

bparees:


bparees:
@deads2k @mfojtik take a look at the screenshot and tell me if you think including those "test" results would be useful to you or not...  this is in an effort to include "install failures" in sippy reporting.

i'm kinda leaning towards "not useful" (because it's too generic, we need specific tests that fail when the install fails in specific ways).

PR #58, opened 2020-08-24
add synthetic junit tests for infra, install, and upgrade

deads2k:
This tracks the known install and upgrade junit tests and follows the know setup container junit to try to determine whether an install failed, whether the infra failed, or whether the upgrade failed.

It seems to be showing a tough set of infra problems on some platforms.

### DIFF
--- a/pkg/util/utils.go
+++ b/pkg/util/utils.go
@@ -99,6 +99,26 @@ type JobRunResult struct {
 	Failed             bool     `json:"failed"`
 	HasUnknownFailures bool     `json:"hasUnknownFailures"`
 	Succeeded          bool     `json:"succeeded"`
+
+	// SetupStatus can be "", "Success", "Failure"
+	SetupStatus      string          `json:"setupStatus"`
+	InstallOperators []OperatorState `json:"installOperators"`
+	UpgradeOperators []OperatorState `json:"upgradeOperators"`
+}
+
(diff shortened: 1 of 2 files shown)


pkg/util/utils.go
```diff
@@ -99,6 +99,26 @@ type JobRunResult struct {
...
+	UpgradeTestName        = `[sig-sippy] upgrade should work`
+
+	Success = "Success"
+	Failure = "Failure"
+)
+
+type OperatorState struct {
+	Name string `json:"name"`
+	// OperatorState can be "", "Success", "Failure"
+	State string `json:"state"`
```
bparees, line 121: was there a compelling reason not to use a boolean here?  anticipating future expansion of the list of states?

deads2k: in kube, we found nothing is a bool.  You can imagine this developing into subdivision of failures for reporting of naughty operators.  It could have caught storage degraded versus something not available.  and I think registry is showing progressing in one :(

bparees: but you're only going to have the boolean info of "test failed" or "test succeeded" so unless you introduce specific tests for things like "is operator foo progressing" and then report state=progressing based on that test success/failure, i'm not sure how you'll expose the operator state through this.

anyway i'm not going to hold it up

PR #60, opened 2020-08-28
refactor the processing pipeline slightly

deads2k:
in preparation for associating jobruns (not tests) with sigs


bparees:
i feel like i'm somehow letting the camel's nose in the tent here, but this looks like a straightforward refactoring that makes the code easier to follow, so:
/lgtm

PR #61, opened 2020-08-28
show how much components are blocking

deads2k:
This aims to show how many jobs a bz component is blocking.  It needs work on presentation and an "expand" button.  But basically every component is blocking someone pretty badly.

Screenshot from 2020-08-28 19-53-21

### DIFF
--- a/main.go
+++ b/main.go
@@ -31,10 +31,13 @@ var (
 )
 
 type RawData struct {
-	ByAll       map[string]util.AggregateTestResult
-	ByJob       map[string]util.AggregateTestResult
-	ByPlatform  map[string]util.AggregateTestResult
-	BySig       map[string]util.AggregateTestResult
+	ByAll      map[string]util.AggregateTestResult
+	ByJob      map[string]util.AggregateTestResult
+	ByPlatform map[string]util.AggregateTestResult
+	BySig      map[string]util.AggregateTestResult
+	// ByBugzillaComponent has keys that are a best guess of the bugzilla component which is causing the failing test.
+	ByBugzillaComponent map[string]util.AggregateTestResult
+
 	JobRuns     map[string]util.JobRunResult
 	JobDetails  []testgrid.JobDetails
 	BugFailures map[string]util.Bug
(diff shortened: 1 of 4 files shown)


main.go
```diff
@@ -390,6 +393,90 @@ func updateGlobalBugCache(failedTestNames sets.String) {
...
 
+func (a *Analyzer) identifyJobRunBugzillaFailer() {
+	for jrrKey, jrr := range a.RawData.JobRuns {
+		failingBugzillaComponents := sets.NewString()
+
+		for _, failedTest := range jrr.FailedTestNames {
+			// some tests are skipped
+			switch {
+			case strings.HasPrefix(failedTest, util.OperatorInstallPrefix) || strings.HasPrefix(failedTest, util.OperatorUpgradePrefix):
+				// these are handled by checking operator, no work on the string.
```
bparees, line 404: there's no reason they can't work on the string (if you open a BZ and put the test name in the BZ, they will be found in the ci-search lookup).

whether that's useful or not i guess i'm less sure (you'll probably have a lot of periods of time when no bug exists associated w/ a particular "install operator foo", but you might at least want to try to do the bug lookup first.

pkg/util/utils.go
```diff
@@ -133,6 +149,33 @@ type JobResult struct {
...
+}
+
+type BugzillaJobResult struct {
+	JobName           string `json:"jobName"`
+	BugzillaComponent string `json:"bugzillaComponent"`
+
+	FailuresByThisBugzillaComponent int `json:"failuresByThisBugzillaComponent"`
+	// This one is phrased as a failure percentage because we don't know a success percentage since we don't know how many times it actually ran
+	// we only know how many times its tests failed.
+	FailPercentageByThisBugzillaComponent float64 `json:"passPercentageByThisBugzillaComponent"`
```
bparees, line 176:
```suggestion
	FailPercentageByThisBugzillaComponent float64 `json:"failPercentageByThisBugzillaComponent"`
```

pkg/util/utils.go
```diff
@@ -133,6 +149,33 @@ type JobResult struct {
...
+	JobsFailed []BugzillaJobResult `json:"jobsFailed"`
+}
+
+type BugzillaJobResult struct {
+	JobName           string `json:"jobName"`
+	BugzillaComponent string `json:"bugzillaComponent"`
+
+	FailuresByThisBugzillaComponent int `json:"failuresByThisBugzillaComponent"`
+	// This one is phrased as a failure percentage because we don't know a success percentage since we don't know how many times it actually ran
+	// we only know how many times its tests failed.
```
bparees, line 175: if we don't know how many times it ran, what is this a percentage of?

deads2k: > if we don't know how many times it ran, what is this a percentage of?

we can know failures percentage out of all runs, but we cannot know success percentages since we don't know how often the test runs

deads2k: My goal here is not to determine how often a particular test itself fails or how frequent a failure of a test is relative to how often it runs.  The goal is to measure impact to the average job run for a given job, so that's a percentage against all runs

bparees: fair enough, but i think calling it a "failure percentage" is what lead to my confusion.  failureCausePercentage?

bparees:
@deads2k has promised to revert this tomorrow and follow it up w/ a cleaned up/reviewer PR, so
/lgtm

PR #62, opened 2020-08-31
refactor the bug cache to remove global

deads2k:
This makes the bug cache local.  It's built on a revert of the ugly impl for BZ components.

/hold

hold until after team leads have a chance to see impact tomorrow morning.


bparees:
what is the motivation to de-globalize the bugcache?  W/ this change you now need to run 8x the bug queries, and bug queries are already the slowest part of doing the analysis.

if you just want an easier way to filter bugs to the correct release or job, let's improve the bugcache datastructure, but i don't think we should be running multiple queries for the same testname (once per release and current/prev period).

deads2k:
> what is the motivation to de-globalize the bugcache? W/ this change you now need to run 8x the bug queries, and bug queries are already the slowest part of doing the analysis.

That isn't what this does.  It shares a single bug cache, but passes it as an instance variable and makes it an interface.  This allows for shared state, but also allows for future mocks and unit tests.

pkg/apis/bugs/v1/types.go
```diff
@@ -0,0 +1,22 @@
+package v1
+
+import "time"
+
+// Bug is used to represent bugs in some serialized content.  It also appears to be used internally for data tracking
```
bparees, line 5:
```suggestion
// Bug is used to represent bugs in some serialized content.  It is also used internally for data tracking
```

pkg/buganalysis/cache.go
```diff
@@ -0,0 +1,170 @@
...
+		r, err := findBugs(batchNames)
+		for k, v := range r {
+			c.cache[k] = v
+		}
+		if err != nil {
+			c.lastUpdateError = err
+		}
+	}
+
+	c.lastUpdateError = nil
```
bparees, line 84: why is this always being set back to nil here?

pkg/buganalysis/internal/types_external.go
```diff
@@ -0,0 +1,20 @@
+package internal
+
+import bugsv1 "github.com/openshift/sippy/pkg/apis/bugs/v1"
+
+// These types are used to decode information from bugzilla, but we don't want to expose these for anyone else.
```
bparees, line 5: they're decoding information from ci-search.

PR #63, opened 2020-08-31
stop rendering with look-aside data

deads2k:
This change eliminates the bugcache from the html rendering flow so that we ensure the data in our API is the data we need.


pkg/html/html.go
```diff
@@ -754,12 +752,9 @@ func WriteLandingPage(w http.ResponseWriter, releases []string) {
 	fmt.Fprintf(w, landingHtmlPageEnd)
 }
 
-func PrintHtmlReport(w http.ResponseWriter, req *http.Request, bugCache buganalysis.BugCache, report, prevReport util.TestReport, endDay, jobTestCount int) {
+func PrintHtmlReport(w http.ResponseWriter, req *http.Request, report, prevReport util.TestReport, endDay, jobTestCount int) {
 	w.Header().Set("Content-Type", "text/html;charset=UTF-8")
 	fmt.Fprintf(w, htmlPageStart, "Release CI Health Dashboard")
-	if err := bugCache.LastUpdateError(); err != nil {
-		fmt.Fprintf(w, bugLookupWarning, err)
-	}
```
bparees, line 1015: am i missing how you're replacing this message on the page?

bparees:
just the question about the BZ warning on this one.

your updates to the bugcache address the lastErr concern from the earlier PR, but the docs comments still apply i think.

pkg/html/html.go
deads2k: > am i missing how you're replacing this message on the page?

I owe you I think. After the final type split, we can add a banner/warning list

PR #64, opened 2020-08-31
ignore tests in one spot

deads2k:
Ignore tests during data gathering instead of trying to compensate across many different locations.


bparees:
lgtm once the things it's built on are updated

PR #65, opened 2020-09-01
straight move of the sippy API

deads2k:
We'll tidy up from here, but this at least separates the types. It's now possible to write linters.


bparees:
lgtm pending updates to earlier commits

PR #66, opened 2020-09-01
separate the test result intermediate types from test result output types

deads2k:
This starts to bring a more principled pipeline and will help developers understand what information is available at what point.  Also, turns out some fields were never filled in before!


pkg/apis/sippyprocessing/v1/types.go
```diff
@@ -10,41 +10,43 @@ import (
...
+	Successes      int     `json:"successes"`
+	Failures       int     `json:"failures"`
+	Flakes         int     `json:"flakes"`
+	PassPercentage float64 `json:"passPercentage"`
+	// BugList shows all applicable bugs for the context.
+	// Inside of a release, only bugs matching the release are present.
+	// TODO Inside a particular job, only bugs matching the job are present.
+	// TODO Inside a platform, only bugs matching the platform are present.
+	BugList []bugsv1.Bug `json:"bugList"`
+	// TODO fix search link to properly take into account release, job, and platform.
```
bparees, line 48: i'm not sure we want that.  it's a bit of a debate in my head how much to filter the ci-search results.  Filtering by release makes reasonable sense to me (i don't care about test failures from another release, they are a different codebase and may not be relevant).  But I probably do care about failures from other platforms/jobs within the same release.  They *may* not have the same cause, but it's informative to know where else the test is failing.

Since the search link already accounts for the release, i'd vote for just removing this todo as i don't think it's an appropriate goal.  but i guess we can debate that elsewhere.

deads2k: > Since the search link already accounts for the release, i'd vote for just removing this todo as i don't think it's an appropriate goal. but i guess we can debate that elsewhere.

you'll get a chance to object if anyone actually makes it do this. Based on context, I think we want different ci-search links.  
0. for a release (like the top failing test view) - restrict by release
1. for a variant (like the by variant view) - restrict by variant and release
2. for a job (like the by-job view and the BZ failure view) - restrict by job and release

It's much easier to remove the restrictions after clicking that knowing the format to create them to begin with.

bparees: actually it already does those things today, so is this change removing that behavior?

deads2k: In master, the only spot I found setting SearchLink is here: https://github.com/openshift/sippy/blob/953f7851ceb79fb03bef4204c6fecc907b359692/main.go#L708-L710

However I think that testresults in the various `ByFoo` should have different SearchLinks.

PR #67, opened 2020-09-01
refine jobrunresult into intermediate and final APIs

deads2k:
this way we have processing APIs and serialization/rendering APIs.


bparees (approved):
looks ok to me

PR #68, opened 2020-09-01
isolate the intermediate processing types

deads2k:
This PR moves the testgrid analysis types into a testgridanalysis package.

Inputs to this package will eventually be the testgrid details with an output of a rendering type (sippyprocessing/v1 today)


bparees (approved):
looks ok to me

PR #69, opened 2020-09-02
Add back bz component

deads2k:
undoes the revert with a much cleaner interface


pkg/html/bugs.go
```diff
@@ -0,0 +1,54 @@
...
+import (
+	"fmt"
+	"net/url"
+
+	bugsv1 "github.com/openshift/sippy/pkg/apis/bugs/v1"
+	"github.com/openshift/sippy/pkg/buganalysis"
+)
+
+func bugLink(bug bugsv1.Bug) string {
+	return fmt.Sprintf("<a target=\"_blank\" href=%s>%d</a> ", bug.Url, bug.ID)
```
mfojtik, line 11: nit: href="%s"

pkg/util/utils.go
```diff
@@ -82,13 +76,13 @@ type AggregateTestResult struct {
...
-	Flakes         int     `json:"flakes"`
-	PassPercentage float64 `json:"passPercentage"`
-	BugList        []Bug   `json:"BugList"`
-	SearchLink     string  `json:"searchLink"`
+	Name           string       `json:"name"`
+	Successes      int          `json:"successes"`
+	Failures       int          `json:"failures"`
+	Flakes         int          `json:"flakes"`
+	PassPercentage float64      `json:"passPercentage"`
+	BugList        []bugsv1.Bug `json:"BugList"`
```
mfojtik: nit: bugList ?

main.go
```diff
@@ -400,30 +362,25 @@ func (a *Analyzer) analyze() {
...
 	failedTestNamesAcrossAllJobRuns := getFailedTestNamesFromJobRuns(a.RawData.JobRuns)
-	updateGlobalBugCache(failedTestNamesAcrossAllJobRuns)
+	a.BugCache.UpdateForFailedTests(failedTestNamesAcrossAllJobRuns.List()...)
 
 	// for every test that failed in some job run, look up the bug(s) associated w/ the test
 	// and attribute the number of times the test failed+flaked to that bug(s)
-	for t := range failedTestNamesAcrossAllJobRuns {
-		if util.IgnoreTestRegex.MatchString(t) {
+	for testName := range failedTestNamesAcrossAllJobRuns {
+		if util.IgnoreTestRegex.MatchString(testName) {
```
mfojtik: can this be just `util.IsTestIgnored()` ?

pkg/apis/bugs/v1/types.go
```diff
@@ -0,0 +1,22 @@
...
+
+// BugzillaBug matches the bugzilla API.  We cannot change this and should consider writing a converter instead of having
+// a serialization we don't own.
+type BugzillaBug struct {
+	ID             int64     `json:"id"`
+	Status         string    `json:"status"`
+	LastChangeTime time.Time `json:"last_change_time"`
+	Summary        string    `json:"summary"`
+	TargetRelease  []string  `json:"target_release"`
+	Component      []string  `json:"component"`
```
mfojtik, line 21: nit: `Components` ?

pkg/apis/bugs/v1/types.go
```diff
@@ -0,0 +1,22 @@
...
+}
+
+// BugzillaBug matches the bugzilla API.  We cannot change this and should consider writing a converter instead of having
+// a serialization we don't own.
+type BugzillaBug struct {
+	ID             int64     `json:"id"`
+	Status         string    `json:"status"`
+	LastChangeTime time.Time `json:"last_change_time"`
+	Summary        string    `json:"summary"`
+	TargetRelease  []string  `json:"target_release"`
```
mfojtik, line 20: wonder why is this an []string, can you have more than 1 target release?

pkg/apis/bugs/v1/types.go
```diff
@@ -0,0 +1,22 @@
...
+	BugzillaBug  `json:",inline"`
+	Url          string `json:"url"`
+	FailureCount int    `json:"failureCount,omitempty"`
+	FlakeCount   int    `json:"flakeCount,omitempty"`
+}
+
+// BugzillaBug matches the bugzilla API.  We cannot change this and should consider writing a converter instead of having
+// a serialization we don't own.
+type BugzillaBug struct {
+	ID             int64     `json:"id"`
```
mfojtik, line 16: in all libs I saw the ID was an `int` :-)

pkg/buganalysis/cache.go
```diff
@@ -0,0 +1,170 @@
...
+		}
+		batchNames = append(batchNames, testName)
+		// we're going to lookup bugs for this test, so put an entry into the map.
+		// if we find a bug for this test, the entry will be replaced with the actual
+		// array of bugs.  if not, this serves as a placeholder so we know not to look
+		// it up again in the future.
+		c.cache[testName] = []bugsv1.Bug{}
+		batchCount++
+
+		if batchCount > 50 {
```
mfojtik, line 63: nit:

```
if batchCount <=50 { continue }
```

makes less nesting

pkg/buganalysis/cache.go
```diff
@@ -0,0 +1,170 @@
...
+		testName = regexp.QuoteMeta(testName)
+		klog.V(4).Infof("Searching bugs for test name: %s\n", testName)
+		v.Add("search", testName)
+	}
+
+	//searchUrl:="https://search.apps.build01.ci.devcluster.openshift.com/search"
+	searchUrl := "https://search.ci.openshift.org/v2/search"
+	resp, err := http.PostForm(searchUrl, v)
+	if err != nil {
+		e := fmt.Errorf("error during bug search against %s: %s", searchUrl, err)
```
mfojtik, line 138: nit: `%v`

pkg/apis/bugs/v1/types.go
bparees: yes.  the BZ api returns it as an array.  though i haven't figured out how set more than one in the UI.

pkg/apis/bugs/v1/types.go
bparees: this is driven by the ci-seach api:
https://search.ci.openshift.org/v2/search?search=storage&type=bug

```
bugInfo": {

    "id": 1867757,
    "status": "VERIFIED",
    "resolution": "",
    "severity": "medium",
    "priority": "unspecified",
    "summary": "Rebase node-registrar sidebar with latest version",
    "keywords": [ ],
    "whiteboard": "",
    "cf_internal_whiteboard": "",
    "creator": "hekumar@redhat.com",
    "component": [
        "Storage"
    ],
    "assigned_to": "hekumar@redhat.com",
    "creation_time": "2020-08-10T17:07:59Z",
    "last_change_time": "2020-08-25T01:42:05Z",
    "cf_environment": "",
    "target_release": [
        "4.6.0"
    ]

}
```

bparees: (which in turn is a more or less direct mapping of the BZ api)

deads2k:
@mfojtik I propose dealing with all those as followups.

pkg/apis/sippyprocessing/v1/types.go
```diff
@@ -22,14 +22,18 @@ type TestReport struct {
...
 // SortedAggregateTestsResult
 type SortedAggregateTestsResult struct {
-	Successes          int          `json:"successes"`
-	Failures           int          `json:"failures"`
-	TestPassPercentage float64      `json:"testPassPercentage"`
-	TestResults        []TestResult `json:"results"`
+	Successes          int     `json:"successes"`
+	Failures           int     `json:"failures"`
+	TestPassPercentage float64 `json:"testPassPercentage"`
+	// TestResults holds the values for individual runs of this test sorted by.....
```
bparees, line 35: don't leave me in suspense......i assume they're storted by pass rate?

it's not really individual runs of the test, right?  isn't it an array of tests, each entry being a testname and the totals/percentages(across all runs of that test) for that test name?

pkg/html/html.go
```diff
@@ -727,6 +731,188 @@ func testImpactingComponents(testImpactingBugs []bugsv1.Bug) string {
...
 }
 
+func summaryJobsFailuresByBugzillaComponent(report, reportPrev sippyprocessingv1.TestReport, endDay int, release string) string {
+	failuresByBugzillaComponent := util.SummarizeJobsFailuresByBugzillaComponent(report)
+	failuresByBugzillaComponentPrev := util.SummarizeJobsFailuresByBugzillaComponent(reportPrev)
+
+	s := fmt.Sprintf(`
+	<table class="table">
+		<tr>
+			<th colspan=4 class="text-center"><a class="text-dark" title="Bugzilla components ranked by maximum fail percentage of any job." id="BZToWorstJobImpact" href="#BZToWorstJobImpact">Bugzilla Components to Worst Job Impact</a></th>
```
bparees, line 741:
```suggestion
			<th colspan=4 class="text-center"><a class="text-dark" title="Bugzilla components ranked by maximum fail percentage of any job." id="BZToWorstJobImpact" href="#BZToWorstJobImpact">Job Impacting BZ Components</a></th>
```

bparees: (and elsewhere)

pkg/html/html.go
```diff
@@ -727,6 +731,188 @@ func testImpactingComponents(testImpactingBugs []bugsv1.Bug) string {
...
+		</tr>
+		<tr>
+			<th>Variant</th><th>Latest %d days</th><th/><th>Previous 7 days</th>
```
bparees, line 744:
```suggestion
			<th>Component</th><th>Latest %d days</th><th/><th>Previous 7 days</th>
```

pkg/html/html.go
```diff
@@ -727,6 +731,188 @@ func testImpactingComponents(testImpactingBugs []bugsv1.Bug) string {
...
+		`
+
+	for _, v := range failuresByBugzillaComponent {
+		safeBZJob := fmt.Sprintf("%s---component", v.Name)
+		safeBZJob = strings.ReplaceAll(safeBZJob, ".", "")
+		safeBZJob = strings.ReplaceAll(safeBZJob, " ", "")
+
+		prev := util.GetPrevBugzillaJobFailures(v.Name, failuresByBugzillaComponentPrev)
+		highestFailPercentage := v.JobsFailed[0].FailPercentage
+		lowestPassPercentage := 100 - highestFailPercentage
```
bparees, line 808: so ultimately the value you're presenting in the table is "the percentage of jobs which did not fail because of this component" right?

i appreciate it is aligned w/ how the other tables work but it also doesn't seem like the most intuitive value.  Need to think about it.

pkg/html/html.go
```diff
@@ -727,6 +731,188 @@ func testImpactingComponents(testImpactingBugs []bugsv1.Bug) string {
...
+			} else {
+				arrow = fmt.Sprintf(flatdown, previousLowestPassPercentage-lowestPassPercentage)
+			}
+
+			s = s + fmt.Sprintf(bzGroupTemplate,
+				rowColor,
+				v.Name,
+				safeBZJob,
+				lowestPassPercentage,
+				prev.JobsFailed[0].TotalRuns,
```
bparees, line 851: something seems off about this value.  i'm seeing pretty low values reported.  if it were the number of failed runs the numbers would make sense, but as a "total" they don't make sense to me.  let's discuss when you're free.

or maybe it's low because you just picked the first entry?  should this be a sum of all the entries?

bparees: also i don't thiink you mean to have `prev` here.  should be the current week

pkg/html/html.go
```diff
@@ -727,6 +731,188 @@ func testImpactingComponents(testImpactingBugs []bugsv1.Bug) string {
...
+				break
+			}
+			count++
+			s = s + fmt.Sprintf(bzFailingJobTemplate,
+				safeBZJob,
+				v.Name,
+				failingJob.JobName,
+				bzJobTuple,
+				100.0-failingJob.FailPercentage,
+				failingJob.NumberOfJobRunsFailed,
```
bparees, line 883: since you inverted the percentage to be "passing" i think you should invert this to be successes(well, non-failures) also.

pkg/apis/sippyprocessing/v1/types.go
deads2k: > don't leave me in suspense......i assume they're storted by pass rate?
> 
> it's not really individual runs of the test, right? isn't it an array of tests, each entry being a testname and the totals/percentages(across all runs of that test) for that test name?

I have no idea. Your API ;)

pkg/html/html.go
deads2k: > i appreciate it is aligned w/ how the other tables work but it also doesn't seem like the most intuitive value. Need to think about it

aligning to the other tables was the what I was going for.  It was super confusing in the reverse and I knew what it was doing.

bparees: hm, another confusing thing about this is that the top line value you're presenting isn't some aggregation of the percentage of failures this component is causing across all jobs.  It's "here's the job where this component performs worst"

which i realize probably aligns with your intent (if a single component is tanking a single job, that's the thing you care about more than anything in the world, more so than another component that might be impacting all jobs but to a lesser degree....which is more or less what the "test impacting components" view is showing).  But i'm pretty sure people won't understand it.  I know i was confused when i started expanding jobs and trying to understand how the top line number made sense in light of the expanded jobs values.

your hover text sort of explains it.  let me see if i can offer a suggestion on that to help.

pkg/html/html.go
bparees:
```
Bugzilla components ranked by maximum fail percentage of any job
```
suggestion:

Bugzilla components ranked by computing which job they contribute the most failure percentage to and then sorting the components by each one's worst job performance.

but i wonder if this would be clearer if you just showed which job was the top failing one for the component, in the top line results, and the expansion just showed "other jobs" or "all jobs"

main.go
deads2k: > can this be just `util.IsTestIgnored()` ?

this shouldn't leak out.  I tightened it up in the interim

PR #70, opened 2020-09-02
provide warnings on out when analysis failures (like BZ lookup) happen

deads2k:


pkg/html/html.go
```diff
@@ -936,6 +935,9 @@ func WriteLandingPage(w http.ResponseWriter, releases []string) {
 func PrintHtmlReport(w http.ResponseWriter, req *http.Request, report, prevReport sippyprocessingv1.TestReport, endDay, jobTestCount int) {
 	w.Header().Set("Content-Type", "text/html;charset=UTF-8")
 	fmt.Fprintf(w, htmlPageStart, "Release CI Health Dashboard")
+	for _, analysisWarning := range report.AnalysisWarnings {
+		fmt.Fprintf(w, bugLookupWarning, analysisWarning)
+	}
```
bparees, line 940: should also include warnings from prevReport for completeness

PR #72, opened 2020-09-04
make job rendering and expansion consistent with failing tests and single format

deads2k:


pkg/html/html.go
```diff
@@ -864,48 +628,61 @@ func summaryJobsFailuresByBugzillaComponent(report, reportPrev sippyprocessingv1
 
 		count := 0
 		for _, failingJob := range v.JobsFailed {
+			if count > 5 { // only show five
```
bparees, line 631:
```suggestion
			if count > 4 { // only show five
```

bparees (commented):
fencepost error

PR #74, opened 2020-09-08
separate infrequent jobs into a different table section

deads2k:
We can create different coloriziation rules too. I'm biased towards making infrequent jobs more forgiving in terms of going yellow or red since an off day has a bigger impact.


bparees:
rebase and lgtm

PR #77, opened 2020-09-09
collapse some types

deads2k:
Brings the types closer together for general aggregation.

### DIFF
--- a/pkg/apis/sippyprocessing/v1/types.go
+++ b/pkg/apis/sippyprocessing/v1/types.go
@@ -64,7 +64,6 @@ type TestResult struct {
 type JobRunResult struct {
 	Job                string   `json:"job"`
 	Url                string   `json:"url"`
-	TestGridJobUrl     string   `json:"testGridJobUrl"`
 	TestFailures       int      `json:"testFailures"`
 	FailedTestNames    []string `json:"failedTestNames"`
 	Failed             bool     `json:"failed"`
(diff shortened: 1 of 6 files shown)


pkg/apis/sippyprocessing/v1/types.go
```diff
@@ -75,15 +74,11 @@ type JobRunResult struct {
...
-	Successes                       int     `json:"successes"`
-	PassPercentage                  float64 `json:"PassPercentage"`
-	PassPercentageWithKnownFailures float64 `json:"PassPercentageWithKnownFailures"`
-	TestGridUrl                     string  `json:"TestGridUrl"`
+	PassPercentageWithKnownFailures float64 `json:"passPercentageWithKnownFailures"`
+	TestGridURL                     string  `json:"TestGridURL"`
 
-	// TestResults holds entries for each test that is a part of this aggregation.  Each entry aggregates the results of all runs of a single test.  The array is sorted from lowest PassPercentage to highest PassPercentage
-	TestResults []TestResult `json:"results"`
+	SortedAggregateTestsResult `json:",inline"`
```
bparees, line 81: i don't understand this.  testresults doesn't replace job pass percentage, they aren't the same thing.

job pass percentage:  how many times the job succeeded over how many times it failed
testresults: information about the test executionst that ran within the executions of that job

bparees: at best this makes it very confusing when "job.TestPassPercentage" contains the "job pass percentage" (if indeed that's what you've done, i haven't looked closely) and not the testpasspercentage as aggregated across all runs of all tests in all runs of the job.

deads2k: That's not what this field should mean.  Maybe this will make it more clear https://github.com/openshift/sippy/pull/78/files#diff-d8ba90165ac3a1ed7d918c9deca1dbb9L152-L154  Those fields are dead.  We should remove them and then talk about the structure we want, which I think is a case where the aggregate links to constituent jobs and lists the success and fails for the aggregate job pass value.   This is equivalent to what you do everywhere you actually display the information.

deads2k:
/hold

I think this will be more clear once you see https://github.com/openshift/sippy/pull/78.  Once we remove the dead fields, we can talk about the correct names for the replacement.

PR #78, opened 2020-09-10
remove dead fields

deads2k:
success, failure, and passpercentage in the sorted aggregate fields are not used and are deceptive.  In fact, several spots explicitly avoid reporting the value, the overall summary was removed, and even the spots showing a pass percentage (questionable value) didn't report success and failure.

this makes the print report consistent with the html one and removes the dead fields.

### DIFF
--- a/pkg/util/utils.go
+++ b/pkg/util/utils.go
@@ -149,9 +149,6 @@ func SummarizeTestResults(
 		}
 
 		s := sorted[k]
-		s.Successes = passedCount
-		s.Failures = failedCount
-		s.TestPassPercentage = Percent(passedCount, failedCount)
 		sorted[k] = s
 
 		// sort from lowest to highest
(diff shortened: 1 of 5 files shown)


pkg/util/utils.go
```diff
@@ -149,9 +149,6 @@ func SummarizeTestResults(
 		}
 
 		s := sorted[k]
-		s.Successes = passedCount
-		s.Failures = failedCount
-		s.TestPassPercentage = Percent(passedCount, failedCount)
```
bparees, line 154: i'm surprised you don't think this is of interest (no, it's not in the html render today, but it could be added).

it provides a nuance that "passing job percentage" does not.  (some platforms may be consistently failing because of a single bad test. other platforms may be consistently failing because of hundreds of test failures)

deads2k:
> > success, failure, and passpercentage in the sorted aggregate fields are not used and are deceptive.
> 
> what is deceptive about them?

In places where a summary of success and failure are presented, the summarized success and failure isn't actually used and gets loaded from a different spot.  We should remove these (we aren't using them) and figure out what (if anything) we want to replace them with.

PR #79, opened 2020-09-10
move testreportconversion to its own package and pull it's single use helpers out

deads2k:
Helps figure out which code is truly shared helpers and which is simply associated with a single purpose.


main.go
```diff
@@ -512,70 +514,22 @@ func downloadData(releases []string, filter string, storagePath string) {
...
-		JobFailuresByBugzillaComponent: bugzillaComponentResults,
-
-		AnalysisWarnings: a.analysisWarnings,
-	}
-
-	if !prev {
-		topFailingTestsWithoutBug, topFailingTestsWithBug := getTopFailingTests(byAll, a.Release, a.BugCache)
-		a.Report.TopFailingTestsWithBug = topFailingTestsWithBug
-		a.Report.TopFailingTestsWithoutBug = topFailingTestsWithoutBug
-	}
```
bparees, line 573: assume that dropping this optimization is intentional (you are valuing the consistency of the report data over the optimization of not computing values we have no current use for) but your PR summary doesn't mention it, so was this intentional?

deads2k: > assume that dropping this optimization is intentional (you are valuing the consistency of the report data over the optimization of not computing values we have no current use for) but your PR summary doesn't mention it, so was this intentional?

Yes.  Knowing the previous list let's you flag things that have stayed in place.  Also, it no longer costs a BZ lookup.

PR #82, opened 2020-09-10
build byPlatform from raw results

deads2k:
This builds the equivalent ByPlatform from the raw job results.  This lets us build one representation of test results and slice them after we can interpret the data into our display format.


pkg/testgridanalysis/testreportconversion/testresult.go
```diff
@@ -10,3 +17,100 @@ func (a testResultsByPassPercentage) Swap(i, j int) { a[i], a[j] = a[j], a[i] }
...
+		// bugs should be the same for now.
+		byTestName[currTestResult.Name] = existing
+	}
+
+	combined := []sippyprocessingv1.TestResult{}
+	for _, currTestResult := range byTestName {
+		combined = append(combined, currTestResult)
+	}
+
+	sort.Stable(testResultsByPassPercentage(combined))
```
bparees, line 50: given the way you're using this (invoking combine for each jobresult) i would have separated the sort so you aren't repeatedly sorting the intermediary results you don't care about as you build up the aggregated list.

pkg/testgridanalysis/testreportconversion/by_platform.go
```diff
@@ -0,0 +1,60 @@
...
+	// TODO deads2k wants to eliminate the successThreshold
+	successThreshold float64, // indicates an upper bound on how successful a test can be before it is excluded
+) map[string]sippyprocessingv1.SortedAggregateTestsResult {
+
+	resultsByPlatform := make(map[string]sippyprocessingv1.SortedAggregateTestsResult)
+	for _, platform := range testidentification.AllPlatforms.List() {
+
+		allPlatformTestResults := []sippyprocessingv1.TestResult{}
+		successfulJobRuns := 0
+		failedJobRuns := 0
```
bparees, line 26: these values don't appear to be getting used for anything (they aren't set anywhere or returned out of this function)

pkg/testgridanalysis/testreportconversion/by_platform.go
```diff
@@ -0,0 +1,60 @@
...
+
+import (
+	sippyprocessingv1 "github.com/openshift/sippy/pkg/apis/sippyprocessing/v1"
+	"github.com/openshift/sippy/pkg/buganalysis"
+	"github.com/openshift/sippy/pkg/testgridanalysis/testgridanalysisapi"
+	"github.com/openshift/sippy/pkg/testgridanalysis/testidentification"
+	"github.com/openshift/sippy/pkg/util/sets"
+)
+
+// convertRawDataToByPlatform takes the raw data and produces a map of platform names to results
```
bparees, line 11:
```suggestion
// convertRawDataToByPlatform takes the raw data and produces a map of platform names to test(not job) results
```

pkg/testgridanalysis/testreportconversion/by_platform.go
deads2k: > these values don't appear to be getting used for anything (they aren't set anywhere or returned out of this function)

they get used one pull later

pkg/testgridanalysis/testreportconversion/by_platform.go
deads2k: IOU built on top?

pkg/testgridanalysis/testreportconversion/testresult.go
deads2k: > given the way you're using this (invoking combine for each jobresult) i would have separated the sort so you aren't repeatedly sorting the intermediary results you don't care about as you build up the aggregated list.

Depends on where and how it's called.  I'd like to keep the data clean when combining two (usually, might even be always now) sorted lists.  if we find ourselves getting two slow, I'm confident we could deal with it then.

pkg/testgridanalysis/testreportconversion/by_platform.go
bparees: sure

PR #84, opened 2020-09-10
 add expanding rows for jobs under platforms

deads2k:
Screenshot from 2020-09-10 13-56-24


bparees:
list of IOUs:

1) https://github.com/openshift/sippy/pull/82#discussion_r486998473

2) https://github.com/openshift/sippy/pull/83#discussion_r487194148

3) resolve the off by one error shown here (628 job runs, 629 sig-sippy infra test runs)
image

suggested folllow-on (i put this comment on the wrong PR) for doing the indenting, quickly + hackily:
https://github.com/openshift/sippy/pull/83#issuecomment-691215646

I also want to resolve
https://github.com/openshift/sippy/issues/85
and
https://github.com/openshift/sippy/issues/86

before embarking on further refactorings

/lgtm

PR #87, opened 2020-09-11
allow failing tests to expand by job so you can see where they fail

deads2k:


pkg/api/api.go
```diff
@@ -275,11 +273,11 @@ func canaryTestFailures(result map[string]v12.SortedAggregateTestsResult) []v1.C
 }
 
 // job runs with failure groups
-func failureGroupList(report v12.TestReport) []v1.FailureGroup {
+func failureGroupList(report sippyprocessingv1.TestReport) []sippyv1.FailureGroup {
```
mfojtik, line 276: @deads2k splitting these in separate commit will reduce the noise... next time :)

pkg/testgridanalysis/testreportconversion/testresult.go
```diff
@@ -84,47 +102,64 @@ func convertRawTestResultsToProcessedTestResults(
...
 		}
-
-		filteredResults = append(filteredResults, testResult)
+		return true
 	}
+}
 
-	return filterSuccessfulTestResults(filteredResults, successThreshold)
+func filterTestResultsByFilters(fns ...testResultFilterFunc) testResultFilterFunc {
+	return func(testResult sippyprocessingv1.TestResult) bool {
```
mfojtik, line 133: Stefan will love this :-)

deads2k:
I've placed an expansion button and refactored decorations to make them more consistent.  There is more work to do on rendering.

mfojtik:
/lgtm

Note that @deads2k won't make a career as web developer :-)

PR #89, opened 2020-09-14
address minor fixup after some refactors

deads2k:
IOUs from #83

### DIFF
--- a/pkg/testgridanalysis/testreportconversion/testresult.go
+++ b/pkg/testgridanalysis/testreportconversion/testresult.go
@@ -112,11 +112,12 @@ func filterTestResults(
 		filteredResults = append(filteredResults, testResult)
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/testgridanalysis/testreportconversion/testresult.go
```diff
@@ -112,19 +112,20 @@ func filterTestResults(
 		filteredResults = append(filteredResults, testResult)
 	}
 
-	return filterSuccessfulTestResults(filteredResults)
+	return filterSuccessfulTestResults(filteredResults, successThreshold)
```
bparees, line 115: i'm still confused by why you're doing this, didn't it already get filtered on line 104 above?

bparees: (alternatively, remove line 104?)

PR #90, opened 2020-09-14
add indenting for expanding sections

deads2k:
makes it easier to read.


bparees:
indent the headers too

and if you can easily add an empty spacer row, i think that would make this look a lot better:


image

PR #92, opened 2020-09-15
replace byAll with ByTest which actually includes all tests

deads2k:
ByAll was filtered and didn't include all the results.  The new ByTest does include them and uses the same datatype as the top failing test calculations.

This completes the main changes to the processing pipeline.  From here it is recutting an sorting.

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -14,25 +14,6 @@ import (
 	"k8s.io/klog"
 )
 
-// summary across all job
-func summaryAcrossAllJobs(result, resultPrev map[string]sippyprocessingv1.SortedAggregateTestsResult, endDay int) *sippyv1.SummaryAcrossAllJobs {
-	all := result["all"]
-	allPrev := resultPrev["all"]
-
-	summary := sippyv1.SummaryAcrossAllJobs{
-		TestExecutions: map[string]int{
-			"latest": all.Successes + all.Failures,
-			"prev":   allPrev.Successes + allPrev.Failures,
-		},
-		TestPassPercentage: map[string]float64{
-			"latest": all.TestPassPercentage,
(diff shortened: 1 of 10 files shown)


pkg/api/api.go
```diff
@@ -98,17 +79,15 @@ func summaryJobsByPlatform(report, reportPrev sippyprocessingv1.TestReport, endD
...
 	var topFailingTests []sippyv1.FailingTestBug
 
-	allPrev := resultPrev["all"]
-
 	for _, test := range topFailingTestsWithBug {
 		encodedTestName := url.QueryEscape(regexp.QuoteMeta(test.TestName))
 
 		testLink := fmt.Sprintf("%s%s", html.BugSearchUrl, encodedTestName)
-		testPrev := util.GetPrevTest(test.TestName, allPrev.TestResults)
+		testPrev := util.GetTestResult(test.TestName, prevTopFailingTestsWithBug)
```
bparees, line 90: there's no guarantee this test was a top failing test last week.  That's why this looked it up the prv results from the global set of data, not the prevTopFailing list.

bparees: looks like maybe you figured that out in your subsequent commit

bparees: hm. well you fixed it for the html but not for the api

deads2k: > hm. well you fixed it for the html but not for the api

tidied up in 94.  We do need to figure out why there's a second API and if it's supposed to be for a single purpose, why it's so broad.

bparees: the api wasn't created solely for pm score.  the api was created to make the data available to anyone who wanted to access it in a programmatic way for reporting purposes (such as @mfojtik is doing now, as well as the ML team who are doing some analysis)

PR #94, opened 2020-09-15
Job Pass Rates By Most Reduced Pass Rate

deads2k:
Adds slicing for jobs that suddenly start failing based on the last two days of results.  This is currently horrifying....

Screenshot from 2020-09-15 16-54-21

### DIFF
--- a/pkg/html/fastest_moving_jobs.go
+++ b/pkg/html/fastest_moving_jobs.go
@@ -1,0 +1,67 @@
+package html
+
+import (
+	"fmt"
+	"sort"
+
+	"github.com/openshift/sippy/pkg/util"
+
+	sippyprocessingv1 "github.com/openshift/sippy/pkg/apis/sippyprocessing/v1"
+)
+
+func summaryTopNegativelyMovingJobs(twoDaysJobs, prevJobs []sippyprocessingv1.JobResult, jobTestCount int, release string) string {
+	type jobPassChange struct {
+		jobName              string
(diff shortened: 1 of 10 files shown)


pkg/html/fastest_moving_jobs.go
```diff
@@ -0,0 +1,66 @@
...
+	`)
+
+	jobDisplayed := 0
+	for _, jobDetails := range jobPassChanges {
+		jobDisplayed++
+		if jobDisplayed > 10 {
+			break
+		}
+		// don't display things moving in the right direction
+		if jobDetails.passPercentageChange > 0 {
```
bparees, line 50:
```suggestion
		if jobDetails.passPercentageChange > 10 {
```

pkg/html/fastest_moving_jobs.go
```diff
@@ -0,0 +1,66 @@
...
+		</tr>
+	`)
+
+	jobDisplayed := 0
+	for _, jobDetails := range jobPassChanges {
+		jobDisplayed++
+		if jobDisplayed > 10 {
+			break
+		}
+		// don't display things moving in the right direction
```
bparees, line 49:
```suggestion
		// don't display things moving in the right direction or that only dropped within the margin of error
```

pkg/html/fastest_moving_jobs.go
deads2k: updated

PR #96, opened 2020-09-17
split vsphere into upi and ipi variants

deads2k:
These jobs have different characteristics.

### DIFF
--- a/pkg/html/jobaggregationresult.go
+++ b/pkg/html/jobaggregationresult.go
@@ -230,7 +230,7 @@ func (b *jobAggregationResultRenderBuilder) toHTML() string {
 			testsCollapseName,
 			60,
 			test.Name,
-			b.currAggregationResult.AggregationName,
+			getCIJobSubstring(b.currAggregationResult.AggregationName),
 			encodedTestName,
 			bugHTML,
 			test.PassPercentage,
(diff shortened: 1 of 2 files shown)


bparees:
/hold

discussing this in slack.  i don't agree w/ the direciton that variants are being taken in.  This slices what is already only 4 jobs, into 2 jobs.  If you need to see how 2 jobs are doing, just look at the individual jobs, you don't need a variant for that.

Variants are to provide a quick high level view of how the product is doing at a coarse scope.

deads2k:
This is fixing a visualization problem faced by the developers trying to stabilize the platform and the managers trying to support us.  By combining two dislike categories of jobs

1. The jobs are different enough that they have different owners and different problem. This means that finding the responsible parties became harder and we cannot relate a platform to an owner
2. we place the platform in the wrong order.  vsphere-upi is green, vsphere-ipi is very read (about 20%).  By combining them, we've placing them in the wrong spot (about 50%)


The ordering is particular deceptive and has us focusing our effort on debugging the wrong problems.  Clicking on expand jobs doesn't fix this problem and is not a substitute for proper segregation.

bparees:
> These jobs have different characteristics.

every job has different characteristics, that's why there are different jobs.

The question is whether they have more similarities or differences.  Cloud platform success is a significant similarity that is of interest at a high level.  If cloud platform success rate is not where we want it to be, it's already trivial to drill into the jobs that make it up to see where we are falling short (if indeed it's unique to a subset of the jobs that make up the variant).

deads2k:
> The question is whether they have more similarities or differences. Cloud platform success is a significant similarity that is of interest at a high level. If cloud platform success rate is not where we want it to be, it's already trivial to drill into the jobs that make it up to see where we are falling short (if indeed it's unique to a subset of the jobs that make up the variant).

At this moment (and for the past several weeks), they have more differences than similarities.

Variants don't need to be forever. See https://github.com/openshift/sippy/pull/97 for example.

deads2k:
removed hold too early.  We weren't done

/hold

bparees:
/hold cancel

i can accept the value this currently brings given that the jobs being aggregate are, today, behaving more differently than similarly (upi install failing).

but the goal (as i think @deads2k stated and agrees to) should be to resolve the deltas and bring these back together, as was done w/ ocp/origin, not to use this as a general precedent to subdivide everything further absent specific evidence of a fundamental delta.

/lgtm

PR #99, opened 2020-09-18
Refactor to make it easier to tie webserver hits to results produced.

deads2k:
This separates the data from a particular run from the configuration used to produce data on a repeating cycle.

### DIFF
--- a/main.go
+++ b/main.go
@@ -14,13 +14,14 @@ import (
 	"strings"
 	"time"
 
+	"github.com/openshift/sippy/pkg/testgridanalysis/testgridconversion"
+
 	"github.com/openshift/sippy/pkg/api"
 	sippyprocessingv1 "github.com/openshift/sippy/pkg/apis/sippyprocessing/v1"
 	testgridv1 "github.com/openshift/sippy/pkg/apis/testgrid/v1"
 	"github.com/openshift/sippy/pkg/buganalysis"
 	"github.com/openshift/sippy/pkg/html"
 	"github.com/openshift/sippy/pkg/testgridanalysis/testgridanalysisapi"
-	"github.com/openshift/sippy/pkg/testgridanalysis/testidentification"
 	"github.com/openshift/sippy/pkg/testgridanalysis/testreportconversion"
 	"github.com/openshift/sippy/pkg/util"
 	"github.com/openshift/sippy/pkg/util/sets"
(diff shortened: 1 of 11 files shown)


pkg/util/utils.go
```diff
@@ -43,6 +39,15 @@ func GetPlatform(platform string, allPlatforms []sippyprocessingv1.PlatformResul
 	return nil
 }
 
+func FindPrevBugzillaJobFailures(bzComponent string, bugzillaJobFailures []sippyprocessingv1.SortedBugzillaComponentResult) *sippyprocessingv1.SortedBugzillaComponentResult {
```
bparees, line 39: not really specific to Prev, right?  is this just "FindBugzillaJobFailures"?  FindJobFailuresForBZComponent?

deads2k: > not really specific to Prev, right? is this just "FindBugzillaJobFailures"? FindJobFailuresForBZComponent?

Yeah, rough name from long before.  Followup?

bparees: i'm aware the legacy is mine, just figured if you're gonna rename it, let's get it right :)

followup is fine.

main.go
```diff
@@ -379,32 +149,31 @@ func getFailedTestNamesFromJobResults(jobResults map[string]testgridanalysisapi.
...
-	for _, details := range a.TestGridJobInfo {
-		klog.V(2).Infof("processing test details for job %s\n", details.Name)
-		a.processJobDetails(details)
-	}
-
-	// now that we have all the JobRunResults, use them to create synthetic tests for install, upgrade, and infra
-	a.createSyntheticTests()
+// updateBugCacheForJobResults looks up all the bugs related to every failing test in the jobResults and returns a list of
+// warnings/errors that happened looking up the data
+func updateBugCacheForJobResults(bugCache buganalysis.BugCache, rawJobResults testgridanalysisapi.RawData) []string {
```
bparees, line 154:
```suggestion
func updateBugCacheForJobResults(bugCache buganalysis.BugCache, rawData testgridanalysisapi.RawData) []string {
```

pkg/testgridanalysis/testgridconversion/to_raw_data.go
```diff
@@ -0,0 +1,196 @@
...
+		}
+	}
+	return start, len(timestamps)
+}
+
+// tagStripRegex removes test markers deemed unhelpful at one point in time.
+// TODO relitigate the value of doing this.  Without these markers, I don't think it is possible to run the failing test back through `openshift-tests run-test <foo>`
+var tagStripRegex = regexp.MustCompile(`\[Skipped:.*?\]|\[Suite:.*\]`)
+
+// ignoreTestRegex is used to strip o ut tests that don't have predictive or diagnostic value.  We don't want to show these in our data.
```
bparees, line 67:
```suggestion
// ignoreTestRegex is used to strip out tests that don't have predictive or diagnostic value.  We don't want to show these in our data.
```

bparees:
all comments can be handled as followups.

/lgtm

PR #101, opened 2020-09-18
refactor analyzer to support re-use across releases

deads2k:
The configuration itself doesn't vary today, but the release data does. Since it is fully independent, we can re-use the configuration and pass the data use to select the data itself. This will allow very easy parallelization in the future.

Also stop using maps for for release+variant to days and track variant days as a struct.

### DIFF
--- a/pkg/sippyserver/analyzer.go
+++ b/pkg/sippyserver/analyzer.go
@@ -1,120 +1,101 @@
 package sippyserver
 
 import (
-	"bytes"
-	"encoding/json"
 	"fmt"
-	"io/ioutil"
 	"regexp"
-	"strings"
 	"time"
 
-	sippyprocessingv1 "github.com/openshift/sippy/pkg/apis/sippyprocessing/v1"
 	testgridv1 "github.com/openshift/sippy/pkg/apis/testgrid/v1"
+
+	"github.com/openshift/sippy/pkg/testgridanalysis/testgridhelpers"
+
+	sippyprocessingv1 "github.com/openshift/sippy/pkg/apis/sippyprocessing/v1"
 	"github.com/openshift/sippy/pkg/buganalysis"
 	"github.com/openshift/sippy/pkg/testgridanalysis/testgridanalysisapi"
 	"github.com/openshift/sippy/pkg/testgridanalysis/testgridconversion"
-	"github.com/openshift/sippy/pkg/testgridanalysis/testgridhelpers"
 	"github.com/openshift/sippy/pkg/testgridanalysis/testreportconversion"
-	"github.com/openshift/sippy/pkg/util"
 	"github.com/openshift/sippy/pkg/util/sets"
 	"k8s.io/klog"
 )
 
-// TODO I think many of these are dead/inappropriate
-type Options struct {
-	LocalData               string
-	Releases                []string
-	StartDay                int
-	EndDay                  int
-	TestSuccessThreshold    float64
-	JobFilter               string
+// TestGridLoadingOptions control the data which is loaded from disk into the testgrid structs
+type TestGridLoadingConfig struct {
+	// LocalData is the directory where the testgrid data is stored
+	LocalData string
+	// JobFilter is a regex run against job names. Only match names are loaded.
+	JobFilter *regexp.Regexp
+}
+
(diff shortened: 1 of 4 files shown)


pkg/sippyserver/server.go
```diff
@@ -171,45 +162,33 @@ func (s *Server) detailed(w http.ResponseWriter, req *http.Request) {
...
-	twoDayAnalyzer := Analyzer{
-		Release:  release,
-		Options:  optCopy,
-		BugCache: s.bugCache,
-	}
-	twoDayReport := twoDayAnalyzer.PrepareTestReport()
-
-	// prior 7 day period
-	optCopy = opt
-	optCopy.StartDay = endDay + 1
```
deads2k, line 203: this isn't consistent.

pkg/sippyserver/server.go
```diff
@@ -13,55 +14,41 @@ import (
...
-			Release:  release,
-			Options:  o,
-			BugCache: server.bugCache,
-		}
-		server.testReportGeneratorOptions[release+"-days-2"] = analyzer
-
-		// prior 7 day period (days 7-14)
-		optCopy = o
-		optCopy.EndDay = 14
-		optCopy.StartDay = 7
```
deads2k, line 48: with this.

pkg/sippyserver/analyzer.go
```diff
@@ -1,122 +1,103 @@
...
-	Release        string
-
-	BugCache buganalysis.BugCache
+// RawJobResultsAnalysisOptions control which subset of data from the testgrid data is analyzed into the rawJobResults
+type RawJobResultsAnalysisConfig struct {
+	StartDay int
+	EndDay   int
 }
 
-func (a *Analyzer) getTestGridData(releases []string, storagePath string) []testgridv1.JobDetails {
```
deads2k, line 48: this moved to https://github.com/openshift/sippy/pull/101/files#diff-9ae0ddd19e4b07eb469a96212c8537c8R83 as a straight move with args passed.

pkg/testgridanalysis/testgridhelpers/testgridhelpers.go
```diff
@@ -78,7 +78,87 @@ func DownloadData(releases []string, filter string, storagePath string) {
 	}
 }
 
-func LoadJobSummaries(dashboard string, storagePath string) (map[string]testgridv1.JobSummary, time.Time, error) {
+// LoadTestGridDataFromDisk reads the requested testgrid data from disk and returns the details and the timestamp of the last
+// modification on disk
+func LoadTestGridDataFromDisk(storagePath string, releases []string, jobFilter *regexp.Regexp) ([]testgridv1.JobDetails, time.Time) {
```
deads2k, line 83: straight move with args added.

pkg/sippyserver/analyzer.go
```diff
@@ -1,122 +1,103 @@
...
+	// JobFilter is a regex run against job names. Only match names are loaded.
+	JobFilter *regexp.Regexp
 }
 
-type Analyzer struct {
-	Options        Options
-	LastUpdateTime time.Time
-	Release        string
-
-	BugCache buganalysis.BugCache
```
deads2k, line 45: became an arg to stay static.  Could move back to an instance var since all are shared and it's threadsafe.

pkg/sippyserver/server.go
```diff
@@ -13,55 +14,41 @@ import (
...
 type Server struct {
-	bugCache                   buganalysis.BugCache
-	testReportGeneratorOptions map[string]Analyzer
-	currTestReports            map[string]sippyprocessingv1.TestReport
-	options                    Options
+	listenAddr string
+	releases   []string
+
+	bugCache                  buganalysis.BugCache
+	testReportGeneratorConfig TestReportGeneratorConfig
```
deads2k, line 44: only one is needed for the entire server since the params are all the same

pkg/sippyserver/server.go
```diff
@@ -13,55 +14,41 @@ import (
...
+	currTestReports           map[string]StandardReport
```
deads2k, line 45: values are now a struct, so we avoid suffixed key names.

PR #102, opened 2020-09-18
add top level indicator for infra, install, upgrade

deads2k:
Screenshot from 2020-09-22 09-39-45


bparees:
@deads2k give me a heading(preferably w/ mouse-over text explaining the categories)+some separation between the TOC and the top level indicator and lgtm.

bparees:
> How do I do that?

the same way the other tables do, you should be able to copy from them.

deads2k:
> > How do I do that?
> 
> the same way the other tables do, you should be able to copy from them.

Oh, this is the `<th>` thing?

PR #103, opened 2020-09-18
add install page linked from top level indicators.

deads2k:
Adds install section

Screenshot from 2020-10-06 12-02-30


deads2k:
simple rebase, reapplying

PR #104, opened 2020-09-23
refine install, upgrade, infra

deads2k:
Makes the infra, install, upgrade calculation more clear. This uncovered calculation problems where certain platforms never had infra report success or failure.  This very slightly pushes the infra percentages, but makes a fairly  large shift in install (vsphere-upi and metal are installing quite well).  It's pushing upgrade down.

Not all containers are named setup anymore.  Some jobs are currently missing an install step.  That's weird.

### DIFF
--- a/pkg/testgridanalysis/testgridconversion/synthentic_tests.go
+++ b/pkg/testgridanalysis/testgridconversion/synthentic_tests.go
@@ -23,14 +23,16 @@ func createSyntheticTests(rawJobResults testgridanalysisapi.RawData) {
 
 			syntheticTests := map[string]*synthenticTestResult{
 				testgridanalysisapi.InstallTestName:        &synthenticTestResult{name: testgridanalysisapi.InstallTestName},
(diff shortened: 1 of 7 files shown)


pkg/testgridanalysis/testgridconversion/synthentic_tests.go
```diff
@@ -42,31 +44,67 @@ func createSyntheticTests(rawJobResults testgridanalysisapi.RawData) {
 				}
 			}
 			setupFailed := jrr.SetupStatus != testgridanalysisapi.Success
+			setupSucceeded := jrr.SetupStatus == testgridanalysisapi.Success
 
-			if installFailed {
```
deads2k, line 46: reasoning through a total runs difference.  Previously, to be counted as install success or failure you had to 

1. have some operator results, with at least one failed, OR
2. have completed the setup container successfully.  There was a bug that prevented the setup container from being found on platforms without setup containers


In the new world
1. setupsucceeded, with the detection bug fixed
2. have some operator results.

To my mind, this means we need to have a way to detect when we aren't finding a setup container.

PR #107, opened 2020-09-25
Add minimumJobPassRatesByComponent API field

sttts:
Add to the JSON API:

```json
"minimumJobPassRatesByComponent": [{
  "name":"apiserver-auth",
  "passRates": {
    "latest": {
      "name": "<job-name>",
      "percentage": 0.33,
      "runs": 126
    },
    "prev": {
      "name": "<job-name>",
      "percentage": 0.27,
      "runs": 101
    }
  }
}]
```


pkg/api/api.go
```diff
@@ -253,6 +253,30 @@ func canaryTestFailures(all []sippyprocessingv1.FailingTestResult) []sippyv1.Can
...
 }
 
+func minimumJobPassRateByBugzillaComponent(report, prev sippyprocessingv1.TestReport) []sippyv1.MinimumPassRatesByComponent {
+	var result []sippyv1.MinimumPassRatesByComponent
+	for c, failures := range report.JobFailuresByBugzillaComponent {
+		passRate := sippyv1.MinimumPassRatesByComponent{
+			Name: c,
+			PassRates: map[string]sippyv1.PassRate{
+				"latest": {
+					Percentage: failures.JobsFailed[0].FailPercentage,
```
bparees, line 263: hm.  you called this field passrate but you're putting a fail rate into it.

bparees: how about putting `1-FailurePercentage` in here?  Though I think @deads2k might say that's a lie because we only know when this thing failed, we don't know when it passed.

sttts: Good catch. Fixed.a

sttts: optimistic pass rates :-)

PR #108, opened 2020-09-28
convert to startday and offset

deads2k:
This makes -1 mean, "most recent available time" and uses an offset instead of endDay.  This allows `./sippy --local-data historical-data/4.5GA --release 4.5 --server --start-day=-1` to allow visiting http://localhost:8080/?release=4.5 to "just work", making these urls very easy to build.

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -15,7 +15,7 @@ import (
 )
 
 // stats on failure groups
-func failureGroups(failureGroups, failureGroupsPrev []sippyprocessingv1.JobRunResult, endDay int) *sippyv1.FailureGroups {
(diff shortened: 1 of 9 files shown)


pkg/api/api.go
```diff
@@ -295,16 +295,16 @@ func formatJSONReport(report, prevReport sippyprocessingv1.TestReport, endDay, j
 	data := html.TestReports{
 		Current:      report,
 		Prev:         prevReport,
-		EndDay:       endDay,
+		NumDays:      endDay,
```
bparees, line 298: should the arg name on this function change?  (seems like yes?)

mfojtik: yes it should :-)

PR #110, opened 2020-09-29
use parens to indicate pass rate if infrastructure worked

deads2k:
Updates the value in parens to indicate what the pass rate is for jobs that started to run the installer and got at least the bootstrap kube-apiserver up and running.  Given infrastructure failures between 10-15%, this has a meaningful impact on the pass rates of many platforms and jobs.  

We still separately track infrastructure/bootstrapping failures to drive them out, but the number in parens can be thought of as, "after a customer gets an install to start, what are the odds he will be successful".

The previous check was, "if all the tests that have bugs associated with them started passing, what would our pass rate be".  In order to be interpreted as "known failures that have been triaged and deferred" 
1. bugs associated with failing tests are mapped correctly
2. bugs associated with failing tests are the only reasons those tests are failing
3. the bugs have been properly assessed for impact to the product (this requires root causing in many cases) versus impact to the test results (this is easy and what most people do with frequency data and "not a regression" arguments)

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -189,12 +189,12 @@ func summaryJobPassRatesByJobName(report, reportPrev sippyprocessingv1.TestRepor
 				PassRates: map[string]sippyv1.PassRate{
 					"latest": sippyv1.PassRate{
 						Percentage:          v.PassPercentage,
-						ProjectedPercentage: v.PassPercentageWithKnownFailures,
+						ProjectedPercentage: v.PassPercentageWithoutInfrastructureFailures,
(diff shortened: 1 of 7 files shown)


bparees:
all your refactoring must have paid off.  i'm impressed how small this change is.

Can you update the descriptive text?  e.g. here:
https://github.com/openshift/sippy/blob/master/pkg/html/html.go#L154
```
 The pass rate in parenthesis is the projected pass rate ignoring runs which failed only due to tests with associated bugs.
```

and probably a couple other spots.

otherwise lgtm.

PR #111, opened 2020-09-29
pkg/testgridanalysis: Generic handling for sad-operator test-case

wking:
Per openshift/release#12298, the test-case is really about whether the operator has any sad conditions at gather-time, regardless of whether those gathers occured after a failed install or not.

[1]: https://github.com/openshift/release/pull/12298

### DIFF
--- a/pkg/testgridanalysis/testgridanalysisapi/types.go
+++ b/pkg/testgridanalysis/testgridanalysisapi/types.go
@@ -2,6 +2,10 @@
 // long term, I think these types become internal under this package.
 package testgridanalysisapi
 
+import (
+	"regexp"
+)
+
 // 1. TestGrid contains jobs
 // 2. Jobs contain JobRuns.  Jobs have associated variants/platforms.
 // 3. JobRuns contain Tests.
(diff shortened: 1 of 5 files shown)


pkg/testgridanalysis/testgridanalysisapi/types.go
```diff
@@ -69,3 +72,7 @@ const (
 	Success = "Success"
 	Failure = "Failure"
 )
+
+var (
+	OperatorConditionsTestCaseName = regexp.MustCompile("operator (install|conditions) (?P<operator>.*)")
```
bparees, line 77: i don't see any failures matching "operator conditions", is this a pending test case name change?
https://search.ci.openshift.org/?search=operator+conditions&maxAge=48h&context=1&type=bug%2Bjunit&name=&maxMatches=5&maxBytes=20971520&groupBy=job

bparees: i see https://github.com/openshift/release/pull/12298 now.

PR #114, opened 2020-09-29
allow a single BZ to be blamed for all job failures

deads2k:
This allows a job with environment containing "job=\<jobName\>=all" to accept blame for every failing test in the job.  This is more consistent than using platforms, while still keeping a list that is tractably long.

https://bugzilla.redhat.com/show_bug.cgi?id=1883693 is my practice bug

/hold  

Hold until I can see proof that this functions.  I think the ci search cache fills every hour or so.

### DIFF
--- a/pkg/buganalysis/cache.go
+++ b/pkg/buganalysis/cache.go
@@ -11,17 +11,18 @@ import (
 	"sync"
 	"time"
 
-	"github.com/openshift/sippy/pkg/buganalysis/internal"
-
 	bugsv1 "github.com/openshift/sippy/pkg/apis/bugs/v1"
+	"github.com/openshift/sippy/pkg/buganalysis/internal"
 	"k8s.io/klog"
 )
 
 // BugCache is a thread-safe way to query about bug status.
 // It is stateful though, so for a time after clearing the data will not be up to date until the Update is called
 type BugCache interface {
+	ListJobBlockingBugs(job string) []bugsv1.Bug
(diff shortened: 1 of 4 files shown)


deads2k:
/hold cancel

worked locally for me.

pkg/buganalysis/cache.go
```diff
@@ -42,21 +45,62 @@ func NewBugCache() BugCache {
...
+	return lastUpdateError
+}
+
+func getJobKey(jobName string) string {
+	return fmt.Sprintf("job=%v=all", jobName)
+}
+
+// updates a global variable with the bug mapping based on current failures.
+func (c *bugCache) UpdateJobBlockers(jobNames ...string) error {
+	failedTestNames := []string{}
```
bparees, line 69:
```suggestion
  jobSearchStrings := []string{}
```

?  or something.  not failedTestNames though, since it's clearly not that.

pkg/buganalysis/cache.go
```diff
@@ -42,21 +45,62 @@ func NewBugCache() BugCache {
...
+	for _, jobName := range jobNames {
+		failedTestNames = append(failedTestNames, getJobKey(jobName))
+	}
+	newBugs, lastUpdateError := findBugsForFailedTests(failedTestNames...)
```
bparees, line 73: so the ci-search api does return you the BZ metadata in such a way that you could restrict this to only match bugs that have this value in the "environment" field instead of in a comment, if you wanted to reduce the risk of false positives.

probably not worth it but food for thought.

PR #115, opened 2020-09-30
don't include job results for tests that aren't run on the job

deads2k:
if a job didn't run a test, that job should not be included in job list. This happens for our synthetic jobs like install, upgrade, and infra.


abhinavdahiya (approved):
/lgtm

PR #116, opened 2020-10-01
refactor display of tests to always show past results for tests

deads2k:
Unifies display for tests so that we always get previous test data and if available we get the nested sections.


deads2k:
replaced by https://github.com/openshift/sippy/pull/117

PR #117, opened 2020-10-01
use the job aggregation display functionality to render the BZ component view

deads2k:
This brings nearly all the html display into a common format.


deads2k:
replaces https://github.com/openshift/sippy/pull/116

PR #119, opened 2020-10-01
add a curated test section

deads2k:
Add a section of curated tests.  We need this at the moment to track certain test which aren't failing frequently, but which are indicative of underlying problems.

### DIFF
--- a/pkg/testgridanalysis/testreportconversion/by_tests.go
+++ b/pkg/testgridanalysis/testreportconversion/by_tests.go
@@ -21,6 +21,35 @@ func getTopFailingTestsWithoutBug(testResultsByName testResultsByName, testResul
 	}, testResultFilterFn)
 }
 
+// curatedTestSubstrings is keyed by release.  This is a list of tests that are important enough to individually watch.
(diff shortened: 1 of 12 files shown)


pkg/testgridanalysis/testreportconversion/by_tests.go
```diff
@@ -21,6 +21,35 @@ func getTopFailingTestsWithoutBug(testResultsByName testResultsByName, testResul
...
+		"[Feature:SCC][Early] should not have pod creation failures during install",
+		"infrastructure should work",
+		"install should work",
+		"Kubernetes APIs remain available",
+		"OAuth APIs remain available",
+		"OpenShift APIs remain available",
+		"Pod Container Status should never report success for a pending container",
+		"pods should never transition back to pending",
+		"pods should successfully create sandboxes",
+		"upgrade should work",
```
bparees, line 38: you've introduced a couple of "curated lists" of things at this point, can you reconcile them all into a single file so that when the next TRT team takes over and wants to change this, it's all in one obvious location?

i'm thinking of things like `customJobSetupContainers` and `jobsNeverStableForPlatforms`

(long term this should probably all become config yaml, in an ideal world, but i'm realistic)

PR #126, opened 2020-10-08
client: add simple sippy api go client

mfojtik:
This add very simple client that can be used to query the release reports from sippy API. It has minimal dependencies and can be used by third party integrations (slack, dashboards, etc.).

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -292,7 +292,7 @@ func failureGroupList(report sippyprocessingv1.TestReport) []sippyv1.FailureGrou
 	return failureGroups
 }
 
-func formatJSONReport(report, prevReport sippyprocessingv1.TestReport, numDays, jobTestCount int) map[string]interface{} {
+func formatJSONReport(report, prevReport sippyprocessingv1.TestReport, numDays, jobTestCount int) *sippyv1.Report {
 	data := releasehtml.TestReports{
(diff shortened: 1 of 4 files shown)


pkg/client/sippy/v1/types.go
```diff
@@ -0,0 +1,21 @@
...
+	bugsv1 "github.com/openshift/sippy/pkg/apis/bugs/v1"
+	sippyv1 "github.com/openshift/sippy/pkg/apis/sippy/v1"
+)
+
+// releaseReport is returned from Sippy /json API endpoint and it contains map with release name and the actual report.
+// TODO: Fix the returned data and move this map to single structure
+type releaseReport map[string]*Report
+
+// Report contain a single release report returned from Sippy API.
+type Report struct {
```
deads2k, line 13: I don't like having a new type unless it's a strict subset of a different type we already have.

mfojtik: @deads2k it is a wrapper for existing types

bparees: I tend to agree with @deads2k, why can't the server and client share a common type for the resource being provided/consumed via the api endpoint?

fabianvf (commented):
Would it be possible to add something to the README to highlight that this exists?

PR #127, opened 2020-10-08
show additional upgrade related tests

deads2k:
we were missing some that were added during 4.6 and refine "where did the upgrade fail".


deads2k:
see the tests in https://github.com/openshift/origin/blob/a62542da495b59a68ef378c8a2298cb2ea3fdb6b/test/e2e/upgrade/upgrade.go#L335

PR #130, opened 2020-10-09
add final operator health tests

deads2k:
Adds final operator health as a top level indicator and corrects the virtual vs real tests for install and upgrade on consumption.

### DIFF
--- a/pkg/html/installhtml/finalhealth_by_operators.go
+++ b/pkg/html/installhtml/finalhealth_by_operators.go
@@ -1,0 +1,83 @@
+package installhtml
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/openshift/sippy/pkg/util"
+
+	"github.com/openshift/sippy/pkg/util/sets"
+
+	"github.com/openshift/sippy/pkg/testgridanalysis/testgridanalysisapi"
+
+	sippyprocessingv1 "github.com/openshift/sippy/pkg/apis/sippyprocessing/v1"
+)
+
+func operatorHealthTests(curr, prev sippyprocessingv1.TestReport) string {
+	dataForTestsByPlatform := getDataForTestsByPlatform(
+		curr, prev,
+		isOperatorHealthRelatedTest,
+		isOperatorHealthOverallTest,
+	)
+	// compute platform columns before we add the special "All" column
+	platformColumns := sets.StringKeySet(dataForTestsByPlatform.aggregationToOverallTestResult).List()
+
+	// we add an "All" column for all platforms. Fill in the aggregate data for that key
+	for _, testName := range sets.StringKeySet(dataForTestsByPlatform.aggregateResultByTestName).List() {
+		dataForTestsByPlatform.testNameToPlatformToTestResult[testName]["All"] = dataForTestsByPlatform.aggregateResultByTestName[testName].toCurrPrevTestResult()
+	}
+
+	// fill in the data for the first row's "All" column
+	var prevTestResult *sippyprocessingv1.TestResult
+	if installTest := util.FindFailedTestResult(testgridanalysisapi.FinalOperatorHealthTestName, prev.ByTest); installTest != nil {
+		prevTestResult = &installTest.TestResultAcrossAllJobs
+	}
+	dataForTestsByPlatform.aggregationToOverallTestResult["All"] = &currPrevTestResult{
+		curr: util.FindFailedTestResult(testgridanalysisapi.FinalOperatorHealthTestName, curr.ByTest).TestResultAcrossAllJobs,
+		prev: prevTestResult,
+	}
+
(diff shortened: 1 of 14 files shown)


pkg/html/installhtml/finalhealth_by_operators.go
```diff
@@ -0,0 +1,83 @@
...
+
+	return dataForTestsByPlatform.getTableHTML("Operator Health by Operator", "OperatorHealthByOperator", "Operator Health by Operator by Platform", columnNames, getOperatorFromTest)
+}
+
+func summaryOperatorHealthRelatedTests(curr, prev sippyprocessingv1.TestReport, numDays int, release string) string {
+	// test name | bug | pass rate | higher/lower | pass rate
+	s := fmt.Sprintf(`
+	<table class="table">
+		<tr>
+			<th colspan=5 class="text-center"><a class="text-dark" title="Install related tests, sorted by passing rate.  The link will prepopulate a BZ template to be filled out and submitted to report a bug against the test." id="InstallRelatedTests" href="#InstallRelatedTests">Install Related Tests</a></th>
```
bparees, line 50: install related?  not operator related?

pkg/html/releasehtml/top_level_indicators.go
```diff
@@ -18,11 +18,13 @@ func topLevelIndicators(report, reportPrev sippyprocessingv1.TestReport, release
 			<th title="How often we get to the point of running the installer.  This is judged by whether a kube-apiserver is available, it's not perfect, but it's very close." class="text-center {{ .infraColor }}">Infrastructure</th>
 			<th title="How often the install completes successfully." class="text-center {{ .installColor }}"><a href="/install?release={{ .release }}">Install</a></th>
 			<th title="How often an upgrade that is started completes successfully." class="text-center {{ .upgradeColor }}"><a href="/upgrade?release={{ .release }}">Upgrade</a></th>
+			<th title="How often CI job runs finish with all operators healthy." class="text-center {{ .finalColor }}"><a href="/operator-health?release={{ .release }}">Operator Health</a></th>
```
bparees, line 21: is this really a critical top level indicator?

i think it's a good thing to aim for, but it doesn't in and of itself tell us anything about our CI stability or ability to ship the product.

pkg/testgridanalysis/testgridanalysisapi/types.go
```diff
@@ -82,5 +81,5 @@ const (
 
 var (
 	OperatorInstallPrefix          = "operator install "
-	OperatorConditionsTestCaseName = regexp.MustCompile("operator (install|conditions) (?P<operator>.*)")
+	OperatorConditionsTestCaseName = regexp.MustCompile("operator install (?P<operator>.*)")
```
bparees, line 84: we really can't get this test renamed to better align with what it is actually reporting on?

pkg/testgridanalysis/testgridconversion/synthentic_tests.go
```diff
@@ -104,14 +125,37 @@ func createSyntheticTests(rawJobResults testgridanalysisapi.RawData) []string {
...
+							name: testName,
+							pass: 1,
+						}
+					}
+
 				} else {
-					jrr.TestFailures++
-					jrr.FailedTestNames = append(jrr.FailedTestNames, testgridanalysisapi.UpgradeTestName)
 					syntheticTests[testgridanalysisapi.UpgradeTestName].fail = 1
+					// if the test succeeded, then the operator upgrade tests should match the operator state
```
bparees, line 139:
```suggestion
					// if the test failed, then the operator upgrade tests should match the operator state
```

pkg/testgridanalysis/testgridanalysisapi/types.go
deads2k: > we really can't get this test renamed to better align with what it is actually reporting on?

There will still be an install test.  it will simply change to be virtual.  But you need to track the old value or you won't be able to read your historical data.  This code handles both properly so when https://github.com/openshift/release/pull/12298 merges, we won't have trouble.

pkg/html/releasehtml/top_level_indicators.go
deads2k: > is this really a critical top level indicator?
> 
> i think it's a good thing to aim for, but it doesn't in and of itself tell us anything about our CI stability or ability to ship the product.

I'm about done in 4.6, so I'm probably willing to comment out these lines and let the 4.7 TRT decide.  However, when you look at the differences between install/upgrade and operator health, you can see cases where we are able to install without issue, but cannot keep a cluster healthy for the duration of an e2e test.  Stand out cases in comparison include

1. azure - healthy less than 90% of the time, but installs over 90%.  Monitoring is particularly prone to failure
2. ovirt - installs more commonly than it ends healthy. authentication and console are particularly bad, which suggests that ingress goes offline at some point

PR #132, opened 2020-10-12
ovirt upgrade and three upgrade test have passed more than 50% of the time

deads2k:
The three day signal looks pretty good.  Opening this to be sure we recheck at the end of this week to see if we passed the bar for treating them as normal jobs.

/hold


deads2k:
ovirt is still in the 40s.  opened https://github.com/openshift/sippy/pull/135 to show triple upgrade results.

PR #134, opened 2020-10-14
Test details for test sections

deads2k:
Adds test details buttons for each aggregation we have.  It allows a button from any level to say, "are the problems on this test, job, or platform different by platform" in a quick chart

### DIFF
--- a/pkg/html/releasehtml/failing_tests.go
+++ b/pkg/html/releasehtml/failing_tests.go
@@ -11,11 +11,11 @@ import (
 )
 
 func summaryTopFailingTestsWithBug(topFailingTestsWithBug, allTests []sippyprocessingv1.FailingTestResult, numDays int, release string) string {
-	// test name | bug | pass rate | higher/lower | pass rate
+	rows, testNames := topFailingTestsRows(topFailingTestsWithBug, allTests, release)
 	s := fmt.Sprintf(`
 	<table class="table">
 		<tr>
-			<th colspan=5 class="text-center"><a class="text-dark" title="Most frequently failing tests with a known bug, sorted by passing rate.  The link will prepopulate a BZ template to be filled out and submitted to report a bug against the test." id="TopFailingTestsWithABug" href="#TopFailingTestsWithABug">Top Failing Tests With A Bug</a></th>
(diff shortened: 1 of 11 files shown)


bparees:
1) errors are being duplicated:
image

2) I think the `no-data` cells should be gray rather than green 

3) including the "test details by platforms" button on the test details page seems inceptionish (the information is already in the table above).

4) why does one of these have buttons and the other does not?
image

pkg/html/releasehtml/failing_tests.go
```diff
@@ -11,74 +11,74 @@ import (
...
-
-	s = s + "</table>"
+	s += rows
+	s += "</table>"
 
 	return s
 }
 
-func topFailingTestsRows(topFailingTests, prevTests []sippyprocessingv1.FailingTestResult, release string) string {
+// returns the rows to display and the names of the tests being show
```
bparees, line 77:
```suggestion
// returns the rows to display and the names of the tests being shown
```

deads2k:
> 1. errors are being duplicated:

prexisting



> 3\. including the "test details by platforms" button on the test details page seems inceptionish (the information is already in the table above).

open to other names



> 4\. why does one of these have buttons and the other does not?

The one without buttons has no tests to expand.  thresholds on failures not met

bparees:
> open to other names

i'm suggesting not having it.  It doesn't take me anywhere more useful than where i already am.

deads2k:
> > open to other names
> 
> i'm suggesting not having it. It doesn't take me anywhere more useful than where i already am.

it does a different thing. It answers the question of "are the tests failing on this platform the same or different than the other platforms" in a way that no other view does.

pkg/testgridanalysis/testreportconversion/testresult.go
```diff
@@ -215,7 +215,7 @@ func StandardTestResultFilter(
 		TestResultFilterFuncs{
 			FilterLowValueTestsByName,
 			FilterTooFewTestRuns(minRuns),
-			FilterSuccessfulTestResults(successThreshold),
+			//FilterSuccessfulTestResults(successThreshold),
```
bparees, line 218: you can drop the TODO on line 210 w/ this change

PR #138, opened 2020-10-15
fix handling for promotion jobs that don't always set up a cluster

deads2k:
the promotion jobs only do work if there is a diff.  When there is no diff, they never install a cluster.  When they don't install a cluster, we don't get a positive, "the infra worked" signal.


pkg/testgridanalysis/testgridconversion/synthentic_tests.go
```diff
@@ -107,6 +107,9 @@ func createSyntheticTests(rawJobResults testgridanalysisapi.RawData) []string {
 
 			// set the infra status
 			switch {
+			case jobsWithKnownBadSetupContainer.Has(jobName):
+				// do nothing.  If we don't have a setup container, we have now way of determining infrastructure
```
wking, line 111: "now" -> "no"?

PR #139, opened 2020-10-20
Make gcp variant regex match 3.11 jobs

bparees:
3.11 job name is periodic-ci-openshift-origin-release-3.11-e2e-gcp so the current regex misses it.

@deads2k as an aside, it's probably worth us thinking about what to do in the case of "0 runs" both current and previous week for a given variant, as happens here:
https://sippy.ci.openshift.org/?release=3.11

Hiding it is a little scary since if something stops running, we lose sight of it.  But it's not hard to imagine that the set of valid variants evolves from release to release and keeping a bunch of scary looking empty red rows around for that reason doesn't seem ideal in terms of bringing visibility to the things we do care about.


bparees:
/assign @abhinavdahiya 

you've been identified as a potential future sippy maintainer :)

PR #141, opened 2020-10-20
add option to disable bugzilla lookups to speed up debug iteration

bparees:
adds --skip-bug-lookup as an arg that will skip over the ci-search bz lookup.

This reduces startup analysis time substantially, useful when iterating on code
that is not related to the bug assocation/reporting.

### DIFF
--- a/pkg/buganalysis/cache.go
+++ b/pkg/buganalysis/cache.go
@@ -29,6 +29,32 @@ type BugCache interface {
 	LastUpdateError() error
 }
 
+// noOpBugCache is a no-op implementation of the bug cache/lookup interface
+// used to opt-out of bug lookups for faster test analysis/reporting.
+type noOpBugCache struct {
+}
+
+func (*noOpBugCache) ListJobBlockingBugs(job string) []bugsv1.Bug {
+	return []bugsv1.Bug{}
+}
+func (*noOpBugCache) ListBugs(release, platform, testName string) []bugsv1.Bug {
+	return []bugsv1.Bug{}
+}
+func (*noOpBugCache) UpdateForFailedTests(failedTestNames ...string) error {
+	return nil
+}
+func (*noOpBugCache) UpdateJobBlockers(jobNames ...string) error {
+	return nil
+}
+func (*noOpBugCache) Clear() {}
+func (*noOpBugCache) LastUpdateError() error {
+	return nil
+}
+
+func NewNoOpBugCache() BugCache {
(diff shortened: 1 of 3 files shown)


bparees:
@deads2k may be of interest to you.

/assign @abhinavdahiya

pkg/buganalysis/cache.go
```diff
@@ -37,7 +59,10 @@ type bugCache struct {
 	lastUpdateError error
 }
 
-func NewBugCache() BugCache {
+func NewBugCache(skipBugLookup bool) BugCache {
+	if skipBugLookup {
+		return &dummyBugCache{}
+	}
```
abhinavdahiya, line 65: I think we should let the caller decide which bug cache they need.
so the caller calling
```
if skipBugCache {
  NewDummyBugCache()
} else {
  NewBugCache()
}
```

this allows the actual bug cache implementation `new` function to add options later on for initialization and not have this unrelated argument in there.

pkg/buganalysis/cache.go
```diff
@@ -29,6 +29,28 @@ type BugCache interface {
 	LastUpdateError() error
 }
 
+// dummyBugCache is a no-op implementation of the bug cache/lookup interface
+// used to opt-out of bug lookups for faster test analysis/reporting.
+type dummyBugCache struct {
```
abhinavdahiya, line 34: instead of dummy maybe we call it noOpBugCache , dummy imo doesn't convey the expected impl

pkg/buganalysis/cache.go
bparees: yeah that was my first version.  i'm fine w/ going back to that.

will update.

pkg/buganalysis/cache.go
bparees: works for me, will update.

bparees:
@abhinavdahiya updated for your comments

pkg/sippyserver/server.go
```diff
@@ -21,11 +21,18 @@ func NewServer(
 	displayDataOptions DisplayDataConfig,
 	releases []string,
 	listenAddr string,
+	skipBugLookup bool,
```
abhinavdahiya, line 24: hmm.. i don't see the call site of `NewServer` being updated in this PR? This change will break it right?

abhinavdahiya: Ah, nvm saw it.

PR #143, opened 2020-10-22
Filter out zero run jobs from negatively moving results

jluhrsen:
No need bring a job to a list of those that are regressing
if they have actually never run.

### DIFF
--- a/pkg/html/releasehtml/fastest_moving_jobs.go
+++ b/pkg/html/releasehtml/fastest_moving_jobs.go
@@ -25,10 +25,14 @@ func summaryTopNegativelyMovingJobs(twoDaysJobs, prevJobs []sippyprocessingv1.Jo
 		if prevJob == nil {
 			continue
 		}
-		jobPassChanges = append(jobPassChanges, jobPassChange{
-			jobName:              job.Name,
-			passPercentageChange: job.PassPercentage - prevJob.PassPercentage,
-		})
+
+		// in case our job actually never ran we don't need to show it as a negatively moving result
+		if len(job.TestResults) != 0 {
+			jobPassChanges = append(jobPassChanges, jobPassChange{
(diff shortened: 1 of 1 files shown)


pkg/html/releasehtml/fastest_moving_jobs.go
```diff
@@ -25,10 +25,14 @@ func summaryTopNegativelyMovingJobs(twoDaysJobs, prevJobs []sippyprocessingv1.Jo
 		if prevJob == nil {
 			continue
 		}
-		jobPassChanges = append(jobPassChanges, jobPassChange{
-			jobName:              job.Name,
-			passPercentageChange: job.PassPercentage - prevJob.PassPercentage,
-		})
+
+		// in case our job actually never ran we don't need to show it as a negatively moving result
```
bparees, line 29: it's sort of concerning if a job went from running the previous week to not running this week, though, right?

jluhrsen: that's something I don't have experience in knowing yet... are there jobs that just don't run for a couple days in a row for some good reasons or not? Maybe 2 days is too short of a window?

Anyway, I wanted to poke around sippy code and I recalled @deads2k making note how we weren't filtering out 0 run
jobs so figured that was a good way to learn and play with the code.

bparees: There shouldn't be any that don't run for good reason.  there may be some that don't run for bad reasons.

Which makes me think this is likely an undesirable change (because it should not happen under normal circumstances anyway)

jluhrsen: fair enough.

jluhrsen:
we probably do want to see zero runs in this case. closing.

PR #145, opened 2020-10-27
Update jobsWithKnownBadSetupContainer to include 4.7 versions of the already known issues

fabianvf:
Since we're seeing the same analysis errors I'm making the assumption that this hasn't changed since 4.6. If these are never expected to work perhaps we should change them to a regex match so that it won't need to be updated for future releases.

Alternatively, if we do expect this to be fixed over time as the comment suggests, is there currently some way we can track/are tracking it in a GH issue/PR/BZ? Then we could at least add that info to the warning so it's clear it can be ignored for the time being.


fabianvf:
closed in favor of #147

PR #147, opened 2020-10-28
Update SetupContainer list; fix setup warnings

jluhrsen:
moves jobsWithBadSetupContainer list to use a regex pattern for each job
to help with matching on multiple releases. May be useful in other
cases later.

- excludes all jobs matching
  promote-release-openshift-machine-os-content-e2e-aws-4.[2-7].*

  These jobs were showing warnings in all other releases except for
  4.6

### DIFF
--- a/pkg/testgridanalysis/testgridconversion/synthentic_tests.go
+++ b/pkg/testgridanalysis/testgridconversion/synthentic_tests.go
@@ -2,6 +2,7 @@ package testgridconversion
 
 import (
 	"fmt"
+	"regexp"
 
 	"github.com/openshift/sippy/pkg/testgridanalysis/testgridanalysisapi"
 	"github.com/openshift/sippy/pkg/util/sets"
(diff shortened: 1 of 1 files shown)


fabianvf (approved):
/lgtm

stevekuznetsov:
The promote jobs still run a cluster, right? I don't think ignoring them is the right thing to do

jluhrsen:
> The promote jobs still run a cluster, right? I don't think ignoring them is the right thing to do

I think this is just to ignore marking them as missing a setup stage, which ends up in a warning you see at the top of
the page, like:

"promote-release-openshift-machine-os-content-e2e-aws-4.7" is missing a test setup job to indicate successful installs"

stevekuznetsov:
Right I guess what I am saying is the fix should be to find their setup container, which does exist, instead of removing the warning by ignoring them

stevekuznetsov:
The job only sometimes actually runs a test - it's polling to see if it needs to promote something new, doing nothing when it doesn't -- so maybe the logic should just check that _some_ jobs have the container?

jluhrsen:
yeah, I think that's the idea down the road. There are other jobs with similar problems I guess. See the comments in the code.
This was just to stay in line with what we had by ignoring it in 4.6 and now was showing up in 4.7 and scaring people. We have
this issue to track doing a better job at this: https://github.com/openshift/sippy/issues/149

deads2k:
I don't speak regex, but if you want it ok.

/approve

PR #152, opened 2020-11-02
add sig name for logging testing

QiaolingTang:
Hi @deads2k, per your suggestion, I opened this PR, could you please help take a look?

### DIFF
--- a/pkg/testgridanalysis/testidentification/component.go
+++ b/pkg/testgridanalysis/testidentification/component.go
@@ -136,6 +136,7 @@ func init() {
 	must(addSigMapping("sig-imageregistry", "Image Registry"))
 	must(addSigMapping("sig-network", "Networking"))
 	must(addSigMapping("sig-node", "Node"))
+	must(addSigMapping("sig-openshift-logging", "Logging"))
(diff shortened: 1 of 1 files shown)


pkg/testgridanalysis/testidentification/component.go
```diff
@@ -136,6 +136,7 @@ func init() {
 	must(addSigMapping("sig-imageregistry", "Image Registry"))
 	must(addSigMapping("sig-network", "Networking"))
 	must(addSigMapping("sig-node", "Node"))
+	must(addSigMapping("sig-openshift-logging", "Logging"))
```
bparees, line 139: this is the sig name that was chosen, and not `sig-logging`?
(if so this seems fine, just want to confirm)

QiaolingTang: In [DPTP-1691](https://issues.redhat.com/browse/DPTP-1691?focusedCommentId=15378281&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-15378281), I was suggested to use `sig-openshift-logging` as the sig name.

PR #154, opened 2020-11-06
Add section for jobs with specific build log msgs

jluhrsen:
There may be a specific build log message we are interested
in knowing all the jobs that contain it that doesn't
neccessarily show up in a junit test case. This commit
is adding the message 'Failed to pull image "busybox"' which
can help to know how prevalent the problem is where the
docker image pull rate limiting is causing image pull
failures. But, the logic could easily be refactored to
have a list of messages to search for and a simple loop
on those could produce however many sections of these
job groupings that would be useful.

### DIFF
--- a/pkg/html/releasehtml/html.go
+++ b/pkg/html/releasehtml/html.go
@@ -2,6 +2,7 @@ package releasehtml
 
 import (
 	"fmt"
+	"io/ioutil"
 	"net/http"
 	"net/url"
 	"regexp"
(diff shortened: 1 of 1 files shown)


jluhrsen:
This is probably just POC level of my idea to include a list of jobs that are having some message in the build-log
that is not showing up in the testgrid data we use for populating sippy (I think). It's essentially just doing the same
search we can do manually at https://search.ci.openshift.org/ so it may not feel all that useful to some. It does put
jobs of interest in the single sippy pane of glass though. It could be useful for multiple things down the line. One
would be to find all the jobs that are hitting the 4h prow timeout.

anyway, it's not very pretty right now. Here is a screenshot, the bottom is where the new job output can be seen.

sippy

bparees:
since this is just in-lining the result of a ci search query, and hopefully it's a short-term condition we resolve and move on from, i'd suggest just going to ci-search directly for this information rather than adding it to sippy.

jluhrsen:
> since this is just in-lining the result of a ci search query, and hopefully it's a short-term condition we resolve and move on from, i'd suggest just going to ci-search directly for this information rather than adding it to sippy.

yeah, I don't really disagree. we can close it, but it could be useful if there are other
searches as well so I figured I'd put it out there. for example, one bug I think is wasteful
to our infra are the 4 hour prow timeouts. There may be others. Anyway, feel free to
close.

PR #155, opened 2020-11-12
allow sippy to wire against kube

deads2k:
Refactors the args to allow execution like `./sippy --server --local-data ../data-dir --dashboard=kube-master=sig-release-master-blocking,sig-release-master-informing=` that allows wiring against kubernetes (or non-openshift releases).

The `--release` flag still work and pipes to this new and specific mechanism.

### DIFF
--- a/main.go
+++ b/main.go
@@ -44,7 +44,6 @@ func main() {
 		FailureClusterThreshold: 10,
 		StartDay:                0,
 		ListenAddr:              ":8080",
-		OpenshiftReleases:       []string{"4.4"},
 	}
 
 	klog.InitFlags(nil)
@@ -138,6 +137,13 @@ func (o *Options) Validate() error {
 	case "json":
 	default:
 		return fmt.Errorf("invalid output type: %s\n", o.Output)
+	}
+
+	for _, dashboard := range o.Dashboards {
+		tokens := strings.Split(dashboard, "=")
+		if len(tokens) != 3 {
+			return fmt.Errorf("must have three tokens: %q", dashboard)
+		}
 	}
 
 	return nil
--- a/main.go
+++ b/main.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"os"
 	"regexp"
+	"strings"
 
 	"github.com/openshift/sippy/pkg/buganalysis"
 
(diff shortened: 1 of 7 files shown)


main.go
```diff
@@ -91,10 +93,45 @@ func (o *Options) Complete() error {
...
 }
 
+func (o *Options) ToTestGridDashboardCoordinates() []sippyserver.TestGridDashboardCoordinates {
+	dashboards := []sippyserver.TestGridDashboardCoordinates{}
+	for _, dashboard := range o.Dashboards {
+		tokens := strings.Split(dashboard, "=")
+		if len(tokens) != 3 {
+			// launch error
+			panic(fmt.Sprintf("must have three tokens: %q", dashboard))
+		}
```
abhinavdahiya, line 110: could we move this to validate function? making sure the Dashboard is always in the format we need?

deads2k: > could we move this to validate function? making sure the Dashboard is always in the format we need?

I would still keep this panic, but sure, adding it to validate is reasonable.

pkg/sippyserver/server.go
```diff
@@ -69,63 +89,86 @@ func (s *Server) refresh(w http.ResponseWriter, req *http.Request) {
...
-		s.currTestReports[release].CurrentTwoDayReport,
-		s.currTestReports[release].PreviousWeekReport,
+		s.currTestReports[dashboard.ReportName].CurrentPeriodReport,
+		s.currTestReports[dashboard.ReportName].CurrentTwoDayReport,
+		s.currTestReports[dashboard.ReportName].PreviousWeekReport,
 		s.testReportGeneratorConfig.RawJobResultsAnalysisConfig.NumDays,
 		15)
 }
 
+func (s *Server) reportNameToDashboardCoordinates(release string) (TestGridDashboardCoordinates, bool) {
```
bparees, line 118:
```suggestion
func (s *Server) reportNameToDashboardCoordinates(reportName string) (TestGridDashboardCoordinates, bool) {
```

?

bparees (commented):
nit and lgtm

main.go
```diff
@@ -91,23 +93,70 @@ func (o *Options) Complete() error {
...
 	case "json":
 	default:
 		return fmt.Errorf("invalid output type: %s\n", o.Output)
 	}
 
+	for _, dashboard := range o.Dashboards {
+		tokens := strings.Split(dashboard, "=")
+		if len(tokens) != 3 {
+			// launch error
+			fmt.Errorf("must have three tokens: %q", dashboard)
```
bparees, line 146: shouldn't this be returned?

PR #158, opened 2020-11-19
allow alternative variants per server

deads2k:
This will allow us to specify different variants for kube.  It just makes an instance variable that is passed around.

### DIFF
--- a/main.go
+++ b/main.go
@@ -9,9 +9,9 @@ import (
 	"strings"
 
 	"github.com/openshift/sippy/pkg/buganalysis"
-
 	"github.com/openshift/sippy/pkg/sippyserver"
 	"github.com/openshift/sippy/pkg/testgridanalysis/testgridhelpers"
+	"github.com/openshift/sippy/pkg/testgridanalysis/testidentification"
 	"github.com/spf13/cobra"
 	"k8s.io/klog"
 )
@@ -177,7 +177,8 @@ func (o *Options) runServerMode() error {
 		o.toDisplayDataConfig(),
 		o.ToTestGridDashboardCoordinates(),
 		o.ListenAddr,
-		o.SkipBugLookup,
+		o.getVariantManager(),
+		o.getBugCache(),
 	)
 	server.RefreshData() // force a data refresh once before serving.
 	server.Serve()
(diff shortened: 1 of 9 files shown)


pkg/testgridanalysis/testidentification/variants.go
```diff
@@ -1,156 +1,16 @@
...
 
-	"github.com/openshift/sippy/pkg/util/sets"
-	"k8s.io/klog"
-)
+// VariantManager identifies and describes different variants
+type VariantManager interface {
+	// allOpenshiftVariants returns a set of all known variants
+	AllVariants() sets.String
 
-var (
```
deads2k, line 11: this all moved to ocp_variants

pkg/testgridanalysis/testidentification/ocp_variants.go
```diff
@@ -0,0 +1,165 @@
...
+
+import (
+	"fmt"
+	"regexp"
+
+	"github.com/openshift/sippy/pkg/util/sets"
+	"k8s.io/klog"
+)
+
+var (
```
deads2k, line 11: a file move and interface alignment.

main.go
```diff
@@ -190,18 +191,38 @@ func (o *Options) runCLIReportMode() error {
...
+	if o.SkipBugLookup || !o.hasOpenshiftRelease() {
+		return buganalysis.NewNoOpBugCache()
+	} else {
+		return buganalysis.NewBugCache()
+	}
+}
+
+func (o *Options) getVariantManager() testidentification.VariantManager {
+	if o.hasOpenshiftRelease() {
+		return testidentification.NewOpenshiftVariantManager()
```
bparees, line 220: it's not immediately obvious to me why the set of variants should be tied to whether or not what we are analyzing includes an openshift release version?

deads2k: > it's not immediately obvious to me why the set of variants should be tied to whether or not what we are analyzing includes an openshift release version?

Depending on what test-grid you're looking at, you have different variants.  See https://testgrid.k8s.io/sig-release-master-informing for instance.  It's likely to have a kube set

bparees: yeah, i get that you might have a reason to want different variants depending what set of testgrid jobs you are analyzing.  i'm just not convinced "we have an openshift release" is the right discriminator for choosing the appropriate set of variants.

deads2k: > yeah, i get that you might have a reason to want different variants depending what set of testgrid jobs you are analyzing. i'm just not convinced "we have an openshift release" is the right discriminator for choosing the appropriate set of variants.

So this is trying to determine: "am I looking at a set of testgrids that includes an openshift".  How would you like to determine that?

bparees:
/lgtm

with iou on "add a --variant-manager" option and do something more explicit in terms of defaulting the variant manager rather than the currently somewhat fragile/limited "do we have an openshift release value"(which doesn't work if someone doesn't specify --release) decision logic.

PR #159, opened 2020-11-19
elide sections without data

deads2k:
This leaves the header at the top in place, but it hides the empty tables at least.

### DIFF
--- a/pkg/html/releasehtml/failing_tests.go
+++ b/pkg/html/releasehtml/failing_tests.go
@@ -11,6 +11,10 @@ import (
 )
 
 func summaryTopFailingTestsWithBug(topFailingTestsWithBug, allTests []sippyprocessingv1.FailingTestResult, numDays int, release string) string {
+	if len(topFailingTestsWithBug) == 0 {
+		return ""
+	}
+
 	rows, testNames := topFailingTestsRows(topFailingTestsWithBug, allTests, release)
 	s := fmt.Sprintf(`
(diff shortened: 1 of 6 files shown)


pkg/html/releasehtml/failing_tests.go
```diff
@@ -11,6 +11,10 @@ import (
 )
 
 func summaryTopFailingTestsWithBug(topFailingTestsWithBug, allTests []sippyprocessingv1.FailingTestResult, numDays int, release string) string {
+	if len(topFailingTestsWithBug) == 0 {
+		return ""
```
bparees, line 15: hm.  i think i'd prefer some sort of "this page(table) intentionally left blank" message over elliding it entirely.

deads2k: > hm. i think i'd prefer some sort of "this page(table) intentionally left blank" message over elliding it entirely.

It reads a little odd when wiring against dashboards that don't have bugzilla or issues wired.

bparees: hm, fair enough.  but i reserve the right to revisit this if i start getting a lot of questions from build-watchers about blank sections :)

deads2k: > hm, fair enough. but i reserve the right to revisit this if i start getting a lot of questions from build-watchers about blank sections :)

Accepted, thanks.

PR #160, opened 2020-11-19
allow synthetic test calculation to be configured

deads2k:
Also updates the openshift detection to inspect the dashboard names.

### DIFF
--- a/main.go
+++ b/main.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/openshift/sippy/pkg/buganalysis"
 	"github.com/openshift/sippy/pkg/sippyserver"
+	"github.com/openshift/sippy/pkg/testgridanalysis/testgridconversion"
 	"github.com/openshift/sippy/pkg/testgridanalysis/testgridhelpers"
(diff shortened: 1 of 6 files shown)


main.go
```diff
@@ -192,37 +194,47 @@ func (o *Options) runCLIReportMode() error {
...
 
-	testReport := analyzer.PrepareTestReport(o.ToTestGridDashboardCoordinates()[0], o.getVariantManager(), o.getBugCache())
+	testReport := analyzer.PrepareTestReport(o.ToTestGridDashboardCoordinates()[0], o.getSynthenticTestManager(), o.getVariantManager(), o.getBugCache())
 	enc := json.NewEncoder(os.Stdout)
 	enc.Encode(testReport)
 	return nil
 }
 
-func (o *Options) hasOpenshiftRelease() bool {
+func (o *Options) hasOpenshiftDashboard() bool {
```
bparees, line 203: you may want to preemptively start referring to this as the OCP dashboard, since there is also an OKD dashboard that might get pulled in at some point.

bparees:
suggestion for future-proofing, otherwise lgtm

PR #162, opened 2020-12-04
Add osd platform

fabianvf:


bparees:
do we have jobs on the 4.7 informing/blocking testgrid that match this?

fabianvf:
@bparees I believe so https://github.com/openshift/release/blob/master/core-services/testgrid-config-generator/_allow-list.yaml#L53-L54

PR #171, opened 2020-12-09
Precompute multiple days of results

fabianvf:
Adds the `endDay` parameter to the normal dashboard, allowing you to specify how many days back results will be accounted for. Precomputes the results for each day before the default `endDay` so that it doesn't need to process so much data each time it runs.

This is just the initial stab/POC

### DIFF
--- a/pkg/sippyserver/server.go
+++ b/pkg/sippyserver/server.go
@@ -66,9 +66,10 @@ type TestGridDashboardCoordinates struct {
 }
 
 type StandardReport struct {
-	CurrentPeriodReport sippyprocessingv1.TestReport
-	CurrentTwoDayReport sippyprocessingv1.TestReport
-	PreviousWeekReport  sippyprocessingv1.TestReport
+	CurrentPeriodReports []sippyprocessingv1.TestReport
+	CurrentPeriodReport  sippyprocessingv1.TestReport
+	CurrentTwoDayReport  sippyprocessingv1.TestReport
+	PreviousWeekReport   sippyprocessingv1.TestReport
 }
 
 func (s *Server) refresh(w http.ResponseWriter, req *http.Request) {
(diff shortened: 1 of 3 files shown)


pkg/sippyserver/server.go
```diff
@@ -66,9 +66,10 @@ type TestGridDashboardCoordinates struct {
 }
 
 type StandardReport struct {
-	CurrentPeriodReport sippyprocessingv1.TestReport
-	CurrentTwoDayReport sippyprocessingv1.TestReport
-	PreviousWeekReport  sippyprocessingv1.TestReport
+	CurrentPeriodReports []sippyprocessingv1.TestReport
+	CurrentPeriodReport  sippyprocessingv1.TestReport
+	CurrentTwoDayReport  sippyprocessingv1.TestReport
```
deads2k, line 71: If you precompute all reports, this become unnecessary, right?

pkg/sippyserver/server.go
```diff
@@ -66,9 +66,10 @@ type TestGridDashboardCoordinates struct {
...
+	PreviousWeekReport   sippyprocessingv1.TestReport
```
deads2k, line 72: seems like this should also be a slice so that you get the correct, "seven days before the prior interval", right?

pkg/sippyserver/server.go
```diff
@@ -66,9 +66,10 @@ type TestGridDashboardCoordinates struct {
 }
 
 type StandardReport struct {
-	CurrentPeriodReport sippyprocessingv1.TestReport
-	CurrentTwoDayReport sippyprocessingv1.TestReport
-	PreviousWeekReport  sippyprocessingv1.TestReport
+	CurrentPeriodReports []sippyprocessingv1.TestReport
+	CurrentPeriodReport  sippyprocessingv1.TestReport
```
deads2k, line 70: this ought not be needed, right?

You could simplify the diff by creating getters, right?

pkg/sippyserver/server.go
```diff
@@ -99,11 +100,20 @@ func (s *Server) printHtmlReport(w http.ResponseWriter, req *http.Request) {
 		return
 	}
 
+	var numDays int
+	t := req.URL.Query().Get("endDay")
```
deads2k, line 104: doesn't this get set on the "detail" page today?

fabianvf: yes, but the `detailed` page has a ton of options and isn't really pre-computable (we could probably cache things after they are computed the first time but that's a little bit more involved I think). This just adds the parameter to the regular page

pkg/sippyserver/server.go
fabianvf: yep this was just a quick POC so I avoided too much refactoring, but we could definitely remove these

pkg/sippyserver/server.go
fabianvf: yeah I guess we should compute the rolling prior 7 days

pkg/sippyserver/server.go
fabianvf: yep will just require more refactoring so wanted to make sure the approach was sound before I went too deep

fabianvf:
#174 fixes the performance issues on the `/detailed` endpoint, so this is no longer necessary

PR #174, opened 2020-12-11
Check bug cache for bugs before searching again

fabianvf:
This should make the `/detailed` endpoint much quicker, as the previous
behavior was searching for every single bug every time it was loaded
with absolutely no caching.

### DIFF
--- a/pkg/buganalysis/cache.go
+++ b/pkg/buganalysis/cache.go
@@ -71,12 +71,14 @@ func NewBugCache() BugCache {
 
 // updates a global variable with the bug mapping based on current failures.
 func (c *bugCache) UpdateForFailedTests(failedTestNames ...string) error {
+	c.lock.RLock()
 	newFailedTestNames := []string{}
 	for _, testName := range failedTestNames {
 		if _, found := c.cache[testName]; !found {
 			newFailedTestNames = append(newFailedTestNames, testName)
 		}
 	}
+	c.lock.RUnlock()
 	newBugs, lastUpdateError := findBugsForFailedTests(newFailedTestNames...)
 
 	c.lock.Lock()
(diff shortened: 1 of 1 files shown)


pkg/buganalysis/cache.go
```diff
@@ -71,15 +71,18 @@ func NewBugCache() BugCache {
 
 // updates a global variable with the bug mapping based on current failures.
 func (c *bugCache) UpdateForFailedTests(failedTestNames ...string) error {
-	newBugs, lastUpdateError := findBugsForFailedTests(failedTestNames...)
+	newFailedTestNames := []string{}
+	for _, testName := range failedTestNames {
+		if _, found := c.cache[testName]; !found {
```
bparees, line 76: to be "safe" this needs to be inside the lock.

pkg/buganalysis/cache.go
```diff
@@ -94,17 +97,17 @@ func getJobKey(jobName string) string {
 func (c *bugCache) UpdateJobBlockers(jobNames ...string) error {
 	jobSearchStrings := []string{}
 	for _, jobName := range jobNames {
-		jobSearchStrings = append(jobSearchStrings, getJobKey(jobName))
+		jobKey := getJobKey(jobName)
+		if _, found := c.jobBlockers[jobKey]; !found {
```
bparees, line 101: same here, needs to be inside the lock

pkg/buganalysis/cache.go
fabianvf: cool, wasn't sure if read operations needed to be inside the lock. the `findBugsForFailedTests` function can be quite slow, should I lock/unlock for reads, run that function, then lock/defer unlock afterward? Presumably we don't want to lock for 10s of seconds while we wait for bug lookups

bparees: take a readlock on it, like ListBugs does.

readLock
  check cache, accumulate list of names not in the cache
unreadlock

bugquery

writelock
  update cache w/ query results
unwritelock

i think that achieves the goal.

fabianvf:
@bparees read locking has been added

bparees:
/lgtm

thanks!

PR #175, opened 2020-12-18
decorate bug links based on bug status

bparees:
also properly filter CLOSED (not just VERIFIED) bugs from consideration
for test failure causality


jluhrsen:
cool. thanks for this @bparees

PR #183, opened 2021-01-05
Add guard for bug.Component to prevent panic when loading /detailed endpoint

fabianvf:
Currently when viewing the /detailed endpoint sippy will panic and the content will not be served


jluhrsen:
I know I have no powers here, but /lgtm

PR #184, opened 2021-01-06
move tooltip text to an explicit icon

bparees:


pkg/html/installhtml/install_by_operators.go
```diff
@@ -51,7 +51,10 @@ func summaryInstallRelatedTests(curr, prev sippyprocessingv1.TestReport, numDays
 	s := fmt.Sprintf(`
 	<table class="table">
 		<tr>
-			<th colspan=5 class="text-center"><a class="text-dark" title="Install related tests, sorted by passing rate.  The link will prepopulate a BZ template to be filled out and submitted to report a bug against the test." id="InstallRelatedTests" href="#InstallRelatedTests">Install Related Tests</a></th>
+			<th colspan=5 class="text-center">
+				<a class="text-dark" id="InstallRelatedTests" href="#InstallRelatedTests">Install Related Tests</a>
+				<i class="fa fa-info-circle" title="Install related tests, sorted by passing rate.  The link will prepopulate a BZ template to be filled out and submitted to report a bug against the test." id="UpgradeJobs" href="#UpgradeJobs">Upgrade Jobs</i>
```
deads2k, line 56: is this supposed to be install or upgrade?

pkg/html/installhtml/upgrade_by_operators.go
```diff
@@ -79,7 +82,10 @@ func summaryUpgradeRelatedJobs(report, reportPrev sippyprocessingv1.TestReport,
 	s := fmt.Sprintf(`
 	<table class="table">
 		<tr>
-			<th colspan=4 class="text-center"><a class="text-dark" title="Passing rate for each job upgrade definition, sorted by passing percentage.  Jobs at the top of this list are unreliable or represent environments where the product is not stable and should be investigated.  The pass rate in parenthesis is the pass rate for jobs that started to run the installer and got at least the bootstrap kube-apiserver up and running." id="UpgradeJobs" href="#UpgradeJobs">Upgrade Jobs</a></th>
+			<th colspan=4 class="text-center">
+				<a class="text-dark" id="UpgradeJobs" href="#UpgradeJobs">Upgrade Jobs</a>
+				<i class="fa fa-info-circle" title="Passing rate for each job upgrade definition, sorted by passing percentage.  Jobs at the top of this list are unreliable or represent environments where the product is not stable and should be investigated.  The pass rate in parenthesis is the pass rate for jobs that started to run the installer and got at least the bootstrap kube-apiserver up and running." id="UpgradeJobs" href="#UpgradeJobs">Upgrade Jobs</i>
```
deads2k, line 87: I don't really know html, is this allowed to have the same ID as the `<a>` block above it?

bparees: oh, this should not have had an id field at all, good catch.

pkg/html/installhtml/install_by_operators.go
bparees: it's supposed to be neither. will remove

PR #191, opened 2021-01-12
Set missing setup job threshold to 95% of jobs

bparees:


deads2k:
/lgtm

This seems ok, it will still catch these eventually.

PR #192, opened 2021-01-12
add kube conformance variant

deads2k:


pkg/testgridanalysis/testidentification/kube_variants.go
```diff
@@ -9,14 +9,16 @@ import (
 
 var (
 	// variant regexes
-	kubeKindRegex    = regexp.MustCompile(`kind-`)
-	kubeKubeadmRegex = regexp.MustCompile(`kubeadm-`)
-	kubeSerialRegex  = regexp.MustCompile(`-serial`)
-	kubeWindowsRegex = regexp.MustCompile(`windows-`)
-	kubeUpgradeRegex = regexp.MustCompile(`upgrade-`)
-	kubeE2eRegex     = regexp.MustCompile(`-parallel`)
+	kubeConformanceRegex = regexp.MustCompile(`onformance-`)
```
deads2k, line 12: they have cap C and lower c and I dont' speak regex, but this is uncommon.

bparees: `[Cc]onformance-` should do it.

or you might at least want a comment on this so some "helpful" person doesn't add a C in the future.

deads2k: Updated.

PR #196, opened 2021-01-14
updates to remove grid=old argument

bparees:
I think the shell script is an improvement in a few ways:

1) don't have to recompile it to make changes to it (or use it)
2) it also covers recreating the symlinks

it does lose the "dry-run" functionality, but the whole thing is a git repo.  Dry run is "git reset --hard HEAD"


historical-data/rename.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/sh
+
+# remove the symlinks first, we'll recreate them later
+rm common/*
+
+# find and rename all the historical files
+find . -name *testgrid* -exec rename  -- "-grid=old&show-stale" "-show-stale" {} \;
```
deads2k, line 7: I don't speak bash, but I'm guessing this is some kind of a replacement thing

deads2k:
tbh, I'd choose basically anything over bash, but it seems to have worked.

/lgtm

historical-data/rename.sh
bparees: "find all files in the current directory tree that contain the string `testgrid`, for each file(`{}`) execute the `rename` command, replacing $1 with $2 in the filename".  The `--` is so that `-grid` doesn't get treated as an argument to rename.

PR #197, opened 2021-01-19
Include testsuite prefix in upgrade test names

fabianvf:
This will allow us to report upgrade pass rates properly again. I also added a guard so that if no jobs report starting upgrade the upgrade page doesn't die

### DIFF
--- a/pkg/testgridanalysis/testidentification/test_identification.go
+++ b/pkg/testgridanalysis/testidentification/test_identification.go
@@ -1,6 +1,7 @@
 package testidentification
 
 import (
+	"regexp"
 	"strings"
 
 	"github.com/openshift/sippy/pkg/testgridanalysis/testgridanalysisapi"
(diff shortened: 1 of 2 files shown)


pkg/testgridanalysis/testidentification/test_identification.go
```diff
@@ -74,9 +74,9 @@ var curatedTestSubstrings = map[string][]string{
 }
 
 var (
-	cvoAcknowledgesUpgrade = "[sig-cluster-lifecycle] Cluster version operator acknowledges upgrade"
-	operatorsUpgraded      = "[sig-cluster-lifecycle] Cluster completes upgrade"
-	machineConfigsUpgraded = "[sig-mco] Machine config pools complete upgrade"
+	cvoAcknowledgesUpgrade = "Cluster upgrade.[sig-cluster-lifecycle] Cluster version operator acknowledges upgrade"
+	operatorsUpgraded      = "Cluster upgrade.[sig-cluster-lifecycle] Cluster completes upgrade"
+	machineConfigsUpgraded = "Cluster upgrade.[sig-mco] Machine config pools complete upgrade"
```
bparees, line 79: this is going to cause some trouble w/ doing reporting on historical files which now won't match this.

Can you instead make the consumers of these constants do a regex or substring comparison instead of an exact equals check?  If you make it a regex we can include a $ terminator to ensure it's still a pretty precise match that's unlikely to get false positives against something else.

fabianvf: :+1:

PR #198, opened 2021-01-20
Remove suite from test name when loading testgrid data

fabianvf:
This makes it so that when searching for bugs/in search.svc.ci results will actually be found. The approach is kind of overkill but I thought the regex test name parser could be generally useful in the future and is not too bad in terms of maintenance burden.


fabianvf:
/hold
this may have broken some things

fabianvf:
/hold cancel

now I think the only issue is that some tests are of the form `[a-z]*\..+` (like `job.initialize`), and we can't really differentiate between that format and a test that's actually namespaced by suite

bparees:
@fabiand lgtm, thanks for seeing it through.  Squash and i'll merge

fabianvf:
@bparees I've gotten lazy with my commits since so many repos do the auto-squash now, done

bparees:
>  @bparees I've gotten lazy with my commits since so many repos do the auto-squash now, done

i know, i know..we could probably set the bot to squash on merge for this repo too.  Personally manually squashing commits scratches some weird itch for me, though :)  and sometimes we do want to retain the separate commits, so having the bot always do it is limiting.

/lgtm

PR #204, opened 2021-02-06
Add favicon

dmage:


dmage:
@bparees this favicon is a simple letter `s` on a red background, let me know if you envision it differently. Once we migrate to 1.16, we can use `embed` for static files.

dmage:
Or we can pick some emoji-favicon, like this https://favicon.io/emoji-favicons/robot/ or https://favicon.io/emoji-favicons/police-car-light/ :)

dmage:
Updated using official logo

Dockerfile
```diff
@@ -1,11 +1,7 @@
...
-#    dnf clean all -y
-RUN mkdir /tmp/src
-COPY . /tmp/src
-RUN cd /tmp/src; make build
-ENTRYPOINT /tmp/src/sippy
-
+WORKDIR /go/src/sippy
+COPY . .
+RUN make build
+ENTRYPOINT ["/go/src/sippy/sippy"]
```
bparees, line 5: fetchdata.sh and fetchdata-kube.sh will need to be updated for this.  and the deploymentconfig yamls

bparees (changes requested):
thanks! :)

Dockerfile
dmage: updated

PR #206, opened 2021-02-10
Add new string for knowing if install worked

jluhrsen:
ovirt and metal-assisted-ipv6 jobs have a slightly different name
for their setup containers:

- e2e-ovirt-ipi-install-ovirt-setup container test
- e2e-metal-assisted-ipv6-baremetalds-packet-setup container test

they were not being recognized and as such, sippy was complaining
with a warning that they were missing a setup container. This
fixes that.


fabianvf (approved):

pkg/testgridanalysis/testidentification/test_identification.go
```diff
@@ -41,7 +41,7 @@ func IsSetupContainerEquivalent(testName string) bool {
 		}
 	}
 
-	if strings.HasSuffix(testName, "container setup") {
+	if strings.HasSuffix(testName, "container setup") || strings.HasSuffix(testName, "setup container test") {
```
bparees, line 44: you can put this in the customJobSetupContainers set a few lines above this.  

The only reason not to do that is if you think other tests will contain this string and you don't want false positives (since the Set is compared on "test name contains" whereas this logic is "test name ends with").

that doesn't seem likely with `setup container test` as the string being looked for.

jluhrsen: that was my first thought, but when I realized both jobs had the same suffix I thought it made more sense to
add here. I'll do it whichever way you think is best though. just let me know.

bparees: the strings in the set don't necessarily only apply to a single job.  it's really more just an alternate way to identify tests that indicate install success at this point.  so putting it there is fine.

jluhrsen: sounds good. I just force pushed it like you suggested. I'm now wondering if we just want to throw that whole
line 46 ("container setup") up in that block of strings too? anyway...

bparees: yeah, i debated, but again there is a slight difference in semantics ("suffix" vs "contains") and given that "container setup" is a pretty generic string, i feel better if it's only matched as a suffix.  `setup container test` seemed specific enough that it didn't need further constraints.

bparees: (honestly my main concern was that copycats add more `||` logic here in the future instead of using the Set where appropriate, so i'd rather avoid even introducing a single `||`, lest it become a slippery slope)

pkg/testgridanalysis/testidentification/test_identification.go
```diff
@@ -17,12 +17,14 @@ var customJobSetupContainers = sets.NewString(
...
 	"e2e-gcp-upgrade-ipi-install-install-stableinitial",
 	"e2e-metal-assisted-baremetalds-assisted-setup",
+	"e2e-metal-assisted-ipv6-baremetalds-packet-setup container test",
 	"e2e-metal-assisted-onprem-baremetalds-assisted-setup",
 	"e2e-metal-ipi-baremetalds-devscripts-setup",
 	"e2e-metal-ipi-ovn-dualstack-baremetalds-devscripts-setup",
 	"e2e-metal-ipi-ovn-ipv6-baremetalds-devscripts-setup",
 	"e2e-metal-ipi-upgrade-baremetalds-packet-setup",
 	"e2e-metal-ipi-virtualmedia-baremetalds-devscripts-setup",
+	"e2e-ovirt-ipi-install-ovirt-setup container test",
```
bparees, line 27: i just meant you could add "setup container test" as the string here(which would cover both jobs).  but this is fine.

pkg/testgridanalysis/testidentification/test_identification.go
jluhrsen: totally understand.

PR #210, opened 2021-03-04
Update patterns for variants

dmage:
Jobs have been recently renamed, so we need to update variant patterns.

### DIFF
--- a/pkg/testgridanalysis/testidentification/ocp_variants.go
+++ b/pkg/testgridanalysis/testidentification/ocp_variants.go
@@ -10,27 +10,27 @@ import (
 
 var (
 	// variant regexes
-	awsRegex   = regexp.MustCompile(`(?i)-aws-`)
-	azureRegex = regexp.MustCompile(`(?i)-azure-`)
-	fipsRegex  = regexp.MustCompile(`(?i)-fips-`)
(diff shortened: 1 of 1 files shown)


pkg/testgridanalysis/testidentification/ocp_variants.go
```diff
@@ -27,10 +27,10 @@ var (
 	// proxy jobs do not have a trailing -version segment
 	proxyRegex   = regexp.MustCompile(`(?i)-proxy`)
 	promoteRegex = regexp.MustCompile(`(?i)^promote-`)
-	ppc64leRegex = regexp.MustCompile(`(?i)-ppc64le-`)
+	ppc64leRegex = regexp.MustCompile(`(?i)-ppc64le`)
 	rtRegex      = regexp.MustCompile(`(?i)-rt-`)
```
petr-muller, line 31: Can we just remove the trailing `-` everywhere? Renaming to the new form will happen very soon for pretty much all jobs, just this morning we merged a PR that migrated the realtime jobs for instance (https://github.com/openshift/release/pull/16434)

dmage: updated

petr-muller (approved):

PR #214, opened 2021-03-12
Add output for canary(aka highly reliable) tests

bparees:
new output produced here is consumed by https://github.com/openshift/origin/pull/25946


deads2k:
suggestion: create an endpoint that uses the current data.  Doing so would also allow you to handle future diffs like https://github.com/openshift/origin/pull/25946#issuecomment-799511302 points out.

Issue #128, opened 2020-10-08
Sippy misses existing bugs for test failures


bgilbert:
It appears Sippy sometimes fails to associate existing bugs with test failures.

## Example 1

In [this page](https://sippy.ci.openshift.org/?release=4.6#TopFailingTestsWithoutABug), Ctrl-F `kubelet terminates kube-apiserver gracefully`.  I used Sippy to file [this bug](https://bugzilla.redhat.com/show_bug.cgi?id=1886563), which was a duplicate of [this one](https://bugzilla.redhat.com/show_bug.cgi?id=1882750).  Sippy finds neither of them (and they don't show up when clicking through to the [search page](https://search.ci.openshift.org/?maxAge=168h&context=1&type=bug%2Bjunit&name=4.6&maxMatches=5&maxBytes=20971520&groupBy=job&search=%5C%5Bsig-api-machinery%5C%5D%5C%5BFeature%3AAPIServer%5C%5D%5C%5BLate%5C%5D+kubelet+terminates+kube-apiserver+gracefully) either).

Update: as of now, Sippy seems to have figured this out, but it hadn't figured it out half an hour ago when I wrote the above.  The search page still hasn't.

## Example 2

In [this page](https://sippy.ci.openshift.org/?release=4.6#TopFailingTestsWithoutABug), Ctrl-F `OpenShift APIs remain available`.  That failure [should](https://github.com/openshift/sippy/issues/23#issuecomment-655289547) link to [this bug](https://bugzilla.redhat.com/show_bug.cgi?id=1845411), but did not do so on the basis of [this comment](https://bugzilla.redhat.com/show_bug.cgi?id=1845411#c8) and continued not to do so when the test names were added to the Environment field.

bparees:
for a given sippy release page (e.g. 4.6), sippy only attempts to match test failures to bugs that are targeted at that release.

In your 2nd example, you linked to the 4.6 release page, but the bugs you think it should have associated are targeted to 4.7.

The intent here is that if the team that owns the bug hasn't targeted a clone to the release where sippy is seeing the failures, it doesn't count.

bgilbert:
The bug was reported against 4.6, so the test failure is known to the owning team, but Sippy is presenting it as though it's actionable.  What's the intended action for a build-watcher here?  File a duplicate bug?

bgilbert:
Reported the search issue as https://github.com/openshift/ci-search/issues/92.

bparees:
[...]

PR #215, opened 2021-03-12
    List bugs targeted to a different release than the one being reported on

bparees:
List test name matching bugs that are targeted to the same release as the report as linked.
List test name matching bugs that are not targeted to the same release as the report as associated

fixes https://github.com/openshift/sippy/issues/128

### DIFF
--- a/pkg/apis/sippy/v1/types.go
+++ b/pkg/apis/sippy/v1/types.go
@@ -54,6 +54,8 @@ type FailingTestBug struct {
 	Url       string              `json:"url"`
 	PassRates map[string]PassRate `json:"passRates"`
 	Bugs      []bugsv1.Bug        `json:"bugs,omitempty"`
+	// AssociatedBugs are bugs that match the test/job, but do not match the target release
+	AssociatedBugs []bugsv1.Bug `json:"associatedBugs,omitempty"`
 }
 
 // JobSummaryVariant describes a single variant and its associated jobs, their pass rates, and failing tests
--- a/pkg/buganalysis/cache.go
+++ b/pkg/buganalysis/cache.go
@@ -22,6 +22,8 @@ import (
 type BugCache interface {
 	ListJobBlockingBugs(job string) []bugsv1.Bug
 	ListBugs(release, variant, testName string) []bugsv1.Bug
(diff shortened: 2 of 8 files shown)


bparees:
sample output:

image


image

pkg/apis/sippy/v1/types.go
```diff
@@ -50,10 +50,11 @@ type MinimumPassRatesByComponent struct {
...
 type FailingTestBug struct {
-	Name      string              `json:"name"`
-	Url       string              `json:"url"`
-	PassRates map[string]PassRate `json:"passRates"`
-	Bugs      []bugsv1.Bug        `json:"bugs,omitempty"`
+	Name           string              `json:"name"`
+	Url            string              `json:"url"`
+	PassRates      map[string]PassRate `json:"passRates"`
+	Bugs           []bugsv1.Bug        `json:"bugs,omitempty"`
+	AssociatedBugs []bugsv1.Bug        `json:"associatedBugs,omitempty"`
```
deads2k, line 57: The distinction between `bugs` and `associatedBugs` isn't immediately obvious at an API level.  `nonCausalBugs`?

pkg/buganalysis/cache.go
```diff
@@ -21,7 +21,7 @@ import (
 // It is stateful though, so for a time after clearing the data will not be up to date until the Update is called
 type BugCache interface {
 	ListJobBlockingBugs(job string) []bugsv1.Bug
-	ListBugs(release, variant, testName string) []bugsv1.Bug
+	ListBugs(release, variant, testName string, invertReleaseQuery bool) []bugsv1.Bug
```
deads2k, line 24: Prefer a second method that can call a private helper with this bool.  That keeps code paths consistent and makes intent in the calling code very clear and searchable.

deads2k:
a couple minor comments.

pkg/apis/sippy/v1/types.go
bparees: i dunno that noncausalbugs is going to be any more obvious.  would you settle for me adding a comment that explains the distinction?

pkg/buganalysis/cache.go
bparees: updated

PR #220, opened 2021-03-30
Add dashboard for jobs statuses

dmage:
A new dashboard (similar to TestGrid) with all jobs.

This PR adds two new endpoints:

* `/api/jobs` that returns a JSON with statuses of all jobs
* `/jobs` with React app that shows data from `/api/jobs`

image

### DIFF
--- a/pkg/sippyserver/server.go
+++ b/pkg/sippyserver/server.go
@@ -12,6 +12,7 @@ import (
 	"github.com/openshift/sippy/pkg/buganalysis"
 	"github.com/openshift/sippy/pkg/html/releasehtml"
 	"github.com/openshift/sippy/pkg/testgridanalysis/testgridconversion"
+	"github.com/openshift/sippy/pkg/testgridanalysis/testgridhelpers"
 	"github.com/openshift/sippy/pkg/testgridanalysis/testidentification"
 	"k8s.io/klog"
 )
@@ -265,6 +266,36 @@ func (s *Server) detailed(w http.ResponseWriter, req *http.Request) {
 
 }
 
+func (s *Server) jobs(w http.ResponseWriter, req *http.Request) {
+	reportName := req.URL.Query().Get("release")
+	jobFilterString := req.URL.Query().Get("jobFilter")
+
+	var jobFilter *regexp.Regexp
+	if len(jobFilterString) > 0 {
+		var err error
+		jobFilter, err = regexp.Compile(jobFilterString)
+		if err != nil {
+			http.Error(w, fmt.Sprintf("jobFilter: %s", err), http.StatusBadRequest)
+			return
+		}
+	}
+
+	dashboardCoordinates, found := s.reportNameToDashboardCoordinates(reportName)
+	if !found {
+		http.Error(w, fmt.Sprintf("release %s not found", reportName), http.StatusBadRequest)
+		return
+	}
+
+	testGridJobDetails, lastUpdateTime := testgridhelpers.LoadTestGridDataFromDisk(s.testReportGeneratorConfig.TestGridLoadingConfig.LocalData, dashboardCoordinates.TestGridDashboardNames, jobFilter)
+
+	api.PrintJobsReport(w, s.syntheticTestManager, testGridJobDetails, lastUpdateTime)
+}
+
+func (s *Server) jobsReport(w http.ResponseWriter, req *http.Request) {
+	reportName := req.URL.Query().Get("release")
+	releasehtml.PrintJobsReport(w, reportName)
+}
+
 func (s *Server) Serve() {
 	http.DefaultServeMux.HandleFunc("/", s.printHtmlReport)
 	http.DefaultServeMux.HandleFunc("/install", s.printInstallHtmlReport)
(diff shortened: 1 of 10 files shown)


pkg/sippyserver/server.go
```diff
@@ -265,6 +269,132 @@ func (s *Server) detailed(w http.ResponseWriter, req *http.Request) {
...
+		var statuses []string
+		for i, result := range overall {
+			var status string
+			switch result {
+			case 4:
+				status = "R" // running
+			case 1:
+				status = "S" // success
+				if !hasOpenShiftTests(results, i) {
+					status = "P" // passed
```
bparees, line 362: any particular reason we care about this distinction?

bparees:
one question about the distinction between success+pass and "openshift tests", otherwise lgtm.

pkg/sippyserver/server.go
bparees: For (1) i guess i can appreciate that the "no-op" successes need to be ignored so they don't provide false indication that the job is generally passing when it reality it's mostly no-op'ing and then failing whenever it tries to do real work.  But if that's the case, once there is real work to do, you should start seeing a string of repeated failures, i think?

For (2) I would still expect them to be interesting to TRT because they still represent an indication about whether we are able to install/run the cluster on a platform we care about.  And more generally if there are jobs TRT does not care about I think i'd rather see TRT maintain an explicit list of "jobs we don't care about and why" and pass that in as some sort of jobexclusion filter to this endpoint.

dmage: (1) Sometimes it has nothing to do, then it fails to check that it has nothing to do, and then it's green again as it has nothing to do. Example: https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/promote-release-openshift-machine-os-content-e2e-aws-4.8-ppc64le/1374136409908056064

That's what I've found using this distinction. This knowledge isn't super useful, so I can ditch it.

(2) I guess I need incorporate bits that identify infra problems and installer failures. That should give us a signal from these tests and we'll be able to see more detailed picture about failures:

* infra
* installer
* upgrade
* openshift-e2e
* other tests (osd, certs, arcconformance, etc.)

pkg/sippyserver/server.go
```diff
@@ -265,6 +269,132 @@ func (s *Server) detailed(w http.ResponseWriter, req *http.Request) {
...
+	}
+
+	type jsonTest struct {
+		Name    string   `json:"name"`
+		Results []string `json:"results"`
+	}
+	type jsonJob struct {
+		Name        string     `json:"name"`
+		Timestamps  []int      `json:"timestamps"`
+		Tests       []jsonTest `json:"tests"`
```
bparees, line 334: this array doesn't align w/ the other arrays in this struct.

All the other arrays contain elements that represent "instance data for an instance of the job"

so it makes the api a bit awkward.

dmage: This is a leftover from the first iterations, I was thinking about returning "Overall" and some selected tests.
Updated.

pkg/sippyserver/server.go
dmage: image

Updated, now it looks like this.

pkg/testgridanalysis/testgridanalysisapi/types.go
```diff
@@ -66,9 +66,10 @@ type OperatorState struct {
 }
 
 const (
-	OperatorUpgradePrefix       = "Operator upgrade "
-	OperatorFinalHealthPrefix   = "operator conditions "
-	FinalOperatorHealthTestName = `[sig-sippy] tests should finish with healthy operators`
+	OperatorUpgradePrefix        = "Operator upgrade "
+	OldOperatorFinalHealthPrefix = "operator conditions "
+	OperatorFinalHealthPrefix    = "Operator results.operator conditions "
```
dmage, line 71: @bparees is there a better way to account this `Operator results.`?

bparees: @deads2k added those so i'll defer to him.  you're looking for a way to identify those tests?  Or you're looking for the best way to assess final operator health?

dmage: I'm looking for a way to distinguish infra failures (the installer wasn't able to do anything: https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-openshift-release-master-ci-4.8-upgrade-from-stable-4.7-e2e-ovirt-upgrade/1378693925845340160) and installation failures (some operators failed to proceed: https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-openshift-release-master-ci-4.8-upgrade-from-stable-4.7-e2e-ovirt-upgrade/1378693951728390144).

The signal isn't very reliable though, sometimes CI says that all operators are healthy when they are not: https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-openshift-release-master-ci-4.8-e2e-aws-serial/1379268056181641216

dmage: I guess I can try to re-use `openshiftSyntheticManager` and all those abstraction layers

dmage:
@bparees I rewrote the API part, now it's more aligned with other Sippy parts, ptal

PR #225, opened 2021-04-22
Print all releases as links from top of PrintHtmlReport

dcbw:
Also default the "detailed" view to the first release in the array (which presumably is sorted correctly).

@wking @smarterclayton


bparees:
can i get a screenshot of the new output?

dcbw:
@bparees http://people.redhat.com/dcbw/sippy.png

PR #228, opened 2021-06-09
remove extraneous whitespace in test name

bparees:
this was causing issues when we sent the search string to ci-search and got no results

### DIFF
--- a/pkg/testgridanalysis/testgridconversion/ocp_synthentic_tests.go
+++ b/pkg/testgridanalysis/testgridconversion/ocp_synthentic_tests.go
@@ -138,7 +138,7 @@ func (openshiftSyntheticManager) CreateSyntheticTests(rawJobResults testgridanal
 					syntheticTests[testgridanalysisapi.UpgradeTestName].pass = 1
 					// if the test succeeded, then the operator install tests should all be passes
 					for _, operatorState := range jrr.FinalOperatorStates {
-						testName := testgridanalysisapi.OperatorUpgradePrefix + " " + operatorState.Name
+						testName := testgridanalysisapi.OperatorUpgradePrefix + operatorState.Name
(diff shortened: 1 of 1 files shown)


bparees:
/assign @deads2k 

i'm a little fuzzy on this, since there appears to be a non-synthetic test w/ this same name.  I think the idea is we generate the synthetic tests when the job didn't complete and thus didn't output the non-synthetic operator tests?  

but then i'm not sure if want those having the same name since in one case they are the state of the operator post job run(?) and in this case they are....well, post job run, but for a failed install instead of a successful one.  Do we want to be able to distinguish them?

pkg/testgridanalysis/testgridconversion/ocp_synthentic_tests.go
```diff
@@ -138,7 +138,7 @@ func (openshiftSyntheticManager) CreateSyntheticTests(rawJobResults testgridanal
 					syntheticTests[testgridanalysisapi.UpgradeTestName].pass = 1
 					// if the test succeeded, then the operator install tests should all be passes
 					for _, operatorState := range jrr.FinalOperatorStates {
-						testName := testgridanalysisapi.OperatorUpgradePrefix + " " + operatorState.Name
+						testName := testgridanalysisapi.OperatorUpgradePrefix + operatorState.Name
```
bparees, line 141: the const already has a trailing space, so this wasn't needed.  I debated changing the const, but this seemed like it would have less unintended consequences.

deads2k:
> i'm a little fuzzy on this, since there appears to be a non-synthetic test w/ this same name. I think the idea is we generate the synthetic tests when the job didn't complete and thus didn't output the non-synthetic operator tests?

I think that's what I tried to do in order to count effectively.

/approve
/lgtm

PR #229, opened 2021-06-11
update prow url

cheesesashimi:
Sippy references the old Prow URL, which results in a dead link on some reports. Updates all references from https://prow.svc.ci.openshift.org to https://prow.ci.openshift.org.


bparees:
/lgtm

thanks!

PR #230, opened 2021-06-14
disable animation to improve rendering speed

deads2k:
This helped, but not enough.

### DIFF
--- a/pkg/html/releasehtml/html.go
+++ b/pkg/html/releasehtml/html.go
@@ -40,6 +40,12 @@ const (
 #table td, #table th {
 	border: 
 }
+
+.collapsing {
+    -webkit-transition: none;
+    transition: none;
+    display: none;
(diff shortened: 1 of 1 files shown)


bparees:
thanks, let's hold this and see how @dmage's change works out.  We can merge this on top of it if we still think we need it.

pkg/html/releasehtml/html.go
```diff
@@ -40,6 +40,12 @@ const (
 #table td, #table th {
 	border: 
 }
+
+.collapsing {
+    -webkit-transition: none;
+    transition: none;
+    display: none;
+}
```
bparees, line 48: if we do it, i'd just put it in the generichtml HTMLPageStart since i think then any page that does this would be covered.

deads2k:
no longer needed.

PR #232, opened 2021-06-14
Add v4.9 release to sippy deployment/data fetch

stbenjam:


scripts/fetchdata.sh
```diff
@@ -1,13 +1,13 @@
...
-# sleep before fetching so that if we're in some sort of fast crashloop/reschedule mode, 
+# sleep before fetching so that if we're in some sort of fast crashloop/reschedule mode,
 # we don't ping testgrid everytime we come back up
 echo "Doing initial sleep before fetching testgrid data"
 sleep 600 # 10 minutes
 while [ true ]; do
   echo "Fetching new testgrid data"
   rm -rf /data/*
-  /go/src/sippy/sippy -v 4 --fetch-data /data --release 3.11 --release 4.4 --release 4.5 --release 4.6 --release 4.7 --release 4.8
+  /go/src/sippy/sippy -v 4 --fetch-data /data --release 3.11 --release 4.4 --release 4.5 --release 4.6 --release 4.7 --release 4.8 --release 4.9
```
bparees, line 10: we're getting into memory pressure w/ all the data and i don't think anyone is looking at 4.4 + 4.5 data these days, let's go ahead and drop those (leave 3.11, it's a special case)

PR #235, opened 2021-06-18
fix inactive bug match for closed bugs

hasbro17:
Closed bugs in bugzilla usually have a two word status that includes the resolution
e.g CLOSED DUPLICATE. This commit adds more cases to accurately mark those bugs as inactive
and striked through on the sippy dashboard.

### DIFF
--- a/pkg/util/utils.go
+++ b/pkg/util/utils.go
@@ -89,7 +89,7 @@ func RelevantJob(jobName, status string, filter *regexp.Regexp) bool {
 
 func IsActiveBug(bug bugsv1.Bug) bool {
 	switch bug.Status {
-	case "VERIFIED", "RELEASE_PENDING", "CLOSED":
+	case "VERIFIED", "RELEASE_PENDING", "CLOSED", "CLOSED WONTFIX", "CLOSED DUPLICATE", "CLOSED NOTABUG":
 		return false
 	default:
 		return true


hasbro17:
There's probably other `CLOSED <RESOLUTION>` cases we might want to consider as inactive (all of them?).

pkg/util/utils.go
```diff
@@ -89,7 +89,7 @@ func RelevantJob(jobName, status string, filter *regexp.Regexp) bool {
 
 func IsActiveBug(bug bugsv1.Bug) bool {
 	switch bug.Status {
-	case "VERIFIED", "RELEASE_PENDING", "CLOSED":
+	case "VERIFIED", "RELEASE_PENDING", "CLOSED", "CLOSED WONTFIX", "CLOSED DUPLICATE", "CLOSED NOTABUG":
```
bparees, line 92: I don't think this is actually going to matter.

sippy gets its bug status information from ci-search, and ci-search mostly ignores closed bugs.  It only uses the "CLOSED" status to mark a bug that has disappeared from its queries as closed, until that bug is completely dropped from its query results:

https://github.com/openshift/ci-search/pull/98/files#diff-b1c84977729de3db2e886f505eb5b0624942a5b9a6177554782cca5269e0b728R171

So sippy should not be seeing bugs with any of these other statuses.

bparees: (if sippy is referencing a bug, but it's "closed wontfix" in bugzilla, check ci-search.  The most likely cause is that ci-search's data is out of sync w/ bugzilla).

bparees:
/hold

i don't think this is necessary nor is likely to fix the issue you're trying to fix

hasbro17:
Yeah I think it's likely a caching issue that's causing a CLOSED bug to not be striked-out. Closing.

PR #238, opened 2021-07-06
Add never-stable jobs to sippy for 4.9

stbenjam:
This adds the list of jobs that are currently below 40% passing rate,
the majority of which are near or at zero percent for at least the
last 14 days. Once jobs exceed 40% pass rate over the last 7 days, they
can "graduate" and be listed under their typical variant in sippy.

### DIFF
--- a/pkg/testgridanalysis/testidentification/ocp_variants.go
+++ b/pkg/testgridanalysis/testidentification/ocp_variants.go
@@ -62,13 +62,45 @@ var (
 		"single-node",
 	)
 
-	// openshiftJobsNeverStableForVariants is a list of jobs that have never been stable (not were stable and broke)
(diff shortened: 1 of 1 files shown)


bparees:
how are we tracking these jobs to ensure we do actually get them above 40%, or get rid of them?  I want to make sure this isn't just another "sweep it under the rug" situation

stbenjam:
> how are we tracking these jobs to ensure we do actually get them above 40%, or get rid of them? I want to make sure this isn't just another "sweep it under the rug" situation

Our idea is to incorporate this into one of the things that TRT is responsible for tracking and reporting on.  We'll report on never stable jobs during the mid-sprint and state of OpenShift meetings.

This is just a first pass but I think there's lots of room for improvement - we can start showing the date it was added, link to the BZ(s) associated with the job, etc.

PR #239, opened 2021-07-07
add compact variant

cheesesashimi:


pkg/testgridanalysis/testidentification/ocp_variants.go
```diff
@@ -141,6 +143,9 @@ func (v openshiftVariants) IdentifyVariants(jobName string) []string {
 	if azureRegex.MatchString(jobName) {
 		variants = append(variants, "azure")
 	}
+	if compactRegex.MatchString(jobName) {
+		variants = append(variants, "compact")
+	}
```
stbenjam, line 148: I *think* we want this below the promote variant and above aws around line 137.

The variants get considered in a sorted order, so I think AWS or azure compact jobs wouldn't be put in the compact bucket

cheesesashimi: It seems to bucket them correctly as-is:
Screen Shot 2021-07-07 at 9 05 26 AM

Though I am open to suggestion.

stbenjam: Oh ok, sorry - you're right. Promote returns early, which is why those jobs get excluded from other buckets.

bparees: this should be fine, jobs can go into multiple variant buckets, with a few exceptions (the cases where we return early: `never-stable` and `promote`)

PR #241, opened 2021-07-12
Return status 200 on main sippy page

wojtek-t:
This seem to be the problem we were facing:
- sippy was returning 404 on the main page
- as a result GCP ingress was marking the backends as unhealthy

@deads2k - assuming you agree with this change, will you be able to also build a new image (so that I can deploy it in k8s cluster)?

### DIFF
--- a/pkg/html/releasehtml/html.go
+++ b/pkg/html/releasehtml/html.go
@@ -364,7 +364,7 @@ type TestReports struct {
 
 func WriteLandingPage(w http.ResponseWriter, displayNames []string) {
 	w.Header().Set("Content-Type", "text/html;charset=UTF-8")
-	w.WriteHeader(http.StatusNotFound)
+	w.WriteHeader(http.StatusOK)
 	fmt.Fprintf(w, generichtml.HTMLPageStart, "Release CI Health Dashboard")
 	releaseLinks := make([]string, len(displayNames))
 	for i := range displayNames {


bparees:
@wojtek-t where are you pulling the image from currently?

wojtek-t:
> @wojtek-t where are you pulling the image from currently?

This is the image we're using now:
https://github.com/kubernetes/k8s.io/blob/main/apps/sippy/deployment.yaml#L21

bparees:
@wojtek-t i've pushed an updated image to that location.

wojtek-t:
> @wojtek-t i've pushed an updated image to that location.

You mean - again with "unstable" tag, right?

bparees:
@wojtek-t yes, i updated the unstable tag.

I've also changed the permissions on registry.ci.openshift.org/bparees/sippy:latest, you should be able to pull from there anonymously now.  That tag is updated any time a merge to master happens on this repo.  So if you always want the latest build, you can use that tag.

wojtek-t:
Hmm - it seems it's not starting now - I'm getting
```
81s         Warning   Failed                   pod/sippy-7f5d87d959-2whjl    Error: failed to create containerd task: OCI runtime create failed: container_linux.go:349: starting container process caused "exec: \"/tmp/src/sippy\": stat /tmp/src/sippy: no such file or directory": unknown
```

Any suggestions? Did you copy that too?

bparees:
Based on the dockerfile, the binary should be in "/go/src/sippy/sippy", which is what the entrypoint is set to in the image.  Are you overriding that in your deployment?


```
FROM golang:1.15.2
WORKDIR /go/src/sippy
COPY . .
RUN make build
ENTRYPOINT ["/go/src/sippy/sippy"]
EXPOSE 8080
```

bparees:
looks like we changed the location for the binary here:
https://github.com/openshift/sippy/commit/13bc3ac578b438f97fbf51886cc290d4cd609be3#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557

wojtek-t:
Yup - I will take care of it on our side. Thanks!

PR #243, opened 2021-07-20
adds tests for testgridconversion package

cheesesashimi:
To progress with https://issues.redhat.com/browse/TRT-28, I wanted to be confident that any changes I make to Sippy would not break existing functionality.

To that end, I've written a set of test cases for the `testgridconversion` package to establish a baseline of coverage. When run with `go test -cover`, it shows that 91.7% of the statements are covered by these tests. I admit quite a bit of naïveté with Sippy and TestGrid, so there are likely edge-cases I've missed or assertions that could be tightened up. So if anythings amiss or I completely misunderstood something, please let me know!


pkg/testgridanalysis/testgridconversion/to_raw_data_test.go
```diff
@@ -0,0 +1,574 @@
...
+
+// The TestGrid job name; hardcoded for simplicity.
+const jobName string = "periodic-ci-openshift-release-master-nightly-4.9-e2e-aws"
+
+// For simplicity, we assume a single test run on a single day.
+// Only the RunLength test uses a different value
+const numOfJobs int = 1
+
+// The testgrid test statuses we're concerned about
+var statuses []testgridv1.TestStatus = []testgridv1.TestStatus{
```
deads2k, line 20: only used in RunRawDataTestCase ?  If so, move locally there I think

pkg/testgridanalysis/testgridconversion/to_raw_data_test.go
```diff
@@ -0,0 +1,574 @@
...
+	rawData testgridanalysisapi.RawData
+	// Any warnings generated by the conversion process
+	warnings []string
+	// The test case used to run the code under test
+	testCase rawDataTestCase
+	// The testgrid status of the generated input data
+	testGridStatus testgridv1.TestStatus
+}
+
+func RunRawDataTestCase(t *testing.T, rawDataTestCase rawDataTestCase) {
```
deads2k, line 52: why cap-R?

cheesesashimi: No reason, I'll make it lower-R.

cheesesashimi:
Looks like my lint fixes got merged before this did, creating a merge conflict. I've addressed the conflict and rebased.

pkg/testgridanalysis/testgridconversion/to_raw_data_test.go
```diff
@@ -0,0 +1,574 @@
...
+			},
+			expectedTestNames: []string{
+				"not a synthetic test",
+			},
+			// We expect these warnings to be present because the "container setup"
+			// or "Overall" tests were not run.
+			expectedWarnings: map[testgridv1.TestStatus][]string{
+				testgridv1.TestStatusSuccess: missingSetupJobWarning,
+				testgridv1.TestStatusFailure: missingSetupJobWarning,
+				testgridv1.TestStatusFlake:   missingSetupJobWarning,
```
bparees, line 302: i'm confused about why this is expected to be emitted as successful, failing, and flaking, simultaneously for a given run?

pkg/testgridanalysis/testgridconversion/to_raw_data_test.go
```diff
@@ -0,0 +1,574 @@
...
+		passingChangelists []string
+		// List of changelists which should not appear in the output given the
+		// startday / numday configuration.
+		excludedChangelists []string
+		// Note: Changelist lists within the test-case are unordered and no order is assumed.
+	}{
+		{
+			name: "minus one both ends",
+			options: testgridconversion.ProcessingOptions{
+				StartDay: -1,
```
bparees, line 414: StartDay==-1 means "start at the first timestamp we find in the data", so I'd expect this test see the `123` changeset in the results, not sure why it doesn't?

cheesesashimi: I'm also confused. I wonder if I didn't set up the test fixture correctly.

pkg/testgridanalysis/testgridconversion/to_raw_data_test.go
cheesesashimi: It's not expecting that from a single run of `ProcessTestGridDataIntoRawJobResults`. The testcase calls `ProcessTestGridDataIntoRawJobResults` three times, once with each of those TestGrid statuses being set. This groups the expected warnings by the TestGrid status.

bparees: why doesn't the `Overall` testcase a little further down also have all 3 of these then?

on `Overall` sucess/flake we don't warn about missing setup, but if there's no `Overall` test we always warn about the missing setup?

cheesesashimi: I can explain why the test passes as-is, but I cannot say if the behavior the test verifies is what is desired. The why behind it is:

In the case where `Overall` is a success, `processTestToJobRunResults` flips the success bit and sets `SetupStatus`: https://github.com/openshift/sippy/blob/master/pkg/testgridanalysis/testgridconversion/to_raw_data.go#L144-L148.

In the case where `Overall` fails, `processTestToJobRunResults` flips the failure bit, but it does not set `SetupStatus`: https://github.com/openshift/sippy/blob/9f8f55fd255647d7f9aacf698e78654e3279ea6c/pkg/testgridanalysis/testgridconversion/to_raw_data.go#L189-L190.

When this code path is hit, the test name "Overall" falls through the subsequent regex checks (https://github.com/openshift/sippy/blob/master/pkg/testgridanalysis/testgridconversion/to_raw_data.go#L196-L206) since it doesn't match any of those cases. If there was another test name that matches one (or more) of those regex tests, `SetupStatus` would get set. For this particular test case, there are no other test names that would pass one of those regex checks, so `SetupStatus` never gets set. Adding another test name that would pass one of those regex checks causes `SetupStatus` to be set, and no warnings are emitted.

Maybe this test is a edge-case that would never actually occur? Should `Overall` also set `SetupStatus`?

bparees: ok, i think that behavior makes sense.

We can always assume there will be an Overall test (it's possible sippy should throw some other warning/error if it can't find an Overall test result in a run).  And when overall is successful, setting setupstatus to successful makes sense, as described in the comment in the code.

so your test case where there is not an `overall` test result is testing for behavior/situations that we should never encounter in practice.  Again it might be better to have sippy completely error out in such cases, because i'm not sure if there are other aspects of the reporting that will become confusing/meaningless if there was no Overall result in the run.

pkg/testgridanalysis/testgridconversion/to_raw_data_test.go
cheesesashimi: I have an answer for this now:

If any of the timestamps is `time.Now()`, and `startDay` is 0, those timestamps are filtered out, because `computeLookback` also calls `time.Now()` and the resulting timestamp(s) fail these (https://github.com/openshift/sippy/blob/master/pkg/testgridanalysis/testgridconversion/to_raw_data.go#L79-L84) checks (they're not less than, they're equal to). For the sake of simplicity, the test fixtures subtract a second from `time.Now()` so they're not unintentionally filtered out.

Even with this being the case, we should still see `123` in the results since that's the first changelist and is aligned with the first timestamp. If `startDay` is set to `0`, the start column computed by `computeLookback` is also 0, hence `123` is present. However, if it is set to -1 (or any other negative number, for that matter) the start column gets set to the absolute value of `startDay` (effectively `n * -1`). This occurs regardless of whether the timestamp we're currently looking at occurs between `startDay` and `startDay + numDays`. Because of this, the first element in the timestamps list is ignored (which in this case, is `123`).

This is another case where the test reflects the current behavior, but I'm not sure if the current behavior is the desired behavoir.

bparees: > However, if it is set to -1 (or any other negative number, for that matter) the start column gets set to the absolute value of startDay (effectively n * -1). 

the intended behavior is that if startDay is set to `-1`, then we effectively set startDay to align with(include) the latest timestamp in the dataset.  

Not sure where the absolute value bit is coming in, you can see that if startDay is -1, then this logic will end up with startTS==mostRecentTimestamp:
https://github.com/openshift/sippy/blob/9f8f55fd255647d7f9aacf698e78654e3279ea6c/pkg/testgridanalysis/testgridconversion/to_raw_data.go#L64-L73

I think the issue is that this logic:
https://github.com/openshift/sippy/blob/9f8f55fd255647d7f9aacf698e78654e3279ea6c/pkg/testgridanalysis/testgridconversion/to_raw_data.go#L79

is always going to exclude the actual column where the timestamp was the most recent timestamp(the one we found in the iteration above), because it's doing a less than check, not a less than or equal to check?

cheesesashimi: The absolute value bit was just the best way to describe what was observed (the code isn't explicitly doing that, but that is the outcome). I didn't mean to imply the code was doing that explicitly.

The less-than-check is why the test fixtures decrement `time.Now()` by 1 second, since I was writing the tests to reflect the current behavior. What confused me was expecting `123` to be present when I couldn't see a way that the code (as-is) could do that. Since the intended behavior is to include those values, I'll change the less-than to less-than-or-equal. Please accept my apologies for overanalyzing.

bparees: > Since the intended behavior is to include those values, I'll change the less-than to less-than-or-equal. 

thanks!  i'd say the original behavior was probably always bugged, it just didn't matter all that much if we missed one column, for the kind of reporting we do

> Please accept my apologies for overanalyzing.

no apologies necessary, it was a productive discussion.  And i'm happy to see some tests around some of this behavior... the code that extracts the "right" columns from testgrid based on the date range is definitely fiddly and subject to some off by one and other bounding issues (again, they don't matter that much in the grand scheme of what it needs to do, but tests that enforce the behavior are much appreciated)

cheesesashimi: Duly noted, thanks! What you just described was my primary motivation for doing this work, especially since I want to use it as a baseline for my step metrics work. Is there anything else we should address at this point?

bparees: i think this comment still needs resolution:
https://github.com/openshift/sippy/pull/243#discussion_r681154968

i think we should enforce the expectation that there is always an Overall test in the jobs we process, and then you can drop your testcase where there isn't one (or at least change the expected outcome of that test)

pkg/testgridanalysis/testgridconversion/to_raw_data_test.go
cheesesashimi: That makes sense. Since we process multiple jobs and their tests within this codepath, we should run a pre-check to ensure that each job has an "Overall" test before we process the individual tests. This way, we can skip jobs that don't have an "Overall" test and still process the ones that do without affecting our statistics. We'll emit one error for each one we find without an "Overall" test since it's possible we could have multiple jobs.

I started coding this up and thankfully, the newly-merged sippyserver smoke tests caught a few edge-cases I need to look into.

cheesesashimi: I've got an initial implementation of this pushed up for review. I was trying to decide what to do with the errors beyond just logging them to the console, so within the UI portion of the server, I emit them using the same path as the synthetic test and bug cache warnings.

cheesesashimi:
This has been rebased.

pkg/testgridanalysis/testgridconversion/to_raw_data.go
```diff
@@ -25,29 +25,58 @@ type SyntheticTestManager interface {
...
+	}
+
+	return false
 }
 
-func processJobDetails(rawJobResults testgridanalysisapi.RawData, job testgridv1.JobDetails, startCol, endCol int) {
+func processJobDetails(rawJobResults testgridanalysisapi.RawData, job testgridv1.JobDetails, startCol, endCol int) error {
+	// Do a first pass to make sure the job has an "Overall" test
+	// If not, return early so we don't taint the rest of the results
+	if !hasOverallTest(job) {
```
bparees, line 76: i'm curious, is this tripping on any of the jobs we're currently analyzing?  (i would expect/hope not).

just want to be sure before we turn it on and start throwing errors

PR #244, opened 2021-07-21
Add a new variants page

stbenjam:
This adds a new page to view all jobs for a specific variant. When
expanding the job details from the main page, the first few results
are shown. This adds a link to view all job results.

If you look at the production sippy, and expand the never-stable
variant, you'll see a "Plus N jobs" -- but there's no way to view them. This
links that message to a new page at `/variants?release=4.9&variant=never-stable`,
which lists all the jobs for that variant.

### DIFF
--- a/pkg/html/generichtml/errors.go
+++ b/pkg/html/generichtml/errors.go
@@ -1,0 +1,35 @@
+package generichtml
+
+import (
+	"html/template"
+	"k8s.io/klog"
+	"net/http"
+)
+
+var messageTemplate = template.Must(template.New("jobs").Parse(`<!DOCTYPE html>
+<html>
+<head>
+  <meta charset="utf-8">
+  <title>{{.StatusCode}} {{.StatusMessage}}</title>
+</head>
+<body>
+  <h1>{{.StatusCode}} {{.StatusMessage}}</h1>
+  <hr>
+  <p>{{.UserMessage}}</p>
+</body>
+</html>
+`))
+
+func PrintStatusMessage(w http.ResponseWriter, code int, message string) {
+	w.WriteHeader(code)
+	w.Header().Set("Content-Type", "text/html;charset=UTF-8")
+
+	err := messageTemplate.Execute(w, map[string]interface{}{
+		"StatusCode":    code,
+		"StatusMessage": http.StatusText(code),
+		"UserMessage":   message,
+	})
+	if err != nil {
+		klog.Error(err)
+	}
+}
--- a/pkg/html/generichtml/errors.go
+++ b/pkg/html/generichtml/errors.go
@@ -20,6 +20,7 @@ var messageTemplate = template.Must(template.New("jobs").Parse(`<!DOCTYPE html>
 </html>
 `))
 
+// PrintStatusMessage is used to display a proper error page to an end user.
 func PrintStatusMessage(w http.ResponseWriter, code int, message string) {
 	w.WriteHeader(code)
 	w.Header().Set("Content-Type", "text/html;charset=UTF-8")
--- a/pkg/html/generichtml/jobaggregationresult.go
+++ b/pkg/html/generichtml/jobaggregationresult.go
@@ -67,7 +67,6 @@ type jobAggregationResultRenderBuilder struct {
 	maxJobResultsToShow  int
 	colors               ColorizationCriteria
 	collapsedAs          string
-	groupby              string
 }
 
 func NewJobAggregationResultRenderer(sectionBlock string, currJobResult jobAggregationDisplay, release string) *jobAggregationResultRenderBuilder {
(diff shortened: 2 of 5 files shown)


pkg/html/generichtml/errors.go
```diff
@@ -0,0 +1,36 @@
...
+	w.WriteHeader(http.StatusNotFound)
+	w.Header().Set("Content-Type", "text/html;charset=UTF-8")
+
+	err := messageTemplate.Execute(w, map[string]interface{}{
+		"StatusCode":    code,
+		"StatusMessage": http.StatusText(code),
+		"UserMessage":   message,
+	})
+	if err != nil {
+		log.Print(err)
```
cheesesashimi, line 34: I noticed that Sippy uses [klog](https://github.com/kubernetes/klog) for logging, so this should probably be `klog.Errorf(err.Error())` (or similar).

pkg/sippyserver/server.go
```diff
@@ -88,6 +89,14 @@ func (s *Server) RefreshData() {
 	klog.Infof("Refresh complete")
 }
 
+func (s *Server) defaultHandler(w http.ResponseWriter, req *http.Request) {
+	if req.URL.Path == "/" {
+		s.printHtmlReport(w, req)
+	} else {
+		generichtml.PrintStatusMessage(w, 404, "Page not found.")
```
cheesesashimi, line 96: Instead of `404`, refer to the constant from the `http` package, `http.StatusNotFound`.

stbenjam: Oops, fixed.

pkg/html/generichtml/errors.go
stbenjam: Ah thanks!

pkg/html/generichtml/jobaggregationresult.go
```diff
@@ -137,95 +138,110 @@ func (b *jobAggregationResultRenderBuilder) StartCollapsedAs(collapsedAs string)
...
-			class,
-			b.currAggregationResult.displayName,
-			button,
-			b.currAggregationResult.displayPercentage,
-			b.currAggregationResult.parenDisplayPercentage,
-			b.currAggregationResult.totalJobRuns,
-		)
+	// TODO either make this a template or make this a builder that takes args and then has branches.
+	// that will fix the funny link that goes nowhere.
+	if b.groupby == "variant" {
```
deads2k, line 166: When we aren't displaying a variant, what does this show?  Are you using this just to create a table of the jobs?

pkg/html/releasehtml/variants.go
```diff
@@ -0,0 +1,56 @@
...
+	Variant string
+	Results template.HTML
+}
+
+
+// PrintVariantsReport shows an aggregated listing of all jobs for a particular variant.
+func PrintVariantsReport(w http.ResponseWriter, release, variant string, currentWeek, previousWeek v1.VariantResults, timestamp time.Time) {
+	w.Header().Set("Content-Type", "text/html;charset=UTF-8")
+
+	results := template.HTML(generichtml.NewJobAggregationResultRendererFromVariantResults("by-job", currentWeek, release).
```
deads2k, line 43: Why not filter out the jobs you want to display and use the job rendering code for the job tables we have?

stbenjam: >  job rendering code for the job tables we have

Can you point me to this?

pkg/html/generichtml/jobaggregationresult.go
stbenjam: Job aggregation result is what the main page uses, I wanted to reuse that same code. This if hides the variant header. Based on your comments, I guess I'm missing something.

pkg/html/releasehtml/variants.go
stbenjam: Oh, do you mean this? https://github.com/openshift/sippy/blob/master/pkg/html/generichtml/jobresult.go#L89

The aggregator already handles everything I want to do, I was trying to avoid duplicating that logic.

stbenjam:
@deads2k Thanks for the comments, can you have another look when you get a chance?

deads2k:
/lgtm
/hold

holding in case you want more rows.  If you don't, feel free to release the hold at will.

pkg/html/releasehtml/variants.go
```diff
@@ -0,0 +1,57 @@
...
+
+// PrintVariantsReport shows an aggregated listing of all jobs for a particular variant.
+func PrintVariantsReport(w http.ResponseWriter, release, variant string, currentWeek, previousWeek *v1.VariantResults, timestamp time.Time) {
+	w.Header().Set("Content-Type", "text/html;charset=UTF-8")
+
+	var s string
+	for _, currJobResult := range currentWeek.JobResults {
+		prevJobResult := util.FindJobResultForJobName(currJobResult.Name, previousWeek.JobResults)
+		jobHTML := generichtml.NewJobResultRendererFromJobResult("by-variant", currJobResult, release).
+			WithMaxTestResultsToShow(10).
```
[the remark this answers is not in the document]
stbenjam: At the moment I'm just interested in seeing the whole never-stable list so I can graduate the ones at the bottom.

stbenjam:
> holding in case you want more rows. If you don't, feel free to release the hold at will.

In a future iteration, I intend to link the "Plus N more test results" to dedicated test result page like I did for variants, so the amount of tests shown don't matter much.

PR #248, opened 2021-07-21
adds golangci-lint and address lint issues

cheesesashimi:
In the interest of improving Sippy's code quality, I ran `golangci-lint` against this repo using a fairly conservative configuration I cribbed from my teams' repository. I fixed what lint issues I felt I could reasonably fix, however there are some other ones that would require refactoring to address. For those, I've added `//nolint` comments where necessary along with a brief description.

I spread this work across multiple commits for easier context and review, but I'm happy to squash everything into a single commit for easier merging.


stbenjam:
Nice!  I tested locally and AFAICT everything is still working. We should try to land this soon otherwise you'll keep having to rebase

/lgtm
/assign @bparees @deads2k

cheesesashimi:
I've enabled some additional linters and made the appropriate changes, so this may need another look.

pkg/api/api.go
```diff
@@ -49,12 +49,12 @@ func summaryJobsByVariant(report, reportPrev sippyprocessingv1.TestReport) []sip
 			jobSummaryVariant = sippyv1.JobSummaryVariant{
 				Variant: v.VariantName,
 				PassRates: map[string]sippyv1.PassRate{
-					"latest": sippyv1.PassRate{
+					"latest": {
```
bparees, line 52: imho this change reduces readability.  again not against it per se, but i am against it being enforced by the linter.

pkg/html/releasehtml/html.go
```diff
@@ -214,7 +210,7 @@ func canaryTestFailures(all, prevAll []sippyprocessingv1.FailingTestResult) stri
 			<th>Test Name</th><th>Pass Rate</th>
 		</tr>
 	`
-	template := `
+	tmpl := `
```
bparees, line 213: what's the motivation for this change?

pkg/testgridanalysis/testgridconversion/to_raw_data.go
```diff
@@ -29,14 +31,19 @@ type ProcessingOptions struct {
 	NumDays              int
 }
 
+type lookback struct {
+	startCol int
+	endCol   int
+}
```
bparees, line 37: was this driven by the linter?  I agree there's a point at which combining multiple args into a struct makes sense.  i'm not positive that "2" is the point at which it becomes important though.

is it specifically tied to the use as a return value to avoid multiple return values?

pkg/testgridanalysis/testgridconversion/to_raw_data.go
```diff
@@ -227,45 +236,50 @@ func processTestToJobRunResults(jobResult testgridanalysisapi.RawJobResult, job
...
+		result.Flakes += testResult.Flakes
+	} else {
+		result = testResult
 	}
-	result.Name = testName
-	result.Successes += passed
-	result.Failures += failed
-	result.Flakes += flaked
 
-	testResults[testName] = result
```
bparees, line 270: i'm struggling to see this as a strict improvement.

previously the code always set the result.* fields based on the testResult.* fields.

now there's an if/else that effectively does the same thing, but in two different ways depending on whether you had an existing value or not.

bparees (changes requested):
I appreciate the huge amount of work that I assume went into this and i think 99% of the changes are positive, so thank you for that.  

that said, i did comment on a couple spots where i think maybe we should dial back the aggressiveness of the linter.

bparees:
/approve
because i do think this is a positive overall, just would like to see some tweaks in a few spots.

main.go
```diff
@@ -129,9 +129,11 @@ func (o *Options) ToTestGridDashboardCoordinates() []sippyserver.TestGridDashboa
...
 func dashboardArgFromOpenshiftRelease(release string) string {
 	const openshiftDashboardTemplate = "redhat-openshift-ocp-release-%s-%s"
-	dashboards := []string{}
-	dashboards = append(dashboards, fmt.Sprintf(openshiftDashboardTemplate, release, "blocking"))
-	dashboards = append(dashboards, fmt.Sprintf(openshiftDashboardTemplate, release, "informing"))
+
+	dashboards := []string{
+		fmt.Sprintf(openshiftDashboardTemplate, release, "blocking"),
+		fmt.Sprintf(openshiftDashboardTemplate, release, "informing"),
+	}
```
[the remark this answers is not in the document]
cheesesashimi: Originally, the linter wanted me to nest the two `append` calls, probably like this:

```golang
dashboards := []string{}
dashboards = append(append(dashboards, fmt.Sprintf(openshiftDashboardTemplate, release, "blocking")), fmt.Sprintf(openshiftDashboardTemplate, release, "informing"))
```

I didn't think that was very readable, so I opted for what you see instead.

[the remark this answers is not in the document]
bparees: can we just turn off that rule entirely, then?

pkg/api/api.go
cheesesashimi: This was from the linter expecting files to be ran through `gofmt` with the `-s` flag. I agree this particular change isn't great for readability, so I'll dial that one back.

pkg/html/releasehtml/html.go
cheesesashimi: This was from the linter catching an import shadow. That package imports `template`, but you're also assigning a value to the symbol `template`. Later, if you try to access anything in the `template` module from within that scope, you won't be able to. This one bit me a couple of times over the years.

pkg/testgridanalysis/testgridconversion/to_raw_data.go
cheesesashimi: This one was entirely on me. After sleeping on it, I agree.

main.go
```diff
@@ -163,7 +165,7 @@ func (o *Options) Validate() error {
 }
 
 func (o *Options) Run() error {
-	if len(o.FetchData) != 0 {
+	if o.FetchData != "" {
```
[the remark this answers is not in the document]
cheesesashimi: Yeah, I agree. I'll switch this one off.

pkg/html/releasehtml/html.go
bparees: ah hah, makes sense, thanks

pkg/testgridanalysis/testgridconversion/to_raw_data.go
cheesesashimi: Yes it was. I agree with your point. It identified this and `utils.ComputeFailureGroupStats` as potential hotspots. In the latter case, I think fixing it makes sense. I'll revert the change for the `lookback` struct and disable that particular linter.

main.go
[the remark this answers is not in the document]
cheesesashimi: Yeah, I'll switch that one off. This was the only spot that particular rule identified.

pkg/api/api.go
cheesesashimi: The linter will now just make sure the file is ran through `gofmt`, with no enforcement that the `-s` flag is used.

bparees: sgtm

cheesesashimi:
@bparees Thanks for your review! I pushed up the changes and though not specifically called out by the linter, I fixed the spelling of the word "synthetic" as it applies to the `SyntheticTestManager` interface.

pkg/testgridanalysis/testgridconversion/ocp_synthentic_tests.go
```diff
@@ -37,11 +37,11 @@ func (openshiftSyntheticManager) CreateSyntheticTests(rawJobResults testgridanal
...
-				testgridanalysisapi.InstallTestName:             {name: testgridanalysisapi.InstallTestName},
-				testgridanalysisapi.InstallTimeoutTestName:      {name: testgridanalysisapi.InstallTestName},
-				testgridanalysisapi.InfrastructureTestName:      {name: testgridanalysisapi.InfrastructureTestName},
-				testgridanalysisapi.FinalOperatorHealthTestName: {name: testgridanalysisapi.FinalOperatorHealthTestName},
-				testgridanalysisapi.OpenShiftTestsName:          {name: testgridanalysisapi.OpenShiftTestsName},
+				testgridanalysisapi.InstallTestName:             &synthenticTestResult{name: testgridanalysisapi.InstallTestName},
+				testgridanalysisapi.InstallTimeoutTestName:      &synthenticTestResult{name: testgridanalysisapi.InstallTestName},
+				testgridanalysisapi.InfrastructureTestName:      &synthenticTestResult{name: testgridanalysisapi.InfrastructureTestName},
+				testgridanalysisapi.FinalOperatorHealthTestName: &synthenticTestResult{name: testgridanalysisapi.FinalOperatorHealthTestName},
+				testgridanalysisapi.OpenShiftTestsName:          &synthenticTestResult{name: testgridanalysisapi.OpenShiftTestsName},
```
bparees: looks like synthetic is still typoed here.

cheesesashimi: Good catch, thanks!

cheesesashimi:
@bparees After fixing the merge conflict, I think this is ready for a final review. If desired, I can squash all my commits before merging.

PR #250, opened 2021-07-28
update common symlinks for historical data

bparees:


bparees:
@dmage fyi i forget to update the symlinks that the historical reporting uses.

bparees:
(running the "sippy/historical-data/rename.sh" script generates these)

PR #252, opened 2021-07-29
adds smoke tests for sippyserver

cheesesashimi:
To get better test coverage, I added some smoke tests to the `sippyserver` module. Right now, it only checks for the HTTP status returned from the HTTP request given certain args. In the future, it could also examine the returned payload. `go test -cover` indicates this covers 93.2% of code in the `sippyserver` module.

### DIFF
--- a/pkg/sippyserver/server.go
+++ b/pkg/sippyserver/server.go
@@ -14,7 +14,6 @@ import (
 	"github.com/openshift/sippy/pkg/buganalysis"
 	"github.com/openshift/sippy/pkg/html/releasehtml"
 	"github.com/openshift/sippy/pkg/testgridanalysis/testgridconversion"
-	"github.com/openshift/sippy/pkg/testgridanalysis/testgridhelpers"
 	"github.com/openshift/sippy/pkg/testgridanalysis/testidentification"
 	"k8s.io/klog"
 )
@@ -57,6 +56,7 @@ type Server struct {
 	bugCache                  buganalysis.BugCache
 	testReportGeneratorConfig TestReportGeneratorConfig
 	currTestReports           map[string]StandardReport
+	httpServer                *http.Server
 }
 
 type TestGridDashboardCoordinates struct {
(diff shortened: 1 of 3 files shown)


pkg/sippyserver/server.go
```diff
@@ -363,22 +358,36 @@ func (s *Server) variantsReport(w http.ResponseWriter, req *http.Request) {
...
+	serveMux.HandleFunc("/operator-health", s.printOperatorHealthHTMLReport)
+	serveMux.HandleFunc("/testdetails", s.printTestDetailHTMLReport)
+	serveMux.HandleFunc("/json", s.printJSONReport)
+	serveMux.HandleFunc("/detailed", s.detailed)
+	serveMux.HandleFunc("/refresh", s.refresh)
+	serveMux.HandleFunc("/canary", s.printCanaryReport)
+	serveMux.HandleFunc("/api/jobs", s.jobs)
+	serveMux.HandleFunc("/jobs", s.jobsReport)
+	serveMux.HandleFunc("/variants", s.variantsReport)
+	serveMux.Handle("/static/", http.StripPrefix("/static/", http.FileServer(http.Dir("./static"))))
```
bparees, line 376: don't you still need to add the defaulthandler?  (for 404s if nothing else)

cheesesashimi: You mean https://github.com/openshift/sippy/pull/252/files#diff-768a5e234f25507b29e6156fd5af2bfb561aef7c7b19612e7e73ea4cb6be07e4L93-L99? It doesn't appear to be wired into anything.

Instead, it looks like `server.printHTMLReport` is what's wired up to the root handler (https://github.com/openshift/sippy/pull/252/files#diff-768a5e234f25507b29e6156fd5af2bfb561aef7c7b19612e7e73ea4cb6be07e4L93-L99) and what handles 404s (which it does by silently redirecting to the landing page) (https://github.com/openshift/sippy/blob/master/pkg/sippyserver/server.go#L101-L107).

bparees: hm, there's some weirdness here but yeah i guess that other code is dead.

PR #259, opened 2021-08-18
Reconsider use of never-stable, add techpreview variant

stbenjam:
- I haven't found being more liberal with adding jobs to the never-stable
list to be particularly helpful, especially with so many failing. I
think each should be investigated, and a bug filed -- and then added to
the never-stable list.

- Exclude never-stable jobs from home page widgets

- Adds a new techpreview variant, for TPNU jobs (being added in https://github.com/openshift/release/pull/19845)

### DIFF
--- a/pkg/testgridanalysis/testidentification/ocp_variants.go
+++ b/pkg/testgridanalysis/testidentification/ocp_variants.go
@@ -32,6 +32,7 @@ var (
 	rtRegex      = regexp.MustCompile(`(?i)-rt`)
 	s390xRegex   = regexp.MustCompile(`(?i)-s390x`)
 	serialRegex  = regexp.MustCompile(`(?i)-serial`)
+	techpreview  = regexp.MustCompile(`(?i)-techpreview`)
 	upgradeRegex = regexp.MustCompile(`(?i)-upgrade`)
 	// some vsphere jobs do not have a trailing -version segment
 	vsphereRegex    = regexp.MustCompile(`(?i)-vsphere`)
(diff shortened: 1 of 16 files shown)


pkg/testgridanalysis/testidentification/ocp_variants.go
```diff
@@ -12,6 +12,7 @@ var (
 	// variant regexes
 	awsRegex     = regexp.MustCompile(`(?i)-aws`)
 	azureRegex   = regexp.MustCompile(`(?i)-azure`)
+	canaryRegex  = regexp.MustCompile(`(?i)-canary`)
```
bparees, line 15: if you want this i won't say no, but there is only a single "canary" job(per release), so it's not really aggregating anything.  You can just look at the canary job directly if you care about it its health.

stbenjam: Yea, I wasn't sure if it was useful -- if it's a variant then it shows up under it's own card  under "Variant status."  Maybe the better thing to do is show a warning if it gets below some high percentage.

stbenjam:
I asked on slack and got no opinions about never-stable, so I think it's better to use it more sparingly.

/hold cancel

pkg/testgridanalysis/testidentification/ocp_variants.go
```diff
@@ -140,10 +107,16 @@ func (v openshiftVariants) IdentifyVariants(jobName string) []string {
 		}
 	}()
 
+	// If a job is in never-stable, it is excluded from other possible variant aggregations
```
bparees, line 110: it seems like this is the main benefit from having a strong list of never-stable jobs:  they don't pollute our variant aggregate results.  Was that discussed?

@deads2k i'd expect you to have strong opinions here.

stbenjam: > it seems like this is the main benefit from having a strong list of never-stable jobs: they don't pollute our variant aggregate results. Was that discussed?

That was the intention, but at the moment [84 of 125 jobs](https://sippy.ci.openshift.org/sippy-ng/jobs/4.9?filters=%7B%22items%22%3A%5B%7B%22id%22%3A3089%2C%22columnField%22%3A%22current_pass_percentage%22%2C%22operatorValue%22%3A%22%3C%3D%22%2C%22value%22%3A%2240%22%7D%5D%7D&sort=asc&sortField=net_improvement) are 40 percent or less. Adding them back into their respective variants isn't going to taint them very much.

I think there's a distinction between "this job is permafailing, TRT has done some initial investigation, we've got a bug and someone's really actually looking at it, let's exclude it from the variant reporting" and "let's sweep these under the rug"

stbenjam: @bparees No feedback from @deads2k here, can you have another look?

What if we meet in the middle and only keep jobs in never-stable if it's been at least investigated and had a bug opened?

That cuts the list in half or so, instead of removing most.

bparees: > What if we meet in the middle and only keep jobs in never-stable if it's been at least investigated and had a bug opened?

that sounds reasonable to me.

stbenjam:
Ok this should be ready now

PR #262, opened 2021-08-19
Add a page for all job runs

stbenjam:
This creates an "all job runs" page, and it's linked from the test
table, so you can for example see a list of all job runs that failed
a particular test, and click through to see it in Prow.

It's live in the sandbox env, from the [tests page](http://sippy-stbenjam-dev.apps.sandbox-m2.ll9k.p1.openshiftapps.com/sippy-ng/tests/4.9?filters=%7B%22items%22%3A%5B%7B%22id%22%3A99%2C%22columnField%22%3A%22name%22%2C%22operatorValue%22%3A%22equals%22%2C%22value%22%3A%22Symptom%20Detection.Undiagnosed%20panic%20detected%20in%20pod%22%7D%5D%7D) click on the running person to see all the job runs that failed that test.

This page is surprisingly performant for me, given the sheer quantity of data you get if you load all the runs [without a filter](http://sippy-stbenjam-dev.apps.sandbox-m2.ll9k.p1.openshiftapps.com/sippy-ng/jobs/4.9/runs).


stbenjam:
/hold cancel

Needs another lgtm, a few minor cosmetic changes

PR #263, opened 2021-08-19
Fix scrollbars on test-by-variant table

stbenjam:
This changes the test-by-variant to use the 'view' class instead of
'table-container' to match the behavior of job details page.


stbenjam:
I'll put this minor fix in #262

PR #264, opened 2021-08-20
Minor fixes: table layouts, sorting, test links

stbenjam:
A handful of minor fixes:

- Fix divs for install and upgrade. This prevents the browser from rendering two sets of horizontal scroll
bars.
- Ensure stable Date.now for consistent snapshots
- Job runs sort before limit
- Tests link to runs sorted by date


stbenjam:
@stevekuznetsov This should fix the other two pages, I think that's all the tables, let me know if you see it on any others

PR #267, opened 2021-08-23
Fix wrapping of test names

### LINKED ISSUE #266: New UI: Test names unhelpfully truncated
The new UI looks great!

Problem: The test names are cut a bit too short, especially since many of them have prefixes.  

Possible solution: 
* Remove the two-column display for more detail (at the cost of more scrolling)
* Squish in the percentages (we probably don't need anything past the decimal point) to give more space to test names.

image

stbenjam:
Data grid tables don't support dynamic row heights, which is why I used
the ellipsis and tooltip, but I realize I don't need dynamic row heights
they can all be the same size, just taller.

This also fixes the test-by-variant table which isn't data grid and can
have dynamic heights.

I also removed the decimal points.

fixes #266 

Release dashboard, compare with [this page](https://sippy.ci.openshift.org/sippy-ng/)
image

Test page, compare with [this page](https://sippy.ci.openshift.org/sippy-ng/tests/4.9?sortField=net_improvement&sort=asc&filters=%7B%22items%22%3A%5B%7B%22id%22%3A1%2C%22columnField%22%3A%22current_runs%22%2C%22operatorValue%22%3A%22%3E%3D%22%2C%22value%22%3A%2210%22%7D%5D%7D)
image

Test-by-variant, compare with [this page](https://sippy.ci.openshift.org/sippy-ng/tests/4.9/details?test=API)
image

### DIFF
--- a/sippy-ng/src/__snapshots__/App.test.js.snap
+++ b/sippy-ng/src/__snapshots__/App.test.js.snap
@@ -1381,719 +1381,7 @@ exports[`app should render correctly 1`] = `
             >
               <a class="MuiTypography-root MuiTypography-h5"
                  style="text-align: center;"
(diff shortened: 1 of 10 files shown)


stbenjam:
Hm, looks like I still need to do something about very very long test names

image

stbenjam:
Is some truncation ok?

image

PR #268, opened 2021-08-25
Add a test analysis page

stbenjam:
I've been looking into the gcp API availability issues, and I wanted an easier way to see when a test started failing, page looks like this, and it's the default destination of clicking on a test name. The example shows clearly when gcp regressed between 8/22 and 8/23:

image


bparees:
/lgtm

i love this new page

PR #277, opened 2021-09-01
bugs: consider version when determining if bug is linked or not

stbenjam:
The difference whether a bug is considered linked to a release, or only
associated is based on the target release. While a bug is untriaged,
target release is blank -- so bugs that are new but definitely linked to
the problem only show up as associated.  Bugs that were "found in" the
current release should be considered linked, not associated.

image


stbenjam:
search.ci needs to collect this info first: https://github.com/openshift/ci-search/pull/104

PR #281, opened 2021-09-01
Calculate and display some new statistical data

stbenjam:
I wanted some additional indicators of overall release health -- this adds a `Tests` widget which shows how often we complete e2e successfully. It also adds a histogram of job pass rates, and a chart of the last 24 hours of job runs.

It's up on http://sippy-stbenjam-dev.apps.sandbox-m2.ll9k.p1.openshiftapps.com/sippy-ng/release/4.9

image


pkg/apis/sippyprocessing/v1/types.go
```diff
@@ -79,6 +91,10 @@ type TopLevelIndicators struct {
 	// Upgrade indicates how successful we are with upgrading onto clusters that have already installed.
 	// Low Upgrade pass rates means clusters are not able to upgrade.  This should stop us from shipping the product.
 	Upgrade FailingTestResult
+	// Tests indicates how successful we are in running the e2e test suites.
+	Tests FailingTestResult
```
bparees, line 95: hm.  This is based on how often the e2e container exits successfully?

presumably only counting job runs where the e2e container was started (i.e. throwing away runs where install failed or something else went wrong early)?

pkg/testgridanalysis/testreportconversion/test_report.go
```diff
@@ -96,6 +104,35 @@ func PrepareTestReport(
...
+		if isNeverStable(job) {
+			continue
+		}
+
+		for _, runs := range job.AllRuns {
+			runCount++
+			if runs.OverallResult == sippyprocessingv1.JobSucceeded {
+				success++
+			} else {
+				failure++
```
bparees, line 122: since this doesn't seem to be excluding job runs where we never even launched the e2e container/step, i think we need a different name for what this represents.

it also doesn't exclude job runs where the e2e step succeeded, but something went wrong during teardown/must-gather.

It's really just the overall pass rate of job runs across the system (Excluding never-stable)?

stbenjam: I think you're right.  I could fix the name, or just take it out. I'm not actually displaying the data anywhere since I didn't think the number told me anything useful.

pkg/apis/sippyprocessing/v1/types.go
stbenjam: I think the utility for TRT is that if we're staying on top of things - filing bugs, and adding exceptions to origin - then we should be holding this percentage steady or improving it.  I think there's also a useful comparison across releases - 4.8 is 70%+ (same for 4.7), but 4.9 is only 55% (but it is higher than last week).

bparees: yeah, i just wonder about how noisy it will be since it's an aggregation across so many factors.  but sure, we can give it a shot.

pkg/testgridanalysis/testreportconversion/test_report.go
stbenjam: I removed this

pkg/apis/sippyprocessing/v1/types.go
stbenjam: Ok, happy to remove it if it turns out not to be useful.  I added more to the tooltip, let me know if that isn't clear enough

sippy-ng/src/releases/ReleaseOverview.js
```diff
@@ -187,9 +205,135 @@ export default function ReleaseOverview(props) {
...
+                    )}&period=hour}`}
+                  >
+                    {dayOffset === 1
+                      ? 'Last 24 hours'
+                      : `${dayOffset * 24} hours ago`}
+                  </Link>
+                  <Tooltip
+                    title={
+                      'This chart shows a 24 hour period of  job runs, excluding never-stable and tech preview. ' +
+                      'The most recent hour will generally look more successful until all jobs finish running. ' +
```
bparees, line 298: we can't excluded running jobs?

stbenjam: Sorry this is worded a little confusingly.  The chart will show running jobs in gray so it's clear we don't have results for those, but the problem is mostly because of jobs that haven't even started.

At the top of the hour the quickest running jobs (e.g. promote) show up green, but later on when everything has run, it settles down and looks more typical. Does that make sense?

bparees: yeah that makes sense.

```suggestion
                      'The most recent hour will generally look more successful until all jobs finish running because it will be heavily biased towards the shortest jobs(more likely to have already completed) which also tend to be the most reliable. ' +
```

stbenjam: Thanks! That's much better

PR #283, opened 2021-09-06
Add 4.10 to Sippy

stbenjam:
Also removes 3.11, there doesn't appear to be any data for it for some time.

### DIFF
--- a/resources/deploymentconfig.yaml
+++ b/resources/deploymentconfig.yaml
@@ -42,6 +42,8 @@ spec:
         - --local-data
         - /data
         - --release
+        - "4.10"
+        - --release
         - "4.9"
         - --release
         - "4.8"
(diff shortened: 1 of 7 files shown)


resources/deploymentconfig.yaml
```diff
@@ -42,15 +42,15 @@ spec:
...
+        - --release
         - "4.9"
         - --release
         - "4.8"
         - --release
         - "4.7"
         - --release
         - "4.6"
-        - --release
-        - "3.11"
```
bparees, line 53: @brenton @nstielau we had added 3.11 at your request and i know at one point we actually had at least one job that was being reported, any idea what happened?  I no longer see the 3.11 jobs on testgrid.

@stbenjam please split this out for now, i don't want to remove 3.11 until we decide we really aren't using sippy to watch it anymore.

brenton: I don't remember anything about the 3.11 job in testgrid unfortunately.

PR #296, opened 2021-09-24
Fetch job data from BigQuery

stbenjam:
This fetches some basic job and run info out of the BigQuery tables and persists it to disk, like we do for TestGrid data.  The new fields are filterable, and show up in some tables, e.g. job runs:

image


deads2k:
The bigquery aspects look well-formed.  This is just a first step towards something else?

stbenjam:
> The bigquery aspects look well-formed. This is just a first step towards something else?

Yea, I just wanted this to get reviewed, and then I can start pulling in more data. Is the schema of your tables fixed at this point? 

And I was going to pull in job run data next, unless you had a preference.

pkg/bigqueryanalysis/insertion.go
```diff
@@ -0,0 +1,58 @@
+package bigqueryanalysis
+
+import (
+	bigqueryv1 "github.com/openshift/sippy/pkg/apis/bigquery/v1"
+	sippyv1 "github.com/openshift/sippy/pkg/apis/sippyprocessing/v1"
+	"github.com/openshift/sippy/pkg/util"
+)
+
+// InsertBigQueryDataToJobs adds data from the bigquery tables to the JobResult data from TestGrid.
+func InsertBigQueryDataToJobs(bigqueryJobs []bigqueryv1.Job, sippyJobs []sippyv1.JobResult) []sippyv1.JobResult {
```
dgoodwin, line 10: Nit: mentally this feels like a merge, suggest MergeBigQueryDataToJobs or similar.

pkg/apis/bigquery/v1/types.go
```diff
@@ -0,0 +1,31 @@
...
+
+type JobRun struct {
+	Name       string
+	JobName    string
+	Status     string
+	StartTime  time.Time
+	EndTime    time.Time
+	ReleaseTag string
+	Cluster    string
+}
```
dgoodwin, line 31: Are these an actual API type we'd return, or something we'd work with only transiently to merge the data with the existing testgrid data?

If it's the latter, suggest taking these out of the apis package. If not then disregard. :)

pkg/sippyserver/analyzer.go
```diff
@@ -5,6 +5,9 @@ import (
 	"regexp"
 	"time"
 
+	"github.com/openshift/sippy/pkg/bigqueryanalysis"
+
+	bigqueryv1 "github.com/openshift/sippy/pkg/apis/bigquery/v1"
```
dgoodwin, line 10: Nit: Looks like it should group with the imports below.

pkg/bigqueryanalysis/bigquery.go
```diff
@@ -0,0 +1,96 @@
...
+
+	return jobs, nil
+}
+
+func (client *Client) GetJobRuns(ctx context.Context, releases []string) ([]v1.JobRun, error) {
+	runs := make([]v1.JobRun, 0)
+	query := client.Query(`
+		SELECT
+			*
+		FROM ` + "`ci_data.JobRuns`")
```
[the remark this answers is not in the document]
dgoodwin: Yeah I'm afraid that's where we're heading as well.

When we get to the really expensive stuff hopefully we can query bigquery when refreshing data, and just store summaries/results rather than all the data?

main.go
```diff
@@ -175,6 +175,7 @@ func (o *Options) Run() error {
 			dashboards = append(dashboards, dashboardCoordinate.TestGridDashboardNames...)
 		}
 		testgridhelpers.DownloadData(dashboards, o.JobFilter, o.FetchData)
+		bigqueryhelpers.DownloadData(o.OpenshiftReleases, o.FetchData)
```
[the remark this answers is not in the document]
stbenjam: We do, there's a k8s instance of sippy: https://sippy.k8s.io/

If GOOGLE_APPLICATION_CREDENTIALS env var isn't set, we'll end up skipping this. It probably makes sense to have a flag to hide all the openshift-isms.

pkg/apis/bigquery/v1/types.go
stbenjam: It's an intermediate type used to marshal the bigquery rows. We have similar for TestGrid, which is why I put it here: https://github.com/openshift/sippy/tree/master/pkg/apis/testgrid/v1

pkg/apis/sippyprocessing/v1/types.go
```diff
@@ -175,19 +175,33 @@ const (
...
 type JobRunResult struct {
+	ID              string           `json:"id"`
 	Job             string           `json:"job"`
 	URL             string           `json:"url"`
 	TestFailures    int              `json:"testFailures"`
 	FailedTestNames []string         `json:"failedTestNames"`
 	Failed          bool             `json:"failed"`
 	Succeeded       bool             `json:"succeeded"`
 	Timestamp       int              `json:"timestamp"`
+	Duration        int64            `json:"duration"`
```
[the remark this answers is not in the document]
stbenjam: It's actually millis! I'll add comments to all these

PR #311, opened 2021-10-19
Update never stable for October 19, 2021

stbenjam:


pkg/testgridanalysis/testidentification/ocp_variants.go
```diff
@@ -96,31 +94,13 @@ var (
...
 		// https://bugzilla.redhat.com/show_bug.cgi?id=2007692
 		"release-openshift-origin-installer-old-rhcos-e2e-aws-4.9",
 		"release-openshift-origin-installer-old-rhcos-e2e-aws-4.10",
 
-		// https://bugzilla.redhat.com/show_bug.cgi?id=2008128
-		"release-openshift-origin-installer-e2e-azure-shared-vpc-4.9",
-		"release-openshift-origin-installer-e2e-azure-shared-vpc-4.10",
-		"release-openshift-origin-installer-e2e-aws-shared-vpc-4.9",
-		"release-openshift-origin-installer-e2e-aws-shared-vpc-4.10",
-
```
dgoodwin, line 123: More going out than going in, awesome.

Would be cool if someday sippy could inform us when never-stable jobs with an associated bug has gone verified.

pkg/testgridanalysis/testidentification/ocp_variants.go
```diff
@@ -160,6 +140,10 @@ var (
 		"release-openshift-ocp-installer-e2e-metal-compact-4.9",
 		"release-openshift-origin-installer-e2e-aws-sdn-network-stress-4.9",
 		"release-openshift-origin-installer-e2e-aws-sdn-network-stress-4.10",
+
+		// https://bugzilla.redhat.com/show_bug.cgi?id=2012844
+		"periodic-ci-openshift-release-master-nightly-4.10-e2e-metal-assisted",
+		"periodic-ci-openshift-release-master-nightly-4.10-e2e-metal-assisted-ipv6",
```
dgoodwin, line 146: This is a shame but it's been over a week with no info on the bz. Looks good to me and we should ping them externally and let them know.

PR #313, opened 2021-10-29
Speedup sippy API startup by pre-generating reports

dgoodwin:
Goal is to move the expensive report generation from server startup, where we're stuck waiting roughly 2 minutes per --release with sippy down entirely, to the slower fetchdata scripts. This wait is extremely painful for developing on sippy.

Adds a new --gen-reports command which is to be run separately from --fetch-data, as we may want to run it against a snapshot of historical restored testgrid data.

With this change and using just --release 4.10, sippy api server startup time goes from 2m to 9s.

### DIFF
--- a/pkg/sippyserver/server.go
+++ b/pkg/sippyserver/server.go
@@ -97,6 +97,7 @@ func (s *Server) refresh(w http.ResponseWriter, req *http.Request) {
 
 func (s *Server) RefreshData() {
 	klog.Infof("Refreshing data")
+	s.bugCache.Clear()
 
 	s.currTestReports = s.testReportGeneratorConfig.TestGridLoadingConfig.ReportLoader(
 		s.testReportGeneratorConfig.TestGridLoadingConfig.LocalData)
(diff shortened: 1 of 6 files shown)


dgoodwin:
Tests fixed now.

pkg/sippyserver/server.go
```diff
@@ -97,10 +97,9 @@ func (s *Server) refresh(w http.ResponseWriter, req *http.Request) {
 
 func (s *Server) RefreshData() {
 	klog.Infof("Refreshing data")
-	s.bugCache.Clear()
```
stbenjam, line 100: Is removing the bug cache clearance intentional?

dgoodwin: It was yes, as far as I can see, there is no bug cache in play anymore, the data is included in the reports file. I will look into removing s.bugCache and see if that surfaces anything I missed on Monday.

dgoodwin: I did find it used here: https://github.com/openshift/sippy/blob/38ca4dec6de0c2b5db3a20c2d15a16cc8fb42cf0/pkg/sippyserver/server.go#L219

Very strange function, seems insanely slow to be hooked up to a live web request. I'm guessing that is somewhat legacy?

In any case, is the best course of action to restore the bug cache clearing and leave it be for now?

dgoodwin: Updated.

PR #315, opened 2021-11-03
Fetch release controller releases, store in postgresql database

stbenjam:
This PR fetches release controller data from BigQuery and exports it to a local postgres DB.

Create a simple DB:

```
$ podman run --name some-postgres -e POSTGRES_PASSWORD=password -p 5432:5432 -d quay.io/enterprisedb/postgresql
```

Fetch data, specifiying a DSN and google creds:

```
$  export GOOGLE_APPLICATION_CREDENTIALS=<my creds.json>
$ ./sippy --fetch-data ../sippy-data --database-dsn="postgresql://postgres:password@localhost/postgres" --release 4.10 
```

Then start up the server:

```
$ ./sippy -v 1 --server --local-data ../sippy-data --release 4.10 --database-dsn="postgresql://postgres:password@localhost/postgres" --skip-bug-lookup
```

And the UI:

```
$ cd sippy-ng/
$ npm start
```

And visit http://localhost:3000/releases/tags


go.mod
```diff
@@ -3,17 +3,21 @@ module github.com/openshift/sippy
...
 	github.com/montanaflynn/stats v0.6.6
 	github.com/pkg/errors v0.9.1
 	github.com/spf13/cobra v0.0.6
 	github.com/tidwall/gjson v1.9.4
 	github.com/valyala/fasttemplate v1.2.1 // indirect
 	go.etcd.io/bbolt v1.3.6 // indirect
-	golang.org/x/sys v0.0.0-20211004093028-2c5d950f24ef // indirect
+	google.golang.org/api v0.60.0
+	gorm.io/driver/postgres v1.2.1
+	gorm.io/gorm v1.22.2
```
dgoodwin, line 21: I'd been tempoted to rely mostly on jsonb columns, but doing a little research, this is probably better. There's a price to pay for use of json in terms of what you can do with the data and for things we control internal to sippy itself, it makes sense to do proper relational db. hopefully gorm makes that fairly easy.

pkg/db/models/releases.go
```diff
@@ -0,0 +1,35 @@
+package models
+
+type ReleaseTag struct {
```
dgoodwin, line 3: Can you clarify why we need two very similar models?

pkg/apis/bigquery/v1/releases.go
```diff
@@ -0,0 +1,35 @@
+package v1
+
+type ReleaseRow struct {
```
dgoodwin, line 3: Name's a bit generic here, ReleasePayload perhaps?

pkg/bigqueryexporter/bigquery.go
```diff
@@ -0,0 +1,57 @@
...
+		if err == iterator.Done {
+			break
+		}
+		if err != nil {
+			return err
+		}
+		rows = append(rows, row)
+	}
+
+	return db.DB.Clauses(clause.OnConflict{UpdateAll: true}).Create(&rows).Error
```
dgoodwin, line 56: This is interesting, on an id conflict it will update the row instead? Could we have an issue with orphaned rows that should be deleted?

pkg/db/db.go
```diff
@@ -0,0 +1,26 @@
...
+	DB *gorm.DB
+}
+
+func New(dsn string) (*DB, error) {
+	db, err := gorm.Open(postgres.Open(dsn), &gorm.Config{})
+	if err != nil {
+		return nil, err
+	}
+
+	if err := db.AutoMigrate(&models.ReleaseTag{}); err != nil {
```
dgoodwin, line 19: Neat.

dgoodwin (commented):
I really think we need to start moving in this direction given the problems I'm hitting elsewhere. I'm a big +1 on moving to use of a db.

go.mod
stbenjam: gorm can do jsonb, https://v1.gorm.io/docs/dialects.html -- so we could always use that to start for existing reports, and move to normalized data over time

dgoodwin: That's really cool, and might be a good idea, we can treat the db as totally throwaway for some time while we sort that all out.

pkg/db/models/releases.go
stbenjam: We don't, I'm not sure why I duplicated it TBH.  I'll remove the bigquery one.

pkg/bigqueryexporter/bigquery.go
stbenjam: This lets me fetch all data from bigquery and update the rows if needed. I should probably change this to only fetch data from BQ with an ID > our current local max ID.  That should avoid needing to pull down everything, every refresh.

It's true that if data disappears from bigquery, we will still keep the data in our DB.  Removing data from bigquery seems unlikely in the short term, but I'll see if I can think about a way to fix this.

dgoodwin: I had it in my head that the db,or most of it, might be transient and could be wiped out and rebuilt. That however may not be a great thing to do on refreshes as it would bring down sippy UI for a bit in the middle. Probably can disregard my concern here.

dgoodwin:
What do you think about one of us doing a priming PR where we get a postgres container into the production pod and sippy connecting to it. Or we can lump it into this PR.

resources/deploymentconfig.yaml
```diff
@@ -56,9 +56,30 @@ spec:
...
+            memory: 2G
+        env:
+          - name: POSTGRES_PASSWORD
+            valueFrom:
+              secretKeyRef:
+                name: postgres
+                key: password
+        volumeMounts:
+          - mountPath: /var/lib/postgresql/data
+            name: db
```
dgoodwin, line 82: Is postgresql going to be *required* or optional?

Presumably the kube instance is using separate manifests, as I doubt they have DeploymentConfig.

However is this something we want to lean in hard, or do we want to preserve both methods for now?

For this PR am I correct that bigquery will keep the data permanently for now, so if we were to blow away the postgres db, it would all get re-imported next time through? Are we targeting postgres being something that is easy to throwaway and rebuild via the fetchdata script, and if so is that likely permanent or temporary? In my mind it might be nice to start transient and throwaway, and in time possibly it becomes something we need to monitor and backup.

We'll want to carefully balance with the historical requirements for loading a directory of testgrid data. That should probably be replaced by something in the db ultimately though if it becomes permanent data.

resources/postgres-secret.yaml.example
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: postgres
+type: Opaque
+stringData:
+  password: password
+  dsn: "postgresql://postgres:password@localhost/postgres"
```
dgoodwin, line 8: Perfect gitops world we'd have a separate repo for kube artifacts, and a Secret there or I guess via some vault solution. That might be possible if we move Sippy to the DPCR cluster. For now, what do we think about manually creating the Secret?

stbenjam: Yea I was manually creating it on the cluster (with a different password), I just had this here as an example

resources/deploymentconfig.yaml
stbenjam: > Is postgresql going to be required or optional?

I think if we move report data to it other than bigquery, we should just always put it in a Postgres DB.  WDYT? It’s probably using less resources to do that.

> Presumably the kube instance is using separate manifests, as I doubt they have DeploymentConfig.

Yea, I don’t think they use anything in here.  Also updating it is a manual process. David knows the details.

> For this PR am I correct that bigquery will keep the data permanently for now, so if we were to blow away the postgres db, it would all get re-imported next time through? Are we targeting postgres being something that is easy to throwaway and rebuild via the fetchdata script, and if so is that likely permanent or temporary? In my mind it might be nice to start transient and throwaway, and in time possibly it becomes something we need to monitor and backup.

I had imagined the Postgres always being ephemeral. It’ll be rebuilt from scratch if needed.  I’m not entirely sure I want to be in the business of being responsible for backups, etc — it’s nicer if we can re-derive everything from BigQuery or TestGrid data.  I guess if we want to have comments or other stateful data, we’ll have to come up with a solution for that.


> We'll want to carefully balance with the historical requirements for loading a directory of testgrid data. That should probably be replaced by something in the db ultimately though if it becomes permanent data.

PR #316, opened 2021-11-03
[wip] Add init container for generating reports

stbenjam:
After #313, we write generated reports directly to disk to speed up sippy start, but there's a catch-22 where we don't run the generate reports for the first time until fetch-data runs, so Sippy will fail to start.


bparees:
just a reminder, you'll need to apply this change directly to the deployed resources (before or after it merges)

PR #322, opened 2021-11-10
Add name to sippy service

stbenjam:
oc apply is complaining:

  The Service "sippy" is invalid: spec.ports[0].name: Required value

### DIFF
--- a/resources/service.yaml
+++ b/resources/service.yaml
@@ -8,7 +8,8 @@ metadata:
   name: sippy
 spec:
   ports:
-  - port: 8080
+  - name: www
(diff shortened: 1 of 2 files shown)


resources/service.yaml
```diff
@@ -8,7 +8,8 @@ metadata:
   name: sippy
 spec:
   ports:
-  - port: 8080
+  - name: www
+    port: 8080
```
bparees, line 12: update the historical service too?

stbenjam: Done

PR #323, opened 2021-11-15
Fix metrics port for sippy.

dgoodwin:
Accidentally added to fetchdata container, not the API.


dgoodwin:
@stbenjam I think I need someone to manually apply this. :( Doesn't seem to have rolled out.

PR #327, opened 2021-11-18
Show period toolbar selector on tests and jobs pages

stbenjam:
Without a default value it isn't shown, this sets one so it appears. This sets period to `default`, and updates the API/test snapshots. 

image


deads2k:
/lgtm
/hold

holding in case someone more javascript aware needs to look.

PR #329, opened 2021-11-19
Postgres Reporting Phase 1

dgoodwin:
This is a reattempt after #319 which had to be reverted after discovering that trying to serialize 5 releases worth of data we hold in memory to a json file would immediately blow out any reasonable amount of memory we could give the pod and crash.

This PR attempts to build on that PR, and the work @stbenjam did recently to bring postgres into sippy.

Goal is the same.

Today sippy has a slowish fetch data script that runs in a sidecar container, pulling testgrid data down to disk.
When sippy's api startup, it does a pile of calculation on that disk data, and stores a pile of data in memory. It takes roughly 2 minutes per --release you're pulling, so each time we roll out a new sippy commit we're down for approximately 10 minutes will sippy recalculates the world. For developers this is a huge pain in testing API changes. We're also using way more memory than necessary.

This patch is a step towards moving us to a world where we run the slow fetch data script, which also populates the database with what we parse out of testgrid, in a more sane form. Then we can use this db in our API handlers. Sippy will start instantly.

At present, nothing changes in the API server, it's still doing all it's analysis up front and will start very slow. This PR adds loading data into postgres, and one ported API handler at a temporary API endpoint so we can verify we're getting the data in there and see how things are working. (rather than a massive refactor and cutover, I'd like to do this slowly as time allows) Sample API can be curled at /api-ex/jobs, and should be equivalent to /api/jobs, though they are returning very slightly different numbers for reasons that are difficult to discern given how testgrid presents it's data.

We continue to assume the db is ephemeral and can easily be destroyed and rebuilt from testgrid data, though at some point it would be nice to begin storing this data longer term and let sippy report on those larger trends.

As written loading the database with 1 release takes about 1.5 minutes. The /api-ex/jobs endpoint takes about 137ms to load the relevant data from postgres and perform all calculations, well within reason for an API request to process.


pkg/api/jobs.go
```diff
@@ -104,9 +109,11 @@ func PrintJobsReport(w http.ResponseWriter, req *http.Request, currReport, twoDa
...
 	// period (typically 7 days) and the last two days.
+	// Otherwise the default of last 7 days vs last 14 days.
 	var current, previous []v1sippyprocessing.JobResult
-	switch req.URL.Query().Get("period") {
-	case "twoDay":
+	period := req.URL.Query().Get("period")
+	switch period {
+	case periodTwoDay:
 		current = twoDayPeriod
 		previous = currentPeriod
```
stbenjam, line 118: I *think* I do this wrong in the new UI, since it compares 2 days to the previousPeriod, but I think both comparisons aren't ideal.  It might be nice to change the reporting to compare 2 days to the previous 7 days (e.g. 9 days ago to 2 days ago) once it's all in a DB.

pkg/db/models/prow.go
```diff
@@ -0,0 +1,61 @@
...
+// IDName is a partial struct to query limited fields we need for caching. Can be used
+// with any type that has a unique name and an ID we need to lookup.
+// https://gorm.io/docs/advanced_query.html#Smart-Select-Fields
+type IDName struct {
+	ID   uint
+	Name string `gorm:"unique"`
+}
+
+type ProwJobRun struct {
+	gorm.Model
```
stbenjam, line 38: Oh nice, I should've included this on the release structs.

pkg/api/jobs.go
```diff
@@ -138,6 +145,152 @@ func PrintJobsReport(w http.ResponseWriter, req *http.Request, currReport, twoDa
...
+	}
+	klog.V(3).Infof("Found %d job runs in period between %s and %s",
+		len(jobRunsInPeriod),
+		startDate.Format(time.RFC3339),
+		now.Format(time.RFC3339))
+
+	// Build a map of job ID to the API job result we'll return.
+	// Iterate all job results, incrementing counters on the job result we'll return.
+	apiJobResults := map[uint]*apitype.Job{}
+	for _, jobRun := range jobRunsInPeriod {
```
stbenjam, line 220: What do you think about doing more of this in SQL? We'd have to loop over jobs, but that's much smaller than job runs (and it might be possible to do that in sql too)

```sql
WITH results AS (
SELECT
	coalesce(count(case when succeeded = true AND timestamp BETWEEN NOW() - INTERVAL '14 DAY' AND NOW() - INTERVAL '7 DAY' then 1 end), 0) as previous_passes,
	coalesce(count(case when succeeded = false AND timestamp BETWEEN NOW() - INTERVAL '14 DAY' AND NOW() - INTERVAL '7 DAY' then 1 end), 0) as previous_failures,
	coalesce(count(case when timestamp BETWEEN NOW() - INTERVAL '14 DAY' AND NOW() - INTERVAL '7 DAY' then 1 end), 0) as previous_total_runs,
	coalesce(count(case when infrastructure_failure = true AND timestamp BETWEEN NOW() - INTERVAL '14 DAY' AND NOW() - INTERVAL '7 DAY' then 1 end), 0) as previous_infrastructure_failures,
	
	coalesce(count(case when succeeded = true AND timestamp > NOW() - INTERVAL '7 DAY' then 1 end), 0) as current_passes,
	coalesce(count(case when succeeded = false AND timestamp > NOW() - INTERVAL '7 DAY' then 1 end), 0) as current_failures,	
	coalesce(count(case when timestamp > NOW() - INTERVAL '7 DAY' then 1 end), 0) as current_total_runs,
	coalesce(count(case when infrastructure_failure = true AND timestamp > NOW() - INTERVAL '7 DAY' then 1 end), 0) as current_infrastructure_failures
FROM
	prow_job_runs JOIN prow_jobs ON prow_jobs.id = prow_job_runs.prow_job_id WHERE prow_jobs.name = 'promote-release-openshift-machine-os-content-e2e-aws-4.10'
)
SELECT *,
	previous_passes * 100.0 / previous_total_runs AS previous_pass_percentage,
	previous_failures * 100.0 / previous_total_runs AS previous_failure_percentage,
	current_passes * 100.0 / current_total_runs AS current_pass_percentage,
	current_failures * 100.0 / current_total_runs AS current_failure_percentage
FROM results;
```

Results:

```
-[ RECORD 1 ]--------------------+-----------------------
previous_passes                  | 591
previous_failures                | 4
previous_total_runs              | 595
previous_infrastructure_failures | 1
current_passes                   | 67
current_failures                 | 0
current_total_runs               | 67
current_infrastructure_failures  | 0
previous_pass_percentage         | 99.3277310924369748
previous_failure_percentage      | 0.67226890756302521008
current_pass_percentage          | 100.0000000000000000
current_failure_percentage       | 0.00000000000000000000
```

pkg/db/models/prow.go
```diff
@@ -0,0 +1,61 @@
...
+	Variants    pq.StringArray `gorm:"type:text[]"`
+	TestGridURL string
+
+	// BugList []bugsv1.Bug `json:"bugList" gorm:"-"`
+	// AssociatedBugList are bugs that match the test/job, but do not match the target release
+	// AssociatedBugList []bugsv1.Bug `json:"associatedBugList" gorm:"-"`
+
+	// TestResults holds entries for each test that is a part of this aggregation.  Each entry aggregates the results
+	// of all runs of a single test.  The array is sorted from lowest PassPercentage to highest PassPercentage
+	// TestResults []TestResult `json:"results" gorm:"foreignKey:Job;References:Name"`
```
stbenjam, line 26: Should this be removed?

pkg/sippyserver/dbloader.go
```diff
@@ -0,0 +1,125 @@
...
+					err := dbc.DB.Clauses(clause.OnConflict{UpdateAll: true}).Create(&ft).Error
+					if err != nil {
+						return errors.Wrapf(err, "error loading test into db: %s", ft.Name)
+					}
+				}
+				failedTests[i] = ft
+			}
+			pjr.FailedTests = failedTests
+
+			err := dbc.DB.Clauses(clause.OnConflict{UpdateAll: true}).Create(&pjr).Error
```
stbenjam, line 114: It's a little faster you make an array, and do `CreateInBatches`

pkg/sippyserver/dbloader.go
```diff
@@ -0,0 +1,125 @@
...
+		dashboard.ReportName, rawJobResults, bugCache, dashboard.BugzillaRelease, variantManager)
+
+	// Load all job and test results into database:
+	klog.Info("loading ProwJobs into db")
+
+	// Build up a cache of all prow jobs we know about to speedup data entry.
+	// Maps job name to db ID.
+	prowJobCache := map[string]uint{}
+	var idNames []models.IDName
+	dbc.DB.Model(&models.ProwJob{}).Find(&idNames)
```
stbenjam, line 47: Can we make sure this is filterable?

pkg/api/jobs.go
dgoodwin: We've got this working for all jobs on slack, will implement with a followup.

pkg/sippyserver/dbloader.go
dgoodwin: Looks like i'll need to redo once we move to the query, will address in followup.

dgoodwin:
Other comments addressed, decent history in the commits so I guess I'll leave them be if that's ok.

main.go
```diff
@@ -179,11 +182,36 @@ func (o *Options) Validate() error {
...
+		return fmt.Errorf("must specify --local-data with --load-database")
+	}
+
+	if o.LoadDatabase && o.DSN == "" {
+		return fmt.Errorf("must specify --database-dsn with --load-database")
+	}
+
+	if !o.Server && !o.LoadDatabase && o.FetchData == "" && o.DSN == "" {
+		return fmt.Errorf("must specify --database-dsn with for cli reports")
+	}
```
stbenjam, line 207: I think I've convinced myself this logic works

PR #330, opened 2021-11-21
Distinguish between no payload and rejected payloads

stbenjam:
This distinguishes between no payload, and rejected payloads.

- Rejected payloads are red.
- Accepted payloads in the last 24 hours are green.
- When we haven't had a payload in several days, but the most recent one
  was accepted, we mark it as a warning.

A tooltip also now tells you how many of the most recent payloads
had that status (e.g. The last 9 payloads were rejected.)

image

### DIFF
--- a/pkg/db/models/release_tag.go
+++ b/pkg/db/models/release_tag.go
@@ -75,3 +75,49 @@ func GetLastAcceptedByArchitectureAndStream(db *gorm.DB, release string) ([]Rele
 
 	return results, nil
 }
+
+// GetLastPayloadStatus returns the most recent payload status for an architecture/stream combination,
+// as well as the count of how many of the last payloads had that status (e.g., when this returns
+// Rejected, 5 -- it means the last 5 payloads were rejected.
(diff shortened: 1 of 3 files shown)


pkg/db/models/release_tag.go
```diff
@@ -75,3 +75,49 @@ func GetLastAcceptedByArchitectureAndStream(db *gorm.DB, release string) ([]Rele
...
+					SUM(change) OVER(ORDER BY id) AS group FROM changes
+			)
+		SELECT
+			phase, COUNT(phase)
+		FROM
+			groups
+		WHERE
+			groups.group = 1
+		GROUP BY
+			phase`, architecture, stream, release).Scan(&count)
```
stbenjam, line 120: This absurd SQL statement brought to you by cobbling together several StackOverflow:tm: posts.  

It uses multiple [CTE](https://www.geeksforgeeks.org/cte-in-sql/) to group islands of release tags in a particular phase, and gets the most recent island. It lets us know the status (i.e. accepted or rejected) of the most recent payload, and also how many of them occurred in a row.

PR #332, opened 2021-11-23
db sync: batch insertions and only fetch what is needed from bigquery

stbenjam:
We're failing to insert job runs because the size of the insert is now
greater than 65,535 (max of int16). This changes the insert to go in
batches.

More importantly, we also only fetch what is exactly needed from
BigQuery.  If our DB is already populated, there's no need to pull
everything from BQ - this gets the highest ID from postgres, and fetches
only the newer records from BQ.

Note: BigQuery does not support autoincrementing primary keys, so we use
`ROW_NUMBER() OVER(...)` -- this produces stable ID's if we do not
remove older records. Currently, we use BQ as append-only and for the
foreseeable future will probably maintain release information
indefinitely (it's not a lot of data).

### DIFF
--- a/pkg/bigqueryexporter/bigquery.go
+++ b/pkg/bigqueryexporter/bigquery.go
@@ -52,10 +52,13 @@ func (client *Client) ExportData(ctx context.Context, dbClient *db.DB) error {
 
 func (client *Client) ExportReleaseTags(ctx context.Context, dbClient *db.DB) error {
 	rows := make([]models.ReleaseTag, 0)
-	query := client.Query(`
-		SELECT
-			ROW_NUMBER() OVER() id, *
-		FROM ` + "`ci_data.ReleaseTags`")
+	query := client.Query(`SELECT * FROM (SELECT ROW_NUMBER() OVER() id, * FROM ` + "`ci_data.ReleaseTags`) WHERE id > @lastID)")
+	query.Parameters = []bigquery.QueryParameter{
+		{
+			Name:  "lastID",
+			Value: dbClient.LastID("release_tags"),
+		},
+	}
 	it, err := query.Read(ctx)
 	if err != nil {
 		return err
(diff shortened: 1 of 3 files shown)


pkg/bigqueryexporter/bigquery.go
```diff
@@ -52,10 +52,13 @@ func (client *Client) ExportData(ctx context.Context, dbClient *db.DB) error {
 
 func (client *Client) ExportReleaseTags(ctx context.Context, dbClient *db.DB) error {
 	rows := make([]models.ReleaseTag, 0)
-	query := client.Query(`
-		SELECT
-			ROW_NUMBER() OVER() id, *
-		FROM ` + "`ci_data.ReleaseTags`")
+	query := client.Query(`SELECT * FROM (SELECT ROW_NUMBER() OVER() id, * FROM ` + "`ci_data.ReleaseTags`) WHERE id > @lastID)")
```
dgoodwin, line 55: Could you clarify the select row_number like you did in the commit msg with a comment on each use here? Will be helpful for future maintenance I think.

Should we be using an order by here? It feels a little bit coincidental that this produces consistent results and if it didn't, thinks might get a little weird? If you're confident this is documented and safe then disregard my concern.

pkg/bigqueryexporter/bigquery.go
```diff
@@ -73,15 +76,18 @@ func (client *Client) ExportReleaseTags(ctx context.Context, dbClient *db.DB) er
...
+	return dbClient.DB.Clauses(clause.OnConflict{UpdateAll: true}).CreateInBatches(&rows, dbClient.BatchSize).Error
 }
 
 func (client *Client) ExportPullRequests(ctx context.Context, dbClient *db.DB) error {
 	rows := make([]models.PullRequest, 0)
-	query := client.Query(`
-		SELECT
-			ROW_NUMBER() OVER() id, *
-		FROM ` + "`ci_data.ReleasePullRequests`")
+	query := client.Query(`SELECT * FROM (SELECT ROW_NUMBER() OVER() id, * FROM ` + "`ci_data.ReleasePullRequests`) WHERE id > @lastID")
```
dgoodwin, line 84: So bigquery literally has no ID we can reuse on these rows in the table?

stbenjam: [Nope](https://cloud.google.com/blog/products/data-analytics/bigquery-and-surrogate-keys-practical-approach). We could create and manage our own, but ROW_NUMBER seems to work ok.

dgoodwin:
Perfect.

/lgtm

PR #337, opened 2021-12-03
Remove job and job run tags from API.

dgoodwin:
These were only set to upgrade if variants contained upgrade, and used
in manual UI filtering. We suspect they may not be needed so removing
for now.


stbenjam:
Seems good to me, not entirely sure why I added this.  You can already use the upgrade variant, or just checking if the name contains variant.

/lgtm

dgoodwin:
/hold

I updated the test data and it's passing locally, not sure what this is.

stbenjam:
I think you'll have to re-run your snapshots with the locale env vars, too

dgoodwin:
@stbenjam still busted, could I get a crash course in what's going on here? What are these snapshots?

Failures now are all kinds of these numbers being off like this: 

```
     -                                 <div class="MuiCardContent-root makeStyles-cardContent-292"
    +                                 <div class="MuiCardContent-root makeStyles-cardContent-289" 
```

stbenjam:
/me curses whoever added these snapshot tests (it was me)

```
     -                   <a class="MuiBox-root MuiBox-root-227"
    +                   <a class="MuiBox-root MuiBox-root-224" 
```

Not sure why this is happening now, those specific ID's used to be deterministic. We can just remove them:

```diff
diff --git a/sippy-ng/src/setupTests.js b/sippy-ng/src/setupTests.js
index a619586..47b56b0 100644
--- a/sippy-ng/src/setupTests.js
+++ b/sippy-ng/src/setupTests.js
@@ -15,6 +15,8 @@ export const withoutMuiID = (wrapper) =>
       .html()
       .replace(/id="mui-[0-9]*"/g, '')
       .replace(/aria-labelledby="(mui-[0-9]* *)*"/g, '')
+      .replace(/makeStyles-.*-[0-9]*/, '')
+      .replace(/MuiBox-root-.*-[0-9]*/, '')
   )
 
 Enzyme.configure({ adapter: new Adapter() })
```

stbenjam:
The snapshot tests render the page and ensure it matches what's expected, it's caught a bunch of problems where I broke a page unknowingly, introduced infinite react re-renders, etc.  They're always a little troublesome though, and some of the behavior seems tied to my + CI environment which I guess are fairly close.

dgoodwin:
Hah! Thank you for the help though.

They do seem really brittle, surprising though that your env matches well with the CI env on pr testing, maybe the problem is my setup somehow.

Pushing another attempt now with the patch you have above.

dgoodwin:
Hah! Thank you for the help though.

They do seem really brittle, surprising though that your env matches well with the CI env on pr testing, maybe the problem is my setup somehow.

/hold cancel

PR #338, opened 2021-12-06
postgres jobs query

dgoodwin:
Uses a very fast query crafted by @stbenjam, serializes to the existing API type, and does in-memory filtering.

This can build a report across 4.6-4.10 in about 50ms on my laptop.

### DIFF
--- a/pkg/api/jobs.go
+++ b/pkg/api/jobs.go
@@ -7,6 +7,7 @@ import (
 	gosort "sort"
 	"strconv"
 	"strings"
+	"text/template"
 	"time"
 
 	apitype "github.com/openshift/sippy/pkg/apis/api"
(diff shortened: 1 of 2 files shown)


dgoodwin:
@stbenjam how much do you hate this in-memory filtering? :) 

I'm not sure we can do views with configurable slicing. 

Another option would be to inject WHERE's into the query.

pkg/api/jobs.go
```diff
@@ -168,115 +168,79 @@ func PrintDBJobsReport(w http.ResponseWriter, req *http.Request,
...
-		finalAPIJobResult = append(finalAPIJobResult, *v)
+		filteredJobReports = append(filteredJobReports, jobReport)
 	}
 
 	elapsed := time.Since(now)
-	klog.Infof("BuildJobResult completed in %s with %d results", elapsed, len(finalAPIJobResult))
+	klog.Infof("BuildJobResult completed in %s with %d results from db, filtered down to %s", elapsed, len(jobReports), len(filteredJobReports))
 
-	return finalAPIJobResult, nil
+	return filteredJobReports, nil
```
stbenjam, line 243: I think you'll want to call `sort(req).limit(req))` on this

dgoodwin: Looks like I accidentally did that already, this func is called by PrintDBJobsReport which has the request param and does the sort/limit.

stbenjam:
> @stbenjam how much do you hate this in-memory filtering? :)
> I'm not sure we can do views with configurable slicing.

This is certainly an improvement since we've still offloaded a bunch to SQL.  This blog post - https://vvvvalvalval.github.io/posts/using-postgresql-temporary-views-for-expressing-business-logic.html -- talks about basically the kind of thing we want.  We do have the ability to create a temporary view and then query from it (i.e. apply our filters via SQL).

We could experiment with that later on.

> Another option would be to inject WHERE's into the query.

I think the issue is mostly where to put the where's.  If you want current percentage it goes in one place but on job name it'd go in a different part of the query, I think.

stbenjam:
I'm happy with this as-is (or at least after lint/unit pass)

dgoodwin:
I'm digging into the tests aspect and with a loaded database, 10 million test rows over last 30 days, indeed you are right it's looking too slow. Query is up to about 5s and I'm not done yet, so we probably will need the materialized views.

Thoughts on how to proceed? We can merge this and interate (my test work builds on this), or hold and bring it all in together.

PR #339, opened 2021-12-07
Update never-stable week of Dec 6th

stbenjam:
Many graduates, few new ones.

### DIFF
--- a/pkg/testgridanalysis/testidentification/ocp_variants.go
+++ b/pkg/testgridanalysis/testidentification/ocp_variants.go
@@ -99,10 +99,6 @@ var (
 		"periodic-ci-openshift-verification-tests-master-ocp-4.10-e2e-baremetal-cucushift-ipi",
 		"periodic-ci-openshift-verification-tests-master-ocp-4.10-e2e-openstack-cucushift-ipi",
 
-		// Reported on slack to SD-CICD
-		"release-openshift-ocp-osd-aws-nightly-4.10",
-		"release-openshift-ocp-osd-gcp-nightly-4.10",
-
 		// https://bugzilla.redhat.com/show_bug.cgi?id=2019375
 		"periodic-ci-openshift-release-master-nightly-4.10-e2e-openstack-proxy",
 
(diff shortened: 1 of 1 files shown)


pkg/testgridanalysis/testidentification/ocp_variants.go
```diff
@@ -153,9 +150,7 @@ var (
 		"periodic-ci-openshift-multiarch-master-nightly-4.9-upgrade-from-nightly-4.8-ocp-remote-libvirt-s390x",
 		"periodic-ci-openshift-release-master-ci-4.9-upgrade-from-stable-4.8-e2e-aws-uwm",
 		"periodic-ci-openshift-release-master-ci-4.9-upgrade-from-stable-4.8-e2e-gcp-ovn-upgrade",
-		"periodic-ci-openshift-release-master-nightly-4.10-e2e-aws-ovn-local-gateway",
-		"periodic-ci-openshift-release-master-nightly-4.10-e2e-azure-fips",
-		"periodic-ci-openshift-release-master-nightly-4.10-e2e-gcp-fips",
+		"periodic-ci-openshift-release-master-ci-4.10-upgrade-from-stable-4.9-e2e-gcp-ovn-upgrade",
```
deads2k, line 153: isn't this used for payload promotion?

pkg/testgridanalysis/testidentification/ocp_variants.go
```diff
@@ -123,6 +119,7 @@ var (
 
 		"periodic-ci-openshift-release-master-ci-4.9-e2e-azure-upgrade",
 		"periodic-ci-openshift-release-master-ci-4.10-e2e-azure-upgrade",
+		"periodic-ci-openshift-release-master-ci-4.10-upgrade-from-stable-4.9-e2e-azure-upgrade",
```
deads2k, line 122: isn't this used for payload promotion?

pkg/testgridanalysis/testidentification/ocp_variants.go
stbenjam: Only through aggregation.  The jobs themselves have dismal overall pass rates, they belong in never-stable. 

never-stable doesn't mean we forget about them, but it does make it easier for us to find new issues.  It may not make it easier for you to find problems, but it does for me.

PR #341, opened 2022-01-04
Sync data to postgres based on timestamps, not artificial ID's

stbenjam:
BigQuery doesn't support auto-incrementing primary keys. I thought that
using ROW_NUMBER OVER() using deterministic ordering would mimic that,
but it appears it's not working reliably.

This changes the fetch from bigquery to be based on the most recent
timestamp of a release tag.

[TRT-78](https://issues.redhat.com/browse/TRT-78)

### DIFF
--- a/main.go
+++ b/main.go
@@ -225,17 +225,17 @@ func (o *Options) Run() error {
 		testgridhelpers.DownloadData(dashboards, o.JobFilter, o.FetchData)
 
 		if o.DSN != "" && os.Getenv("GOOGLE_APPLICATION_CREDENTIALS") != "" {
-			bge, err := bigqueryexporter.New(context.Background())
-			if err != nil {
-				return err
-			}
-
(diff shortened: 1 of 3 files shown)


stbenjam:
@dgoodwin Could you take a look at this? This should fix the immediate issue of us missing releases from Sippy.

I think I also want to normalize the data and add the gorm time stamps, too but I'll save that for a separate PR.  That'll require us dropping the ephemeral postgres in production and resyncing everything from bigquery.

stbenjam:
/hold

Hm, if we have an arm64 release from 11:52, and an amd64 release from 11:50, we could push the arm64 one to bigquery first while the amd64 is still running tests. In that case, we'll never get the amd64 tag locally.

I probably want "created at" time stamps on the big query tables and use that instead of the release tag time.

dgoodwin (commented):
Looks good just one question.

PR #343, opened 2022-01-21
trim unnecessary tests

deads2k:
the `.Overall` tests are simply summaries of the data we're looking at, so they don't provide unique value to us for viewing.  It fills the "most failing test", with entries we cannot investigate.

### DIFF
--- a/pkg/testgridanalysis/testreportconversion/test_report.go
+++ b/pkg/testgridanalysis/testreportconversion/test_report.go
@@ -3,6 +3,7 @@ package testreportconversion
 import (
 	"fmt"
 	"sort"
+	"strings"
 	"time"
 
 	bugsv1 "github.com/openshift/sippy/pkg/apis/bugs/v1"
@@ -33,6 +34,8 @@ func PrepareTestReport(
 	// allJobResults holds all the job results with all the test results.  It contains complete frequency information and
 	allJobResults := ConvertRawJobResultsToProcessedJobResults(
 		reportName, rawData, bugCache, bugzillaRelease, variantManager)
+
+	removeUninterestingTests(allJobResults)
 
 	stats := calculateJobResultStatistics(allJobResults)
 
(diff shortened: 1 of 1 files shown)


pkg/testgridanalysis/testreportconversion/test_report.go
```diff
@@ -105,6 +108,19 @@ func PrepareTestReport(
...
+		result := results[i]
+		for j := len(result.TestResults) - 1; j >= 0; j-- {
+			testResult := result.TestResults[j]
+			switch {
+			case strings.HasSuffix(testResult.Name, ".Overall"): // .Overall is a suffix for a test that doesn't add value
+				result.TestResults = append(result.TestResults[:j], result.TestResults[j+1:]...)
+			}
+		}
+	}
+}
```
stbenjam, line 122: The increase in processing time you mentioned on Slack is because of for loop creep like this. Can you move this into one of the many other for loops over allJobResults?

https://github.com/openshift/sippy/pull/344 fixes npm deps.

deads2k: > The increase in processing time you mentioned on Slack is because of for loop creep like this. Can you move this into one of the many other for loops over allJobResults?

This logically has to happen before the testresults are built.  And the function above that builds the results seems responsible for building something with accuracy.  So we can do it
1. after we have the unfitered data and before we use the results (that's this PR) or 
2. we can do it in every place  iterating over it here 
3. or we can build a required ordering to the processing

I'm open to other suggestions, but this seems to be the correct spot in the pipeline.

stbenjam: It’s up to you, but you pointed out the length it takes to start up and prime the initial data and this is one of the reasons why. Have you benchmarked how long it takes before and after this with multiple releases? 

> And the function above that builds the results seems responsible for building something with accuracy

If it’s not used anywhere else I don’t see why this matters.

deads2k: > It’s up to you, but you pointed out the length it takes to start up and prime the initial data and this is one of the reasons why. Have you benchmarked how long it takes before and after this with multiple releases? There’s literally dozens of deep, nested for loops in the report generation. Each one may only take a few seconds but it adds up quickly.

The startup time isn't an issue for me.  I don't find that 2-3 minutes of startup slows me down overall.  Maybe two restarts per change and there's always something else to look at in the meantime.

> If it’s not used anywhere else I don’t see why this matters.

I don't know where else the methods could be used.  Pre-filtering may constrain future choice, so I prefer to separate the conversion itself from the filtering before building this particular report.

dgoodwin: I did a test and I can't detect a perf hit with this PR.

/hold cancel

For the 2-3 min startup, I do not believe that is healthy or sustainable, it's slowing us down and preventing us from solving problems, and we plan to keep working on getting this fixed. Hope you're back in time for Feb Hack week @stbenjam :)

deads2k:
the linter is wrong.  The change is bad because future cases will not read well.  But I am a slave to the machine.

deads2k:
reapplying lgtm

dgoodwin:
The linter strictness on this project is vastly out of proportion to the actual quality of the code.

PR #346, opened 2022-01-25
*: Add 4.11 to Sippy

wking:
Like 1b71c2d44789f (#283), but for 4.11.  In preparation for the pending branch cut: openshift/release#25557

Diverging from #283, I've left `openshiftJobsNeverStableForVariants` unpopulated for 4.11.  Filling those 4.11 entries in seems like a useful time for an audit to confirm that the entries still have open bugs that weren't fixed in 4.10.


dgoodwin:
/lgtm

Very real possibility this will take sippy down due to additional RAM so we gotta watch that.

PR #347, opened 2022-01-26
Add a few missing setup test for metal ipi

xueqzhan:
The missing tests in customJobSetupContainers cause sippy to fail to identify some setup test properly. This might result in incorrect report on install health chart.


deads2k:
setup is equivalent to "install"?  Mind adding a link to the description showing where it calls openshift-install?

/lgtm
/approve
/hold

holding for that link showing the call to openshift-install. Once you've added it, use `/hold cancel` to release this PR.

xueqzhan:
> setup is equivalent to "install"? Mind adding a link to the description showing where it calls openshift-install?
> 
> /lgtm /approve /hold
> 
> holding for that link showing the call to openshift-install. Once you've added it, use `/hold cancel` to release this PR.

"[sig-sippy] install should work" is a synthetic test within sippy, not part of the openshift-install. Sippy is using the setup status deduced while processing testGrid result. That includes checking the above list [here](https://github.com/openshift/sippy/blob/ec9d620b398d4522423d45ba9f5e427878792e51/pkg/testgridanalysis/testgridconversion/to_raw_data.go#L205). And then in CreateSyntheticTests, the status of this synthetic test [is set based on that value](https://github.com/openshift/sippy/blob/ec9d620b398d4522423d45ba9f5e427878792e51/pkg/testgridanalysis/testgridconversion/ocp_synthetic_tests.go#L92).

deads2k:
This is correct because that script calls the installer https://github.com/openshift/release/blob/be8c4512b32e6a5125d972046aaf755ce0507b7a/ci-operator/step-registry/baremetalds/devscripts/setup/baremetalds-devscripts-setup-commands.sh#L100

/hold cancel

Fixing naming and comments to reflect intent better instead of my original choice of mechanism (we used to call the container or pod "Setup"), will likely help future you.

PR #350, opened 2022-01-28
Add Pass Rate By Variant to test analysis page.

dgoodwin:
This will allow us to see trends by variant instead of just by job
previously.

### DIFF
--- a/sippy-ng/src/tests/TestAnalysis.js
+++ b/sippy-ng/src/tests/TestAnalysis.js
@@ -105,7 +105,27 @@ export function TestAnalysis(props) {
     return (100 * (1 - failures / runs)).toFixed(2)
   }
 
-  const chart = {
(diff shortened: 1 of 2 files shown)


sippy-ng/src/tests/TestAnalysis.js
```diff
@@ -204,8 +268,33 @@ export function TestAnalysis(props) {
   })
 
   let index = 0
+  variants.forEach((variant) => {
+    byVariantChart.datasets.push({
```
xueqzhan, line 272: Can this be put into the same loop on line 258?

dgoodwin: it can, good idea. Updated.

dgoodwin:
Updated with unit test fix.

PR #351, opened 2022-01-31
Add all bare metal jobs to never stable due to eol for centos8

xueqzhan:


deads2k:
somewhere to start

/lgtm

PR #358, opened 2022-03-02
Fix upstream kube job processing.

dgoodwin:
We updated for testgrid change in Oct from displaying "Overall" to what
we thought was "jobname.Overall". This is generally consistent but it
turns out prow jobs can specify a testgrid-tab-name annotation, which
changes the test name to that value.Overall, causing sippy to think it
has mass infrastructure failures.

Fix by just looking for suffix ".Overall", this looks consistent across
all our tests.


stbenjam:
/test unit

Well that's weird, the tests panicked, but I can't reproduce it locally.

PR #359, opened 2022-03-03
Use new install test to identify install step

xueqzhan:


deads2k:
/hold

let's get a sense of pass/fail first.  I have a prediction that things previously considered infrastructure will fail here. I also suspect that with proper help and explanations, we can likely get changes to the installer to have certain types of errors produce different exit codes.

pkg/testgridanalysis/testidentification/test_identification.go
```diff
@@ -66,25 +68,47 @@ var customJobInstallNames = sets.NewString(
...
+	totalJobRunWithNewInstallIdentificationMetric = promauto.NewCounter(prometheus.CounterOpts{
+		Name: "sippy_total_runs_with_new_install_identification",
+		Help: "Total job runs with new style install identification",
+	})
+)
+
 // TODO We should instead try to detect whether we fail in a pre-step to determine whether install succeeded
 // Install steps have different names in different jobs. This is heavily dependent on the actual UPI jobs, but they turn out to be different.
 // When this needs updating,  it shows up as installs timing out in weird numbers
 func IsInstallStepEquivalent(testName string) bool {
```
deads2k, line 88: Where do we compute the infrastructure failures?  I think we should add the install failures on infra to the infra failed percentage.

We should also add the bootstrapping failure and infrastructure failure to the list of interesting install tests on the install details page.  I think that vsphere is failing bootstrapping a lot (as a for instance).

dgoodwin:
/hold cancel

Let's see what this shows.

dgoodwin:
/hold

Sorry something looks off.

dgoodwin:
Yeah I think this is broken, the IsInstallStepEquivalent is run when parsing raw data, this is only done in the fetchdata container which does not expose metrics (only the sippy container does), so the data would not appear in the metrics endpoint. This likely is only since the db changeover, I suspect it would have shown up previously.

Additionally it looks like we duplicate the increments every refresh for any job we saw before that's still in the testgrid data. I think this means the numbers aren't useful other than to indicate we saw one more than the other.

I think a better way to do this would be to:
1. see if you can view the info direct from sippy UI, [I'm not sure it tells us enough](https://sippy.dptools.openshift.org/sippy-ng/tests/4.11?filters=%257B%2522items%2522%253A%255B%257B%2522columnField%2522%253A%2522name%2522%252C%2522operatorValue%2522%253A%2522contains%2522%252C%2522value%2522%253A%2522cluster%2520install.install%2520should%2520succeed%2522%257D%252C%257B%2522columnField%2522%253A%2522name%2522%252C%2522operatorValue%2522%253A%2522contains%2522%252C%2522value%2522%253A%2522%255Bsig-sippy%255D%2520install%2520should%2520work%2522%257D%255D%252C%2522linkOperator%2522%253A%2522or%2522%257D)
2. If not ^^, then you could go to server.go refreshMetricsDB, and tie in a new db query to count total job runs, and total job runs with your new tests. I would break it down by release as well so we get a clear picture.

PR #361, opened 2022-03-14
Snapshot 4.10 at GA for historical sippy

stbenjam:


bparees:
@stbenjam add the new release to the args on the resource too, please:
https://github.com/openshift/sippy/blob/fa159e17ed4ad6f1bb3ccc3c3b56e746b676cade/resources/deploymentconfig-historical.yaml#L31-L42

(it's not auto deployed, but at least it keeps a copy of what the DC needs to look like)

stbenjam:
> @stbenjam add the new release to the args on the resource too, please:

It's there, it's just at the very end of the diff

bparees:
> It's there, it's just at the very end of the diff

ah, sorry.  I had more confidence in the github ui to sort that to the top than apparently was warranted

/lgtm

PR #362, opened 2022-03-15
Add a new 'assisted' variant

omertuc:
See commit message

### DIFF
--- a/pkg/testgridanalysis/testidentification/ocp_variants.go
+++ b/pkg/testgridanalysis/testidentification/ocp_variants.go
@@ -10,12 +10,13 @@ import (
 
 var (
 	// variant regexes
-	arm64Regex   = regexp.MustCompile(`(?i)-arm64`)
-	awsRegex     = regexp.MustCompile(`(?i)-aws`)
-	azureRegex   = regexp.MustCompile(`(?i)-azure`)
-	compactRegex = regexp.MustCompile(`(?i)-compact`)
-	fipsRegex    = regexp.MustCompile(`(?i)-fips`)
-	metalRegex   = regexp.MustCompile(`(?i)-metal`)
+	arm64Regex    = regexp.MustCompile(`(?i)-arm64`)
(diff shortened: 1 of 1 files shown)


eranco74 (commented):
/lgtm

omertuc:
We've decided to look for alternative ways to get Sippy to work as we expect, closing this for now

stbenjam:
> We've decided to look for alternative ways to get Sippy to work as we expect, closing this for now

Is there any reason you don't want a variant for assisted? This change looked correct to me.

omertuc:
> > We've decided to look for alternative ways to get Sippy to work as we expect, closing this for now
> 
> Is there any reason you don't want a variant for assisted? This change looked correct to me.

It's not that I don't want, it's just that the original goal was to get *all* assisted jobs into sippy (so we could get a nice UI / overview) but we later realized sippy only consumes informing testgrids and 95% of our jobs are not informing, so we decided we should probably create our own instance of Sippy pointing at non-informing test grids for assisted.

The change still makes sense though, assisted is a better variant than "metal-assisted", so if it's fine by you we can merge this

dgoodwin:
Sippy's about to change pretty dramatically so I would not recommend setting up a new instance just now, but what kind of data are we missing? I thought we got all periodics.

omertuc:
>Sippy's about to change pretty dramatically so I would not recommend setting up a new instance just now, 

Thanks for the heads up. 

> but what kind of data are we missing? I thought we got all periodics.

We have a lot of assisted related periodic jobs that are not part of "release-controller", i.e. non-informing periodics. We want to track them with sippy. You can see them at `#assisted-deployment-ci`, but one arbitrary example `periodic-ci-openshift-assisted-service-master-e2e-metal-assisted-operator-ztp-periodic-upgrade`

dgoodwin:
Ok let me check with the team this morning, but I think we can probably find a way to get these into sippy for you.

PR #363, opened 2022-03-15
Sippy Backed by Postgresql

dgoodwin:
This PR introduces a major change to sippy where we will no longer load testgrid files from disk and do analysis storing all reports in memory, and instead build out a postgresql db with raw job and test results, and query it for reporting. This change is being undertaken to allow us to bypass the 8-10 minute startup times, the growing memory concerns and crashes, allow us to track trends over more than 2 weeks, and implement several new features having better access to data.

In this PR the new postgresql work is presented as an alternative server mode, launched with --db-only-mode. We plan to run a parallel instance of this while the work is completed, so we can compare db mode vs in-memory mode and make sure it's correct. Once we're confident we will remove the in-memory mode, and all legacy sippy endpoints/UI pages.

The database is populated by a new command to be run in the fetch-data script, ex:

```
$ sippy --release 4.7 --local-data /opt/sippy-testdata --load-database -v 5 --database-dsn="postgresql://postgres:password@localhost:5432/postgres"
```

Cross test queries are too slow to be done live, these use materialized views refreshed hourly.

Remaining: roughly 6 API endpoints still need to be ported. The UI won't load yet due to some of these, but the api endpoints are live for a number of calls.

Work has been evolving over months of hack days and just recently a more dedicated effort. For this reason git merge has been used to keep in sync with master over a long lived branch, and I would strongly advocate for this over trying to rebase. Commits should be reasonably clean and show the development of the feature over months.

May be some sub-optimal use of gorm here, falling back to raw queries etc, as my knowledge of it has evolved since starting and still is.

When comparing data, you will sometimes see small discrepancies between new and old API endpoints, usually just off by a couple results, but in some cases we discovered that the existing implementation is wildly off base. (specifically the test reporting has some unexpected assumptions given we did not track successful test results)


pkg/api/tests.go
```diff
@@ -132,3 +154,129 @@ func PrintTestsJSON(release string, w http.ResponseWriter, req *http.Request, cu
...
+SELECT *,
+       current_successes * 100.0 / NULLIF(current_runs, 0) AS current_pass_percentage,
+       current_failures * 100.0 / NULLIF(current_runs, 0) AS current_failure_percentage,
+       previous_successes * 100.0 / NULLIF(previous_runs, 0) AS previous_pass_percentage,
+       previous_failures * 100.0 / NULLIF(previous_runs, 0) AS previous_failure_percentage,
+       (current_successes * 100.0 / NULLIF(current_runs, 0)) - (previous_successes * 100.0 / NULLIF(previous_runs, 0)) AS net_improvement
+FROM results;
+`
+	if period == "twoDay" {
+		q = strings.ReplaceAll(q, "prow_test_report_7d_matview", "prow_test_report_2d_matview")
```
stbenjam, line 222: Can we make table/view names constants?

pkg/api/tests.go
```diff
@@ -132,3 +154,129 @@ func PrintTestsJSON(release string, w http.ResponseWriter, req *http.Request, cu
...
+		RespondWithJSON(http.StatusInternalServerError, w, map[string]interface{}{"code": http.StatusInternalServerError, "message": "Error building job report:" + err.Error()})
+		return
+	}
+
+	RespondWithJSON(http.StatusOK, w, testsResult.
+		sort(req).
+		limit(req))
+	RespondWithJSON(http.StatusOK, w, tests.
+		sort(req).
+		limit(req))
```
stbenjam, line 190: We should only be rendering once here? The logs show an error:

```
2022/03/16 10:02:50 http: superfluous response.WriteHeader call from github.com/openshift/sippy/pkg/api.RespondWithJSON (api.go:335)
```

pkg/api/tests.go
```diff
@@ -132,3 +154,129 @@ func PrintTestsJSON(release string, w http.ResponseWriter, req *http.Request, cu
...
+		if err := json.Unmarshal([]byte(queryFilter), filter); err != nil {
+			RespondWithJSON(http.StatusBadRequest, w, map[string]interface{}{"code": http.StatusBadRequest, "message": "Could not marshal query:" + err.Error()})
+			return
+		}
+	}
+
+	// If requesting a two day report, we make the comparison between the last
+	// period (typically 7 days) and the last two days.
+	period := req.URL.Query().Get("period")
+	if period != "current" && period != "twoDay" {
```
stbenjam, line 174: The UI sends `default` I think, not `current`.  It should probably allow a blank period, too, to mean default.

stbenjam (commented):
A few comments.  Looks like I broke lint, and array column filtering isn't working (fixed in https://github.com/dgoodwin/sippy/pull/2)

Since this is all optional I think we can land this soon and iterate. It'll also let us split up some of the remaining work.

pkg/api/tests.go
dgoodwin: Did you have specifics in mind? Were you thinking prow_job_test_report_... ?

Here's the full list today:

```
postgres=# \d+
                                                     List of relations
 Schema |                   Name                    |       Type        |  Owner   | Persistence |    Size    | Description 
--------+-------------------------------------------+-------------------+----------+-------------+------------+-------------
 public | bug_jobs                                  | table             | postgres | permanent   | 8192 bytes | 
 public | bug_tests                                 | table             | postgres | permanent   | 40 kB      | 
 public | bugs                                      | table             | postgres | permanent   | 96 kB      | 
 public | bugs_id_seq                               | sequence          | postgres | permanent   | 8192 bytes | 
 public | job_runs                                  | table             | postgres | permanent   | 8192 bytes | 
 public | job_runs_id_seq                           | sequence          | postgres | permanent   | 8192 bytes | 
 public | prow_job_run_failed_tests                 | table             | postgres | permanent   | 0 bytes    | 
 public | prow_job_run_tests                        | table             | postgres | permanent   | 756 MB     | 
 public | prow_job_run_tests_id_seq                 | sequence          | postgres | permanent   | 8192 bytes | 
 public | prow_job_runs                             | table             | postgres | permanent   | 3144 kB    | 
 public | prow_job_runs_id_seq                      | sequence          | postgres | permanent   | 8192 bytes | 
 public | prow_jobs                                 | table             | postgres | permanent   | 120 kB     | 
 public | prow_jobs_id_seq                          | sequence          | postgres | permanent   | 8192 bytes | 
 public | prow_test_analysis_by_job_14d_matview     | materialized view | postgres | permanent   | 665 MB     | 
 public | prow_test_analysis_by_variant_14d_matview | materialized view | postgres | permanent   | 201 MB     | 
 public | prow_test_report_2d_matview               | materialized view | postgres | permanent   | 37 MB      | 
 public | prow_test_report_7d_matview               | materialized view | postgres | permanent   | 37 MB      | 
 public | pull_requests                             | table             | postgres | permanent   | 8192 bytes | 
 public | pull_requests_id_seq                      | sequence          | postgres | permanent   | 8192 bytes | 
 public | release_tags                              | table             | postgres | permanent   | 8192 bytes | 
 public | release_tags_id_seq                       | sequence          | postgres | permanent   | 8192 bytes | 
 public | repositories                              | table             | postgres | permanent   | 8192 bytes | 
 public | repositories_id_seq                       | sequence          | postgres | permanent   | 8192 bytes | 
 public | suites                                    | table             | postgres | permanent   | 16 kB      | 
 public | suites_id_seq                             | sequence          | postgres | permanent   | 8192 bytes | 
 public | tests                                     | table             | postgres | permanent   | 1640 kB    | 
 public | tests_id_seq                              | sequence          | postgres | permanent   | 8192 bytes | 
(27 rows)

```

As an aside, is it worth trying to rename job_runs which is presently used for the release payload jobs?

stbenjam:
Do variants for jobs get updated in the DB? I was specifically thinking of what happens when a job gets removed or added to never-stable?

pkg/api/tests.go
dgoodwin: Whoops!

pkg/api/tests.go
dgoodwin: Hmm weird, I see a couple refs to 'default' in the UI code, but I don't recall ever seeing that respected in the api server side code. I'll try to respect that broadly throughout with some util methods.

dgoodwin:
I think not, I've got a comment in here somewhere that that likely isn't working, I will go get that sorted before we attempt a merge.

pkg/api/tests.go
dgoodwin: I may have missed or removed period entirely in the jobs report, gonna fix that as well to respect it the same as above. I see the old jobs code did if twoDay, else default, so that is probably how it was working without ever calling out "default".

pkg/api/tests.go
dgoodwin: I totally misread, will do!

dgoodwin:
PR updated with these and a couple other fixes.

pkg/sippyserver/dbloader.go
```diff
@@ -13,115 +21,380 @@ import (
...
+		tokens := strings.Split(jobRun.JobRunURL, "/")
+		prowID, _ := strconv.ParseUint(tokens[len(tokens)-1], 10, 64)
+
+		if _, ok := prowJobRunCache[uint(prowID)]; ok {
+			// skip job runs we already have:
+			continue
+		}
+
+		// TODO: copy whatever's happening in jobresults.go
+		// knownFailure := jobRun.Failed && areAllFailuresKnown(jrr, testResults)
```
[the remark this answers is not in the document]
stbenjam: > I'd kinda like to do away with this concept entirely

I am good with that, the new UI doesn't show this information anyway.  The old UI does, though.

PR #364, opened 2022-03-21
Move from bigquery to postgres for release controller data

stbenjam:
This removes the bigqueryexporter, and adds the logic for syncing
release controller data to Sippy for storage directly in postgres.

Example command to sync release controller data:

```
$ ./sippy
    --release 4.11
    --local-data ../sippy-data \
    --load-database \
    -v 5 \
    --database-dsn="postgresql://postgres:password@localhost:5432/postgres" \
    --architecture amd64 \
    --architecture arm64 \
    --architecture s390x
```


main.go
```diff
@@ -95,6 +96,7 @@ func main() {
 	flags.StringVar(&opt.LocalData, "local-data", opt.LocalData, "Path to testgrid data from local disk")
 	flags.StringVar(&opt.DSN, "database-dsn", os.Getenv("SIPPY_DATABASE_DSN"), "Database DSN for storage of some types of data")
 	flags.StringArrayVar(&opt.OpenshiftReleases, "release", opt.OpenshiftReleases, "Which releases to analyze (one per arg instance)")
+	flags.StringArrayVar(&opt.OpenshiftArchitectures, "architecture", opt.OpenshiftArchitectures, "Which architectures to analyze (one per arg instance)")
```
dgoodwin, line 99: Nit: I'd suggest --arch for the arg just for brevity.

Should this be plugged into the fetchdata script?

pkg/releasesync/releasesync.go
```diff
@@ -0,0 +1,232 @@
...
+		architectures: architectures,
+		httpClient:    &http.Client{Timeout: 60 * time.Second},
+	}
+
+	return o.Run()
+}
+
+func (r *releaseSyncOptions) Run() error {
+	for _, release := range r.releases {
+		fmt.Fprintf(os.Stderr, "Fetching release %s from release controller...\n", release)
```
dgoodwin, line 39: Suggest klog for consistency. 

I'd like to replace them all with logrus someday.

pkg/releasesync/releasesync.go
```diff
@@ -0,0 +1,232 @@
...
+		allTags := r.fetchReleaseTags(release)
+
+		for _, tags := range allTags {
+			for _, tag := range tags.Tags {
+				c := int64(0)
+				r.db.DB.Table("release_tags").Where("\"releaseTag\" = ?", tag.Name).Count(&c)
```
dgoodwin, line 45: Are those escaped quotes required?

stbenjam: Camel case column names needs to be quoted for postgres, I should just fix this to use snake case to match the other tables.

dgoodwin:
/lgtm

We'll have to wipe out the old tables in the official db once this is in, as I launched it this morning and it populated the old.

PR #370, opened 2022-03-23
Job analysis page from postgres db

stbenjam:


pkg/api/jobs.go
```diff
@@ -418,3 +361,57 @@ func PrintPerfscaleWorkloadMetricsReport(w http.ResponseWriter, req *http.Reques
 	RespondWithJSON(http.StatusOK, w, filteredScaleJobs)
 
 }
+
+func jobResultsFromDB(req *http.Request, dbc *gorm.DB, release string) (*gorm.DB, error) {
```
dgoodwin, line 365: We're about to collide hard here with the PR I just posted. Going to approve this one and then merge into mine.

PR #372, opened 2022-03-23
Add canary endpoint to DB-backed sippy

stbenjam:
The /canary endpoint is moved under /api, and retrieves the list of
tests with a pass rate above 99% for the most recent 7 day reporting
period. This list is consumed when updating the e2e-aws-canary job.


stbenjam:
FYI @bparees

PR #373, opened 2022-03-25
Port /api/health to db.

dgoodwin:
Skipped two portions of this that appear unused, the two day statistics,
and the variant health (number of variants whose jobs fully pass over
80%, 60%, or less (fail)).

Refactored filtering out to it's own package due to cyclical imports.

Trending towards moving DB query functions to the db package.

Refactored job report query to be more usable without an http request.

### DIFF
--- a/pkg/api/health.go
+++ b/pkg/api/health.go
@@ -1,20 +1,46 @@
 package api
 
 import (
+	"math"
 	"net/http"
+	"strings"
 	"time"
 
+	"github.com/montanaflynn/stats"
+	apitype "github.com/openshift/sippy/pkg/apis/api"
 	sippyv1 "github.com/openshift/sippy/pkg/apis/sippy/v1"
 	sippyprocessingv1 "github.com/openshift/sippy/pkg/apis/sippyprocessing/v1"
+	"github.com/openshift/sippy/pkg/db"
+	"github.com/openshift/sippy/pkg/db/query"
+	"github.com/openshift/sippy/pkg/filter"
+	"github.com/openshift/sippy/pkg/testgridanalysis/testgridanalysisapi"
+	"github.com/openshift/sippy/pkg/testgridanalysis/testreportconversion"
+	"k8s.io/klog"
 )
+
+type indicator struct {
+	Current  sippyv1.PassRate `json:"current"`
+	Previous sippyv1.PassRate `json:"previous"`
+}
+
+type variants struct {
+	Current  sippyprocessingv1.VariantHealth `json:"current"`
+	Previous sippyprocessingv1.VariantHealth `json:"previous"`
+}
+
+type health struct {
+	Indicators  map[string]indicator         `json:"indicators"`
+	Variants    variants                     `json:"variants"`
+	LastUpdated time.Time                    `json:"last_updated"`
+	Promotions  map[string]time.Time         `json:"promotions"`
+	Warnings    []string                     `json:"warnings"`
+	Current     sippyprocessingv1.Statistics `json:"current_statistics"`
+	Previous    sippyprocessingv1.Statistics `json:"previous_statistics"`
+}
 
 // PrintOverallReleaseHealth gives a summarized status of the overall health, including
 // infrastructure, install, upgrade, and variant success rates.
 func PrintOverallReleaseHealth(w http.ResponseWriter, curr, twoDay, prev sippyprocessingv1.TestReport) {
-	type indicator struct {
-		Current  sippyv1.PassRate `json:"current"`
-		Previous sippyv1.PassRate `json:"previous"`
-	}
 	indicators := make(map[string]indicator)
 
 	// Infrastructure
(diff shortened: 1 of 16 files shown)


dgoodwin:
The overview page now loads, but a few panels still missing.

dgoodwin:
Very dicey merge, we both did similar thing in diff ways, I kept both, converted some, but the use in job analysis API had me a little worried, but it *looks* like it's working to me locally. I'll try to fix that now and get down to one solution.

dgoodwin:
Nope, I broke it, two versions for now.

pkg/api/health.go
```diff
@@ -129,8 +139,184 @@ func PrintOverallReleaseHealth(w http.ResponseWriter, curr, twoDay, prev sippypr
...
+				// NOTE: looks unused in ReleaseOverview and of questionable use in general. This field would show
+				// the number of variants whose jobs fully pass over 80% of the time, 60% of the time, or less than that.
+				// skipping for now.
+					Variants: variants{
+						Current:  curr.TopLevelIndicators.Variant,
+						Previous: prev.TopLevelIndicators.Variant,
+					},
+
+			    // TODO: do we have need of these anymore with the new installer tests coming from junit?
+					Warnings: append(curr.AnalysisWarnings, prev.AnalysisWarnings...),
```
stbenjam, line 234: That's just one warning, but there are some others. For example we show one when the last 3 promotion jobs failed.  We need to figure out how to generate and display warnings from the DB-side of things, but I guess it can be done separately. Additional warnings will need to include things like OpenShift/RHCOS release mismatch (there's a card for that one)

pkg/api/job_analysis.go
```diff
@@ -11,6 +11,7 @@ import (
 	"github.com/openshift/sippy/pkg/db"
 
 	v1sippyprocessing "github.com/openshift/sippy/pkg/apis/sippyprocessing/v1"
+	filter2 "github.com/openshift/sippy/pkg/filter"
```
stbenjam, line 14: Can we just make this `filter`? Looks like they all got renamed to filter2.

pkg/api/job_runs.go
```diff
@@ -10,6 +10,7 @@ import (
 	apitype "github.com/openshift/sippy/pkg/apis/api"
 	v1sippyprocessing "github.com/openshift/sippy/pkg/apis/sippyprocessing/v1"
 	"github.com/openshift/sippy/pkg/db"
+	filter2 "github.com/openshift/sippy/pkg/filter"
```
stbenjam, line 13: Ditto

pkg/api/health.go
```diff
@@ -129,8 +139,184 @@ func PrintOverallReleaseHealth(w http.ResponseWriter, curr, twoDay, prev sippypr
...
+		return indicator{}, err
+	}
+	currentPassRate := sippyv1.PassRate{
+		Percentage: testReport[0].CurrentPassPercentage,
+		Runs:       testReport[0].CurrentRuns,
+	}
+	previousPassRate := sippyv1.PassRate{
+		Percentage: testReport[0].PreviousPassPercentage,
+		Runs:       testReport[0].PreviousRuns,
+	}
```
stbenjam, line 252: This needs some bound check, it panics on me when the DB is empty.

```
TestReportExcludeVariants completed in 554.385µs with 0 results from db
2022/03/28 07:20:00 http: panic serving 127.0.0.1:54248: runtime error: index out of range [0] with length 0
```

pkg/api/health.go
```diff
@@ -129,8 +139,184 @@ func PrintOverallReleaseHealth(w http.ResponseWriter, curr, twoDay, prev sippypr
...
+	currPercentages := []float64{}
+	prevPercentages := []float64{}
+	currStats.Histogram = make([]int, 10)
+	prevStats.Histogram = make([]int, 10)
+
+	/* I don't think this is necessary...
+	sort.Slice(results, func(i, j int) bool {
+		return results[i].CurrentPassPercentage > results[j].CurrentPassPercentage
+	})
+	*/
```
stbenjam, line 269: I'd just go ahead and remove this.  It doesn't seem to be needed.

stbenjam (commented):
A few comments below. We'll need to figure out what to do with the data we're getting from  the old `/json` endpoint -- the variants summary widget uses that one.  I think there's a few other pages that make use of that endpoint, too.

And golint needs fixing

dgoodwin:
Updated for review, as well as a fix for the mis-reporting 100% install success. Unable to squash commits due to the use of the merge, on account of the conflict that was resolved.

dgoodwin:
@stbenjam should i delete everything in the prow_ tables and rebuild so we get install history for at least as long as testgrid has? Right now it would only start getting correct going forward, whereas if we rebuild we'd get at least a few days/weeks.

stbenjam:
This looks good to me. 

/lgtm

PR #374, opened 2022-03-28
Include test failures in api job run result

stbenjam:
I missed this last week when fixing this API.


dgoodwin:
/lgtm

bot got a problem today?

PR #379, opened 2022-03-30
Syncing jobs direct from prow

stbenjam:
Sync presubmit data direct from prow, and sync junit data from GCS bucket.

Example usage:

```
$ ./sippy -v 4 --load-periodics=false --load-presubmits --load-database --local-data ../sippy-data --release 4.11 --database-dsn="postgresql://postgres:password@localhost:5432/postgres" --skip-bug-lookup --google-service-account-credential-file ~/Downloads/openshift-ci-data-analysis-1b68cb387203.json
```

There are some additional todo's , I'll handle as follow-ups
- Add a formal JSON output to Prow deck?
- Handle creating sippy synthetic tests from junits


main.go
```diff
@@ -23,11 +23,13 @@ import (
 	"github.com/openshift/sippy/pkg/buganalysis"
 	"github.com/openshift/sippy/pkg/db"
 	"github.com/openshift/sippy/pkg/perfscaleanalysis"
+	prowanalysis "github.com/openshift/sippy/pkg/prowloader"
```
dgoodwin, line 26: Could we do either a prowanalysis as actual package name, or use prowloader as the import name? May as well name it what we'd like to call it.

main.go
```diff
@@ -116,10 +122,16 @@ func main() {
 	flags.BoolVar(&opt.Server, "server", opt.Server, "Run in web server mode (serve reports over http)")
 	flags.BoolVar(&opt.DBOnlyMode, "db-only-mode", opt.DBOnlyMode, "Run web server off data in postgresql instead of in-memory")
 	flags.BoolVar(&opt.SkipBugLookup, "skip-bug-lookup", opt.SkipBugLookup, "Do not attempt to find bugs that match test/job failures")
+	flags.BoolVar(&opt.LoadPresubmits, "load-presubmits", opt.LoadPresubmits, "Fetch presubmit job and job run data from prow")
```
dgoodwin, line 125: Suggest --load-prow-presubmits

main.go
```diff
@@ -116,10 +122,16 @@ func main() {
...
+	flags.BoolVar(&opt.LoadPeriodics, "load-periodics", true, "Fetch periodic job and job run data from testgrid")
```
dgoodwin, line 126: Suggest --load-testgrid-periodics

main.go
```diff
@@ -319,7 +333,18 @@ func (o *Options) Run() error {
 		elapsed := time.Since(start)
 		klog.Infof("Database loaded in: %s", elapsed)
 
-		return err
+		gcsClient, err := gcs.NewGCSClient(context.TODO(),
```
dgoodwin, line 336: Suggest moving this into the o.LoadPresubmits block.

pkg/api/jobs.go
```diff
@@ -408,10 +408,10 @@ func jobResultsFromDB(req *http.Request, dbc *gorm.DB, release string) (*gorm.DB
 		}
 	} else if req.URL.Query().Get("period") == periodTwoDay {
 		// twoDay report period starts 9 days ago, (comparing last 2 days vs previous 7)
-		start = time.Now().Add(-9 * 24 * time.Hour)
+		start = time.Now().UTC().Add(-9 * 24 * time.Hour)
```
dgoodwin, line 411: Nice, ty for fixing these!

pkg/apis/junit/types.go
```diff
@@ -0,0 +1,110 @@
...
+// field for test cases is omitted, as this concept does not directly apply to Go.
+// For XML specifications see http://help.catchsoftware.com/display/ET/JUnit+Format
+// or view the XSD included in this package as 'junit.xsd'
+
+// TestSuites represents a flat collection of jUnit test suites.
+type TestSuites struct {
+	XMLName xml.Name `xml:"testsuites"`
+
+	// Suites are the jUnit test suites held in this collection
+	Suites []*TestSuite `xml:"testsuite"`
```
dgoodwin, line 18: If we switch wholesale over to this approach, I would look forward to proper suite support. There's really nothing I could find that I can do in testgrid to reliably determine suite names other than hardcoding the ones we know about and hoping nobody adds more without us knowing beforehand. "."'s are in use in all kinds of test names.

pkg/prowloader/prow.go
```diff
@@ -0,0 +1,292 @@
...
+		if _, ok := prowJobCache[j.Name]; !ok {
+			prowJobCache[j.Name] = j
+		}
+	}
+	klog.V(4).Infof("job cache created with %d entries from database", len(prowJobCache))
+	return prowJobCache
+}
+
+// Cache the IDs of all known ProwJobRuns for this job. Will be used to skip job run and test
+// results we've already processed.
```
dgoodwin, line 75: Comment says "for this job" but looks like we're loading all ProwJobRun IDs. Going per job might be a good idea at some point to help with memory.

pkg/prowloader/prow.go
```diff
@@ -0,0 +1,292 @@
...
+
+func fetchJobJSON(uri string) ([]byte, error) {
+	resp, err := http.Get(uri)
+	if err != nil {
+		return nil, err
+	}
+	return ioutil.ReadAll(resp.Body)
+}
+
+func jobJSONToProwJob(jobJSON []byte) ([]prow.ProwJob, error) {
```
dgoodwin, line 125: Suggest "ToProwJobs" to clarify this is loading lots of jobs not just one.

pkg/prowloader/prow.go
```diff
@@ -0,0 +1,292 @@
...
+
+	return prowJobRunCache
+}
+
+func (pl *ProwLoader) LoadProwJobsToDB() error {
+	allowedJobRegex := []*regexp.Regexp{
+		regexp.MustCompile(`pull-ci-openshift-.*-(master|main)-e2e-.*`), // For now let's just get master/main presubmits in the openshift org
+	}
+
+	jobJSON, err := fetchJobJSON(prowURL)
```
dgoodwin, line 93: Suggest jobsJSON

pkg/prowloader/prow.go
```diff
@@ -0,0 +1,292 @@
...
+				}
+				break
+			}
+		}
+	}
+
+	return nil
+}
+
+func fetchJobJSON(uri string) ([]byte, error) {
```
dgoodwin, line 117: Suggest fetchJobsJSON

pkg/prowloader/prow.go
```diff
@@ -0,0 +1,292 @@
...
+		for _, re := range allowedJobRegex {
+			if re.MatchString(pj.Spec.Job) {
+				err := pl.prowJobToJobRun(pj)
+				if err != nil {
+					return err
+				}
+				break
+			}
+		}
+	}
```
dgoodwin, line 112: Could you give a test with this running and watching the res memory use?

pkg/prowloader/prow.go
```diff
@@ -0,0 +1,292 @@
...
+}
+
+func (pl *ProwLoader) findOrAddSuite(name string) *uint {
+	if name == "" {
+		return nil
+	}
+
+	if id, ok := pl.suiteCache[name]; !ok {
+		suite := &models.Suite{}
+		pl.dbc.DB.Where("name = ?", name).Find(&suite)
```
dgoodwin, line 232: How many suites did you end up with in your db? Curious how many are out there.

dgoodwin (commented):
Looking awesome.

pkg/prowloader/prow.go
stbenjam: 387, but there are a lot with random names like `cluster_events_c0cb8a41-2c0b-4146-a000-36d4f7bfbeb5`

main.go
stbenjam: Yea this is just an artifact of me not being able to decide on a name. I'm still in the process of trying to organize the code so ignore names and paths for now :)

pkg/prowloader/prow.go
stbenjam: There are some spikes of up to 1G:

rss

stbenjam:
@dgoodwin This should be ready for another review.

I still need to handle synthetic tests, but will do that separately.

dgoodwin (commented):
Just one nit and then we can lgtm and iterate.

dbmigration/20220517200748_prow_job_run_test_duration.sql
```diff
@@ -0,0 +1,9 @@
+-- +goose Up
+-- +goose StatementBegin
+ALTER TABLE prow_job_run_tests ADD COLUMN duration double precision;
+-- +goose StatementEnd
+
+-- +goose Down
+-- +goose StatementBegin
+ALTER TABLE prow_job_run_tests DROP COLUMN duration double precision;
```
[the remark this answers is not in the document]
stbenjam: Oh good catch! Fixed

PR #380, opened 2022-03-31
Load only one testgrid job at a time for memory use.

dgoodwin:
Previously we processed all testgrid job data for a dashboard, then
loaded each into the database. This change adjusts so we read one job at
a time, load it into the db, and then move on to the next.

In local testing this cuts memory use from about 2GB down to about
550MB.


stbenjam:
Does it take longer now?

PR #382, opened 2022-04-06
Add cleanTestName and remove random operator names

neisw:
This is to address [TRT-174](https://issues.redhat.com/browse/TRT-174) where test names with random characters create an ever growing list of test names.  Checking for known pattern and replacing the random bits with 'random'.

### DIFF
--- a/pkg/testgridanalysis/testgridconversion/to_raw_data.go
+++ b/pkg/testgridanalysis/testgridconversion/to_raw_data.go
@@ -116,20 +116,20 @@ func isOverallTest(testName string) bool {
 	return testName == overall || strings.HasSuffix(testName, ".Overall")
 }
 
-//Specific set of test names that include random characters in them
-//if they match the known start then look to replace the random chars with 'random'
+// specific set of test names that include random characters in them
+// if they match the known start then look to replace the random chars with 'random'
 var matchRandomStart = regexp.MustCompile("^\"Installing \"Red Hat Integration")
 var matchRandom = regexp.MustCompile("operator in test-[a-z]+")
 var matchRandomReplace = "operator in test-random"
 
 func removeRandomOperatorTestNames(testName string) string {
-	//not necessary but narrowing the scope we apply the replace all to
-	//by verifyting the name starts with our known case
+	// not necessary but narrowing the scope we apply the replace all to
+	// by verifyting the name starts with our known case
 	match := matchRandomStart.MatchString(testName)
 	if match {
 		return matchRandom.ReplaceAllString(testName, matchRandomReplace)
 	}
-	//essentially an else
+	// essentially an else
 	return testName
 }
 
@@ -146,20 +146,20 @@ func fixOldStyleTestNames(testName string) string {
 	return testName
 }
 
-//function to group together calls for cleansing the test names as needed
+// cleanTestName groups together calls for cleansing the test names as needed
 func cleanTestName(testName string) string {
-	//initialize and then pass through each cleaner function
+	// initialize and then pass through each cleaner function
 	cleanedName := testName
 
-	//is this an old test name?
-	//we may not need this check any longer based on the comments around the check / fix
+	// is this an old test name?
+	// we may not need this check any longer based on the comments around the check / fix
 	cleanedName = fixOldStyleTestNames(cleanedName)
 
-	//look for the specific case related to red hat integration operator in test random names
+	// look for the specific case related to red hat integration operator in test random names
 	cleanedName = removeRandomOperatorTestNames(cleanedName)
 
-	//if there are more conditions to check for we can add them as we find them
-	//...
+	// if there are more conditions to check for we can add them as we find them
+	// ...
 
 	return cleanedName
 }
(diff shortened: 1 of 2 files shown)


pkg/testgridanalysis/testgridconversion/to_raw_data.go
```diff
@@ -116,6 +116,54 @@ func isOverallTest(testName string) bool {
 	return testName == overall || strings.HasSuffix(testName, ".Overall")
 }
 
+//Specific set of test names that include random characters in them
```
DennisPeriquet, line 119: I prefer a space after the `//` and other code I've seen has the space too.

pkg/testgridanalysis/testgridconversion/to_raw_data.go
```diff
@@ -116,6 +116,54 @@ func isOverallTest(testName string) bool {
...
+	// we have to keep it to interpret historical results from 4.6.
+	if testidentification.IsOldInstallOperatorTest(testName) || testidentification.IsOldUpgradeOperatorTest(testName) {
+		operatorName := testidentification.GetOperatorNameFromTest(testName)
+		testName = testgridanalysisapi.OperatorFinalHealthPrefix + " " + operatorName
+	}
+
+	return testName
+}
+
+//function to group together calls for cleansing the test names as needed
```
DennisPeriquet, line 149: The convention is that the comment for a function begins with the name of the function.  So you can phrase it like: `// cleanTestName groups together calls for ...`

pkg/testgridanalysis/testgridconversion/to_raw_data.go
```diff
@@ -116,6 +116,54 @@ func isOverallTest(testName string) bool {
 	return testName == overall || strings.HasSuffix(testName, ".Overall")
 }
 
+// specific set of test names that include random characters in them
+// if they match the known start then look to replace the random chars with 'random'
+var matchRandomStart = regexp.MustCompile("^\"Installing \"Red Hat Integration")
+var matchRandom = regexp.MustCompile("operator in test-[a-z]+")
+var matchRandomReplace = "operator in test-random"
```
dgoodwin, line 123: Could you flip this to "operator in test namespace", that should be slightly more descriptive for the real test name.

pkg/testgridanalysis/testgridconversion/to_raw_data_test.go
```diff
@@ -0,0 +1,318 @@
...
+	nonMatchString := "\"Installing \"Red Hat Integration - 3scale\" operator in test-.Installing \"Red Hat Integration - 3scale\" operator in test- Installs Red Hat Integration - 3scale operator in test- and creates 3scale Backend Schema operand instance\""
+	testStrings = append(testStrings, nonMatchString)
+
+	// will not trigger the first match for the starts with
+	skipString := "\"Doesn'tStartWith Installing \"Red Hat Integration - 3scale\" operator in test-.Installing \"Red Hat Integration - 3scale\" operator in test- Installs Red Hat Integration - 3scale operator in test- and creates 3scale Backend Schema operand instance\""
+	testStrings = append(testStrings, skipString)
+
+	fixedNames := 0
+
+	for _, s := range testStrings {
```
dgoodwin, line 83: There's a very common sub-test pattern used in go here that's probably worth playing with this PR. It would replace the testStrings below (can probably cut down on those to just a handful), as well as these additional testStrings you're adding. You define a temporary struct that represents a test case, including a name, the inputs, and the expected output. Then you define an array of those, iterate them, and use go's sub-test functionality to turn each into a separate sub-test output. Here's a good example: https://github.com/openshift/origin/blob/master/pkg/synthetictests/duplicated_events_test.go#L13. This blog explains a little too: https://medium.com/@sebdah/introduction-to-go-sub-tests-5fca2b9f6983

dgoodwin: This will make it a little easier to see inputs and expected output when looking over what's tested.

neisw: Got it, thanks for the pointer.  Updated the tests

pkg/testgridanalysis/testgridconversion/to_raw_data.go
neisw: definitely. done.

pkg/testgridanalysis/testgridconversion/to_raw_data_test.go
```diff
@@ -0,0 +1,137 @@
...
+			vercount: 3,
+		},
+		{
+			name:     "generic 4",
+			input:    "\"Installing \"Red Hat Integration - 3scale\" operator in test-piiov.Installing \"Red Hat Integration - 3scale\" operator in test-piiov Installs Red Hat Integration - 3scale operator in test-piiov and creates 3scale Backend Schema operand instance\"",
+			vercount: 3,
+		},
+		{
+			name:     "skip 1",
+			input:    "\"Doesn'tStartWith Installing \"Red Hat Integration - 3scale\" operator in test-.Installing \"Red Hat Integration - 3scale\" operator in test- Installs Red Hat Integration - 3scale operator in test- and creates 3scale Backend Schema operand instance\"",
```
dgoodwin, line 72: Looking great, is it worth doing an output comparison here as well to make sure we didn't modify anything?
Does vercount provide value if we're comparing output strings?

dgoodwin (commented):
I'd say this looks good otherwise, thanks!

PR #384, opened 2022-04-07
Disable bugzilla syncing.

dgoodwin:
We're spending 20 minutes every hour doing this due to the increase in
number of tests. We'll be moving to Jira soon regardless, so no point
investing in this right now.


dgoodwin:
/hold

Want to see what #383 reveals before we disable.

dgoodwin:
Linked PR shows about 220 search.ci queries for bugs, each taking an avg of maybe 6-8s, which totals up to 20-30 minutes. Disabling.
/hold cancel

PR #385, opened 2022-04-08
port upgrade api

dgoodwin:
- Port /api/upgrade to postgres.
- Simplify /api/install and fill in Overall + All.


dgoodwin:
I simplified here a bit after fighting with each, and now just show the test name without any massaging. Maybe not as pretty, but much more clear what we're looking at, and simpler safer code.

PR #386, opened 2022-04-12
Remove some unused secondary UI tabs on Install/Upgrade/Tests pages.

dgoodwin:
We may bring some of these back at some point, but for now they are
unweildy, non-functional, and we suspect largely unused.

### DIFF
--- a/sippy-ng/src/tests/Tests.js
+++ b/sippy-ng/src/tests/Tests.js
@@ -70,7 +70,6 @@ export default function Tests(props) {
               <Route path={path + '/details'}>
                 <TestDetails release={props.release} />
               </Route>
-
               <Route exact path={path}>
                 <Container size="xl">
                   <TestTable release={props.release} />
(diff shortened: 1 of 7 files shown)


sippy-ng/src/tests/Tests.js
```diff
@@ -57,20 +57,10 @@ export default function Tests(props) {
                     component={Link}
                     to={url}
                   />
-                  <Tab
-                    label="Tests by variant"
-                    value="details"
-                    component={Link}
-                    to={url + '/details'}
-                  />
```
stbenjam, line 65: I'm not sure I want to remove this page. It may be a little buggy, but the view isn't shown anywhere else this way, e.g.:

e..g, https://psql-sippy.dptools.openshift.org/sippy-ng/tests/4.11/details?test=remains%20available

stbenjam:
I think we can remove "Jobs by variant" which relies on the legacy /json API: https://psql-sippy.dptools.openshift.org/sippy-ng/jobs/4.11/variant

sippy-ng/src/tests/Tests.js
dgoodwin: This was the one that worried me the most, it's thousands and thousands of rows and seemed almost unusable. It did load however, so if you think it's useful I can restore.

stbenjam: The above link loads almost instantly for me -- with a filter I think the page is pretty usable

dgoodwin:
Updated for those three changes, thanks!

PR #389, opened 2022-04-22
Fix running jobs getting recorded in db.

dgoodwin:
If a job was running when we first saw it, it would be recorded as a
failure and never updated. Running jobs should be skipped.

Our job runs are totally hosed in the db and we will have to rebuild. 111474 of 115769 got recorded like this. Basically everything after the first import.


dgoodwin:
/assign @stbenjam

This one sucks. :(

PR #392, opened 2022-04-23
Add postgres API for variant cards

stbenjam:
This restores the variant cards on psql-based sippy. This data was previously retrieved from a legacy `/json/` API.  Note, the API format has changed so this will not work on legacy sippy, but I think we're close to removing all that code anyway.

image

This also removes the JS snapshot-based tests (functional js tests are still here).  I think in the ~6 months I've had this, it's only caused pain and hasn't detected much useful bugs.  These were generated off the historical data and I don't have a good idea yet of how to do that consistently with the DB data.  I hope that's ok.


pkg/api/jobs.go
```diff
@@ -147,6 +148,71 @@ func PrintJobsReport(w http.ResponseWriter, req *http.Request, currReport, twoDa
...
+	} else if req.URL.Query().Get("period") == periodTwoDay {
+		// twoDay report period starts 9 days ago, (comparing last 2 days vs previous 7)
+		start = time.Now().Add(-9 * 24 * time.Hour)
+	} else {
+		// Default start to 14 days ago
+		start = time.Now().Add(-14 * 24 * time.Hour)
+	}
+
+	// TODO: currently we're assuming dates use the 00:00:00, is it more logical to add 23:23 for boundary and end? or
+	// for callers to know to specify one day beyond.
```
dgoodwin, line 177: Tough call, my first instinct is for us to assume 23:59. If I make that API call with and end date, I'd expect those results from that day included.

dgoodwin (commented):
/lgtm

pkg/api/jobs.go
stbenjam: This is your comment, FWIW, this is just copy and pasted

dgoodwin: LOL I am old.

stbenjam: It was a good comment :-) Here's an attempt to fix it by adding +24 hours (and move this handling code to it's own function)

https://github.com/openshift/sippy/pull/393/files#diff-f12f97ebaac66a4fd3a647fe92524c5de1bca1f72340d2829891d118516195f7R35

PR #393, opened 2022-04-25
Move start/boundary/end param handler to util function

stbenjam:
Move this to a function we can use in multiple API's, and also adds +24 hours to boundary and end params to make them inclusive of the entire day.

### DIFF
--- a/pkg/api/health.go
+++ b/pkg/api/health.go
@@ -6,6 +6,8 @@ import (
 	"time"
 
 	"github.com/montanaflynn/stats"
+	"k8s.io/klog"
+
 	apitype "github.com/openshift/sippy/pkg/apis/api"
 	sippyv1 "github.com/openshift/sippy/pkg/apis/sippy/v1"
 	sippyprocessingv1 "github.com/openshift/sippy/pkg/apis/sippyprocessing/v1"
@@ -14,7 +16,6 @@ import (
 	"github.com/openshift/sippy/pkg/filter"
 	"github.com/openshift/sippy/pkg/testgridanalysis/testgridanalysisapi"
 	"github.com/openshift/sippy/pkg/testgridanalysis/testreportconversion"
-	"k8s.io/klog"
 )
 
 type indicator struct {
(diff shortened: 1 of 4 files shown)


stbenjam:
@dgoodwin  The other option is to specify a full ISO8601 timestamp, which would be consistent in how the UI passes other time parameters and avoid this problem all together.

dgoodwin:
> > also adds +24 hours to boundary and end params to make them inclusive of the entire day
> 
> @dgoodwin The other option is to specify a full ISO8601 timestamp, which would be consistent in how the UI passes other time parameters and avoid this problem all together.

Hmm I was not aware of that precedent, maybe that's the better option then? I like the CLI curlability of YYYY-MM-DD but I really doubt that's gonna be used very often.

stbenjam:
> > > also adds +24 hours to boundary and end params to make them inclusive of the entire day
> > 
> > 
> > @dgoodwin The other option is to specify a full ISO8601 timestamp, which would be consistent in how the UI passes other time parameters and avoid this problem all together.
> 
> Hmm I was not aware of that precedent, maybe that's the better option then? I like the CLI curlability of YYYY-MM-DD but I really doubt that's gonna be used very often.

I mispoke, it's actually sending the unix time, e.g. [this page](https://psql-sippy.dptools.openshift.org/sippy-ng/jobs/4.11/analysis?filters=%7B%22items%22%3A%5B%7B%22columnField%22%3A%22name%22%2C%22operatorValue%22%3A%22equals%22%2C%22value%22%3A%22periodic-ci-openshift-release-master-nightly-4.11-e2e-metal-ipi-serial-ovn-dualstack%22%7D%2C%7B%22columnField%22%3A%22timestamp%22%2C%22operatorValue%22%3A%22%3E%22%2C%22value%22%3A%221650802365268%22%7D%2C%7B%22columnField%22%3A%22timestamp%22%2C%22operatorValue%22%3A%22%3C%3D%22%2C%22value%22%3A%221650888765268%22%7D%5D%7D&period=hour) 

It's probably easiest if we keep the same format everywhere, so we can reuse the same UI components

stbenjam:
That makes sense, let's hold off on changing anything for now. We can revisit if/when we add a UI for start/end/boundary

/hold

stbenjam:
Will revisit later

PR #395, opened 2022-04-25
Add all openshift-verification 4.11 jobs to never stable.

dgoodwin:


stbenjam:
/lgtm
/hold

Will remove hold once I'm done with DB, I don't want it to start over on importing data when it restarts

PR #399, opened 2022-04-26
logrus

dgoodwin:
- Switch from klog to logrus.
- Add --log-level CLI argument
- Log basic info on each request and elapsed time.

This offers structured and contextual logging, timestamps, formatting, and well defined levels instead of arbitrary numbers.
Ported some to use structured logging but not all in the interest of time.

New format looks like:

```
INFO[2022-04-26T14:58:35.705-03:00] TestReportsByVariant completed in 341.352301ms with 26 results from db 
INFO[2022-04-26T14:58:35.705-03:00] responded to request                          elapsed=341.62761ms method=GET uri="/api/tests/details?release=4.10&test=Kubernetes%20APIs%20remain%20available"
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -11,6 +11,7 @@ require (
 	github.com/montanaflynn/stats v0.6.6
 	github.com/pkg/errors v0.9.1
 	github.com/prometheus/client_golang v1.11.0
+	github.com/sirupsen/logrus v1.8.1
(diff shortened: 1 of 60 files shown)


stbenjam:
This is beautiful.  I tried it locally and looked at the changes and I think it's all reasonable.

/lgtm

PR #400, opened 2022-04-26
Allow filtering test results by variant

stbenjam:
- Make filtering on tests all by SQL
- Allow filtering test results by variant
- Exclude never-stable/techpreview on dashboard views


dgoodwin (commented):
This is fantastic.

/lgtm

PR #404, opened 2022-04-27
[WIP] Exclude aggregated jobs from variants

stbenjam:
Jobs that make up an aggregated run are already tracked in Sippy. We are
injecting unwanted hysteresis that will cause changes to pass rates to
lag behind. The effect is probably small, but I think it's worth being
correct.

Let's take a simple example:

The follow test runs 10x and passes 7.

    [sig-pizza] Pizza should be delivered within 30 minutes

It has a pass rate of 70%, in line with historical results.

If it is aggregated, we have an 11th entry in Sippy of a pass.  Sippy
actually thinks the test passed 8 times out of 11 -- a pass rate of
72.7%.


dgoodwin:
Scary but I think you're correct, it looks like this is the best play.

/lgtm

dgoodwin:
Still has WIP and looks to be in need of conflicts.

PR #409, opened 2022-04-28
Show working percentage instead of pass percentage for tests

stbenjam:
Based on a discussion from Slack about the difference between showing passes only vs "working" (passes + flakes) for tests.  I'm not entirely convinced this is the best view long term, but I could believe it's the right thing to do right now.


pkg/api/tests.go
```diff
@@ -239,9 +239,16 @@ func BuildTestsResults(dbc *db.DB, release, period string, fil *filter.Filter) (
...
 				current_flakes * 100.0 / NULLIF(current_runs, 0) AS current_flake_percentage,
+				(current_successes + current_flakes) * 100.0 / NULLIF(current_runs, 0) AS current_working_percentage,
+
 				previous_successes * 100.0 / NULLIF(previous_runs, 0) AS previous_pass_percentage,
 				previous_failures * 100.0 / NULLIF(previous_runs, 0) AS previous_failure_percentage,
 				previous_flakes * 100.0 / NULLIF(previous_runs, 0) AS previous_flake_percentage,
+				(previous_successes + previous_flakes) * 100.0 / NULLIF(previous_runs, 0) AS previous_working_percentage,
+
+				(current_failures * 100.0 / NULLIF(current_runs, 0)) - (previous_failures * 100.0 / NULLIF(previous_runs, 0)) AS net_failure_improvement,
+				(current_flakes * 100.0 / NULLIF(current_runs, 0)) - (previous_flakes * 100.0 / NULLIF(previous_runs, 0)) AS net_flake_improvement,
```
dgoodwin, line 250: Using _improvement for these two above looks wrong to me, if current fail or flake percentage is 70%, and it used to be 50% previously, we're reporting 20% growth as an "improvement" when things are actually getting worse. Maybe just need to make it previous - current instead of current - previous?

sippy-ng/src/releases/ReleaseOverview.js
```diff
@@ -28,12 +28,10 @@ import VariantCards from '../jobs/VariantCards'
...
 export const REGRESSED_TOOLTIP =
   'Shows the most regressed items this week vs. last week, for those with more than 10 runs, excluding never-stable and techpreview.'
-export const NOBUG_TOOLTIP =
-  'Shows the list of tests ordered by least successful and without a bug, for those with more than 10 runs'
-export const TRT_TOOLTIP =
-  'Shows a curated list of tests selected by the TRT team'
 export const TWODAY_WARNING =
   'Shows the last 2 days compared to the last 7 days, sorted by most regressed, excluding never-stable and techpreview.'
+export const TOP_FAILERS_TOOLTIP =
+  'Shows the list of tests ordered by least working percentage.'
```
dgoodwin, line 34: least working = failing? I like the use of working meaning pass + flake, but for least working I am thinking "failing" is clearer. Let me know if that sounds wrong though.

pkg/api/tests.go
stbenjam: Makes sense, I flipped them

sippy-ng/src/releases/ReleaseOverview.js
stbenjam: Make sense, I used "failing"

PR #411, opened 2022-04-29
Use custom serializer for use-query-params

stbenjam:
A follow-up to ensure square brackets are quoted in JSON encoded
URI params.

### DIFF
--- a/sippy-ng/src/helpers.js
+++ b/sippy-ng/src/helpers.js
@@ -2,6 +2,19 @@
 // and not supported in all browsers, and it's not in node yet.
 import React from 'react'
 
+export const SafeJSONParam = {
+  encode: (j) => {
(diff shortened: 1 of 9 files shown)


sippy-ng/src/helpers.js
```diff
@@ -2,6 +2,19 @@
...
+    return safeEncodeURIComponent(JSON.stringify(j))
+  },
+  decode: (j) => {
+    try {
+      return JSON.parse(decodeURIComponent(j))
+    } catch (e) {
+      // return undefined
+    }
+  },
+}
```
stbenjam, line 16: Basically the same as the typescript implementation, except using our encoder https://github.com/pbeshai/serialize-query-params/blob/de5792ba6c3c0ae1c6ad5b586fb7143981e39c15/src/serialize.ts#L273-L313

PR #414, opened 2022-05-03
Enable autocomplete for filters

stbenjam:
Add autocomplete on tests, jobs, and variants. Part of making the test pages more useful.

autocomplete


dgoodwin:
I can't break it. This is really cool, I'm quite surprised it's possible to do in a timely fashion.

/lgtm

PR #417, opened 2022-05-04
trt-217: refresh metrics from db

neisw:
per trt-217, when dbOnlyMode is true added new method to refresh metrics via db query with minimal refactoring

### DIFF
--- a/pkg/api/jobs.go
+++ b/pkg/api/jobs.go
@@ -28,13 +28,14 @@ import (
 type jobsAPIResult []apitype.Job
 
 const periodTwoDay = "twoDay"
+const currentPassPercentage = "current_pass_percentage"
 
 func (jobs jobsAPIResult) sort(req *http.Request) jobsAPIResult {
 	sortField := req.URL.Query().Get("sortField")
 	sort := apitype.Sort(req.URL.Query().Get("sort"))
 
 	if sortField == "" {
-		sortField = "current_pass_percentage"
+		sortField = currentPassPercentage
 	}
 
 	if sort == "" {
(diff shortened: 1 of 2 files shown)


pkg/api/jobs.go
```diff
@@ -295,6 +296,41 @@ func PrintJobsReportFromDB(w http.ResponseWriter, req *http.Request,
...
+
+	var start time.Time
+	var boundary time.Time
+	var end time.Time
+
+	var filterOpts = filter.FilterOptions{}
+	filterD := filter.Filter{}
+	filterOpts.Filter = &filterD
+	filterOpts.SortField = currentPassPercentage
+	filterOpts.Sort = apitype.SortDescending
```
dgoodwin, line 309: No need to sort here as we're just publishing so prom can scrape, is the filter required for any reason or can we skip it and save a few cpu cycles.

pkg/api/jobs.go
```diff
@@ -295,6 +296,41 @@ func PrintJobsReportFromDB(w http.ResponseWriter, req *http.Request,
 	RespondWithJSON(http.StatusOK, w, jobsResult)
 }
 
+func JobReportsFromDB(dbc *db.DB, release, period string) ([]apitype.Job, error) {
```
dgoodwin, line 299: This looks super close to part of PrintJobsReportFromDB, could we break that up and separate out JobReportsFromDB, and re-use it in both cases?

pkg/sippyserver/server.go
```diff
@@ -505,17 +545,11 @@ func (s *Server) jsonReleasesReport(w http.ResponseWriter, req *http.Request) {
...
-	type Release struct {
-		Release string
-	}
-
-	response := jsonResponse{}
+type Release struct {
+	Release string
+}
 
+func (s *Server) releasesFromDB() []Release {
```
dgoodwin, line 552: Could this func be moved to pkg/db/query/release_queries.go (new file)

dgoodwin (commented):
Just a couple refactors but this looks great, thanks!

pkg/api/jobs.go
neisw: I actually had to add the sort to get the scan to return results

pkg/api/jobs.go
neisw: yeah, I was on the fence with how much refactoring we wanted to do and how much of a gain it was but let me give it a go

pkg/api/jobs.go
dgoodwin: Ah thats a little odd, probably a bug in there for another day. Thanks.

pkg/sippyserver/server.go
neisw: yep

dgoodwin:
/lgtm

Thanks!

PR #418, opened 2022-05-05
Manage database schema and migrations with goose

dgoodwin:
Previously we were letting gorm do it's automigrations, which was ok for simple things but had no support for managing materialized view schema, and did have some bugs. 

This PR adds goose for db migrations in a style similar to rails. An init container will be configured once this merges to automatically run goose every time we startup the sippy pod. Initial schema was generated from the prod db with a couple hand edits to let it work reliably (certain settings being set that we don't really care about). I have tested this change both on fresh db's, and against a copy of prod db. You'll see the manual sql required for prod to "fake" that the initial schema was populated with goose, after which we can start rolling out incremental changes. (also has been tested)

We've loosely discussed a possibility of a very dynamic fluctuating set of matviews for small windows, few hours a day. If we pursue this, we'll likely have to redo the materialized view portion of this and start managing it in code again. However we're not sure yet we'll be taking that approach, and this is still safer for schema management in general, so I think it should go ahead.

For rollout, this should be fine with existing schema we just won't be automatically making any edits until I merge the init container. Once running we can do the manual db hack with the sql in this PR, then add the init container to the gitops repo, and we should be in business.


dgoodwin:
Ready for review I think.

/assign @stbenjam

DEVELOPMENT.md
```diff
@@ -0,0 +1,64 @@
...
+See the TRT team drive in Google, periodically there is a gzip'd backup stored there. Restore it locally with:
+
+```bash
+psql -h localhost -U postgres -p 5432 postgres < sippy-backup-2022-05-02.sql
+```
+
+## Launch Sippy API
+
+```bash
+go build -mod=vendor . && ./sippy --server --local-data /opt/sippy-testdata --release 4.11 -v 4 --skip-bug-lookup --database-dsn="postgresql://postgres:password@localhost:5432/postgres" --db-only-mode
```
DennisPeriquet, line 43: that `-v 4` gave me some trouble; is that a typo?

DEVELOPMENT.md
```diff
@@ -0,0 +1,64 @@
...
+## Populating Data
+
+### From Scratch
+
+Populate db schema with goose, may need to run periodically if new migrations come in, fetch data from testgrid, and load it into the db. This process can take some time for recent OpenShift releases with substantial number of jobs. Using older releases like 4.7 can be fetched and loaded in just a couple minutes.
+
+```bash
+goose --dir dbmigration postgres "user=postgres password=password dbname=postgres sslmode=disable" up
+go build -mod=vendor . && ./sippy --local-data /opt/sippy-testdata --release 4.11 --fetch-data /opt/sippy-testdata --log-level=debug
+go build -mod=vendor . && ./sippy --release 4.11 --local-data /opt/sippy-testdata --load-database --log-level=debug --database-dsn="postgresql://postgres:password@localhost:5432/postgres" --skip-bug-lookup
```
DennisPeriquet, line 29: for line 28-29, you mean that you run line 27 and then you run one of 28 or 29 right?

DEVELOPMENT.md
```diff
@@ -0,0 +1,64 @@
...
+````
+
+### From a Prod Sippy Backup
+
+See the TRT team drive in Google, periodically there is a gzip'd backup stored there. Restore it locally with:
+
+```bash
+psql -h localhost -U postgres -p 5432 postgres < sippy-backup-2022-05-02.sql
```
DennisPeriquet, line 37: can we run this from inside the container?

DEVELOPMENT.md
dgoodwin: Oops, not updated for the recent log change, --log-level=debug, fixing now, thanks!

DEVELOPMENT.md
dgoodwin: You can definitely exec into the container and run psql that way, the pipe might be slightly more tricky but perhaps it would work fine too.

DEVELOPMENT.md
dgoodwin: No you actually need all of the above. --fetch-data pulls testgrid files down to disk, and --load-database populates the db using those files.

DEVELOPMENT.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Create and Populate PostgreSQL Database
+
+Launch postgresql: 
+
+```bash
+podman run --name sippy-postgres -e POSTGRES_PASSWORD=password -p 5432:5432 -d quay.io/enterprisedb/postgresql
+```
+
+TODO: podman run postgres as a regular user may require disabling selinux on Fedora 35. Needs investigation.
```
stbenjam, line 18: Do you know what the selinux denials were? I don't seem to have that problem:

```
[stbenjam@stbenjam sippy]$ getenforce
Enforcing
[stbenjam@stbenjam sippy]$ cat /etc/redhat-release 
Fedora release 35 (Thirty Five)
```

DEVELOPMENT.md
```diff
@@ -0,0 +1,64 @@
...
+## Modifying Database Schema
+
+```bash
+goose -dir dbmigration create test
+```
+
+Edit the resulting file and add your schema to migrate "up" to, and "down" from.
+
+In the case of materialized views I believe you will need to drop the old view, recreate with the entire 
+new schema, and issue a refresh command. We should maintain the old schema in the "down" section of the migration to
```
stbenjam, line 61: Should we really issue the refresh in a migration? That sounds like the wrong thing to do to me, won't you do it multiple times during the migration process if the DB schema is out of date by several versions?

It could maybe be a final step in some hack script that migrates instead?

DEVELOPMENT.md
```diff
@@ -0,0 +1,64 @@
...
+TODO: podman run postgres as a regular user may require disabling selinux on Fedora 35. Needs investigation.
+
+## Populating Data
+
+### From Scratch
+
+Populate db schema with goose, may need to run periodically if new migrations come in, fetch data from testgrid, and load it into the db. This process can take some time for recent OpenShift releases with substantial number of jobs. Using older releases like 4.7 can be fetched and loaded in just a couple minutes.
+
+```bash
+goose --dir dbmigration postgres "user=postgres password=password dbname=postgres sslmode=disable" up
```
stbenjam, line 27: Goose supports doing this inline with embedded migrations: https://github.com/pressly/goose#embedded-sql-migrations

Is the standalone CLI command better for some reason?

DEVELOPMENT.md
dgoodwin: do you run podman as root? I've been doing it as a regular user.

May 17 08:32:07 berlinetta audit[22918]: AVC avc:  denied  { read } for  pid=22918 comm="docker-entrypoi" path="/usr/lib64/librt-2.28.so" dev="dm-0" ino=3829261 scontext=system_u:system_r:container_t:s0:c7,c741 tcontext=unconfined_u:object_r:user_home_t:s0 tclass=file permissive=0
May 17 08:32:07 berlinetta sippy-kube-postgres[22915]: /usr/bin/coreutils: error while loading shared libraries: librt.so.1: cannot change memory protections

DEVELOPMENT.md
dgoodwin: that's a good point yes, I'll try to adjust.

DEVELOPMENT.md
dgoodwin: Hmm that looks interesting. CLI does give us admin level control to work with the migrations, but we could probably break glass to do that anyhow even if we embed and manage the schema automatically.

Is it ok if I take this as an action item for a next pass? I'm in a pickle with landing the kube work which is based on this ASAP and a bit underwater with watcher stuff on top. I will file a jira.

DEVELOPMENT.md
dgoodwin: Updated guidance to NOT do this. The api will background refresh when it starts, this seems like it's probably ok to me. fetch-data can still run I believe as that's just inserting into the underlying tables.

DEVELOPMENT.md
stbenjam: Yes, always as a regular user.

stbenjam: Maybe try https://github.com/containers/podman/issues/3234#issuecomment-1043188614 ?

PR #419, opened 2022-05-10
rhcos ver check

dgoodwin:
Refactor for release health query re-use.

Display warnings on releaes overview pages if we detect an RHCOS version that doesn't seem to match the release it's in a payload for. (i.e. release 4.11 should match RHCOS version 411.86.202205030351-0)

Adds a gauge metric for the count of warnings per release. It won't tell you what's wrong, but it will tell you something is and can link you to the overview where you get the details easily.

### DIFF
--- a/pkg/api/health.go
+++ b/pkg/api/health.go
@@ -214,9 +214,7 @@ func PrintOverallReleaseHealthFromDB(w http.ResponseWriter, dbc *db.DB, release 
 	if err != nil {
 		warnings = append(warnings, fmt.Sprintf("error checking release health, see logs: %v", err))
 	} else {
-		for _, rw := range releaseWarnings {
-			warnings = append(warnings, rw)
-		}
+		warnings = append(warnings, releaseWarnings...)
 	}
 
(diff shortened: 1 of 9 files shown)


pkg/api/health.go
```diff
@@ -208,23 +209,20 @@ func PrintOverallReleaseHealthFromDB(w http.ResponseWriter, dbc *db.DB, release
 	}
 	currStats, prevStats := calculateJobResultStatistics(jobReports)
 
+	warnings := make([]string, 0)
+	releaseWarnings, err := ScanReleaseHealth(dbc, release)
+	if err != nil {
+		warnings = append(warnings, fmt.Sprintf("error checking release health, see logs: %v", err))
+	} else {
+		warnings = append(warnings, releaseWarnings...)
+	}
```
stbenjam, line 218: If we plan to have more warnings, it might be nice to extract this to it's own function so we don't start inserting a bunch of warning generation inline

stbenjam (commented):
Do we know if this works on kube sippy? I think it should return empty for releases and then it will not generate a warning but I'm not 100% sure.

PR #421, opened 2022-05-11
Open sidebar and expand first release by default.

dgoodwin:
Opening these is a very common first two clicks that are probably not
necessary and we can spare the real estate by default.

### DIFF
--- a/sippy-ng/src/App.js
+++ b/sippy-ng/src/App.js
@@ -110,7 +110,7 @@ export default function App(props) {
   const theme = useTheme()
 
   const [lastUpdated, setLastUpdated] = React.useState(null)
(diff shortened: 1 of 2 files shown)


dgoodwin:
@stbenjam what do you think about this? I almost always make these two clicks and it seems like it's worth just having it open by default.

stbenjam:
I almost never open the side bar, since I'm clicking on release page widgets to go to filtered views.  If we're going to always open the sidebar, perhaps we want the links to Tests and Jobs to have some defaults so the view is more useful (e.g. filtering never-stable, techpreview, etc)

stbenjam:
Let's land this and we can make some adjustments later

/lgtm

PR #422, opened 2022-05-11
kube sippy

dgoodwin:
Fix db load bug caused by kube jobs having 30k tests, far more than openshift does, and hitting param limits for postgres. Will now load tests for a job run in batches of 1000, but within same transaction we create the job run so we can always abort the loader and restart it without losing any job runs or partial data for the run.

Fix health api for kube mode, was querying tests that don't exist and then erroring.

Fixed server capabilities assumption that was showing payloads in kube mode.

Hid many portions of the UI sidebar when we don't have the openshift capability.

Show the correct testgrid link depending on server mode.


stbenjam:
The changes look pretty straight forward to me.

/lgtm

PR #424, opened 2022-05-11
Use new install tests for release overview page

xueqzhan:


sippy-ng/src/releases/TopLevelIndicators.js
```diff
@@ -13,6 +16,21 @@ import PropTypes from 'prop-types'
 import React, { Fragment } from 'react'
 import SummaryCard from '../components/SummaryCard'
 
+function useNewInstallTests(release) {
+  let digits = release.split('.', 2)
+  const major = parseInt(digits[0])
+  const minor = parseInt(digits[1])
```
stbenjam, line 22: I'd like to see an early return of len(digits) < 2

pkg/api/health.go
```diff
@@ -67,6 +67,72 @@ func PrintOverallReleaseHealth(w http.ResponseWriter, curr, twoDay, prev sippypr
...
+	previousPassRate = sippyv1.PassRate{
+		Percentage: passPercent,
+		Runs:       total,
+	}
+
+	indicators["installOther"] = indicator{
+		Current:  currentPassRate,
+		Previous: previousPassRate,
+	}
+
```
stbenjam, line 135: This function is only for the legacy Sippy stuff, which is going to be removed at some point.

This should all be in the `PrintOverallReleaseHealthFromDB` function below, currently the UI doesn't load with this PR for me.

stbenjam (changes requested):

sippy-ng/src/releases/TopLevelIndicators.js
```diff
@@ -59,22 +81,98 @@ export default function TopLevelIndicators(props) {
...
+      {newInstall ? (
+        <Grid item md={3} sm={6}>
+          <SummaryCard
+            key="infrastructure-summary"
+            threshold={INFRASTRUCTURE_THRESHOLDS}
+            name="Infrastructure"
+            link={
+              '/tests/' +
+              props.release +
+              '/details?test=cluster install.install should succeed: infrastructure'
```
stbenjam, line 93: These links don't work since the test details page was changed to show the NURP's.

I'd suggest linking to them like this: https://github.com/xueqzhan/sippy/commit/d949d851c92192bb04fd496305381c1d47f132cb

stbenjam:
From the sippy design meeting, we'd like to see the non-overall install gauges moved to the install page (https://sippy.dptools.openshift.org/sippy-ng/install/4.11)

xueqzhan:
Screen Shot 2022-06-03 at 11 51 34 AM
Screen Shot 2022-06-03 at 11 52 00 AM

PR #432, opened 2022-05-19
Keep sippy operational when refreshing materialized views.

dgoodwin:
Our refreshes run for about 12 minutes total, some of them taking up to
5 minutes on a particular view. During this time reads were locked and
  we noticed pages stop responding in sippy.

Postgresql allows concurrent refreshes which do not lock the table under
some conditions. This PR adds the required unique indexes on each table,
this may actually speedup the app in general.

Concurrent refresh cannot be used on the first refresh, so we always
attempt it, if it fails we fall back to the regular refresh which will
be run on first startups for developers, or when we update the view
schema in a migration.

Added a test id column to one of the matviews that was using name so we
don't have to index on huge strings.

Updated to refresh metrics before and after matview refreshes as we
think this is causing gaps that triggers alerts to clear and resurface.

On startup, we will now also only refresh if the tables look empty to avoid
expensive refreshed on developer systems.


pkg/sippyserver/server.go
```diff
@@ -223,20 +226,52 @@ func (s *Server) refreshMaterializedViews() {
 
 	for _, pmv := range db.PostgresMatViews {
 		start := time.Now()
+		tmpLog := log.WithField("matview", pmv)
+
+		// If requested, we only refresh the materialized view if it has no rows
+		if refreshMatviewOnlyIfEmpty {
+			var count *int
```
stbenjam, line 233: Shouldn't this just be an int?

main.go
```diff
@@ -386,7 +386,7 @@ func (o *Options) runServerMode() error {
 
 	// force a data refresh in the background. This is important to initially populate the db's materialized views
 	// if this is the first time starting sippy.
-	go server.RefreshData()
+	go server.RefreshData(true)
```
neisw, line 389: Just a thought that if the only time we need to refreshMetrics prior to refreshing data is on startup to avoid the scrape gap, we could do it before the RefreshData call and that way we don't do it everytime in the RefreshData call.  Also if it is only needed for dbOnlyMode (non dbOnlyMode won't have data presumably) then it could also be an explicit check and call to RefreshMetricsDB

dgoodwin:
I have updated with the int fix and Forrest's suggestion around metrics refresh.

I also merged in manually the matview change that @stbenjam did in the middle of the time since I forked, Stephen could you verify that commit 50ad938a9976a4310f0590eb28a3506bdeadfa4d contains everything in that change, as I had to eyeball that for the differences.

I then tested again both from prod backup, fresh db, and kube fresh db. Everything seems to be have as we'd expect.

main.go
```diff
@@ -384,9 +384,15 @@ func (o *Options) runServerMode() error {
 		o.DBOnlyMode,
 	)
 
+	// Initial metrics refresh to get the endpoint scrapable ASAP and prevent prom gaps, before
+	// we start the lengthy mat view refreshes.
+	if err := server.RefreshMetricsDB(); err != nil {
```
neisw, line 389: Do we need to check dbOnlyMode here or do we see that mode going away? that check is in the regular server.refreshMetrics call

stbenjam:
Looks correct to me.  Lint needs fixing, but otherwise lgtm

dgoodwin:
Sorry about that, updated now and should be good to go.

PR #434, opened 2022-05-23
Safe-encode string param for test analysis page

stbenjam:
This ensures that square brackets are encoded. While not required, it
makes the links compatible with markdown and jira.


stbenjam:
@dgoodwin Can you have a look at this? This is the last page that has a problem with brackets in URL's.

PR #439, opened 2022-05-26
Fix a crash in some situations due to missing UTC() usage.

dgoodwin:
This surfaced when I was working locally with a backup of the prod db,
not sure why it doesn't show up in prod initially but it looks like this
can crash in some cases due to the missing UTC in key.


dgoodwin:
/assign @stbenjam 

Nice easy one.

stbenjam:
I think it was my fault, I had a stray commit in my Prow PR to fix UTC() usage and looks like I missed a place.

Thanks!

PR #440, opened 2022-05-26
Fix job analysis preselecting all time most failing tests instead of last two weeks.

dgoodwin:
- Fix job analysis preselecting all time most failing tests.


dgoodwin:
Dropped to 7 days after talking with David.

PR #441, opened 2022-05-30
payload report

dgoodwin:
- Convert release job run name column from text to int and rename.
- Vendor and use gorilla/mux for routing.
- Refactor API to list payload job runs for reusability.
- Add preliminary payload stream analysis API.


dbmigration/20220527150154_payload_job_run_name_to_bigint.sql
```diff
@@ -0,0 +1,13 @@
...
+ALTER TABLE release_job_runs ALTER COLUMN name TYPE bigint USING (name::bigint);
+ALTER TABLE release_job_runs RENAME COLUMN name to prow_job_run_id
+-- +goose StatementEnd
+
+-- +goose Down
+-- +goose StatementBegin
+ALTER TABLE release_job_runs RENAME COLUMN prow_job_run_id to name
+ALTER TABLE release_job_runs ALTER COLUMN name TYPE text USING (name::text);
+UPDATE release_job_runs SET name='.' WHERE name='0';
+-- +goose StatementEnd
```
dgoodwin, line 13: This was causing a problem with some queries that casting seems to fail on. Figured it was best to get the column correct.

pkg/sippyserver/server.go
```diff
@@ -935,7 +998,10 @@ func (s *Server) Serve() {
 		serveMux.HandleFunc("/api/releases/health", s.jsonReleaseHealthReport)
 		serveMux.HandleFunc("/api/releases/tags", s.jsonReleaseTagsReport)
 		serveMux.HandleFunc("/api/releases/pull_requests", s.jsonReleasePullRequestsReport)
-		serveMux.HandleFunc("/api/releases/job_runs", s.jsonReleaseJobRunsReport)
+		serveMux.HandleFunc("/api/releases/job_runs", s.jsonListPayloadJobRuns)
+
+		serveMux.HandleFunc("/api/releases/{release}/{stream}/{arch}/analysis",
```
dgoodwin, line 1003: This could easily fit into the release health API we have today, but it's a little slow (1.5s), and the release health is called multiple times on startup when we refresh metrics, slowing down development quite a bit.

pkg/api/releases.go
```diff
@@ -43,31 +51,155 @@ func PrintPullRequestsReport(w http.ResponseWriter, req *http.Request, dbClient
...
 	}
-	q, err = filter.FilterableDBResult(q, filterOpts, nil)
+
+	res := q.Find(&jobRuns)
+	return jobRuns, res.Error
+}
+
+// GetPayloadAnalysis loads the most recent payloads  for a stream and attempts to search for most commonly
+// failing tests, possible perma-failing blockers, etc.
+func GetPayloadAnalysis(dbc *db.DB, release, stream, arch string) (*apitype.PayloadStreamAnalysis, error) {
```
dgoodwin, line 73: I'm drifting towards calling these "payloads" not release tags, as I feel like release is a bit too overloaded and payload is the term we tend to use. Let me know what you think, I'd ultimately like to rename the tables as well if this sounds good to you as well.

dgoodwin:
I have not attempted the UI much yet, focused on the API for now and will add UI support once this is more formalized.

pkg/sippyserver/server.go
stbenjam: I'm a little concerned about introducing any gorilla packages, given they've been searching for maintainers for a long time.  https://github.com/gorilla/mux/issues/659

It may not be as pretty but this is also deviating from the existing API's which use query params for just about everything.

dgoodwin: That's too bad about gorilla, I can remove.

I probably read into our discussion about moving to things like test IDs and assumed that included moving to (I think) more correct URL structures and less reliance on query params. I can change that too but at some point we should talk about improving those.

dgoodwin:
Updated to remove mux and use query params.
I am tempted to merge it into release health API, accept that it will be slower, and refactor how and when we do metrics updates to accomodate that on startup. Maybe a follow up though.

pkg/api/releases.go
```diff
@@ -43,31 +47,157 @@ func PrintPullRequestsReport(w http.ResponseWriter, req *http.Request, dbClient
...
+		consecPayloadFailsCtr := 0
+		for _, payload := range result.ConsecutiveFailedPayloads {
+			if util.StrSliceContains(v.FailedPayloads, payload) {
+				consecPayloadFailsCtr++
+			}
+		}
+		v.ConsecutiveFailedPayloadsCount = consecPayloadFailsCtr
+		// If at least the last three payloads are in rejected state, and this test is broken in all of them,
+		// flag that we may have a blocker.
+		// TODO: refine these criteria? 75% of last failures?
```
dgoodwin, line 190: Need to refine this, if the last 10 payloads failed and we're in the 3 most recent, it's still likely a payload blocker. Frankly latest 2 is probably a likely blocker. It would also be good to filter out infra/install failures and not let those prevent us calling it a blocker.

stbenjam:
/lgtm

Looks like everything for payloads keeps working in the UI with these changes, so I think this is fine to merge.

PR #442, opened 2022-05-31
Add fetchdata-prow.sh to Dockerfile

stbenjam:


stbenjam:
@dgoodwin Could you take a look? Thanks

PR #443, opened 2022-05-31
Use full path to sippy binary

stbenjam:


dgoodwin:
/lgtm

It's basically Monday for Americans, less so for the idiot Canadians who lgtm their PRs and miss things as well.

PR #447, opened 2022-06-01
Correctly handle flakes in prow import

stbenjam:
The existing logic of adding test statuses together is bugged. Make one
pass and any number of failures always make a flake.

I think this is behind [TRT-270](https://issues.redhat.com//browse/TRT-270)

### DIFF
--- a/pkg/prowloader/prow.go
+++ b/pkg/prowloader/prow.go
@@ -20,6 +20,7 @@ import (
 	"github.com/openshift/sippy/pkg/apis/junit"
 	"github.com/openshift/sippy/pkg/apis/prow"
 	sippyprocessingv1 "github.com/openshift/sippy/pkg/apis/sippyprocessing/v1"
+	v1 "github.com/openshift/sippy/pkg/apis/testgrid/v1"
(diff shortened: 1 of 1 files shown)


pkg/prowloader/prow.go
```diff
@@ -268,21 +269,23 @@ func (pl *ProwLoader) extractTestCases(suite *junit.TestSuite, testCases map[str
...
-				Status:   status,
+				Status:   int(status),
 				Duration: tc.Duration,
 			}
-		} else {
-			existing.Status += status
+		} else if (existing.Status == int(v1.TestStatusFailure) && status == v1.TestStatusSuccess) ||
+			(existing.Status == int(v1.TestStatusSuccess) && status == v1.TestStatusFailure) {
+			// One pass among failures makes this a flake
+			existing.Status = int(v1.TestStatusFlake)
```
stbenjam, line 288: Actually I think this isn't correct, because I need to consider the suite, too

stbenjam: Ah past me did consider the suite, the map key prefixes the suite name.  So this should be correct.

PR #449, opened 2022-06-01
Remove in-memory sippy impl and the legacy API

dgoodwin:
Remove in-memory sippy impl and the legacy API.
    
    PostgreSQL sippy has been in the wild for some time now and running
    well, we are ready to jettison all the old code.
    
    Unfortunately required removing the server smoke test as we can no
    longer run without a db, and we do not yet have a way to fire up a
    temporary db for a server. This gap will need to be handled separately
    but this did not catch much anyhow.
    
    Includes removal of a duplicated testidentification directory that got
    missed somehow.


main.go
```diff
@@ -226,10 +226,6 @@ func (o *Options) Validate() error {
 		return fmt.Errorf("must specify --database-dsn with --load-database")
 	}
 
-	if o.DBOnlyMode && o.DSN == "" {
-		return fmt.Errorf("must specify --database-dsn with --db-only-mode")
-	}
```
stbenjam, line 231: What do you think about having an error if DBOnlyMode is false?

main.go
```diff
@@ -226,10 +226,6 @@ func (o *Options) Validate() error {
 		return fmt.Errorf("must specify --database-dsn with --load-database")
```
stbenjam, line 226: Won't database-dsn always be required now? Maybe we should error if not specified?

main.go
dgoodwin: Yeah that should work, and still possible to edit the manifests as we'll be defaulting to true per the above recommendation.

main.go
dgoodwin: I don't think we need it for fetch data, but I can expand this to server and loaddatabase. Seem right?

dgoodwin:
Should I nuke the CLI report mode too? Does anyone use that? (running binary without --server) It looks busted anyhow, I'm going to add another commit to remove that as well.

dgoodwin:
I've made the requested edits, also found a bunch more code that could be removed by following the CLI report path and removing that as well. I think it was already broken anyhow.

stbenjam:
Checked again and I think it's all working correctly in db-only mode.

stbenjam:
@dgoodwin Tests are failing. I think because you need to rebase. Your payload pr used strconv in server.go, but this PR removed it. Once this is rebased on master it breaks

dgoodwin:
Thanks, I think I've addressed and squashed into appropriate commit.

stbenjam:
Oof, something else merged again that broke this.  Let's fix this and not merge anything else until this merges :)

dgoodwin:
Ok updated again, lets hope this is it :)

PR #450, opened 2022-06-02
It seems with RHEL 8.6 UBI8, npm has been moved to a module

stbenjam:
Sippy image building is currently broken.


stbenjam:
I *think* because this has a change to the buildroot image, the PR needs to be merged so it can be rebuilt, the presubmits aren't picking up the change.

stbenjam:
From the [CI docs](https://docs.ci.openshift.org/docs/architecture/ci-operator/#build-root-image), this is indeed the case:

```
Dockerfile will always be obtained from current HEAD of the given branch, even if ci-operator runs in the context of a PR that updates that Dockerfile.
```

/override ci/prow/build
/override ci/prow/images
/override ci/prow/lint
/override ci/prow/unit

PR #451, opened 2022-06-02
Save PR data for prow imported job runs

stbenjam:
[TRT-278](https://issues.redhat.com//browse/TRT-278)

This saves the PR data for a prow job run.  This will eventually allow
us to do analysis on how many retests  it takes to land certain PR's.
It also lets us know which org/repo is linked to a presubmit job.

### DIFF
--- a/pkg/db/models/prow.go
+++ b/pkg/db/models/prow.go
@@ -120,7 +120,10 @@ type Bug struct {
 	Jobs           []ProwJob `gorm:"many2many:bug_jobs;"`
 }
 
-// ProwPullRequest represents a GitHub pull request.
+// ProwPullRequest represents a GitHub pull request, there can be multiple entries
+// for a pull request, if it was tested with different HEADs (SHA). This lets us
+// track jobs at a more granular level, allowing us differentiate between code pushes
+// and retests.
 type ProwPullRequest struct {
 	Model
 
@@ -131,9 +134,10 @@ type ProwPullRequest struct {
 
 	Number int    `json:"number"`
 	Author string `json:"author"`
-	SHA    string `json:"sha"`
 	Title  string `json:"title,omitempty"`
 
+	// SHA is the specific commit at HEAD.
+	SHA string `json:"sha" gorm:"index:pr_link_sha,unique"`
 	// Link links to the pull request itself.
-	Link string `json:"link,omitempty" gorm:"unique"`
+	Link string `json:"link,omitempty" gorm:"index:pr_link_sha,unique"`
 }
--- a/pkg/db/models/prow.go
+++ b/pkg/db/models/prow.go
@@ -46,6 +46,7 @@ type ProwJobRun struct {
 	URL          string
 	TestFailures int
 	Tests        []ProwJobRunTest
+	PullRequests []ProwPullRequest `gorm:"many2many:prow_job_run_prow_pull_requests;"`
 	Failed       bool
 	// InfrastructureFailure is true if the job run failed, for reasons which appear to be related to test/CI infra.
 	InfrastructureFailure bool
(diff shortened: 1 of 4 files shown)


dbmigration/20220602182139_add_prow_pull_requests.sql
```diff
@@ -0,0 +1,60 @@
+-- +goose Up
+-- +goose StatementBegin
+CREATE TABLE public.prow_pull_requests (
```
stbenjam, line 3: This migration was challenging to create, I ended up using AutoMigrate from gorm to generate the schema.

pkg/db/models/prow.go
```diff
@@ -114,3 +115,21 @@ type Bug struct {
 	Tests          []Test    `gorm:"many2many:bug_tests;"`
 	Jobs           []ProwJob `gorm:"many2many:bug_jobs;"`
 }
+
+// ProwPullRequest represents a GitHub pull request.
+type ProwPullRequest struct {
```
stbenjam, line 120: Eventually I'd like to drop the release pull requests table, and have them point to this one.  This will let us click through from a release payload to see the job run history on a PR!

pkg/prowloader/prow.go
```diff
@@ -201,6 +203,39 @@ func (pl *ProwLoader) prowJobToJobRun(pj prow.ProwJob) error {
...
+			continue
+		}
+
+		pull := models.ProwPullRequest{}
+		pl.dbc.DB.Where("link = ?", pr.Link).Find(&pull)
+		pull.Org = refs.Org
+		pull.Repo = refs.Repo
+
+		pull.Link = pr.Link
+		pull.SHA = pr.SHA
```
stbenjam, line 223: We can have multiple SHA's tested, so I'm wondering if it shouldn't be unique on SHA instead of link. Or SHA + Link.

pkg/prowloader/prow.go
```diff
@@ -201,6 +203,39 @@ func (pl *ProwLoader) prowJobToJobRun(pj prow.ProwJob) error {
...
+	}
+	pulls := make([]models.ProwPullRequest, 0)
+
+	for _, pr := range refs.Pulls {
+		if pr.Link == "" {
+			continue
+		}
+
+		pull := models.ProwPullRequest{}
+		pl.dbc.DB.Where("link = ?", pr.Link).Find(&pull)
```
dgoodwin, line 218: Little bit confusing here, I don't see this ever getting created, so am I correct that we're relying on the fact that we try to find one with this link (you may want an index on link), if it exists we update it. If it doesn't exist, is this just working magically via not error checking, and getting an empty "pull" struct, then populating it, and saving it?

Neat if working but I'd go a little more explicit, could be some weirdness here if we got an error that wasn't "this doesn't exist" for example, and it might be helpful for future folks to see a little more clearly where we create vs update.

pkg/prowloader/prow.go
dgoodwin: Is SHA always going to be the last one we saw? Or do you want one entry per SHA being tested?

Do we want to track SHA at all?  I'd assume we're mostly just interested in linking to the PR itself.

pkg/db/models/prow.go
dgoodwin: +1

pkg/prowloader/prow.go
stbenjam: It saves it down below. Save updates or creates.  I did it that way in case the PR title changes, we'll end up with the latest one.

pkg/prowloader/prow.go
stbenjam: The way it's written now, yea we'll have the last SHA we saw.  

If we track it by SHA (i.e. possibly multiple entries per PR), we can differentiate between: "it took a lot of testing to get lucky on this sha" vs. "developer kept pushing changes, that's why this pr had so many test runs."

pkg/prowloader/prow.go
stbenjam: I see what you're saying though, I can fix it to be more explicit.

stbenjam:
@dgoodwin This is ready for another look

pkg/prowloader/prow.go
```diff
@@ -215,6 +217,44 @@ func (pl *ProwLoader) prowJobToJobRun(pj prow.ProwJob) error {
...
+	}
+	pulls := make([]models.ProwPullRequest, 0)
+
+	for _, pr := range refs.Pulls {
+		if pr.Link == "" {
+			continue
+		}
+
+		pull := models.ProwPullRequest{}
+		log.Infof("%+v", pr)
```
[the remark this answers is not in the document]
stbenjam: Done

PR #453, opened 2022-06-03
Disable test table name autocomplete.

dgoodwin:
As with #420, filtering on test substrings is also broken and this is a
fairly useful feature. Disabling as well until it's known how to make it
work for both use cases.


stbenjam:
Thanks!

/lgtm

PR #454, opened 2022-06-03
Hookup -start-day and --num-days for load-database, improve help.

dgoodwin:
Also drops --end-day which was can be accomplished with --num-days.

This change makes it possible to load historical data into the sippy db, even if we can't yet view the UI once done.

### DIFF
--- a/main.go
+++ b/main.go
@@ -49,7 +49,6 @@ type Options struct {
 	// TODO perhaps this could drive the synthetic tests too
 	Variants                           []string
 	StartDay                           int
-	endDay                             int
 	NumDays                            int
 	TestSuccessThreshold               float64
 	JobFilter                          string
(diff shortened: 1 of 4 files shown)


stbenjam:
The API's are hardcoded to return now - 14 days, right? Is that part of the fix for the historical pages?

dgoodwin:
/hold

Looks like --end-day is used in our manifests.

And yes a whole bunch of things are hard coded to use now, the worst being the matviews. Going to be working on that next week. This just gets so we can actually load historical data into a db, even if we can't show it yet.

dgoodwin:
Can unhold after https://github.com/openshift/continuous-release-jobs/pull/1156 merges.

dgoodwin:
/hold cancel

Linked PR has merged (well on that replaces it anyhow)

Ready for review.

PR #455, opened 2022-06-03
Improve default test table filters.

dgoodwin:
Bumps the >= 1 current runs filter up to roughly once per day for both tests and jobs on the overview, and the tests sidebar link.

Also filters out openshift-tests should work, I'm less sure about this one but looking by variant, it does fill up the results and hide the actual tests we'd go after. It's somewhat interesting, but I think more interesting is what tests are causing the overall tests should work to change it's pass rate.

### DIFF
--- a/sippy-ng/src/components/Sidebar.js
+++ b/sippy-ng/src/components/Sidebar.js
@@ -141,7 +141,7 @@ export default function Sidebar(props) {
                   to={withSort(
                     pathForTestsWithFilter(release, {
                       items: [
-                        BOOKMARKS.RUN_1,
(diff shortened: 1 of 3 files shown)


dgoodwin:
I honestly can't think of a time we've cared about a once per day, but we could go to 5 or 7 if that would be safer. I just don't have a lot of confidence in what it's telling us without a good number of runs.

I'll hit the overview widgets too, in general I'm much more interested in the frequently run stuff.

dgoodwin:
Updated to 7, the view is about identical to 10 so this works for me. 2 runs for 2 day view, and lower some of the job widget's from 10 to also 7/2.

PR #456, opened 2022-06-03
Job run matview should LEFT JOIN on failed/flakes

stbenjam:
[TRT-286](https://issues.redhat.com//browse/TRT-286)

Small tweak to the matview to LEFT JOIN instead of just plan JOIN on
flaked/failed tests.

Join alone requires an entry in both tables, meaning the mat view table
currently only has job runs where we have BOTH failure and flaked tests.
This excludes a ton of jobs, but it was noticed because we have no
install or upgrade tests which don't have flakes on those, thus they're
entirely omitted.


dgoodwin:
Nice find.

/lgtm

PR #457, opened 2022-06-06
Result should be overall result

stbenjam:
The DB columns are all named overall_result, but we were returning
`result` in the JSON.  They need to match for filtering to work,
and this is the least invasive way to fix it.  This has been broken for
a while so I don't think many people use this filtering. I doubt
we're breaking any URL's in the wild by changing this.


stbenjam:
@dgoodwin Could you have a look at this one? Thanks

PR #458, opened 2022-06-06
Show run line on job stacked chart

stbenjam:
Follow-up for [TRT-259](https://issues.redhat.com//browse/TRT-259)

This plots a second axis on the stacked chart for job run results
that shows the total count of job runs for that period.

image


deads2k:
I like the output of this change.

PR #459, opened 2022-06-07
Synthetic tests for prow-imported jobs

stbenjam:
[TRT-268](https://issues.redhat.com//browse/TRT-268)

This significantly refactors synthetic test processing, and makes it work for Prow.  This was a difficult refactor, since the original code was so tightly coupled to test grid.

- Extract synthetic test processing to be as generic as possible
- Convert Prow test suites to a RawJobRunResult (which contains OpenShift-specific stuff like operator, upgrade, install, etc)
- Creates synthetic tests based on that result
- Calculate overall job result based on the RawJobRunResult

Todo's:
- [x] Spot check prow results to confirm correctness
- [x] Confirm results are still the same for TestGrid job runs
- [x] Monitor memory usage during fetching to see impact
- [x] Improve efficiency of the conversion to RawJobRunResult (my test handling is weird from DB)


stbenjam:
@dgoodwin This should be ready for a review.  In my testing, testgrid results are similar between this PR (left) and production (right).  Small difference is due to slight difference in age of the testgrid data.

image

pkg/prowloader/testconversion/testconversion.go
```diff
@@ -0,0 +1,85 @@
...
+	testgridv1 "github.com/openshift/sippy/pkg/apis/testgrid/v1"
+	"github.com/openshift/sippy/pkg/db/models"
+	"github.com/openshift/sippy/pkg/synthetictests"
+	"github.com/openshift/sippy/pkg/testidentification"
+)
+
+func ConvertProwJobRunToSyntheticTests(pj prow.ProwJob, tests map[string]*models.ProwJobRunTest, manager synthetictests.SyntheticTestManager) (*junit.TestSuite, v1.JobOverallResult) {
+	jrr := v1.RawJobRunResult{
+		Job:       pj.Spec.Job,
+		Failed:    pj.Status.State != prow.SuccessState,
```
dgoodwin, line 16: Even if filtered elsewhere this looks a little risky, Pending and Triggered could both trigger as "failed" here. I'd go more explicit. Is aborted a failure as well? (probably yes for our purposes)

dgoodwin (commented):
This is about all I can see. I'd say we're ok to try it out.

pkg/prowloader/testconversion/testconversion.go
stbenjam: Fair enough, fixed in last commit.

PR #460, opened 2022-06-13
matviews in code again

dgoodwin:
- Fix a bug in a db down migration.
- Restore matview creation code with updated schemas.
- Update matviews if schema hash has changed.
- Refactor to a function for view management.
- Update indexes if columns changed.
- Small refactor to function names.
- Update functions if definition changed.
- Refactor to separate go files.
- Fix bug with not updating to a new hash once one exists.
- Tidy up.
- Refactor to common code for creating all types of resources.
- Add a new function missed in schema.


dgoodwin:
@stbenjam one piece remaining, what to do with the existing migrations. I can:

1. delete all views/functions from the schema migrations, leave the view only migrations as empty but still existing to prevent weirdness.
2. leave everything be as it is, views will be applied on all new db's forever. (no refresh though as there's no data at that point, so it's no real time hit)

stbenjam:
> @stbenjam one piece remaining, what to do with the existing migrations. I can:
>    delete all views/functions from the schema migrations, leave the view only migrations as empty but still existing to prevent weirdness.
>    leave everything be as it is, views will be applied on all new db's forever. (no refresh though as there's no data at that point, so it's no real time hit)

Is there a reason to continue to use goose at all? I find it very difficult to write migrations to add new columns, relations, etc. In fact for my Prow work I did a gorm automigrate, saved it's output and copied it over to a goose migration.

dgoodwin:
> > @stbenjam one piece remaining, what to do with the existing migrations. I can:
> > delete all views/functions from the schema migrations, leave the view only migrations as empty but still existing to prevent weirdness.
> > leave everything be as it is, views will be applied on all new db's forever. (no refresh though as there's no data at that point, so it's no real time hit)
> 
> Is there a reason to continue to use goose at all? I find it very difficult to write migrations to add new columns, relations, etc. In fact for my Prow work I did a gorm automigrate, saved it's output and copied it over to a goose migration.

Tough one, I find it safer to take explicit control than let goose do it's magic, I've also seen goose unable to automatically manage a tinyint column, had to change it to int, but I could be convinced otherwise. It's definitely looking less useful if we pull this out.

dgoodwin:
TRT-329 filed to remove goose and will come as a separate PR.

I think this one is ready for full review.

stbenjam:
Lint needs fixing up:

```
kg/db/functions.go:27:3: var-naming: var dropSql should be dropSQL (revive)
		dropSql := fmt.Sprintf("DROP FUNCTION IF EXISTS %s", pgFunc.Name)
		^
pkg/db/matviews.go:84:3: var-naming: var dropSql should be dropSQL (revive)
		dropSql := fmt.Sprintf("DROP MATERIALIZED VIEW IF EXISTS %s", pmv.Name) 
```

dgoodwin:
Lint updated.

stbenjam:
In testing locally on a new DB, it appears this drops all matviews and functions, is it worth it to put the hashes that are in production now in the migration?

dgoodwin:
Just found a mistake and pushed a fix.

As for the hashes, I was assuming not worth the effort, but we would eat 20 mins or so of downtime.

stbenjam:
I'm just concerned I can't easily verify this PR creates the same views and funcs that we have now.  All the tables are temporary we're touching here so it's not a big risk....:yolo:

/lgtm
/hold

I'll leave the hold so you can decide when to merge.

dgoodwin:
Yeah I took the current definitions from the db, but there's always a risk I made a mistake. I'll let it roll out tomorrow morning when I get in, less folks around for the downtime.

Thanks!

dgoodwin:
/hold cancel

I'll be watching.

PR #461, opened 2022-06-15
Adding variant for heterogeneous(multi)

deepsm007:
Based on discussion adding a new variant to support heterogeneous CI jobs in sippy

/cc @stbenjam @dgoodwin


stbenjam:
What do you think about multiarch for the variant name instead?`multi` is probably not clear to someone unfamiliar

stbenjam:
And could you add multi to https://github.com/openshift/sippy/blob/master/scripts/fetchdata.sh#L13 so we collect the release payloads?

deepsm007:
Heterogeneous is not counted as Multiarch, Multiarch is default for arm64/s390x/ppc64le. Can I name it as `hetero` or `heterogeneous` to be more distinct? 
I'll update the fetchdata scripts

stbenjam:
> Heterogeneous is not counted as Multiarch, Multiarch is default for arm64/s390x/ppc64le. Can I name it as hetero or heterogeneous to be more distinct?
> I'll update the fetchdata scripts

My main concern was multi didn't reflect that it was an architecture-related variant, heterogeneous is the same problem I think.

If the release controller is `multi`, isn't in appropriate to say these are multiarch? e.g. potential for multiple arches in one cluster.  Maybe too overloaded a term, I guess.

PR #462, opened 2022-06-15
Remove golangci-lint rpm from buildroot

stbenjam:
https://github.com/openshift/release/pull/29517 changes us to use the golangci-lint container we already have available. We're being throttled by GitHub, so we can drop the curl after that PR is merged.


stbenjam:
Ah lint runs npm lint, so not so easy to fix

PR #466, opened 2022-06-16
Fix some things in prow-based sippy

stbenjam:
- Oof, some bad handling of prow job run ID's needed serious refactoring.  Prow job runs table has duplicates :facepalm: 
- Reduced some memory usage
- Create tests in batches like we do for testgrid
- Skip running jobs - we only want jobs in a terminal state


stbenjam:
@dgoodwin This is a doozy.  I totally funked up prow job importing :facepalm:   I wan't using the prow build ID in the db record.  So many duplicates.I consistently create each run 10x.


```
sippy_prow=> select count(distinct(url)) as unique_runs, count(*) as total_runs from prow_job_runs;
 unique_runs | total_runs 
-------------+------------
       16152 |     163390
```

dgoodwin:
Ouch, but glad you found it now and not after we attempt a full switchover.

/lgtm

PR #467, opened 2022-06-16
drop goose

dgoodwin:
- Drop use of goose for sql schema management.
- Specify explicit columns in job_results function.
- Remove goose from the sippy container.

[TRT-329](https://issues.redhat.com//browse/TRT-329)


pkg/db/functions.go
```diff
@@ -96,7 +96,25 @@ WITH results AS (
...
+       current_infra_fails,
+       id,
+       created_at,
+       updated_at,
+       deleted_at,
+       name,
+       release,
+       variants,
+       test_grid_url,
+       kind,
```
dgoodwin, line 117: This was interesting, it broke on fresh db's once I switched back to gorm automigrate.

dgoodwin:
/hold 

Depends on https://github.com/openshift/continuous-release-jobs/pull/1162

Dockerfile
```diff
@@ -13,7 +12,6 @@ COPY --from=builder /go/src/sippy/scripts/fetchdata.sh /bin/fetchdata.sh
 COPY --from=builder /go/src/sippy/scripts/fetchdata-prow.sh /bin/fetchdata-prow.sh
 COPY --from=builder /go/src/sippy/scripts/fetchdata-kube.sh /bin/fetchdata-kube.sh
 COPY --from=builder /go/src/sippy/historical-data /historical-data/
-COPY --from=builder /go/bin/goose /bin/goose
 COPY ./dbmigration /usr/share/sippydbmigrations
```
stbenjam, line 15: This one needs to go too

dgoodwin: Thanks, updated.

PR #469, opened 2022-06-17
metrics

dgoodwin:
- Refactor metrics code into a separate package.
- Add metric for number of consecutive rejected payloads per stream.
- Add metric for number of hours since last accepted payload for each stream.


stbenjam:
Nice!  I tested locally and it looks correct.  Just holding in case you're not ready to merge it yet.

/pony fancy
/lgtm
/hold

dgoodwin:
Thanks! Sadly I forgot lint as usual. Needs another lgtm.

PR #470, opened 2022-06-21
Add UI mechanism for customized views

stbenjam:
[TRT-301](https://issues.redhat.com//browse/TRT-301)

This adds a "View" toolbar item, that allows customizing which columns
are shown. This lets us on the test page create working and flakes
views, for example.

sippy-views

### DIFF
--- a/pkg/api/tests.go
+++ b/pkg/api/tests.go
@@ -50,8 +50,8 @@ const (
 		previous_flakes * 100.0 / NULLIF(previous_runs, 0) AS previous_flake_percentage,
 		(previous_successes + previous_flakes) * 100.0 / NULLIF(previous_runs, 0) AS previous_working_percentage,
 
-		(previous_failures * 100.0 / NULLIF(previous_runs, 0)) - (current_failures * 100.0 / NULLIF(current_runs, 0)) AS net_failure_improvement,
-		(previous_flakes * 100.0 / NULLIF(previous_runs, 0)) - (current_flakes * 100.0 / NULLIF(current_runs, 0)) AS net_flake_improvement,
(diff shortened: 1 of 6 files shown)


stbenjam:
/hold

Pending Trevor's comments on the jira.  I also need to think about how this is wired up,  I wish things were a little more encapsulated.

pkg/api/tests.go
```diff
@@ -50,8 +50,8 @@ const (
 		previous_flakes * 100.0 / NULLIF(previous_runs, 0) AS previous_flake_percentage,
 		(previous_successes + previous_flakes) * 100.0 / NULLIF(previous_runs, 0) AS previous_working_percentage,
 
-		(previous_failures * 100.0 / NULLIF(previous_runs, 0)) - (current_failures * 100.0 / NULLIF(current_runs, 0)) AS net_failure_improvement,
-		(previous_flakes * 100.0 / NULLIF(previous_runs, 0)) - (current_flakes * 100.0 / NULLIF(current_runs, 0)) AS net_flake_improvement,
+		(current_failures * 100.0 / NULLIF(current_runs, 0) - (previous_failures * 100.0 / NULLIF(previous_runs, 0))) AS net_failure_improvement,
+		(current_flakes * 100.0 / NULLIF(current_runs, 0) - (previous_flakes * 100.0 / NULLIF(previous_runs, 0))) AS net_flake_improvement,
```
stbenjam, line 54: This was reversed, IMHO.  If a test went from a 5% flake rate to a 15% flake, we were returning "-10%" and the UI showed a down arrow.  Now it'll show a red up arrow showing flakiness increased by +10%.

Perhaps `improvement` should be `change` but I'm hesitant to change the name (it dates back all the way to legacy sippy), and it would break URL's.

stbenjam:
/hold cancel

This should be ready for a review

stbenjam:
@dgoodwin This one is ready for review

PR #472, opened 2022-06-23
Store build cluster and duration from prow job

stbenjam:
TRT-325

This stores the duration of the job, and the build cluster it was run
on, in the DB for prow sippy.


dgoodwin:
Beautiful, can't wait to see what the job rates look like across clusters, and if it's stable enough we could alert on it.

/lgtm

PR #475, opened 2022-06-27
trt-349: track forced phases

neisw:
[TRT-349](https://issues.redhat.com//browse/TRT-349): Refresh existing Payload entries when Phase status changes and calculate if the status was forced

### DIFF
--- a/pkg/releasesync/releasesync.go
+++ b/pkg/releasesync/releasesync.go
@@ -17,6 +17,12 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
+const (
+	releaseTagsTable = "release_tags"
+	succeeded        = "Succeeded"
+	failed           = "Failed"
+)
+
 type releaseSyncOptions struct {
 	db            *db.DB
 	httpClient    *http.Client
@@ -43,27 +49,35 @@ func (r *releaseSyncOptions) Run() error {
 
 		for _, tags := range allTags {
 			for _, tag := range tags.Tags {
-				c := int64(0)
-				r.db.DB.Table("release_tags").Where(`"release_tag" = ?`, tag.Name).Count(&c)
-				if c > 0 {
+				mReleaseTag := models.ReleaseTag{}
+				r.db.DB.Table(releaseTagsTable).Where(`"release_tag" = ?`, tag.Name).Find(&mReleaseTag)
+				// expect Phase to be populated if the record is present
+				if len(mReleaseTag.Phase) > 0 {
+
+					if mReleaseTag.Phase != tag.Phase {
+						log.Infof("Fetching updated tag %s from release controller...\n", tag.Name)
+						releaseTag := r.buildReleaseTag(tags.Architecture, release, tag)
+
+						if releaseTag == nil {
+							continue
+						}
+
+						// sync up the ids and created at
+						releaseTag.ID = mReleaseTag.ID
+						releaseTag.CreatedAt = mReleaseTag.CreatedAt
+
+						if err := r.db.DB.Table(releaseTagsTable).Save(releaseTag).Error; err != nil {
+							return err
+						}
+					}
 					continue
 				}
 
 				log.Infof("Fetching tag %s from release controller...\n", tag.Name)
-				releaseDetails := r.fetchReleaseDetails(tags.Architecture, release, tag)
-				releaseTag := releaseDetailsToDB(tags.Architecture, tag, releaseDetails)
-				// We skip releases that aren't fully baked (i.e. all jobs run and changelog calculated)
-				if releaseTag == nil || (releaseTag.Phase != api.PayloadAccepted && releaseTag.Phase != api.PayloadRejected) {
-					continue
-				}
+				releaseTag := r.buildReleaseTag(tags.Architecture, release, tag)
 
-				// PR is many-to-many, find the existing relation. TODO: There must be a more clever way to do this...
-				for i, pr := range releaseTag.PullRequests {
-					existingPR := models.ReleasePullRequest{}
-					result := r.db.DB.Table("release_pull_requests").Where("url = ?", pr.URL).Where("name = ?", pr.Name).First(&existingPR)
-					if result.Error == nil {
-						releaseTag.PullRequests[i] = existingPR
-					}
+				if releaseTag == nil {
+					continue
 				}
 
 				if err := r.db.DB.Create(&releaseTag).Error; err != nil {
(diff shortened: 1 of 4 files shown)


pkg/releasesync/releasesync.go
```diff
@@ -43,27 +49,35 @@ func (r *releaseSyncOptions) Run() error {
...
-				r.db.DB.Table("release_tags").Where(`"release_tag" = ?`, tag.Name).Count(&c)
-				if c > 0 {
+				mReleaseTag := models.ReleaseTag{}
+				r.db.DB.Table(release_tags_table).Where(`"release_tag" = ?`, tag.Name).Find(&mReleaseTag)
+				// expect Phase to be populated if the record is present
+				if len(mReleaseTag.Phase) > 0 {
+
+					if mReleaseTag.Phase != tag.Phase {
+						log.Infof("Fetching updated tag %s from release controller...\n", tag.Name)
+						releaseTag := r.buildReleaseTag(tags.Architecture, release, tag)
```
stbenjam, line 59: Do we need to fetch it again in this case?


```go
if mReleaseTag.Phase != tag.Phase {
  mReleaseTag.Phase = tag.Phase
  mReleaseTag.Forced = true
  if err := r.db.DB.Table(release_tags_table).Save(mReleaseTag).Error; err != nil {
  [...]
```

sippy-ng/src/releases/ReleasePayloadTable.js
```diff
@@ -179,6 +179,12 @@ function ReleasePayloadTable(props) {
         return ''
       },
     },
+    {
+      field: 'forced',
+      headerName: 'Forced',
+      flex: 1.5,
+      hide: props.briefTable,
```
stbenjam, line 186: Could we show this as a warning icon instead? And what do you think about making the row a warning color?

Patch: [0001-Show-forced-as-icon.patch.txt](https://github.com/openshift/sippy/files/8996557/0001-Show-forced-as-icon.patch.txt)


image

pkg/releasesync/releasesync.go
stbenjam: Although your version is safer, in case anything else changes.

stbenjam (commented):
Looks great! Just a few nits. I like the comparison of jobs for detection.

pkg/releasesync/releasesync.go
```diff
@@ -17,6 +17,12 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
+const (
+	release_tags_table = "release_tags"
```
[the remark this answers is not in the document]
neisw: yeah, missed that earlier but just fixed it

sippy-ng/src/releases/ReleasePayloadTable.js
neisw: Yeah, that is great!

pkg/releasesync/releasesync.go
neisw: I think I did it that way to force it to reprocess and set the forced flag via the api.  But if performance is a concern we could definitely do it this way.  We could even just update the individual columns.  I started to play with that but seemed like we would have to code in the column names or go through some funky reflection to get it out of gorm tag.

stbenjam: We  only get a handful of releases a day, performance isn't too much of a concern.  Your version is good

pkg/releasesync/releasesync.go
```diff
@@ -169,6 +204,24 @@ func releaseDetailsToDB(architecture string, tag ReleaseTag, details ReleaseDeta
...
+				failedBlocking = true
+				break
+			}
+		}
+	}
+
+	if release.Phase == "Accepted" {
+		release.Forced = failedBlocking
+	} else if release.Phase == "Rejected" {
+		release.Forced = !failedBlocking
```
dgoodwin, line 222: Oof this is a tough one. Technically we could have force rejected before any job actually completed, and virtually all force rejections would be while jobs are still running. This is arguably best effort and probably good enough, force accepted is all we really care about I think. We could change it to just ForceAccepted, but this is also pretty good as is I think.

Or should we try to detect all blocking jobs not completed and phase Rejected on the payload?

sippy-ng/src/releases/ReleasePayloadTable.js
```diff
@@ -58,7 +68,24 @@ function ReleasePayloadTable(props) {
...
+    {
+      field: 'forced',
+      headerName: 'Forced',
+      align: 'center',
+      flex: 0.75,
+      hide: props.briefTable,
+      renderCell: (params) => {
+        if (params.value === true) {
+          return (
+            <Tooltip title="This payload was forced.">
```
dgoodwin, line 80:
```suggestion
            <Tooltip title="This payload was manually force accepted/rejected.">
```

Is this an improvement?

pkg/releasesync/releasesync.go
```diff
@@ -76,6 +90,27 @@ func (r *releaseSyncOptions) Run() error {
 	return nil
 }
 
+func (r *releaseSyncOptions) buildReleaseTag(architecture, release string, tag ReleaseTag) *models.ReleaseTag {
+	releaseDetails := r.fetchReleaseDetails(architecture, release, tag)
+	releaseTag := releaseDetailsToDB(architecture, tag, releaseDetails)
+
+	// We skip releases that aren't fully baked (i.e. all jobs run and changelog calculated)
+	if releaseTag == nil || (releaseTag.Phase != api.PayloadAccepted && releaseTag.Phase != api.PayloadRejected) {
```
dgoodwin, line 98: I think there's a third state, PayloadFailed, which I think happens when something goes wrong in the ART pipeline or similar. Do we need to handle that here? (I see that this is just code being moved so maybe not a concern for this PR)

dgoodwin:
Pro-tip: If you edit your original comment and put TRT-349 somewhere, a bot will auto-link the two.

pkg/releasesync/releasesync.go
neisw: Thinking about this some, if our phase is 'Rejected'  and failedBlocking === false (we didn't have any blocking jobRuns with a state of 'Failed') we would set the release.Forced flag to true which is what I think we want in this case.  If the release phase is 'Rejected' we consider it forced if we have no failed blocking jobs.

pkg/releasesync/releasesync.go
neisw: Only if we want those to show up in Sippy.  Right now they would be skipped which seems reasonable unless we want historical context for pipeline failures

dgoodwin: Gotcha, thanks!

pkg/releasesync/releasesync.go
dgoodwin: Ok that would lead to one more possible issue, would we get Forced=true if *some* blocking jobs failed, but others did not complete? this might not even be worth worrying about, it doesn't matter if it was force rejected at that point really, a blocker filed, it was going to get rejected anyhow. I think this might be good to go.

PR #476, opened 2022-06-29
Retrieve job/release mapping from configuration file

stbenjam:
[TRT-321](https://issues.redhat.com//browse/TRT-321)

This uses a configuration file generated by the ci-tools repo to know
which prow to communicate with, and which release to map jobs to.

Currently manually generated, we can automate it like testgrid does
at some point.

Generation tool: https://github.com/openshift/ci-tools/pull/2879


stbenjam:
@dgoodwin Could you have a look at this one?

Thanks!

main.go
```diff
@@ -121,7 +125,9 @@ func main() {
 	flags.BoolVar(&opt.SkipBugLookup, "skip-bug-lookup", opt.SkipBugLookup, "Do not attempt to find bugs that match test/job failures")
 	flags.StringVar(&opt.LogLevel, "log-level", defaultLogLevel, "Log level (trace,debug,info,warn,error)")
 	flags.BoolVar(&opt.LoadTestgrid, "load-testgrid", true, "Fetch job and job run data from testgrid")
+
 	flags.BoolVar(&opt.LoadProw, "load-prow", opt.LoadProw, "Fetch job and job run data from prow")
+	flags.StringVar(&opt.Config, "config", opt.Config, "Configuration for Prow Sippy")
```
dgoodwin, line 130: Anything more we can say here, is it optional/required, what it should contain, and do we want to link it to prow sippy? Seems like it might be generally useful in future but I'm not sure.

Any particular reason to use a directory with assumed filenames vs cli-arg per config file?

pkg/apis/config/v1/types.go
```diff
@@ -0,0 +1,19 @@
+package v1
+
+type SippyConfig struct {
+	Prow     ProwConfig               `yaml:"prow"`
+	Releases map[string]ReleaseConfig `yaml:"releases"`
+}
+
+type ProwConfig struct {
+	// URL to the prowjob.js endpoint of the prow instance.
```
dgoodwin, line 9: Could you note some details on what this endpoint contains and how we use it.

dgoodwin (commented):
Looks great just hoping we can get that committed file out of the picture and avoid a manual task we have to remember.

config/openshift.yaml
```diff
@@ -0,0 +1,1304 @@
+# Generated by openshift/ci-tools sippy-config-generator on 2022-06-29 13:43:25.337602371 -0400 EDT m=+1.855124459
+prow:
+  url: https://prow.ci.openshift.org/prowjobs.js
```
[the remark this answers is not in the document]
stbenjam: I had intended to do it like testgrid, there's already framework in ci-tools to have it automatically open a PR to sippy and merge it on a recurring basis.

pkg/prowloader/prow.go
```diff
@@ -100,22 +103,41 @@ func (pl *ProwLoader) LoadProwJobsToDB(filters []*regexp.Regexp) error {
...
 		}
 	}
 
 	return nil
 }
 
-func fetchJobsJSON() ([]byte, error) {
-	resp, err := http.Get(prowURL)
+func fetchJobsJSON(prowURL string) ([]byte, error) {
+	resp, err := http.Get(prowURL) // #nosec G107
```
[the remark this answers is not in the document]
stbenjam: It's about 100M but it's only fetched once.

config/openshift.yaml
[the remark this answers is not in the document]
dgoodwin: I can live with that.

main.go
stbenjam: So if mode is ocp it looks for the config file of a known name? I kind of prefer it to be explicit.  Imagine having to do something like the sdn/ovn comparison again, it's easy to customize a list of jobs.

pkg/apis/config/v1/types.go
stbenjam: Done

main.go
stbenjam: Yea sorry it's just one config file.  

The generator can take a customization file to override the generated config. There's some docs in [config/README.md](https://github.com/openshift/sippy/pull/476/files#diff-6e8c7fdec58c9cb55af12d91063af268d43c753cf477bc41ec4e938f1b429b6f), but happy to make it clearer or refactor that if it's not clear

PR #478, opened 2022-06-30
Improvements for experimental payload stream analysis API.

dgoodwin:
Added better insight into number of times a test failed across the
requested number of most recent payloads, and a blocker score indicating
how likely this is to be a payload blocker.


stbenjam:
@dgoodwin Is this one ready to go? Feel free to unhold if ready

/lgtm
/hold

PR #484, opened 2022-07-12
Track PR state for prow-based Sippy

stbenjam:
[TRT-485](https://issues.redhat.com//browse/TRT-485)

Adds new repository pages that show GitHub repos and the worst performing jobs.

image

image

---
Todo's before merging

- [x] Get a github token available on the DPCR cluster


pkg/db/query/repository_queries.go
```diff
@@ -0,0 +1,32 @@
...
+		Joins("INNER JOIN prow_job_run_prow_pull_requests ON prow_job_run_prow_pull_requests.prow_pull_request_id = prow_pull_requests.id").
+		Joins("INNER JOIN prow_job_runs on prow_job_run_prow_pull_requests.prow_job_run_id = prow_job_runs.id").
+		Joins("INNER JOIN prow_jobs on prow_job_runs.prow_job_id = prow_jobs.id").
+		Joins("LEFT JOIN (?) premerge_failures ON premerge_failures.prow_job_ID = prow_jobs.id", averageByJob).
+		Where("prow_jobs.release = ?", release).
+		Group("prow_pull_requests.org, prow_pull_requests.repo").
+		Select("ROW_NUMBER() OVER() as id, prow_pull_requests.org, prow_pull_requests.repo, coalesce(max(average_premerge_job_failures), 0) as premerge_job_failures, count(distinct(prow_jobs.id)) as job_count")
+
+	results := make([]api.Repository, 0)
+	q, err := filter.FilterableDBResult(dbc.DB.Table("(?) as repos", repos), filterOpts, api.Repository{})
```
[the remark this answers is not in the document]
stbenjam: A trick to make filtering work, `SELECT * FROM (subquery)` makes it so the filtering layer can append WHERE clauses arbitrarily on the api-named fields.

stbenjam:
/hold

One more lint thing to fix, and need to get the token in place

stbenjam:
/hold cancel

@dgoodwin This is ready for another look

pkg/prowloader/prow.go
```diff
@@ -136,6 +145,30 @@ func (pl *ProwLoader) LoadProwJobsToDB() error {
...
 }
 
+func (pl *ProwLoader) syncPRStatus() error {
+	if pl.githubClient == nil {
+		log.Infof("No GitHub client, skipping PR sync")
+		return nil
+	}
+
+	pulls := make([]models.ProwPullRequest, 0)
+	pl.dbc.DB.Table("prow_pull_requests").Where("merged_at IS NULL").Scan(&pulls)
```
dgoodwin, line 155: Looks like a missing error check here.

pkg/prowloader/prow.go
```diff
@@ -136,6 +145,30 @@ func (pl *ProwLoader) LoadProwJobsToDB() error {
...
+	for _, pr := range pulls {
+		mergedAt, err := pl.githubClient.GetPRMerged(pr.Org, pr.Repo, pr.Number, pr.SHA)
+		if err != nil {
+			log.WithError(err).Warningf("could not fetch pull request status from GitHub; org=%q repo=%q number=%q sha=%q", pr.Org, pr.Repo, pr.Number, pr.SHA)
+			return err
+		}
+		pr.MergedAt = mergedAt
```
dgoodwin, line 162: Should we check if it changed here before doing the Save to limit writes?

dgoodwin (commented):
Just a couple small ones, otherwise looks good.

pkg/prowloader/prow.go
stbenjam: Thanks, fixed

pkg/prowloader/prow.go
stbenjam: Good call, fixed

stbenjam:
@dgoodwin Addressed your feedback in [7ed9f70](https://github.com/openshift/sippy/pull/484/commits/7ed9f709ab1104094ac14dd4e44a37173c13eaee)

PR #485, opened 2022-07-15
Store failure output from junit test cases

stbenjam:


dgoodwin:
Worth a shot. YOLO.
/lgtm

PR #486, opened 2022-07-15
Add Release Streams UI

dgoodwin:
Previously we linked to a full list of all payloads for all streams, which was difficult to navigate and see trends on, usually requiring us to manually add two filters for what we were actually interested in.

This PR introduces the UI section for Payload Streams, listing each stream and then linking to a page with several tabs for a specific stream. The current tabs are an overview with info about payload acceptance rates, a list of payloads for that stream (re-using the old table, just now pre-filtered for a specific arch and stream), and a new list of test failures over the last two weeks. This list of test failures is accompanied by a blocker score which indicates how likely this test is to be fully blocking the payload. Logic for this pre-existing API has been restructured. 

A new alert has been added for when we have identified possible payload blocking tests.

Adds a new materialized view for test failures from payload blocking jobs over the past two weeks, the query live was too slow on our production server. (40+ seconds)


stbenjam (commented):
I'll look closer on monday but this is so cool

dgoodwin:
Thanks!

Addressed those first couple comments.

sippy-ng/src/releases/PayloadStreamOverview.js
```diff
@@ -0,0 +1,170 @@
...
+          throw new Error('server returned ' + response.status)
+        }
+        return response.json()
+      })
+      .then((json) => {
+        // Find our specific stream's health in the list:
+        for (const stream of json) {
+          if (
+            stream.architecture == props.arch &&
+            stream.stream == props.stream
```
stbenjam, line 60: `===` not `==` is generally preferred for this kind of comparison (https://codeahoy.com/javascript/2019/10/12/==-vs-===-in-javascript/)

sippy-ng/src/releases/PayloadStreamOverview.js
```diff
@@ -0,0 +1,170 @@
...
+
+  return (
+    <Container>
+      <Grid item md={12}>
+        <Typography variant="h3" style={{ textAlign: 'center' }}>
+          Payload Stream Overview
+        </Typography>
+        <Grid container spacing={3} alignItems="stretch">
+          <Grid item md={3}>
+            <SummaryCard
```
stbenjam, line 94: SummaryCard takes a `tooltip` prop.  It actually should be required since we always render the info icon:

image

Suggest a tooltip prop explaining what "Overall" means (what's the time span?).

sippy-ng/src/releases/PayloadStreamOverview.js
```diff
@@ -0,0 +1,170 @@
...
+  if (!isLoaded) {
+    return <p>Loading...</p>
+  }
+
+  return (
+    <Container>
+      <Grid item md={12}>
+        <Typography variant="h3" style={{ textAlign: 'center' }}>
+          Payload Stream Overview
+        </Typography>
```
stbenjam, line 91: Suggest removing this, it's redundant since the tab already says `Overview`

sippy-ng/src/releases/PayloadStreamOverview.js
```diff
@@ -0,0 +1,170 @@
...
+                      % Accepted
+                    </span>
+                  </Tooltip>
+                </Fragment>
+              }
+              fail={streamHealth.phase_counts.total.rejected}
+            />
+          </Grid>
+          <Grid item md={3}>
+            <SummaryCard
```
stbenjam, line 121: Likewise needs a tooltip prop.

The other option is fixing SummaryCard so it doesn't render the Info icon.

sippy-ng/src/releases/PayloadStreamOverview.js
```diff
@@ -0,0 +1,170 @@
...
+              fail={streamHealth.phase_counts.current_week.rejected}
+            />
+          </Grid>
+          <Grid item md={6}>
+            <Typography variant="h6" style={{ textAlign: 'center' }}>
+              Last {streamHealth.count} payloads have been{' '}
+              {streamHealth.last_phase}
+            </Typography>
+            <Typography variant="h6" style={{ textAlign: 'center' }}>
+              Last payload accepted: {streamHealth.release_time}
```
stbenjam, line 153: There's a relative time helper. Suggest:

```jsx
              <Tooltip title={streamHealth.release_time}>
                <Fragment>
                  Last payload accepted:{' '}
                  {relativeTime(new Date(streamHealth.release_time))}
                </Fragment>
              </Tooltip>
```

renders:

`Last payload accepted: 6 hours ago`

(with a tooltip showing the actual timestamp)

sippy-ng/src/releases/PayloadStreamOverview.js
```diff
@@ -0,0 +1,170 @@
...
+    return <p>Loading...</p>
+  }
+
+  return (
+    <Container>
+      <Grid item md={12}>
+        <Typography variant="h3" style={{ textAlign: 'center' }}>
+          Payload Stream Overview
+        </Typography>
+        <Grid container spacing={3} alignItems="stretch">
```
stbenjam, line 92: I'd suggest adding `justifyContent="center"`, and also putting the text in a separate grid:


image

sippy-ng/src/releases/PayloadStreamTestFailures.js
```diff
@@ -0,0 +1,296 @@
...
+      .then(([analysis]) => {
+        if (analysis.status !== 200) {
+          throw new Error('server returned ' + analysis.status)
+        }
+
+        return Promise.all([analysis.json()])
+      })
+      .then(([analysis]) => {
+        if (analysis.length === 0) {
+          return <Typography variant="h5">Analysis not found.</Typography>
```
stbenjam, line 179: You can't return like this in a promise, I'd remove this and put it below line 207

sippy-ng/src/releases/PayloadStreamOverview.js
stbenjam: Example of what I mean in https://github.com/openshift/sippy/commit/64d2caafd6ea609143eb5e2ad0b5713f245da306

sippy-ng/src/releases/PayloadStreamTestFailures.js
stbenjam: Example of what I mean in https://github.com/openshift/sippy/commit/64d2caafd6ea609143eb5e2ad0b5713f245da306

pkg/api/releases.go
```diff
@@ -185,50 +184,53 @@ func GetPayloadAnalysis(dbc *db.DB, release, stream, arch string, numPayloadsToA
...
 // certainly a blocker) based on a number of criteria.
+//
+// consecutiveFailedPayloadTags is the list of our current streak of rejected payload tags. If most recent payload
+// was accepted, this list will be empty, and we don't have much processing to do.
 func calculateBlockerScore(consecutiveFailedPayloadTags []string, ta *apitype.TestFailureAnalysis) {
 	if len(consecutiveFailedPayloadTags) == 0 {
 		// our most recent state is Accepted, could be intermittent, but for the purposes of a blocker
 		// we have to assume 0.
-		ta.BlockerScore = 0.0
+		ta.BlockerScore = 0
```
stbenjam, line 200: Suggest adding a blocker score reason explaining the above comment.  Currently we render a tooltip but without a reason.

stbenjam (commented):
Some small UI comments

sippy-ng/src/releases/PayloadStreamOverview.js
dgoodwin: TIL, thanks!

sippy-ng/src/releases/PayloadStreamOverview.js
dgoodwin: Cherry picking, ty!

dgoodwin:
Added your commit and one of my own for the blocker score reason on 0.

PR #491, opened 2022-07-19
Show additional release payload statistics

stbenjam:
[TRT-264](https://issues.redhat.com//browse/TRT-264)

Adding the min/mean/max stats that Brenton had mentioned in the JIRA ticket.  Open to opinions about how this is laid out, and whether we should show the min/max values too.

image

image


stbenjam:
Oops, I had to fix a test - needs another lgtm.

stbenjam:
Sorry, I pushed up extra stuff... should be good now.

PR #495, opened 2022-07-21
Add some missing indexes on ID's

stbenjam:
I tried to do some direct queries on prow_job_run_tests by test_id and
wow is it slow.

We should merge this early morning or at the end of the day, takes about 30 minutes to run.


stbenjam:
/hold
To merge at end of the day.

dgoodwin:
Is the prow job run timestamp worth an index? I had been wondering if, because our matviews filter down on time, an index here would help us at all to more quickly discard everything we don't care about and focus on a managable set of data.

PR #496, opened 2022-07-22
Identify bad nurp for tests

xueqzhan:


sippy-ng/src/releases/ReleaseOverview.js
```diff
@@ -424,6 +424,56 @@ export default function ReleaseOverview(props) {
...
+                    BOOKMARKS.NO_STEP_GRAPH,
+                    BOOKMARKS.HIGH_DELTA_FROM_AVERAGE,
+                    BOOKMARKS.HIGH_STANDARD_DEVIATION
+                  )}&sortField=delta_from_average&sort=asc`}
+                  style={{ textAlign: 'center' }}
+                  variant="h5"
+                >
+                  Top failing test NURPs
+                  <Tooltip title={TOP_FAILERS_TOOLTIP}>
+                    <InfoIcon />
```
dgoodwin, line 445: We could describe in the tooltip here that these are tests that are performing significantly worse on particular variant combos than the test in general.

pkg/db/query/test_queries.go
```diff
@@ -117,3 +157,26 @@ FROM results;
 	log.Infof("TestReportExcludeVariants completed in %s", elapsed)
 	return testReport, nil
 }
+
+// TestsByNURPAndStandardDeviation returns a test report for every test in the db matching the given substrings, separated by variant.
```
dgoodwin, line 161: Can you expand on the standard deviation aspect in the godoc here.

pkg/db/query/test_queries.go
```diff
@@ -117,3 +157,26 @@ FROM results;
...
+		Select(`id as test_id, variants as pass_rate_variants, `+QueryTestPercentages).
+		Where(`release = ?`, release)
+
+	return dbc.DB.
+		Table(table).
+		Select("*, (current_working_percentage - working_average) as delta_from_average").
+		Joins(fmt.Sprintf(`INNER JOIN (?) as pass_rates on pass_rates.test_id = %s.id AND pass_rates.pass_rate_variants = %s.variants`, table, table), passRates).
+		Joins(fmt.Sprintf(`JOIN (?) as stats ON stats.test_id = %s.id`, table), stats).
+		Where(`release = ?`, release).
+		Where(fmt.Sprintf("NOT ('never-stable'=any(%s.variants))", table))
```
dgoodwin, line 181: Have you tested this will a full copy of the prod db from sippy-exports in s3 Ken? I'm curious about how fast it is. I tried locally with my last import and it seems totally fine.

And my followup question, would it be possible to also be able to do this such that we could also see the oddities in just passing percentage, and especially flaking percentage. (i.e. the three working views Stephen added for us on the tests table)

dgoodwin: @stbenjam does this sql look ok to you? It's a bit beyond my abilities.

Actually in that line of thinking, Ken could you expand a little on how the sql all fits together with a comment for people like me. :)

sippy-ng/src/tests/TestTable.js
```diff
@@ -322,6 +342,37 @@ function TestTable(props) {
         </div>
       ),
     },
+    delta_from_average: {
+      field: 'delta_from_average',
+      headerName: 'Delta_From_Average',
```
dgoodwin, line 347: These should be human readable, so headerName maybe just Delta or "Delta From Avg"? We've got a screen real estate issue I think so brief is going to be important, and then tooltips for a full description perhaps.

sippy-ng/src/tests/TestTable.js
```diff
@@ -322,6 +342,37 @@ function TestTable(props) {
...
+            {params.value
+              ? Number(params.value).toFixed(1).toLocaleString() + '%'
+              : ''}
+          </div>
+        )
+      },
+    },
+    working_average: {
+      field: 'working_average',
+      headerName: 'Working average for test on all NURPs',
```
dgoodwin, line 361: Lets hide this column, it looks like it's always just the inverse of the delta from average.

sippy-ng/src/tests/TestTable.js
```diff
@@ -322,6 +342,37 @@ function TestTable(props) {
...
+        <div className="percentage-cell">
+          {params.value
+            ? Number(params.value).toFixed(1).toLocaleString() + '%'
+            : ''}
+        </div>
+      ),
+    },
+    working_standard_deviation: {
+      field: 'working_standard_deviation',
+      headerName: 'Working standard deviation',
```
dgoodwin, line 373: Suggest just StdDev or something equally brief?

dgoodwin (commented):
This looks awesome Ken, very high value change.

Main themes from review, could we get this for flaking/passing and tie in with Stephen's working views, tightening up the table columns, and I'd love the git commit to have a descriptive message of what the change contains.

pkg/db/query/test_queries.go
stbenjam: It's ok. I think I wrote something like it in my original z-score PR (I like this version with deltas better).

+1 for showing passing and flaking.

sippy-ng/src/releases/ReleaseOverview.js
```diff
@@ -424,6 +424,56 @@ export default function ReleaseOverview(props) {
...
+                    hideControls={true}
+                    sortField="delta_from_average"
+                    sort="asc"
+                    limit={10}
+                    rowsPerPageOptions={[5]}
+                    filterModel={{
+                      items: [
+                        BOOKMARKS.RUN_7,
+                        BOOKMARKS.NO_NEVER_STABLE,
+                        BOOKMARKS.NO_TECHPREVIEW,
```
stbenjam, line 462: We can remove techpreview, it can be considered as stable as anything else now (https://github.com/openshift/sippy/pull/493 is going to merge soon)

sippy-ng/src/tests/TestTable.js
```diff
@@ -685,7 +740,7 @@ function TestTable(props) {
         autoHeight={true}
         rowHeight={100}
         disableColumnFilter={props.briefTable}
-        disableColumnMenu={true}
+        disableColumnMenu={false}
```
stbenjam, line 743: I would suggest leaving this true.  The column menu settings doesn't get saved in links so it's harder to share data.  The `views` feature covers this by having static views stored in the URL.

sippy-ng/src/tests/TestTable.js
xueqzhan: This is the average from which delta is calculated. It is only the inverse when rate is 0. I think it makes sense to keep.

sippy-ng/src/releases/ReleaseOverview.js
```diff
@@ -424,14 +419,65 @@ export default function ReleaseOverview(props) {
...
+                        BOOKMARKS.WITHOUT_OVERALL_JOB_RESULT,
+                        BOOKMARKS.NO_STEP_GRAPH,
+                        BOOKMARKS.HIGH_DELTA_FROM_WORKING_AVERAGE,
+                        BOOKMARKS.HIGH_STANDARD_DEVIATION,
+                      ],
+                    }}
+                    pageSize={5}
+                    briefTable={true}
+                    release={props.release}
+                  />
```
stbenjam, line 469: The home page widgets a little crowded.

image

stbenjam: I'd suggest just showing the test name, variants, current working percentage, and delta.  That should fit.

stbenjam (commented):
I think it looks good, just the overview widget needs some tweaks I think (see below).

sippy-ng/src/releases/ReleaseOverview.js
dgoodwin: You'll notice in my PR I slipped suite name in, so it's still very tight. :(

xueqzhan: Maybe even remove the current working percentage?

PR #503, opened 2022-08-03
Cluster health

stbenjam:
[TRT-237](https://issues.redhat.com//browse/TRT-237)

First pass at a build health page.  Since cluster health isn't really constrained by release, this adds a new side bar item outside of Releases.

Unfortunately the data is noisy, but sharp drops (like build02 today) are visible.

image

### DIFF
--- a/pkg/api/build_clusters.go
+++ b/pkg/api/build_clusters.go
@@ -1,0 +1,47 @@
+package api
+
+import (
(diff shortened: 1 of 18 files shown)


dgoodwin:
/lgtm

Would like this rolled out for today's upgrade. Nice work.

PR #506, opened 2022-08-04
Add some docs about working with prow/gcs/release controller data

stbenjam:
[TRT-403](https://issues.redhat.com//browse/TRT-403)

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -1,3 +1,10 @@
+## Building Sippy
+
+Running `make` will build an all-in-one binary that contains both the go
+app and frontend.
+
+To build just the backend, run `go build -mod=vendor .`
+
 ## Create PostgreSQL Database
 
 Launch postgresql:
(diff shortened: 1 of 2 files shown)


stbenjam:
@xueqzhan Could you have a look?

DEVELOPMENT.md
```diff
@@ -6,38 +12,121 @@ Launch postgresql:
...
+  --database-dsn="postgresql://postgres:password@localhost:5432/postgres" \
+  --skip-bug-lookup
 ````
 
-### From a Prod Sippy Backup
+### From Prow and GCS buckets
 
-See the TRT team drive in Google, periodically there is a gzip'd backup stored there. Restore it locally with:
+Fetching data from prow directly gives us access to more raw data that TestGrid doesn't have, such as build cluster
+data, raw junit files, test durations, and more -- but it requires more configuration that using TestGrid.
```
xueqzhan, line 70: "than using TestGrid"

xueqzhan (commented):
Lots of good information. Thanks

PR #516, opened 2022-08-09
Don't serve metrics until data has been refreshed

stbenjam:


dgoodwin:
I think there's a decent chance this should work, I can't see alertmanager clearing alerts when there's no endpoint as it would when there are no metrics but endpoint is up.

/lgtm

PR #517, opened 2022-08-09
Add index for ReleaseJobRun, and OnConflict clause to help avoid duplicate prow jobs in DB

DennisPeriquet:
I'm hoping that making `prow_job_run_id` an index will enforce that it is unique.

I'm expecting the `OnConflict` clause when used to `Save` or `Create` will update records instead of creating new ones.

The `results[id] = result` fix was added because it didn't look like the return value was ever updated.


pkg/db/models/releases.go
```diff
@@ -112,7 +112,7 @@ type ReleaseJobRun struct {
 
 	ReleaseTag     ReleaseTag `json:"release_tag" gorm:"foreignKey:release_tag_id"`
 	ReleaseTagID   string     `gorm:"column:release_tag_id"`
-	Name           uint       `json:"name" gorm:"column:prow_job_run_id"` // TODO: this could use a rename to ProwJobRunID
+	Name           uint       `json:"name" gorm:"column:prow_job_run_id,index:prow_job_run_id_idx"` // TODO: this could use a rename to ProwJobRunID
```
stbenjam, line 115: A plain index is just a tool to help lookups, this won't guarantee uniqueness.  You want a unique index: https://gorm.io/docs/indexes.html

DennisPeriquet: change pushed to use `index:unique`.

stbenjam:
I believe this works.  Let's try it.

/lgtm

stbenjam:
@DennisPeriquet Actually have you tried this PR on a backup of the production database? I think we may need to clear the duplicates first.

DennisPeriquet:
> @DennisPeriquet Actually have you tried this PR on a backup of the production database? I think we may need to clear the duplicates first.

@stbenjam

I tried these two sets of parameters (on a but did not notice any errors:

```
--server --local-data /opt/sippy-testdata --release 4.11 --log-level=debug --skip-bug-lookup --database-dsn=postgresql://postgres:redacted@192.168.1.52:5432/postgres --db-only-mode

--local-data /opt/sippy-testdata --release 4.12 --log-level=debug --skip-bug-lookup --database-dsn=postgresql://postgres:redacted@192.168.1.52:5432/postgres --db-only-mode --load-database --load-testgrid=true --arch amd64
```

Is there a particular one you'd like me to try?

I can work on getting the duplicates removed.

Thanks.

DennisPeriquet:
The duplicates have been removed from the sippy database as of Aug 22, 2022 11am.

dgoodwin:
/lgtm

Keep an eye on sippy logs when this rolls out, make sure we don't see any errors logged that might get missed, they would be from gorm on startup I think.

PR #518, opened 2022-08-09
Differentiate on test suite in NURPs test tables.

dgoodwin:
Often when hunting for tests that are degraded or worse than others, the
suite can make a dramatic difference in surfacing the problem, typically
during upgrade but not regular non-upgrade e2e.

Update the materialized view for test results to include suite name as a
characteristic.

Update the API to use this data when collapse=false, and ignore it when
collapse = true.

Similarly update the JS TestTable to display the column when
collapse=false.

[TRT-448](https://issues.redhat.com//browse/TRT-448)


dgoodwin:
Note the commit I just added filters our NURPs showing up with 0 current and previous runs, these show up first and I think have no meaning. Not totally sure why they're there at all ,possibly variant combos that used to run but don't anymore.

dgoodwin:
/hold

 #496 should go in first and I will need a rebase after.

dgoodwin:
Merged #496 into this branch. Might be best to review after Ken's work lands. (cleaner diff)

stbenjam:
> Note the commit I just added filters our NURPs showing up with 0 current and previous runs

Yay thank you, I keep meaning to do that

dgoodwin:
Ready for review.

dgoodwin:
/assign @xueqzhan 

Ken would you have time to take a look at this?

PR #527, opened 2022-08-18
Make prow the default source of data for sippy

stbenjam:
[TRT-440](https://issues.redhat.com//browse/TRT-440)

This will make the default deployment of sippy in production use Prow as
a data source, instead of TestGrid.

Despite the diff, this PR is just:
```
$ git mv fetchdata.sh fetchdata-testgrid.sh
$ git mv fetchdata-prow.sh fetchdata.sh
```


stbenjam:
/hold

For merge tomorrow

DennisPeriquet:
I noticed this:

```
$ curl -L -v --fail https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp/stable/openshift-client-linux.tar.gz --output oc.tgz 2>&1 |tee out.log
...
$ wc -l out.log  
6811 out.log
```

Do you want to remove the `-v`? as it adds a lot of lines to the log.

A lot of lines (that might not aid in debugging) can be removed if you wanted to play with it:

```
$ cat out.log |grep -v '\[no content\]' | grep -v '\[1 bytes data\]' | wc -l
70
```

stbenjam:
I wasn't able to spot the problem without seeing the headers, would prefer to keep them in case something odd like this happens again.

stbenjam:
@DennisPeriquet Was this meant for https://github.com/openshift/release/pull/31457 ?

DennisPeriquet:
> @DennisPeriquet Was this meant for [openshift/release#31457](https://github.com/openshift/release/pull/31457) ?

yes.  deleted. sorry for the confusion.

dgoodwin:
/lgtm

as well

PR #528, opened 2022-08-18
Revert "Differentiate on test suite in NURPs test tables."

stbenjam:
Reverts openshift/sippy#518

Any test that doesn't have a suite name ends up excluded from the matview, resulting in a bunch of missing data:

```
sippy_openshift=> SELECT * FROM prow_test_report_7d_matview WHERE release = '4.12' AND name LIKE '%cluster install%';
 id | name | suite_name | previous_successes | previous_flakes | previous_failures | previous_runs | current_successes | current_flakes | current_failures | current_runs | variants | release 
----+------+------------+--------------------+-----------------+-------------------+---------------+-------------------+----------------+------------------+--------------+----------+---------
(0 rows)
```

I think it needs to be a LEFT JOIN [here](https://github.com/openshift/sippy/pull/518/files#diff-7fa1b9fbef139e2c0d9eadf51d97fa898c9360bcfcbdcc407207629fcdca60cfR201).

All the top-level indicators on sippy are blank right now: https://sippy.dptools.openshift.org/sippy-ng/.  I think this PR is safe to revert, so gives us time to find a proper fix.

### DIFF
--- a/pkg/api/tests.go
+++ b/pkg/api/tests.go
@@ -165,7 +165,7 @@ func BuildTestsResults(dbc *db.DB, release, period string, collapse, includeOver
 		rawQuery = rawQuery.Select(`name,` + query.QueryTestSummer).Group("name")
 	} else {
 		rawQuery = query.TestsByNURPAndStandardDeviation(dbc, release, table)
-		variantSelect = "suite_name, variants, " +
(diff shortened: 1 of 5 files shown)


dgoodwin:
My apologies!

/lgtm

PR #532, opened 2022-08-19
Create indexes with materialized views

stbenjam:
[TRT-480](https://issues.redhat.com//browse/TRT-480)

Dropping a materialized view, drops its indexes -- but the schema hash table for the indexes remains unchanged, making us think the indexes are still in place.

The tests matviews are currently missing their indexes since we've updated them.  This change makes indexes created together with the matviews.

### DIFF
--- a/pkg/db/matviews.go
+++ b/pkg/db/matviews.go
@@ -89,15 +89,10 @@ func syncPostgresMaterializedViews(db *gorm.DB) error {
 		}
 		dropSQL := fmt.Sprintf("DROP MATERIALIZED VIEW IF EXISTS %s", pmv.Name)
 		schema := fmt.Sprintf("CREATE MATERIALIZED VIEW %s AS %s WITH NO DATA", pmv.Name, viewDef)
-		if err := syncSchema(db, hashTypeMatView, pmv.Name, schema, dropSQL); err != nil {
-			return err
-		}
-
-		// Sync index for the materialized view:
 		indexName := fmt.Sprintf("idx_%s", pmv.Name)
 		index := fmt.Sprintf("CREATE UNIQUE INDEX %s ON %s(%s)", indexName, pmv.Name, strings.Join(pmv.IndexColumns, ","))
-		dropSQL = fmt.Sprintf("DROP INDEX IF EXISTS %s", indexName)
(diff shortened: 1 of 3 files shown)


stbenjam:
Not sure if this is the right solution.  Maybe we should try to delete schema hashes for all related resources? i.e. if we update the matview, delete the hashes for the indexes too.

There's no urgency on this, I manually created the missing indexes to buy us some time.

dgoodwin:
I think I was just chasing this, sippy seems down in some pages, and found non-concurrent refreshes running because presumably we're missing indicies.

pkg/db/matviews.go
```diff
@@ -89,15 +89,10 @@ func syncPostgresMaterializedViews(db *gorm.DB) error {
...
-			return err
-		}
-
-		// Sync index for the materialized view:
 		indexName := fmt.Sprintf("idx_%s", pmv.Name)
 		index := fmt.Sprintf("CREATE UNIQUE INDEX %s ON %s(%s)", indexName, pmv.Name, strings.Join(pmv.IndexColumns, ","))
-		dropSQL = fmt.Sprintf("DROP INDEX IF EXISTS %s", indexName)
-		if err := syncSchema(db, hashTypeMatViewIndex, indexName, index, dropSQL); err != nil {
+		indexDropSQL := fmt.Sprintf("DROP INDEX IF EXISTS %s", indexName)
+		if err := syncSchema(db, hashTypeMatView, pmv.Name, schema, dropSQL, &index, &indexDropSQL); err != nil {
```
dgoodwin, line 95: Great find. 

This looks like we lose the ability to adjust an index in absence of a mat view change. That might be ok, but would it be simple to just return a bool for changed from the matview syncSchema call, and if true, assume to recreate ALL indexes?

stbenjam: That solution has the same problem, if you add just an index it won't get updated if the underlying matview doesn't change.

dgoodwin: I thought I was hashing indicies, so if they changed in isolation, we'd update them.

stbenjam:
/close

@dgoodwin's going to put up a better fix than this one

PR #535, opened 2022-08-22
test suite nurp

dgoodwin:
- Revert "Revert "Differentiate on test suite in NURPs test tables.""
- Add left join for suites to fix bug.


dgoodwin:
I think your suggestion looks right @stbenjam , the left join fixes the indicators, test tables look correct as far as I can see across 100k+ rows.

stbenjam:
I'll unhold before I leave today, rebuilding the matview takes a while.

PR #538, opened 2022-08-23
Fix loss of indicies when matviews change.

dgoodwin:
Possible alternative to #532 which would preserve ability to update indicies without requiring a matview update.

This change returns a bool when we update schema that will be true if we detected a schema change and did an update. This can be passed to as a force to the index schema update telling it to sync regardless if we see a hash change or not.


dgoodwin:
/hold 

 Still testing locally

PR #539, opened 2022-08-23
Add suite name to test matview indicies.

dgoodwin:


dgoodwin:
I did this manually on the db but forgot to include in the fix PR. Either seems to work, I can't say why. This however seems more correct now.

PR #548, opened 2022-08-25
Remove use of unique index on prow_job_run_id

DennisPeriquet:
This is causing sippy problems:

```
2022/08/25 19:17:33 /go/src/sippy/vendor/gorm.io/driver/postgres/migrator.go:140 ERROR: could not create unique index "idx_release_job_runs_name" (SQLSTATE 23505)
[20.931ms] [rows:0] CREATE UNIQUE INDEX "idx_release_job_runs_name" ON "release_job_runs" ("prow_job_run_id")
time="2022-08-25T19:17:33.263Z" level=fatal msg="error running command" error="ERROR: could not create unique index \"idx_release_job_runs_name\" (SQLSTATE 23505)"
```


stbenjam:
Deleting the dupes worked, I think we can close this

/close

PR #549, opened 2022-08-26
Update variant logic for 4.12

stbenjam:
[TRT-498](https://issues.redhat.com//browse/TRT-498)

- Add hypershift variant
- Metal single node are incorrectly identified as UPI, should be metal-assisted
- Infer network type based on >= 4.12 default being OVN


dgoodwin:
/lgtm

unit tests :100:

PR #559, opened 2022-08-31
Add silenced label to differentiate between zero runs and zero pass rate

xueqzhan:
After this deploy, I will change the alert to include the newly added label.


dgoodwin:
/lgtm

Thanks Ken!

PR #561, opened 2022-08-31
[wip] support pinnedDateTime for historical sippy

neisw:
Main, Server and matviews contain the majority of the logic / changes.  The rest just allow a specific time instance to be used instead of time.Now() calls.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 report.sh
 /sippy
 /data
+.vscode/
--- a/main.go
+++ b/main.go
@@ -138,7 +138,7 @@ func main() {
 	flags.StringVar(&opt.GoogleServiceAccountCredentialFile, "google-service-account-credential-file", os.Getenv("GOOGLE_APPLICATION_CREDENTIALS"), "location of a credential file described by https://cloud.google.com/docs/authentication/production")
 	flags.StringVar(&opt.GoogleOAuthClientCredentialFile, "google-oauth-credential-file", opt.GoogleOAuthClientCredentialFile, "location of a credential file described by https://developers.google.com/people/quickstart/go, setup from https://cloud.google.com/bigquery/docs/authentication/end-user-installed#client-credentials")
 
-	flags.StringVar(&opt.PinnedDateTime, "pinnedDateTime", opt.PinnedDateTime, "optional value to use instead of the current time specified in RFC3339 format - 2006-01-02 15:04:05+00:00")
+	flags.StringVar(&opt.PinnedDateTime, "pinnedDateTime", opt.PinnedDateTime, "optional value to use in a historical context with a fixed date / time value specified in RFC3339 format - 2006-01-02 15:04:05+00:00")
 
 	if err := cmd.Execute(); err != nil {
 		log.Fatalf("error: %v", err)
@@ -313,7 +313,7 @@ func (o *Options) Run() error { //nolint:gocyclo
 			if err != nil {
 				return err
 			}
-			err = perfscaleanalysis.DownloadPerfScaleData(scaleJobsDir, util.GetTimeNow(pinnedTime))
+			err = perfscaleanalysis.DownloadPerfScaleData(scaleJobsDir, util.GetReportEnd(pinnedTime))
 			if err != nil {
 				return err
 			}
@@ -326,12 +326,12 @@ func (o *Options) Run() error { //nolint:gocyclo
 	}
 
 	if o.InitDatabase {
-		_, err := db.New(o.DSN, util.GetTimeNow(pinnedTime))
+		_, err := db.New(o.DSN, util.GetReportEnd(pinnedTime))
 		return err
 	}
 
 	if o.LoadDatabase {
-		dbc, err := db.New(o.DSN, util.GetTimeNow(pinnedTime))
+		dbc, err := db.New(o.DSN, util.GetReportEnd(pinnedTime))
 		if err != nil {
 			return err
 		}
@@ -348,7 +348,7 @@ func (o *Options) Run() error { //nolint:gocyclo
 			loadBugs := !o.SkipBugLookup && len(o.OpenshiftReleases) > 0
 			for _, dashboard := range o.ToTestGridDashboardCoordinates() {
 				err := trgc.LoadDatabase(dbc, dashboard, o.getVariantManager(), o.getSyntheticTestManager(),
-					o.StartDay, o.NumDays, util.GetTimeNow(pinnedTime))
+					o.StartDay, o.NumDays, util.GetReportEnd(pinnedTime))
 				if err != nil {
 					log.WithError(err).Error("error loading database")
 					return err
@@ -421,7 +421,7 @@ func (o *Options) runServerMode(pinnedDateTime *time.Time) error {
 	var dbc *db.DB
 	var err error
 	if o.DSN != "" {
-		dbc, err = db.New(o.DSN, util.GetTimeNow(pinnedDateTime))
+		dbc, err = db.New(o.DSN, util.GetReportEnd(pinnedDateTime))
 		if err != nil {
 			return err
 		}
@@ -450,7 +450,7 @@ func (o *Options) runServerMode(pinnedDateTime *time.Time) error {
 
 	// Initial metrics refresh to get the endpoint scrapable ASAP and prevent prom gaps, before
 	// we start the lengthy mat view refreshes.
-	if err := metrics.RefreshMetricsDB(dbc, server.GetTimeNow()); err != nil {
+	if err := metrics.RefreshMetricsDB(dbc, server.GetReportEnd()); err != nil {
 		log.WithError(err).Error("error refreshing metrics")
 	}
 
--- a/main.go
+++ b/main.go
@@ -22,6 +22,7 @@ import (
 	v1 "github.com/openshift/sippy/pkg/apis/config/v1"
 	"github.com/openshift/sippy/pkg/prowloader/github"
 	"github.com/openshift/sippy/pkg/sippyserver/metrics"
+	"github.com/openshift/sippy/pkg/util"
 
 	"github.com/openshift/sippy/pkg/buganalysis"
 	"github.com/openshift/sippy/pkg/db"
@@ -76,6 +77,7 @@ type Options struct {
 	Config                             string
 	GoogleServiceAccountCredentialFile string
 	GoogleOAuthClientCredentialFile    string
+	PinnedDateTime                     string
 }
 
 func main() {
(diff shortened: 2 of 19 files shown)


.gitignore
```diff
@@ -1,3 +1,7 @@
 report.sh
 /sippy
 /data
+!sippy-ng/build/.gitignore
+sippy-ng/build/
+node_modules/
```
stbenjam, line 6: There shouldn't be any root node_modules directory, it would be under sippy-ng

stbenjam, line 6: There shouldn't be any root node_modules directory, it would be under sippy-ng

.gitignore
```diff
@@ -1,3 +1,7 @@
 report.sh
 /sippy
 /data
+!sippy-ng/build/.gitignore
+sippy-ng/build/
```
stbenjam, line 5: Nit, I would put this in the gitignore in sippy-ng

main.go
```diff
@@ -136,6 +137,8 @@ func main() {
 	flags.StringVar(&opt.GoogleServiceAccountCredentialFile, "google-service-account-credential-file", os.Getenv("GOOGLE_APPLICATION_CREDENTIALS"), "location of a credential file described by https://cloud.google.com/docs/authentication/production")
 	flags.StringVar(&opt.GoogleOAuthClientCredentialFile, "google-oauth-credential-file", opt.GoogleOAuthClientCredentialFile, "location of a credential file described by https://developers.google.com/people/quickstart/go, setup from https://cloud.google.com/bigquery/docs/authentication/end-user-installed#client-credentials")
 
+	flags.StringVar(&opt.PinnedDateTime, "pinnedDateTime", opt.PinnedDateTime, "optional value to use instead of the current time specified in RFC3339 format - 2006-01-02 15:04:05+00:00")
```
stbenjam, line 140: How will this work with historical Sippy? Ideally we would have multiple dates, i.e. the GA date for each release since it's different.  I'd expect this to be become more configurable, what if I wanted to see how things were a month ago?

Perhaps we could tackle it later and replace this with an API parameter, but I think we need a way to remember the GA date for each release and an idea of how it'll be wired up so that we can show different dates

main.go
```diff
@@ -273,6 +276,18 @@ func (o *Options) Run() error { //nolint:gocyclo
 		}
 	}
 
+	timeNow := time.Now()
+
+	if len(o.PinnedDateTime) > 0 {
+		timeNow, err = time.Parse(time.RFC3339, o.PinnedDateTime)
+
+		if err != nil {
+			log.Errorf("Error parsing pinnedDateTime: %#v", err)
```
stbenjam, line 285: Does Errorf exit? I don't think it does, so it might be a little surprising to a user who provides an invalid pinned date but sippy starts without it.

main.go
```diff
@@ -273,6 +276,18 @@ func (o *Options) Run() error { //nolint:gocyclo
 		}
 	}
 
+	timeNow := time.Now()
```
stbenjam, line 279: Having this hard coded seems problematic to me, because the server will forever be pinned to the time you started it with.  You'll never get new data without restarting

main.go
```diff
@@ -413,6 +428,15 @@ func (o *Options) runServerMode() error {
...
 	}
 
+	var sippyserverPinnedTime time.Time
+
+	// the pinnedDateTime that is passed in is always set
+	// so we have to check the input parameter to verify if
+	// we want to pass it along or provide the empty sippyserverPinnedTime
+	if len(o.PinnedDateTime) > 0 {
+		sippyserverPinnedTime = pinnedDateTime
+	}
```
stbenjam, line 438: As far as I can tell, you always pass a value in for o.PinnedDateTime that's > 0

stbenjam (commented):
A few comments, mostly about how we handle flexibility in the pinned date.

I also noticed there's several places we're doing the math to create the boundaries (an existing problem). I think there's an opportunity to centralize the date handling to a set of pkg/util functions.  What do you think?

main.go
neisw: That is the input parameter value, if it isn't provided then the len(..) should be 0, seems to work but can review more

main.go
neisw: No, but I can make it a hard error

main.go
neisw: image

It could be confusing with me passing in the none empty pinnedDateTime and then referencing the string option value but that was because I needed the value for the db.New call but wanted to leave the server value unpinned when the option wasn't specified.

.gitignore
neisw: Yeah, I think what I really needed was to run make clean to remove the extraneous sippy-ng/build files and not add to .gitignore

main.go
stbenjam: I see, I think I was confusing o.PinnedDateTime vs. the function parameter pinnedDateTime.  It's just a little confusing, since you ALWAYS pass a value for the time.Time function parameter.  Why do you need to pass it if a user didn't provide o.PinnedDateTime?

main.go
stbenjam: Were there discussions about how the historical sippy deployment would work? Is it something we'd only deploy on an as-needed basis? Or would we run a historical sippy for each release?

neisw: https://issues.redhat.com/browse/TRT-274 has a few alternatives, this implementation would fall under the 'Another Alternative' heading.  Seems the simplest to implement and most straightforward.  But we could build off of it and do the duplicate matviews approach and pass along the timestamps for each date / release.

main.go
neisw: The materialized views always need a dateTime now (no longer use NOW in the sql) (we can discuss this approach) in main.Run we need the dateTime where the dbs are created, there is also a call in runServerMode that calls db.New if needed, so I didn't want to parse the input over and over (but could) however that left me needing to flag when the server needed to be passed a specific dateTime (pinned) vs when it should get an empty one so the server knows to call time.Now() in that case.

stbenjam: Why not pass a `*time.Time` and check for nil? Use time.Now if it's nil, otherwise use the pinned value?

stbenjam: Yes, I think it's important to be more clear in the code about what's going on.

If there's an argument called pinnedDateTime, it's reasonable for a developer to assume if it has a non-zero value then there's something pinned.  That's not the case here, you always pass in a non-zero value -- and a user has to backtrack and look at the original raw options string to figure it out.  

You can hold all that state clearly by making time.Time a pointer and checking for nil, and if nil using `time.Now().`

I'm also not convinced passing a date to db.New is the right solution. I would personally separate the concerns of creating a client, and syncing the schema/generating mat views/etc.

stbenjam: Latest version is much better

main.go
stbenjam: It's not clear to me from that alternative how the historical instance will work.  The proposal says `This would make it simpler to keep a historical db live and dormant, and a sippy instance running to point to it`, but the reality is we would need a sippy instance + DB for each and every release.  I think we should clarify our plans about how we'll use this before adding this complexity.

main.go
```diff
@@ -136,6 +138,8 @@ func main() {
 	flags.StringVar(&opt.GoogleServiceAccountCredentialFile, "google-service-account-credential-file", os.Getenv("GOOGLE_APPLICATION_CREDENTIALS"), "location of a credential file described by https://cloud.google.com/docs/authentication/production")
 	flags.StringVar(&opt.GoogleOAuthClientCredentialFile, "google-oauth-credential-file", opt.GoogleOAuthClientCredentialFile, "location of a credential file described by https://developers.google.com/people/quickstart/go, setup from https://cloud.google.com/bigquery/docs/authentication/end-user-installed#client-credentials")
 
+	flags.StringVar(&opt.PinnedDateTime, "pinnedDateTime", opt.PinnedDateTime, "optional value to use instead of the current time specified in RFC3339 format - 2006-01-02 15:04:05+00:00")
```
dgoodwin, line 141: Let's add a note that this is for historical sippy instances.

dgoodwin: However do we actually need a CLI option? My thought here is this value can be read from the db, find the last prow job run timestamp.

pkg/db/db.go
```diff
@@ -28,7 +29,7 @@ type DB struct {
 	BatchSize int
 }
 
-func New(dsn string) (*DB, error) {
+func New(dsn string, timeNow time.Time) (*DB, error) {
```
dgoodwin, line 32: If you saved the "timeNow" on the actual DB struct, would that eliminate a whole lot of argument passing? I think by definition pretty much anything that will use it already has a *db.DB available.

pkg/db/query/payload_queries.go
```diff
@@ -57,13 +57,13 @@ func GetLastOSUpgradeByArchitectureAndStream(db *gorm.DB, release string) ([]mod
 }
 
 // GetLastPayloadTags returns payloads tags for last two weeks, sorted by date in descending order.
-func GetLastPayloadTags(db *gorm.DB, release, stream, arch string) ([]models.ReleaseTag, error) {
+func GetLastPayloadTags(db *gorm.DB, release, stream, arch string, timeNow time.Time) ([]models.ReleaseTag, error) {
```
dgoodwin, line 60: Ah to my earlier comment, here we're passing gorm.DB not our db.DB, but we could likely adjust it to take a db.DB pretty easily.

pkg/db/db.go
dgoodwin: If we decide to instead go with querying this value from the db last prow job run timestamp instead of a CLI arg, that could become a function on DB, or perhaps just a standalone func over in query package the that takes a DB.

main.go
neisw: Agree, thanks for the nudge :-)

main.go
neisw: I agree, this could well be a first pass / POC and if it isn't going in the right direction we could even drop it.  I wonder if we want to start creating more materialized views or should we start to consider summarizing data (possibly outside of postgress) for reporting purposes.  How much detailed data do we need for this goal, if we need all the same functionality then we keep it in postgress or move to a more scalable data store if need be, if we can live with summary data can we provide it in JSON or Redshift / Other.  I was thinking that providing a helper to fetch the time value would move us in the right direction and being able to pin a time value would be a first step in our goal but it may well be we need to step back and clarify the long term goal.

dgoodwin (commented):
Overall this looks great, I do suspect you could avoid some arg passing if you stored this on db.DB if you wanted to. 

I don't know if it's worth reading from db instead of a CLI flag, you'd probably need a --historical=true, or maybe we just default to ALWAYS using the last job run timestamp? I can live without this.

Lastly timeNow, is a little ambiguous to me. Could we come up with a better name to refer to this throughout? reportEnd is the best I can come up with.

main.go
dgoodwin: Mmm one historical we might be able to live with, but it will get expensive even with that. I guess the long term dream goal would be to be able to select GA snapshots in the UI/API, and if specified we'd lookup the timestamp for that and adjust everything accordingly. A UI dropdown for 4.11 GA would be cool.

If we decide to work towards that this PR is still very helpful in that it's identifying everywhere we need to adjust, and pushing it towards common code to get the value. I now see the logic to a CLI arg for now rather than looking up from a DB. My suggestion to store on db.DB might also be poor in that case too, and some kind of live query, based on requests params, might be best.

We'd have no solution for matviews though. They are programatically managed though, so we could perhaps dupe them and NOT update them all the time if they're for snapshots.

stbenjam: For historical sippy, we could tolerate it being a little bit slow.  Maybe if there's a pinned date/time, the query goes against a function (like jobs) instead of matview? Not sure *how* slow it'd currently be with that approach.

dgoodwin: Also the looming card to look at partitioning, that could make a function feasible.

neisw: Do we always have all the data we will need for historical purposes or would we ever need a separate db with older records?  

I could see how we could have a new table that defines a key like 4.12GA and a time value then we could drive the creation of matviews for the keys in the table from that.  They only get created / refreshed on startup or explicit historical refresh api call.  Would need to check for ones that have been deleted and cleanup as well.  When passed in on a query param we would consult those matviews.  Or something like that...

Or just go the function route if time isn't an issue.

Renamed timeNow, need to see if we think changing the apis to take the input param is a step forward or we prefer a different approach and hold this back for now.

neisw:
Setting this aside for now while we investigate https://issues.redhat.com/browse/TRT-490 more

/close

PR #566, opened 2022-09-02
Add metric for last data refresh

stbenjam:
Sippy stopped refreshing data recently because of TRT-533, it took 19
hours to spot it, so I think we need an alert to tell us when it's been
more than a couple of hours.  Job run data should get updated hourly.


dgoodwin:
Great idea.

PR #567, opened 2022-09-02
Only update phase when change detected from release controller

stbenjam:
TRT-533

If a release gets force accepted/rejected, the phase will change.
Previously we tried to refetch and regenerate the entire release tag
from the release controller when that happens, but the problem is Sippy
was re-creating all the job runs.  To avoid that, you would have to set
the ID on all the generated job runs to match what was in the DB, so
they would get updated and not created.

A simpler solution I believe is just to update the phase, as that's all
that's likely to change (and really all that we care about in TRT).

### DIFF
--- a/pkg/releasesync/releasesync.go
+++ b/pkg/releasesync/releasesync.go
@@ -11,11 +11,12 @@ import (
 	"strings"
 	"time"
 
+	log "github.com/sirupsen/logrus"
+	"gorm.io/gorm/clause"
+
 	"github.com/openshift/sippy/pkg/apis/api"
 	"github.com/openshift/sippy/pkg/db"
 	"github.com/openshift/sippy/pkg/db/models"
-	log "github.com/sirupsen/logrus"
-	"gorm.io/gorm/clause"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


pkg/releasesync/releasesync.go
```diff
@@ -54,20 +55,10 @@ func (r *releaseSyncOptions) Run() error {
...
-						}
-
-						// sync up the ids and created at
-						releaseTag.ID = mReleaseTag.ID
-						releaseTag.CreatedAt = mReleaseTag.CreatedAt
-
-						if err := r.db.DB.Clauses(clause.OnConflict{UpdateAll: true}).Table(releaseTagsTable).Save(releaseTag).Error; err != nil {
+						log.Warningf("Phase change detected (%q to %q) -- updating tag %s...\n", mReleaseTag.Phase, tag.Phase, tag.Name)
+						mReleaseTag.Phase = tag.Phase
+						if err := r.db.DB.Clauses(clause.OnConflict{UpdateAll: true}).Table(releaseTagsTable).Save(mReleaseTag).Error; err != nil {
```
dgoodwin, line 61: Nice job pinning this down. I was wondering how that happened. We only pull in a release tag that has completed, so job runs should never change, phase should be it. I think it makes sense to me.

dgoodwin (approved):
/lgtm

PR #576, opened 2022-09-08
A couple of UI tweaks

stbenjam:
- Bug counts are showing up on the widgets, making them too wide. 
- Icons in the link column on tests get pushed outside the column due to narrower width


dgoodwin:
My bad, thanks.

/lgtm

PR #580, opened 2022-09-09
Add timing logging to prowloader

stbenjam:
Devan added this in many places in Sippy, and the data is nice to help troubleshoot.  I spotted a case where fetchdata's last log entry was this:

```
2022/09/09 00:44:34 /go/src/sippy/pkg/prowloader/prow.go:155
[21.578ms] [rows:2416] SELECT * FROM "prow_pull_requests" WHERE merged_at IS NULL
```

That's a very quick query -- but without additional logging, I don't really know where Sippy got stuck.


stbenjam:
/hold

To fix devan's feedback

PR #581, opened 2022-09-09
install upgrade metrics

dgoodwin:
- Add install and upgrade success metrics.
- Publish install/upgrade success metrics for all variants.
- Add install and upgrade metrics for change in success rate vs previous week.

These metrics will allow us to add alerts on global install/upgrade success rates, per specific variants, and changes from one week to the next.

Includes a refactor to some duplicated code.


pkg/api/install.go
```diff
@@ -1,17 +1,110 @@
...
+		return
+	}
+
+	jsonStr := string(result)
 	RespondWithJSON(http.StatusOK, w, jsonStr)
 }
+
+// VariantTestsReport returns a set of all variant columns plus "All", and a map of testName to variant column to test results for that variant.
+// Caller can provide exact test names to match, test name prefixes, or test substrings.
+func VariantTestsReport(dbc *db.DB, release string, testNames, testPrefixes, testSubStrings sets.String) (sets.String, map[string]map[string]apitype.Test, error) {
```
dgoodwin, line 53: moved here from the *html packages and made generic

dgoodwin:
One small downside of this, startup is a little slow as we do this query per release. I tried to batch them but the api currently doesn't expose release, so that would need to be added.

stbenjam:
This is great! I don't think the metrics takes that much longer with this. My only concern is the naming:

```
sippy_install_success_delta_last_7d_vs_prev_7d{release="4.10",variant="All"}
```

This doesn't really match how we name other things -- are we able to do it like this instead? And include 2 day values?

```
sippy_install_success_delta_last{release="4.10",variant="All",period="default"}
sippy_install_success_delta_last{release="4.10",variant="All",period="twoDay"}
```

dgoodwin:
> This is great! I don't think the metrics takes that much longer with this. My only concern is the naming:
> 
> ```
> sippy_install_success_delta_last_7d_vs_prev_7d{release="4.10",variant="All"}
> ```
> 
> This doesn't really match how we name other things -- are we able to do it like this instead? And include 2 day values?
> 
> ```
> sippy_install_success_delta_last{release="4.10",variant="All",period="default"}
> sippy_install_success_delta_last{release="4.10",variant="All",period="twoDay"}
> ```

I started out with period but found that the TestReportsByVariant function we use for the indicators is 7 days only. That said I could adapt it for two day pretty easily. I'll try to get this in today.

dgoodwin:
Startup is now 52 seconds on my laptop, this feels too slow for development change testing, I'm going to keep working on that for a followup, or move the metrics refresh back to background (as we took another path for alert flapping), or add a dev flag to skip this.

dgoodwin:
ok here's the metric name and period change, I'm going to hold until startup speed I think.

dgoodwin:
/hold cancel

Pushed a commit to refresh in parallel and let the server start immediately. We took another path for preventing alert flapping.

pkg/db/query/test_queries.go
```diff
@@ -57,6 +58,7 @@ const (
 func TestReportsByVariant(
 	dbc *db.DB,
 	release string,
+	reportType v1.ReportType, // defaults to "current" or last 7 days vs prev 7 days
```
stbenjam, line 61: We should agree on the proper value for this, we both seem to use different terms - https://issues.redhat.com/browse/TRT-548. I noticed the metrics are now inconsistent since the ones I added recently use `default` as the period.

The UI always called this `default` and sent that to the API.  I tend to prefer `default`, as current is overloaded (the twoDay results are also "current"), but perhaps it should really be called sevenDay.   (The API interprets anything other than `twoDay` as the seven day report, so none of these discrepancies cause bugs yet.)

PR #590, opened 2022-09-14
Enable job runs pagination to be server-side

stbenjam:
Part of [TRT-314](https://issues.redhat.com//browse/TRT-314)

Loading the job runs page takes a lot of memory, since we have 80K rows
and we're sending them all to the client. It causes a big spike in Sippy
memory usage.

This enables server-side pagination for job runs. I only did one page since
it's the worst offender, but we should convert most API's I think.  Perhaps
a good issue for someone not familiar with sippy to tackle, since the blueprint
is in this PR for how to do the rest of the pages.


dgoodwin:
This is great, thanks.

/lgtm

PR #596, opened 2022-09-19
Remove build directory from the git repo

stbenjam:
`sippy-ng/build` holds the artifacts from `make build`, so we can
produce an all-in-one binary used in production. `go:embed` embeds the
build dir in the sippy binary but freaks out if the directory is empty.
So we have to have something.

I originally put an index.html so you get something helpful to
understand why the all-in-one binary wasn't working, but it causes tons
of problems since it's the same file that `make build` writes to and
people keep overwriting it.

You can't ignore changes with `.gitignore` to a file that's already in
the repo.

Therefore, this deletes the build directory and creates it with a
Makefile.  A user trying to do `go build .` directly will need to make
the directory themselves and populate it with at least 1 file.


stbenjam:
Well, this doesn't work.  Once you run `make build` it deletes `sippy-ng/build/README.md`

*sigh*

That's why I used index.html.

stbenjam:
@dgoodwin Any idea what we could do?

- If build directory is empty or doesn't exist, `go build .` explodes
- If build contains index.html, it gets overwritten
- If build contains README.md, it gets deleted

dgoodwin:
Looks like go build needs it there for static file embedding? Can we remove the whole dir from git, add to gitignore, create it in make targets if it doesn't exist. Raw go build will fail first time you checkout, but as soon as you've run a make build once it would always be there and populated with the generated files.

stbenjam:
> Looks like go build needs it there for static file embedding? Can we remove the whole dir from git, add to gitignore, create it in make targets if it doesn't exist. Raw go build will fail first time you checkout, but as soon as you've run a make build once it would always be there and populated with the generated files.

That's probably the least-bad option, updated.  I think this works...

PR #601, opened 2022-09-22
Make test name clickable from job anaylsis page

stbenjam:
[TRT-82](https://issues.redhat.com//browse/TRT-82)

In the select/view tests modal, this adds a link to the test analysis page for that test.


dgoodwin:
/lgtm

THanks!

PR #604, opened 2022-09-26
TRT-274:   Support Historical Sippy Views

neisw:
[TRT-274](https://issues.redhat.com//browse/TRT-274): Add support for setting a fixed date to show results.

### DIFF
--- a/pkg/api/job_runs.go
+++ b/pkg/api/job_runs.go
@@ -4,6 +4,7 @@ import (
 	"net/http"
 	gosort "sort"
 	"strconv"
+	"time"
 
 	apitype "github.com/openshift/sippy/pkg/apis/api"
 	"github.com/openshift/sippy/pkg/db"
@@ -44,7 +45,7 @@ func (runs apiRunResults) limit(req *http.Request) apiRunResults {
 type apiRunResults []apitype.JobRun
 
 // JobsRunsReportFromDB renders a filtered summary of matching jobs.
-func JobsRunsReportFromDB(dbc *db.DB, filterOpts *filter.FilterOptions, release string, pagination *apitype.Pagination) (*apitype.PaginationResult, error) {
+func JobsRunsReportFromDB(dbc *db.DB, filterOpts *filter.FilterOptions, release string, pagination *apitype.Pagination, reportEnd time.Time) (*apitype.PaginationResult, error) {
 	jobsResult := make([]apitype.JobRun, 0)
 	table := "prow_job_runs_report_matview"
 	q, err := filter.FilterableDBResult(dbc.DB.Table(table), filterOpts, apitype.JobRun{})
@@ -55,6 +56,8 @@ func JobsRunsReportFromDB(dbc *db.DB, filterOpts *filter.FilterOptions, release 
 	if len(release) > 0 {
 		q = q.Where("release = ?", release)
 	}
+
+	q = q.Where("timestamp < ?", reportEnd.UnixMilli())
 
 	// Get the row count before pagination
 	var rowCount int64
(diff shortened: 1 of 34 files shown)


pkg/api/job_runs.go
```diff
@@ -44,14 +45,19 @@ func (runs apiRunResults) limit(req *http.Request) apiRunResults {
...
 	table := "prow_job_runs_report_matview"
 	q, err := filter.FilterableDBResult(dbc.DB.Table(table), filterOpts, apitype.JobRun{})
 	if err != nil {
 		return nil, err
 	}
-	q = q.Where("release = ?", release)
+
+	if len(release) > 0 {
+		q = q.Where("release = ?", release)
+	}
```
stbenjam, line 58: Would you mind moving this to a separate PR? I would rather land just this right away.

pkg/db/db.go
```diff
@@ -28,7 +29,7 @@ type DB struct {
 	BatchSize int
 }
 
-func New(dsn string) (*DB, error) {
+func New(dsn string, reportEnd *time.Time) (*DB, error) {
```
stbenjam, line 32: I think it'd be cleaner if we separated the concerns of creating a DB client, and handling migrations in a Migrate() function.  You don't have to do it since this is a pre-existing thing but it would be nicer I think.

pkg/db/matviews.go
```diff
@@ -3,29 +3,34 @@ package db
 import (
 	"fmt"
 	"strings"
+	"time"
 
 	"gorm.io/gorm"
 )
 
+const ReplaceTimeNow = "|||TIMENOW|||"
+const TimestampFormat = "2006-01-02 15:04:05"
```
stbenjam, line 12: Are these only used here? They should start with a lower case character, otherwise they're exported

pkg/db/query/payload_queries.go
```diff
@@ -87,7 +89,7 @@ func GetLastPayloadStatus(db *gorm.DB, architecture, stream, release string) (st
 				FROM
 					release_tags
 				WHERE
-					architecture = ? AND stream = ? AND release = ?
+					architecture = ? AND stream = ? AND release = ? AND release_time < ?
```
stbenjam, line 92: The matviews do `<=` and `>=`, should these all be inclusive too?

sippy-ng/src/App.js
```diff
@@ -120,26 +120,38 @@ export default function App(props) {
...
   const [fetchError, setFetchError] = React.useState('')
 
   const fetchData = () => {
     Promise.all([
       fetch(process.env.REACT_APP_API_URL + '/api/releases'),
       fetch(process.env.REACT_APP_API_URL + '/api/capabilities'),
+      fetch(process.env.REACT_APP_API_URL + '/api/report_date'),
     ])
-      .then(([releases, capabilities]) => {
+      .then(([releases, capabilities, reportDate]) => {
```
stbenjam, line 132: Instead of wiring this in many places, I think it might be more appropriate to use a context and wrap the whole app like Capabilities.  What do you think?

stbenjam (commented):
A few comments but looks pretty good.

pkg/db/db.go
neisw: I'm definitely open to some refactoring to consolidate / cleanup.  I could create a story about this and the utility methods for start, end and boundary we touched on in the previous pr that we closed without merging.  Would prefer to get this in an agreeable state first though then refactor after merging if ok.

pkg/db/matviews.go
neisw: yep, thanks!

sippy-ng/src/App.js
neisw: Agree, thanks for the pointer.  Let me review & refactor.

PR #615, opened 2022-10-03
Only call GitHub once per PR

stbenjam:
[TRT-585](https://issues.redhat.com//browse/TRT-585)

We track PR's based off the PR and the HEAD SHA.  We used to query GitHub for each SHA, instead of just once per PR. This adjusts the caching and syncing logic to ensure we only hit the API once for a PR.

GitHub limits us to 5,000 requests per hour. We know more SHA's than that now, so we're hitting the limits.  We have much fewer distinct PR's, so this should fix the rate limiting.

### DIFF
--- a/pkg/prowloader/github/github.go
+++ b/pkg/prowloader/github/github.go
@@ -7,7 +7,7 @@ import (
 	"time"
 
 	gh "github.com/google/go-github/v45/github"
-	"github.com/sirupsen/logrus"
+	log "github.com/sirupsen/logrus"
 	"github.com/tcnksm/go-gitconfig"
 	"golang.org/x/oauth2"
 )
@@ -24,12 +24,10 @@ type prentry struct {
 }
 
 type Client struct {
-	ctx   context.Context
-	cache map[prlocator]*prentry
+	ctx     context.Context
+	cache   map[prlocator]*prentry
+	prFetch func(org, repo string, number int) (*gh.PullRequest, error)
 }
-
-// PRFetch as a global allows tests to override.
-var PRFetch func(org, repo string, number int) (*gh.PullRequest, error)
 
 func New(ctx context.Context) *Client {
 	client := &Client{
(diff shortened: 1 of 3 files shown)


pkg/prowloader/github/github.go
```diff
@@ -39,39 +46,73 @@ func New(ctx context.Context) *Client {
...
+	} else if ok {
+		// If it's in the cache, but this isn't the sha, this sha wasn't merged.
+		return nil, nil
 	}
 
-	pr, _, err := c.client.PullRequests.Get(c.ctx, org, repo, number)
+	// Get PR from GitHub
+	pr, err := PRFetch(org, repo, number)
 	if err != nil {
+		if resp, ok := err.(*gh.ErrorResponse); ok && resp.Response != nil && resp.Response.StatusCode == http.StatusNotFound {
```
dgoodwin, line 93: Worth logging as an error or warning?

pkg/prowloader/github/github.go
```diff
@@ -39,39 +46,73 @@ func New(ctx context.Context) *Client {
...
+//   - the PR is unmerged (cache as nil)
+//   - the PR is merged with a different SHA (cache with the merged sha, return nil)
+//   - the PR is merged with the same SHA (cache with the merged sha, return merged time)
+func (c *Client) GetPRSHAMerged(org, repo string, number int, sha string) (*time.Time, error) {
+	prl := prlocator{org: org, repo: repo, number: number}
+	if val, ok := c.cache[prl]; ok && val != nil && val.sha == sha {
+		// If it's in the cache, and sha matches, this sha was merged.
+		return val.mergedAt, nil
+	} else if ok {
+		// If it's in the cache, but this isn't the sha, this sha wasn't merged.
```
dgoodwin, line 86:
```suggestion
		// If it's in the cache, but this isn't the sha, the PR merged but not with this sha.
```

pkg/prowloader/github/github.go
```diff
@@ -39,39 +46,73 @@ func New(ctx context.Context) *Client {
...
 
 	return client
 }
 
-func (c *Client) GetPRMerged(org, repo string, number int, sha string) (*time.Time, error) {
-	prl := prlocator{org: org, repo: repo, number: number, sha: sha}
-	if val, ok := c.cache[prl]; ok {
-		return val, nil
+// GetPRSHAMerged returns the merge time for a PR/SHA combination. The caching is designed
+// to minimize queries to GitHub. We basically have to handle these cases:
```
dgoodwin, line 75: Did you consider just using the db as the cache? Too slow perhaps? (individual queries per call here)

The cache looks unbounded, it would grow forever as long as sippy was running as new PRs come and go. Surely we'd restart often enough to prevent an OOM though so probably good for now. We could expire someday if it becomes a problem.

dgoodwin (commented):
The logic around reducing calls seems like it should work to me.

pkg/prowloader/github/github.go
stbenjam: We only use this with --load-database so it's lifetime is limited

pkg/prowloader/github/github.go
```diff
@@ -15,19 +16,25 @@ type prlocator struct {
...
 type Client struct {
-	ctx    context.Context
-	client *gh.Client
-	cache  map[prlocator]*time.Time
+	ctx   context.Context
+	cache map[prlocator]*prentry
 }
 
+// PRFetch as a global allows tests to override.
+var PRFetch func(org, repo string, number int) (*gh.PullRequest, error)
```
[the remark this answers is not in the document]
stbenjam: Sounds better

pkg/prowloader/github/github.go
stbenjam: The DB also has the old state, not the new state, so using it as a cache would be tough.

pkg/prowloader/prow.go
```diff
@@ -152,12 +152,14 @@ func (pl *ProwLoader) syncPRStatus() error {
 	}
 
 	pulls := make([]models.ProwPullRequest, 0)
-	if res := pl.dbc.DB.Table("prow_pull_requests").Where("merged_at IS NULL").Scan(&pulls); res.Error != nil && !errors.Is(res.Error, gorm.ErrRecordNotFound) {
+	if res := pl.dbc.DB.
+		Table("prow_pull_requests").
+		Where("merged_at IS NULL").Scan(&pulls); res.Error != nil && !errors.Is(res.Error, gorm.ErrRecordNotFound) {
```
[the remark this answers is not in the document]
stbenjam: Yea we're only looking at unmerged PR's.  There's about 1,200.  With the bot closing rotten ones, this shouldn't grow too quickly, but we may need to figure out what to do once the number gets higher.

stbenjam:
@dgoodwin I addressed your comments

pkg/prowloader/github/github.go
```diff
@@ -15,63 +16,106 @@ type prlocator struct {
...
+//   - the PR is unmerged (cache as nil)
+//   - the PR is merged with a different SHA (cache with the merged sha, return nil)
+//   - the PR is merged with the same SHA (cache with the merged sha, return merged time)
+func (c *Client) GetPRSHAMerged(org, repo string, number int, sha string) (*time.Time, error) {
+	prl := prlocator{org: org, repo: repo, number: number}
+	if val, ok := c.cache[prl]; ok && val != nil && val.sha == sha {
+		// If it's in the cache, and sha matches, this sha was merged.
+		return val.mergedAt, nil
+	} else if ok {
+		// If it's in the cache, but this isn't the sha, the PR merged but not with this sha.
```
xueqzhan, line 82: This should also have the unmerged case, right? We set the cache to nil below at line 113. Also, if during the database loading, subsequent PR updates will not be fetched until next run. How often do we run the db load now?

stbenjam: We run `--load-database` hourly, it's a one-shot thing, it runs and then exits -- the cache only lives that long.

I fixed the comment, thanks!

stbenjam:
/retest

CI got stuck somehow

PR #616, opened 2022-10-03
Sync release tag associations in batches

stbenjam:
[TRT-585](https://issues.redhat.com//browse/TRT-585)

When creating release tags that include multiple revendors of libraries
with large numbers of commits (say, kube), we're trying to create it
all in one single INSERT which doesn't work well. We need to tell gorm
to create things in batches.

`CreateBatchSize` works for releations (in this case PR's).


stbenjam:
@dgoodwin This one as well if you have a chance

PR #619, opened 2022-10-04
Get issues by variants

xueqzhan:
Jira issues are annotated with incident=variants=xxxx. We will get jira cards with this annotation and parse the variants. Any jobs whose variants is a superset of the issue variant will include the corresponding issue in bug_jobs table.

### DIFF
--- a/pkg/db/loader/issue_types_internal.go
+++ b/pkg/db/loader/issue_types_internal.go
@@ -19,6 +19,8 @@ type Result struct {
 
 type Match struct {
 	Bug bugsv1.Bug `json:"bugInfo"`
+	// Context contains a list of all matched instances for the search regex
+	Context []string `json:"context"`
 	// Issues contains data on the jira issue. While plural, it can only contain one result from search.ci.
 	Issues gojira.Issue `json:"issues"`
 }
--- a/pkg/db/loader/issues.go
+++ b/pkg/db/loader/issues.go
@@ -13,12 +13,20 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
+var (
+	// VariantSearchRegex defines the search regex for search.ci
+	VariantSearchRegex = "sippy-link=\\[variants=(\\S+)\\]"
+	// by default maxMatches is 1 for search.ci API. Since we are doing regex match, we pick 100 as the default.
+	// This should be decided by the number of combination of variants.
+	regexMaxMatches = "100"
+)
+
 // FindIssuesForTests queries search.ci for Jira issues mapping based to the given test names.
 func FindIssuesForTests(testNames ...string) (map[string][]jira.Issue, error) {
 
 	issues := map[string][]jira.Issue{}
 
-	newBugs, lastUpdateError := findBugsForSearchStrings(testNames...)
+	newBugs, lastUpdateError := findBugsForSearchStrings(false, testNames...)
 
 	for testName, bug := range newBugs {
 		issues[testName] = bug
@@ -38,7 +46,7 @@ func FindIssuesForJobs(jobNames ...string) (map[string][]jira.Issue, error) {
 		jobSearchStrings[i] = fmt.Sprintf("job=%s=all", jn)
 	}
 
-	newBugs, lastUpdateError := findBugsForSearchStrings(jobSearchStrings...)
+	newBugs, lastUpdateError := findBugsForSearchStrings(false, jobSearchStrings...)
 
 	for jobKey, bug := range newBugs {
 		issues[jobKey] = bug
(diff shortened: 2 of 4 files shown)


pkg/db/loader/issue_types_internal.go
```diff
@@ -18,7 +18,8 @@ type Result struct {
 }
 
 type Match struct {
-	Bug bugsv1.Bug `json:"bugInfo"`
+	Bug     bugsv1.Bug `json:"bugInfo"`
+	Context []string   `json:"context"`
```
stbenjam, line 22: Can you add godoc that explains what a context is?

stbenjam (commented):
Will took a closer look later but a couple of nits

pkg/db/loader/issues.go
```diff
@@ -13,12 +13,20 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
+var (
+	// VariantSearchRegex defines the search regex for search.ci
+	VariantSearchRegex = "incident=variants=(\\S+)"
```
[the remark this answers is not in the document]
dgoodwin: Am I correct it is illogical to have variant support for a test=? Jobs are tied to variants, tests less so, there can still be a problem with a test on specific variants but it doesn't seem like we'd care too much, just link the test in general and be clear in the bug?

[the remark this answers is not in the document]
dgoodwin: Couple thoughts on the above idea: https://issues.redhat.com/browse/TRT-597, some of my examples above probably make no sense to combine.

pkg/sippyserver/dbloader.go
```diff
@@ -334,6 +335,67 @@ func getOrCreateTestID(
...
+	copy(v, variants)
+	sort.Strings(v)
+	return strings.Join(v, ",")
+}
+
+func appendJobIssuesFromVariants(jobCache map[string]*models.ProwJob, jobIssues map[string][]jira.Issue) error {
+	// variantSetMap maps a sorted names of variants to a set of the variants for easy comparison
+	variantsSetMap := map[string]sets.String{}
+	// variantsIssuesMap maps a sorted names of variants to a slice of issues
+	variantsIssuesMap := map[string][]jira.Issue{}
```
dgoodwin, line 349: I'm thinking this code needs to check the AffectsVersion on the Jira issue, and only match jobs in that release. I believe it's multi-valued so if the jira had multiple versions affected, we can tie them to all.

pkg/sippyserver/dbloader.go
```diff
@@ -334,6 +335,67 @@ func getOrCreateTestID(
...
 }
 
+func variantsKey(variants []string) string {
+	v := make([]string, len(variants))
+	copy(v, variants)
+	sort.Strings(v)
+	return strings.Join(v, ",")
+}
+
+func appendJobIssuesFromVariants(jobCache map[string]*models.ProwJob, jobIssues map[string][]jira.Issue) error {
```
dgoodwin, line 345: Can we cover this function with unit tests, no db required.

pkg/db/loader/issue_types_internal.go
xueqzhan: I did some digging, this shows that the context should be the list of matches: https://github.com/openshift/ci-search/blob/30244af43eb695ff156d2d27b9a509a117156017/cmd/search/httpsearch.go#L165. I simply used it without finding the source. Now we know. I will document this.

pkg/db/loader/issues.go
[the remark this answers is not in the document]
xueqzhan: Agree with your comment on variant for tests. Tests themselves do not have association with variants and therefore they are linked to issues only by the test= key words.

pkg/sippyserver/dbloader.go
```diff
@@ -383,13 +385,23 @@ func appendJobIssuesFromVariants(jobCache map[string]*models.ProwJob, jobIssues
...
-					jobSearchStrings := fmt.Sprintf("job=%s=all", job.Name)
-					if _, ok := jobIssues[jobSearchStrings]; !ok {
-						jobIssues[jobSearchStrings] = []jira.Issue{}
+				if !variantsSetMap[variantsKey].IsSuperset(variantsSetMap[key]) {
+					continue
+				}
+				candidates := []jira.Issue{}
+				for _, issue := range issues {
+					for _, version := range issue.Fields.AffectsVersions {
+						if job.Release == version.Name {
```
dgoodwin, line 394: AffectsVersions can be 4.12, 4.12.z, 4.12.1

We should be covered with something like:

if job.Release == version.Name || strings.HasPrefix(version.Name, job.Release + ".") 

Also trying to prevent 4.1 matching 4.12.

pkg/sippyserver/dbloader_test.go
```diff
@@ -0,0 +1,212 @@
...
+)
+
+func findIssuesForVariants() (map[string][]jira.Issue, error) {
+	result := map[string][]jira.Issue{
+		"variants=azure,ovn": []jira.Issue{
+			{
+				ID: "1",
+				Fields: &jira.IssueFields{
+					AffectsVersions: []*jira.AffectsVersion{
+						{Name: "4.12"},
```
dgoodwin, line 20: With above comment in mind, maybe change this to 4.12.z and make sure we still get our match.

dgoodwin (commented):
I think that's it for me, @stbenjam do you want another look?

dgoodwin:
Do we want to discuss the logistics of doing something like https://issues.redhat.com/browse/TRT-597, and if we're going to proceed, settle on a syntax and apply it in this PR just so we don't have to go update references once this rolls out? I'll raise after scrum today.

PR #635, opened 2022-10-14
Cascade delete for all relevant DB models

stbenjam:
[TRT-599](https://issues.redhat.com//browse/TRT-599)

Note: proposed process after merging is 1) scale openshift and kube sippys down, 2) run scripts/clear-constraints-for-cascading.sql, 3) wait for image to be rebuilt, 4) scale back up

When deleting a DB row that has relations, these are not automatically deleted and the foreign key constraints prevent you from actually deleting the row. We need to cascade the deletes, however gorm does not support patching constraints[1], therefore we have to delete them.

There's a one-time SQL script in scripts/ that we can run to clear the constraints, which gorm wil re-add back. I've tested this on a production database restore and it seems to do what I expect. We could write our own migrator, but it looks somewhat challenging, and this is probably a one-ish time thing.

[1] https://github.com/go-gorm/gorm/issues/5559#issuecomment-1200718658

### DIFF
--- a/pkg/db/models/prow.go
+++ b/pkg/db/models/prow.go
@@ -23,7 +23,8 @@ type ProwJob struct {
 	Release     string         `gorm:"varchar(10)"`
 	Variants    pq.StringArray `gorm:"type:text[]"`
 	TestGridURL string
-	Bugs        []Bug `gorm:"many2many:bug_jobs;"`
+	Bugs        []Bug        `gorm:"many2many:bug_jobs;"`
+	JobRuns     []ProwJobRun `gorm:"constraint:OnDelete:CASCADE;"`
 }
 
 // IDName is a partial struct to query limited fields we need for caching. Can be used
@@ -46,8 +47,8 @@ type ProwJobRun struct {
 
 	URL          string
 	TestFailures int
-	Tests        []ProwJobRunTest
-	PullRequests []ProwPullRequest `gorm:"many2many:prow_job_run_prow_pull_requests;"`
+	Tests        []ProwJobRunTest  `gorm:"constraint:OnDelete:CASCADE;"`
+	PullRequests []ProwPullRequest `gorm:"many2many:prow_job_run_prow_pull_requests;constraint:OnDelete:CASCADE;"`
 	Failed       bool
 	// InfrastructureFailure is true if the job run failed, for reasons which appear to be related to test/CI infra.
 	InfrastructureFailure bool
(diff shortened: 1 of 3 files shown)


pkg/db/models/prow.go
```diff
@@ -82,12 +83,12 @@ type ProwJobRunTest struct {
...
 	// ProwJobRunTestOutput collect the output of a failed test run. This is stored as a separate object in the DB, so
 	// we can keep the test result for a longer period of time than we keep the full failure output.
-	ProwJobRunTestOutput *ProwJobRunTestOutput
+	ProwJobRunTestOutput *ProwJobRunTestOutput `gorm:"constraint:OnDelete:CASCADE;"`
 }
 
 type ProwJobRunTestOutput struct {
 	gorm.Model
-	ProwJobRunTestID uint `gorm:"constraint:OnDelete:SET NULL;"`
+	ProwJobRunTestID uint `gorm:"index"`
```
dgoodwin, line 91: Is it intentional to remove the index here?

pkg/db/models/prow.go
```diff
@@ -132,8 +133,8 @@ type Bug struct {
 	FixVersions     pq.StringArray `json:"fix_versions" gorm:"type:text[]"`
 	Components      pq.StringArray `json:"components" gorm:"type:text[]"`
 	URL             string         `json:"url"`
-	Tests           []Test         `json:"-" gorm:"many2many:bug_tests;"`
-	Jobs            []ProwJob      `json:"-" gorm:"many2many:bug_jobs;"`
+	Tests           []Test         `json:"-" gorm:"many2many:bug_tests;constraint:OnDelete:CASCADE;"`
+	Jobs            []ProwJob      `json:"-" gorm:"many2many:bug_jobs;constraint:OnDelete:CASCADE;"`
```
dgoodwin, line 137: Does this *only* affect the bug_tests/bug_jobs entries or will it flow to actual bugs and jobs? If it does we've got an issue, stale bugs from search.ci that are no longer in the results there but are in our db will get cleaned out, and right now I think that happens often if the bug hasn't been touched in 3 weeks. (we need to talk to Hoxna about that actually) Disappearing from the query would be fairly common.

dgoodwin (commented):
I know we keep flip flopping but I'm still nervous about letting gorm manage that schema.

pkg/db/models/prow.go
stbenjam: I'm removing the constraint (it's in the wrong place) and adding an index.

stbenjam:
> I know we keep flip flopping but I'm still nervous about letting gorm manage that schema.

If we're using gorm then it's going to manage foreign keys, but one of it's limitations is it won't actually update FK's after their created. So it's not really even managing it, just creating it once.  We'll have to run my script in this PR to clear the FK's and then gorm will create them with cascades.

The other option is to write a constraint migrator ourselves, but this is basically a one or two-time thing, and there's a lot of complexity to get it right.

pkg/db/models/prow.go
dgoodwin: Perfect thanks.

PR #639, opened 2022-10-17
Add API for analyzing risk of test failures in a job run

dgoodwin:
Done for [TRT-591](https://issues.redhat.com//browse/TRT-591) as part of epic TRT-601. 

This API will analyze the test failures within a job run, classify each as high/medium/low risk based on their pass rates over the past week, and set an overall risk for the jobrun itself. Future logic may be added to make this more intelligent in the future. No test failures = no risk. Mass test failures = high risk, regardless if they were all low risk failures. (we do not analyze each test failure if more than 20) 

Risk levels are represented as an integer for programatic interpretation, as well as a human readable string. Determinations are also accompanied by human readable "reasons" for display.

If we cannot find test results over past week for a test failure, we classify as Unknown, and that is rated in risk between medium and high.

Goal here is to codify knowledge about how serious a test failure was to help openshift engineers across the org decipher job run failures, especially on presubmits. We aim to have this included as an artifact, displayed in spyglass, and ultimately commented on PRs if we can pull it off.

### DIFF
--- a/pkg/api/job_runs.go
+++ b/pkg/api/job_runs.go
@@ -171,8 +171,8 @@ func JobRunAnalysis(dbc *db.DB, jobRunID int64) (apitype.ProwJobRunFailureAnalys
 					Risk: apitype.FailureRisk{
 						Level: getSeverityLevelForPassRate(tr.CurrentPassPercentage),
 						Reasons: []string{
-							fmt.Sprintf("This test has passed %.2f%% of %d runs on %s %v in the last week.",
-								tr.CurrentPassPercentage, tr.CurrentRuns, jobRun.ProwJob.Release, tr.Variants),
+							fmt.Sprintf("This test has passed %.2f%% of %d runs on release %s %v in the last week.",
+								tr.CurrentPassPercentage, tr.CurrentRuns, release, tr.Variants),
 						},
 					},
 				})
--- a/pkg/api/job_runs.go
+++ b/pkg/api/job_runs.go
@@ -116,6 +116,8 @@ func JobRunAnalysis(dbc *db.DB, jobRunID int64) (apitype.ProwJobRunFailureAnalys
 	// TODO: Fail out with a high sev failure if we see more than X failed tests, we don't want to do 3k queries below
 	// for effectively no purpose. 10 or 20 perhaps.
 
+	// TODO: Return early if no tests failed, this API shouldn't have been called?
+
 	// Get the test failures
 	for _, ft := range jobRun.Tests {
 		// TODO: filter test names we don't care about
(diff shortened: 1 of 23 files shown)


dgoodwin:
cc @neisw @stbenjam

pkg/api/job_runs.go
```diff
@@ -81,3 +86,182 @@ func JobsRunsReportFromDB(dbc *db.DB, filterOpts *filter.FilterOptions, release
...
+		// Get latest release from the DB:
+		ar, err := query.ReleasesFromDB(dbc)
+		if err != nil {
+			return apitype.ProwJobRunFailureAnalysis{}, err
+		}
+		if len(ar) == 0 {
+			return apitype.ProwJobRunFailureAnalysis{}, fmt.Errorf("no releases found in db")
+		}
+
+		compareRelease = ar[0].Release
```
neisw, line 120: Any issues here when we are setting up a new release but haven't switched yet>

dgoodwin: Excellent question, that did not occur to me. 

I think the way that works (IIRC), dptp tends to make release-4.13 git branches in all repos first, and they're automatically kept in sync with master. New jobs get created with 4.13 in the name, but sippy won't start scraping them until we explicitly tell it about 4.13. During that time, presubmits are technically 4.13 if it's against a master branch, but we'd be comparing to test pass rates for 4.12. I think that's probably actually ok and the best we could hope for. Seems like it will be ok but it's not guaranteed.

pkg/api/job_runs.go
```diff
@@ -81,3 +86,182 @@ func JobsRunsReportFromDB(dbc *db.DB, filterOpts *filter.FilterOptions, release
...
+	// Return early if we see mass test failures:
+	case len(jobRun.Tests) > 20:
+		response.OverallRisk.Level = apitype.FailureRiskLevelHigh
+		response.OverallRisk.Reasons = append(response.OverallRisk.Reasons,
+			fmt.Sprintf("%d tests failed in this run: High", len(jobRun.Tests)))
+		return response, nil
+	}
+
+	// Iterate each failed test, query it's pass rates by NURPs, find a matching variant combo, and
+	// see how often we've passed in the last week.
```
neisw, line 206: When we get to the bot scoring failed tests before they are in sippy won't we need the logic within this loop broken out?  Wondering if we should do that now, pass in the testID and Test.Name (bot will have to lookup the ID presumably)

pkg/api/job_runs.go
```diff
@@ -81,3 +86,182 @@ func JobsRunsReportFromDB(dbc *db.DB, filterOpts *filter.FilterOptions, release
...
+						fmt.Sprintf("Unable to find matching test results for variants: %v",
+							jobRun.ProwJob.Variants),
+					},
+				},
+			})
+		}
+	}
+
+	// Set the overall risk level for this job run to the highest we encountered:
+	var maxTestRiskReason string
```
neisw, line 245: If we could calculate this each time we call the scoring api (if we add that) then we wouldn't have to post process

pkg/api/job_runs.go
dgoodwin: Definitely needs to happen, but I'd probably go for doing that refactor once I can find the time to dive into the next card more deeply.

pkg/api/job_runs.go
dgoodwin: Good idea, done!

PR #642, opened 2022-10-18
Allow use of environment variables for sippy e2e test

DennisPeriquet:
This helps out [TRT-246](https://issues.redhat.com//browse/TRT-246) and let's us remove the sed in the sippy-e2e-commands.sh script in [this PR](https://github.com/openshift/release/pull/32818/files).

If no environment variables are set, you get the existing behavior.

If you do this, you will get to use those values:

```
export SIPPY_ENDPOINT=sippy-server-postgres.apps-crc.testing
export SIPPY_API_PORT=80
```

### DIFF
--- a/test/e2e/util/e2erequest.go
+++ b/test/e2e/util/e2erequest.go
@@ -4,7 +4,10 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
+	"net"
 	"net/http"
+	"os"
+	"strconv"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


test/e2e/util/e2erequest.go
```diff
@@ -17,7 +19,20 @@ const (
...
+	if len(env_sippy_api_port) > 0 {
+		val, err := strconv.Atoi(env_sippy_api_port)
+		if err == nil {
+			port = val
+		}
+	}
+	if len(env_sippy_endpoint) == 0 {
+		env_sippy_endpoint = "localhost"
+	}
+	return fmt.Sprintf("http://%s:%d%s", env_sippy_endpoint, port, apiPath)
```
stbenjam, line 35: Please don't construct URL host:ports with Sprintf -- it's an antipattern that results in bugs with IPv6 (https://github.com/stbenjam/no-sprintf-host-port#no-sprintf-host-port), use https://pkg.go.dev/net#JoinHostPort instead

test/e2e/util/e2erequest.go
```diff
@@ -17,7 +19,20 @@ const (
 )
 
 func buildURL(apiPath string) string {
-	return fmt.Sprintf("http://localhost:%d%s", APIPort, apiPath)
+	env_sippy_api_port := os.Getenv("SIPPY_API_PORT")
+	env_sippy_endpoint := os.Getenv("SIPPY_ENDPOINT")
```
stbenjam, line 23: nit, go style is mixed caps

https://go.dev/doc/effective_go#mixed-caps

test/e2e/util/e2erequest.go
DennisPeriquet: oh, 100% agree thanks -- will fix.

test/e2e/util/e2erequest.go
DennisPeriquet: LOL, too much bash lately -- will fix.

stbenjam:
Looks like lint is even pickier than me:
>  test/e2e/util/e2erequest.go:23:2: var-naming: var envSippyApiPort should be envSippyAPIPort (revive)

PR #650, opened 2022-10-21
[WIP] Estimate changepoint for job analysis

stbenjam:
Changepoint analysis allows us to detect roughly when a change occurred, which is not always simple to spot for an unreliable job. It uses a go package to do [ED-PELT](https://aakinshin.net/posts/edpelt/) changepoint analysis on our job pass rates.  The UI will plot dashed white lines on the job analysis showing the changepoints.

Limitations: the current approach only looks at success %.  We won't detect a transition from e2e failure to install failure, if the rates end up being similar. I tried to look at all possible job states (install failure, upgrade, etc) but it detected too many transitions.

If this ends up being useful, we could also consider trying [the FCML-CP algorithm](https://www.sciencedirect.com/science/article/pii/S037704271630276X) which might be a more accurate fit for our data (but requires writing a library to implement it).

Some examples:
- periodic-ci-openshift-release-master-ci-4.11-upgrade-from-stable-4.10-e2e-azure-upgrade: This one is interesting because I probably would've expected a line around 8/22 as well, but it didn't.
image

- periodic-ci-shiftstack-shiftstack-ci-main-periodic-4.12-e2e-openstack-csi-cinder
image

- periodic-ci-openshift-release-master-nightly-4.12-e2e-metal-ipi-sdn-bm
image


dgoodwin:
This is awesome, is it computationally expensive?

stbenjam:
It's not computationally expensive, this library is pretty quick, but it's not a great algorithm. I may fiddle with it on another hack day, and try to grok the statistical stuff better. 

/close

PR #657, opened 2022-10-26
transient risk analysis

dgoodwin:
- Rename API to risk_analysis for clarity.
- Working POST requests for transient risk analysis of jobs we don't have.
- Hookup bug data for transient risk analysis requests.
- Stick to GET instead of POST.


dgoodwin:
/assign @neisw 

Forrest could you take a pass at this sometime this week and let me know what you think?

PR #664, opened 2022-10-31
Get sippy ready for 4.13

DennisPeriquet:
Regarding TRT-648.

I wonder if deploymentconfig.yaml is used because the one in [continuous-release-jobs repo](https://github.com/openshift/continuous-release-jobs/blob/ad0c0f453828ddbd5a96a55bcbd38a0b0e67b1ea/config/clusters/dpcr/services/sippy/deploymentconfig.yaml) seems the be the in use when we deploy on the DPCR cluster.  I include it in this PR and if it makes sense to delete it, I will.

this PR is related: https://github.com/openshift/continuous-release-jobs/pull/1237

### DIFF
--- a/resources/deploymentconfig.yaml
+++ b/resources/deploymentconfig.yaml
@@ -46,6 +46,10 @@ spec:
         - --local-data
         - /data
         - --release
+        - "4.13"
+        - --release
+        - "4.12"
+        - --release
         - "4.11"
         - --release
         - "4.10"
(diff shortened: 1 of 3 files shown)


stbenjam:
The deplyomentconfig and related manifests aren't used anymore, perhaps we can delete it all, and replace it with a README that links to the DPCR repo?

DennisPeriquet:
> The deplyomentconfig and related manifests aren't used anymore, perhaps we can delete it all, and replace it with a README that links to the DPCR repo?

Can I just delete the entire resources contents and add a blurb about deployment in README.md.

PR #669, opened 2022-11-02
Handle case where release controller does not report a URL

stbenjam:
Currently this 4.13 payload
(https://amd64.ocp.releases.ci.openshift.org/api/v1/releasestream/4.13.0-0.ci/release/4.13.0-0.ci-2022-11-02-023326) is not reporting a URL for one of the jobs, resulting in follow-on effects such as trying to insert a row where the job run id is 0.


stbenjam:
/override e2e

I tested this locally and sippy is broken

PR #673, opened 2022-11-04
Fix risk analysis by prow job ID.

dgoodwin:
Bad logic here, we didn't fork properly and instead tried to find a prow
job in the body of the request right after, when you gave us the id to
use.

### DIFF
--- a/pkg/sippyserver/server.go
+++ b/pkg/sippyserver/server.go
@@ -837,26 +837,26 @@ func (s *Server) jsonJobRunRiskAnalysis(w http.ResponseWriter, req *http.Request
 				"code": http.StatusBadRequest, "message": res.Error.Error()})
 			return
 		}
+	} else {
+		err := json.NewDecoder(req.Body).Decode(&jobRun)
+		if err != nil {
+			api.RespondWithJSON(http.StatusBadRequest, w, map[string]interface{}{
+				"code":    http.StatusBadRequest,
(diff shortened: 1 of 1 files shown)


pkg/sippyserver/server.go
```diff
@@ -837,27 +837,27 @@ func (s *Server) jsonJobRunRiskAnalysis(w http.ResponseWriter, req *http.Request
...
+		if err != nil {
+			api.RespondWithJSON(http.StatusBadRequest, w, map[string]interface{}{
+				"code":    http.StatusBadRequest,
+				"message": fmt.Sprintf("error decoding prow job run json in request body: %s", err)})
+			return
+		}
+		// We don't expect the caller to fully populate the ProwJob, just it's name,
+		// override the input by looking up the actual ProwJob so we have access to release and variants.
+		job := &models.ProwJob{}
+		res := s.db.DB.Where("name = ?", jobRun.ProwJob.Name).First(&job)
```
stbenjam, line 851: Why not just make job `models.ProwJob{}`?

Job is already a pointer, so I'm wondering if this code works as expected? I think you're passing in a pointer to a pointer (two levels of indirection).

dgoodwin: Good catch. It works, but it doesn't look good. I've removed and pointer here and kept it on the job variable.

PR #679, opened 2022-11-09
TRT-668: linter updates

neisw:


pkg/prowloader/prow.go
```diff
@@ -181,6 +181,8 @@ func fetchJobsJSON(prowURL string) ([]byte, error) {
 	if err != nil {
 		return nil, err
 	}
+
```
neisw, line 184: Slipped this in, I think we need it but if not let me know

pkg/prowloader/prow.go
```diff
@@ -181,6 +181,8 @@ func fetchJobsJSON(prowURL string) ([]byte, error) {
...
+	defer resp.Body.Close()
```
neisw, line 185: Slipped this in, I think we need it but if not let me know

stbenjam:
I think I'd prefer to understand why it's failing, disabling type checks and renaming imports points to something having changed that needs understanding before we merge this.

PR #681, opened 2022-11-10
TRT-668: Update linter version

neisw:
Upgrade golangci-lint version and make minor changes for the linter rules

### DIFF
--- a/pkg/perfscaleanalysis/perfscalehelpers.go
+++ b/pkg/perfscaleanalysis/perfscalehelpers.go
@@ -120,7 +120,7 @@ func DownloadPerfScaleData(storagePath string, reportEnd time.Time) error {
 	if err != nil {
 		return errors.Wrap(err, "error marshalling scale jobs")
 	}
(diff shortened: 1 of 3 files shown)


pkg/perfscaleanalysis/perfscalehelpers.go
```diff
@@ -120,7 +120,7 @@ func DownloadPerfScaleData(storagePath string, reportEnd time.Time) error {
 	if err != nil {
 		return errors.Wrap(err, "error marshalling scale jobs")
 	}
-	err = ioutil.WriteFile(filepath.Join(storagePath, ScaleJobsFilename), file, 0600)
+	err = ioutil.WriteFile(filepath.Join(storagePath, ScaleJobsFilename), file, 0o600)
```
stbenjam, line 123: Looks like this is because of https://github.com/golang/go/issues/33363

Such a weird syntax.  0 as a prefix has denominated octals for decades :-\

stbenjam:
/override ci/prow/lint
/lgtm
/approve

We have to merge this for the buildroot to get updated with the golangci-lint version, a bit of a catch 22 here

PR #684, opened 2022-11-14
Update the download/restore instructions for sippy backups in parts

DennisPeriquet:
[TRT-677](https://issues.redhat.com//browse/TRT-677)

There are no more backups that end in .sql and we need instructions on how to assemble the full backup file before restoring.

related to: https://github.com/openshift/continuous-release-jobs/pull/1239

I verified the restore part by taking the backup "partX" files, concatenating them, and restoring the resulting .dump file without error.


DennisPeriquet:
The backups are no longer split up so this PR is no longer needed.

PR #685, opened 2022-11-15
Extract metadata from flake/failure test output for certain backstop tests

dgoodwin:
[TRT-665](https://issues.redhat.com//browse/TRT-665)

Begin extracting a little metadata from test flakes and failures on very specific set of backstop tests when we're importing job results. This will allow us to track trends in what alerts and pathological events specifically are causing problems over time. 

Result is stored in a jsonb column:

```sql
postgres=# select rt.test_id, rt.status, md.metadata from prow_job_run_test_output_metadata md, prow_job_run_test_outputs o, prow_job_run_tests rt where md.prow_job_run_test_output_id = o.id and o.prow_job_run_test_id = rt.id limit 1;
 test_id | status |                                          metadata                                           
---------+--------+---------------------------------------------------------------------------------------------
   11497 |     12 | {"alert": "TargetDown", "state": "fired", "namespace": "openshift-machine-config-operator"}
(1 row)
```

And can be queried and grouped by normally:
```sql
SELECT                                                                                                                                                                                                                                 
    t.id as "test_id",                                                                                                                                                                                                                 
    md.metadata,                                                                                                                                                                                                                       
    count(md.metadata)                                                                                                                                                                                                                 
FROM                                                                                                                                                                                                                                   
    prow_job_run_test_output_metadata md,                                                                                                                                                                                              
        prow_job_run_test_outputs o,                                                                                                                                                                                                   
        prow_job_run_tests rt,                                                                                                                                                                                                         
        prow_job_runs r,                                                                                                                                                                                                               
        prow_jobs j,                                                                                                                                                                                                                   
        tests t                                                                                                                                                                                                                        
WHERE                                                                                                                                                                                                                                  
        md.prow_job_run_test_output_id = o.id AND                                                                                                                                                                                      
        o.prow_job_run_test_id = rt.id AND                                                                                                                                                                                             
        rt.status = 12 AND                                                                                                                                                                                                             
        rt.prow_job_run_id = r.id AND                                                                                                                                                                                                  
        r.prow_job_id = j.id AND                                                                                                                                                                                                       
        rt.test_id = t.id AND                                                                                                                                                                                                          
                t.name LIKE '%pathological%'                                                                                                                                                                                           
GROUP BY                                                                                                                                                                                                                               
    t.id, md.metadata                                                                                                                                                                                                                  
ORDER BY                                                                                                                                                                                                                               
    count DESC;                                                                                                                                                                                                                        
```


pkg/prowloader/testoutputmetadata.go
```diff
@@ -0,0 +1,128 @@
...
+	matches := regex.FindAllStringSubmatch(testOutput, -1)
+	allMatchMap := []map[string]string{}
+	for _, m := range matches {
+		results := map[string]string{}
+		for i, name := range m {
+			// Skip the empty group name to full string matched:
+			if i == 0 {
+				continue
+			}
+			//if results[regex.SubexpNames()[i]] != "" {
```
neisw: Looks like the linter objects to this line...

pkg/prowloader/testoutputmetadata.go
```diff
@@ -0,0 +1,128 @@
...
+			results[regex.SubexpNames()[i]] = name
+			//}
```
neisw: and likely this one too.  I guess either add spaces or just pull em out if you don't need them.

pkg/prowloader/testoutputmetadata.go
```diff
@@ -0,0 +1,126 @@
...
+// matchFirstRegexPerLine will return the results from the first regex in the given list that matches each
+// test output line.
+// If you wish to "and" the results from multiple regexes, call this function multiple times.
+func matchFirstRegexPerLine(regexes []*regexp.Regexp, testOutput string) []map[string]string {
+	allMetadata := []map[string]string{}
+	for _, line := range strings.Split(testOutput, "\n") {
+		for _, re := range regexes {
+			matchMaps := findAllNamedMatches(re, line)
+
+			fmt.Printf("%v\n", matchMaps)
```
neisw, line 87: left over debugging?

pkg/db/models/prow.go
```diff
@@ -92,6 +93,17 @@ type ProwJobRunTestOutput struct {
...
+
+	// Metadata optionally contains metadata extracted from a select few generic backstop tests
+	// we use to catch problems. This metadata helps us identify developing problems in these broad
+	// tests and figure out what next needs to be broken out into its own test.
+	Metadata []ProwJobRunTestOutputMetadata `gorm:"constraint:OnDelete:CASCADE;"`
+}
+
+type ProwJobRunTestOutputMetadata struct {
+	gorm.Model
+	ProwJobRunTestOutputID uint
```
stbenjam, line 105: What do you think about an index on this?

stbenjam:
Were you able to do JSONB queries? Something like `SELECT * FROM  prow_job_run_test_output_metadata WHERE metadata @> '{"namespace": "openshift-machine-config-operator"}';`

dgoodwin:
> Were you able to do JSONB queries? Something like `SELECT * FROM prow_job_run_test_output_metadata WHERE metadata @> '{"namespace": "openshift-machine-config-operator"}';`

I have not tried with gorm yet just in psql.

pkg/prowloader/testoutputmetadata.go
dgoodwin: Yup, thanks!

pkg/db/models/prow.go
dgoodwin: Will do.

dgoodwin:
/hold

This needs more work.

dgoodwin:
Updated, I took a new approach of regex + string tokens, I want a flexible "parse this tag if it's there" approach (tag meaning ns/foobar or ns=foobar), and regexes seem quite poor at this, so now if the simple regex for things that should always match if we're interested in a line hits, we'll proceed to token parsing and pull those out too, whatever we can find. 

Some of this will pickup new values coming in https://github.com/openshift/origin/pull/27559

I think this is ready for another review.

/hold cancel

pkg/prowloader/prow.go
```diff
@@ -420,6 +422,24 @@ func (pl *ProwLoader) extractTestCases(suite *junit.TestSuite, testCases map[str
...
+		if failureOutput != nil {
+			// Check if this test is configured to extract metadata from it's output, and if so, create it
+			// in the db.
+			extractedMetadata := testOutputMetadataExtractor.ExtractMetadata(testNameWithKnownSuite, failureOutput.Output)
+			if len(extractedMetadata) > 0 {
+				failureOutput.Metadata = make([]models.ProwJobRunTestOutputMetadata, 0, len(extractedMetadata))
+				for _, m := range extractedMetadata {
+					jsonb := pgtype.JSONB{}
+					if err := jsonb.Set(m); err != nil {
+						panic(err)
```
stbenjam, line 434: Can we log a warning  instead? I don't know if we'll really ever panic on `jsonb.Set` but the places where we panic during syncing risks data loss since it'll stop syncing and any further tasks.

stbenjam:
Looks good as best I can tell, the text processing is a little complex. I really have just one comment about the panic above. 

I think I would prefer is openshift-tests just returned serializable failure output (YAML?) for tests we want to store metadata about but we can tackle something like that later if we want to do it.

stbenjam:
> The json output occurred to me too but I couldn't see an easy way to do it. We'd need to either find a way to parse json out of test output, or correlate different output files with each junit file and figure out how to import them at different times. It would be a cleaner way to import for sure if we could figure out how best to do it. Maybe we could add an additional tag to the junit and parse it out somehow.

Ah I was just thinking sippy would know which tests it expects might have YAML output and try to unmarshal it (and fail quietly if it can't)

PR #687, opened 2022-11-16
add load-timeout and pull-policy override options and small https get…

neisw:
… resource cleanup


pkg/prowloader/prow.go
```diff
@@ -181,6 +181,7 @@ func fetchJobsJSON(prowURL string) ([]byte, error) {
 	if err != nil {
 		return nil, err
 	}
+	defer resp.Body.Close()
```
neisw, line 184: I think this is needed / best practice but if not let me know

DennisPeriquet: i saw [this](https://golangbyexample.com/resposne-body-closed-golang) and I agree

e2e-scripts/sippy-e2e-sippy-e2e-setup-commands.sh
```diff
@@ -233,9 +233,9 @@ ${KUBECTL_CMD} -n postgres describe job sippy-load-job
 # We set +e to avoid the script aborting before we can retrieve logs.
 set +e
 # This takes under 3 minutes so 5 minutes (300 seconds) should be plenty.
-TIMEOUT=300s
```
DennisPeriquet, line 236: you took this one out and renamed it `SIPPY_LOAD_TIMEOUT` -- cool.
But looking at the test run, you have to define that somewhere in the scripts.

e2e-scripts/run-e2e.sh
```diff
@@ -48,6 +57,20 @@ if [ ! -f ${DOCKERCONFIGJSON} ]; then
   exit 1
 fi
 
+if [ -z ${SIPPY_LOAD_TIMEOUT} ]; then
+  # Allow timeout override in case it takes longer.
+  echo "SIPPY_LOAD_TIMEOUT is not set; assuming 300s"
+  export SIPPY_LOAD_TIMEOUT='300s'
```
DennisPeriquet, line 63: Hm... I see you define `SIPPY_LOAD_TIMEOUT` here but it's blank in the script that's. called later.

DennisPeriquet: there's this syntax in bash where you can define a variable to be something if it's not already defined.  Let me find a ref.

neisw: Is that not similar to how you have DOCKERCONFIGJSON setup?

DennisPeriquet: yea, that is similar. hm...

Does this work when running it locally with both defining `SIPPY_LOAD_TIMEOUT` and not? it does seem like it should be defined.

DennisPeriquet: ah, I think I know why -- that `run-e2e.sh` is not run in CI.

DennisPeriquet: To get past the error in the CI, you'll need to define that new variable [in release repo](https://github.com/openshift/release/blob/5fb1845e3e2d5f75488d97e8cc94e760818a0971/ci-operator/step-registry/sippy-e2e/sippy-e2e-setup/sippy-e2e-sippy-e2e-setup-ref.yaml).

neisw: Thanks, I just added defaults in the scripts.  Thanks for pointing out that run-e2e.sh wasn't getting called via CI.

e2e-scripts/sippy-e2e-sippy-e2e-setup-commands.sh
```diff
@@ -195,7 +195,7 @@ spec:
       containers:
       - name: sippy
         image: ${SIPPY_IMAGE}
-        imagePullPolicy: Always
+        imagePullPolicy: ${SIPPY_IMAGE_PULL_POLICY:-Always}
```
DennisPeriquet, line 198: this is cool

e2e-scripts/sippy-e2e-sippy-e2e-setup-commands.sh
```diff
@@ -233,9 +233,9 @@ ${KUBECTL_CMD} -n postgres describe job sippy-load-job
 # We set +e to avoid the script aborting before we can retrieve logs.
 set +e
 # This takes under 3 minutes so 5 minutes (300 seconds) should be plenty.
-TIMEOUT=300s
-echo "Waiting up to ${TIMEOUT} for the sippy-load-job to complete..."
-${KUBECTL_CMD} -n postgres wait --for=condition=complete job/sippy-load-job --timeout ${TIMEOUT}
+
+echo "Waiting up to ${SIPPY_LOAD_TIMEOUT:=300s} for the sippy-load-job to complete..."
+${KUBECTL_CMD} -n postgres wait --for=condition=complete job/sippy-load-job --timeout ${SIPPY_LOAD_TIMEOUT}
```
dgoodwin, line 238: Looks like we may log one thing and do another as the use of the var doesn't have the defaulting? Suggest actually setting the value with default and then just use the var in both of these lines.

neisw: I used the := in that case to set SIPPY_LOAD_TIMEOUT per the link Dennis had shared https://unix.stackexchange.com/questions/594841/how-do-i-assign-a-value-to-a-bash-variable-if-that-variable-is-null-unassigned-f/594845

But can refactor for readability if you like

dgoodwin: Ah this actually *sets* the var? I had no idea, this is all good then, thanks.

PR #689, opened 2022-11-17
Move lint to container for consistency

stbenjam:
Local developers may have various different versions of go and golangci-lint installed. For consistency, we should run it from a container that matches the versions used in CI.

This does require a user to have working rootless containers on their machine.


stbenjam:
/hold

CI itself calls `make lint`, so I'll have to fix the CI job before we do something like this.

dgoodwin:
/lgtm

This is nice, thanks!

PR #690, opened 2022-11-17
Add a sippy test watchlist

dgoodwin:
Add a sippy test watchlist
    
Tests in the db will get flagged with watchlist = true if they match one
of the hardcoded regexes for our backstop tests, or if they have a
linked jira with the label sippy-watchlist.
    
Adds a new Tests > Watchlist tab for viewing the NURPs for just these
tests.
    
Removal of the jira label will cause the watchlist entry to go away
within a few hours. (search.ci must sync first, then sippy must)

This will help us keep an eye on tests we're concerned with or interested in for a period of time, then remove the label when we're confident fixes worked and rolled out.


dgoodwin:
/hold

More testing tomorrow but I think it's close.

pkg/api/tests.go
```diff
@@ -209,6 +216,12 @@ func BuildTestsResults(dbc *db.DB, release, period string, collapse, includeOver
 		rawQuery = rawFilter.ToSQL(rawQuery, apitype.Test{})
 	}
 
+	// should we filter down to only tests on the watchlist or not?
+	if watchlist {
+		log.Info("WATCHLIST REQUESTED")
+		rawQuery = rawQuery.Where("watchlist = ?", true)
+	}
```
stbenjam, line 223: Could this not just be part of a filter?

dgoodwin: Ooo yes, did not even think of that. I'll change it, thanks.

stbenjam:
I am not sure if it's too much of a refactor, but I'm not a big fan of special casing filters into function signatures, especially as booleans, when it could just be part of the existing filter infrastructure.  

And one more suggestion, what if tests had tags string array instead of just a single column for watchlist? That would let us tag tests differently if we needed to instead of ending up with a bunch of boolean columns.

stbenjam:
> I also wonder if we'll use this. [One can already filter tests but bug count](https://sippy.dptools.openshift.org/sippy-ng/tests/4.13?filters=%257B%2522items%2522%253A%255B%257B%2522columnField%2522%253A%2522current_runs%2522%252C%2522operatorValue%2522%253A%2522%253E%253D%2522%252C%2522value%2522%253A%25227%2522%257D%252C%257B%2522columnField%2522%253A%2522variants%2522%252C%2522not%2522%253Atrue%252C%2522operatorValue%2522%253A%2522contains%2522%252C%2522value%2522%253A%2522never-stable%2522%257D%252C%257B%2522columnField%2522%253A%2522name%2522%252C%2522not%2522%253Atrue%252C%2522operatorValue%2522%253A%2522contains%2522%252C%2522value%2522%253A%2522.Overall%2522%257D%252C%257B%2522columnField%2522%253A%2522name%2522%252C%2522not%2522%253Atrue%252C%2522operatorValue%2522%253A%2522contains%2522%252C%2522value%2522%253A%2522step%2520graph.%2522%257D%252C%257B%2522columnField%2522%253A%2522name%2522%252C%2522not%2522%253Atrue%252C%2522operatorValue%2522%253A%2522contains%2522%252C%2522value%2522%253A%2522openshift-tests%2520should%2520work%2522%257D%252C%257B%2522columnField%2522%253A%2522open_bugs%2522%252C%2522operatorValue%2522%253A%2522%253E%2522%252C%2522value%2522%253A%25220%2522%257D%255D%252C%2522linkOperator%2522%253A%2522and%2522%257D&sort=asc&sortField=current_working_percentage), and if slightly altered that so it was only showing open bugs, you'd get a similar result.

Hm, I guess it's a different use case, since you can tag tests manually.  So we'd be able to watch tests say over a whole release.

dgoodwin:
Do you folks think this should be uncollapsed, i.e. every variant combo? or collapsed, one row per test.

pkg/sippyserver/dbloader.go
```diff
@@ -411,6 +416,51 @@ func appendJobIssuesFromVariants(jobCache map[string]*models.ProwJob, jobIssues
...
+		}
+
+		for _, bug := range test.Bugs {
+			if util.StrSliceContains(bug.Labels, watchlistLabel) {
+				return true
+			}
+		}
+	}
+	return false
+}
```
stbenjam, line 441: This would be better in `pkg/testidentification`, and having the regexes outside the function would be more efficient.  These get recompiled on each function call.

pkg/sippyserver/dbloader.go
```diff
@@ -411,6 +416,51 @@ func appendJobIssuesFromVariants(jobCache map[string]*models.ProwJob, jobIssues
 	return nil
 }
 
+const watchlistLabel = "sippy-watchlist"
+
+// TestIsOnWatchlist returns true if the test matches one of our hardcoded regexes, or
+// has an associated bug with the sippy-watchlist label.
+func TestIsOnWatchlist(test *models.Test) bool {
```
stbenjam, line 423: Style nit, all our bool-returning test identifying functions begin with `Is`

pkg/api/tests.go
```diff
@@ -236,6 +236,7 @@ func BuildTestsResults(dbc *db.DB, release, period string, collapse, includeOver
 			ID:   math.MaxInt32,
 			Name: "Overall",
 		}
+		// TODO: column open_bugs does not exist here?
```
stbenjam, line 239: Yea I noticed this always causes a warning in the logs

pkg/sippyserver/dbloader.go
stbenjam: Another reason to have these outside the function call like this [1] is because you'll always compile them when sippy starts, meaning you WILL get a panic if you mess up the regex.  Having them  in a function means you won't know until the function is actually called.

[1] https://github.com/openshift/sippy/blob/master/pkg/testidentification/test_identification.go#L163-L173

sippy-ng/src/constants.js
```diff
@@ -119,6 +119,12 @@ export const BOOKMARKS = {
     operatorValue: '>',
     value: '1',
   },
+  WATCHLIST: {
+    columnField: 'watchlist',
+    not: false,
```
stbenjam, line 124: You can omit `not: false`

sippy-ng/src/constants.js
```diff
@@ -119,6 +119,12 @@ export const BOOKMARKS = {
...
+    operatorValue: 'equals',
+    value: 'true',
```
stbenjam, line 126: Looks like this gets treated as a string? Does it work in the UI (I haven't tested it yet)?

Do we use the go-based filtering anywhere anymore? We should probably make sure watchlist is included in https://github.com/openshift/sippy/blob/master/pkg/apis/api/types.go#L435.  There's no bool type, but maybe string  works.

dgoodwin: I did add it to the Test type in that file, unless I'm missing what you meant just let me know.

String does appear to work for the filtering.

stbenjam: It's added to the type but not the go filtering. There's helpers that categorize the field and return the value like https://github.com/openshift/sippy/blob/2c79ecc2739fb34d21477e4390ceb793ec9b0893/pkg/apis/api/types.go#L480.  It's used if filters are processed by go and not sql.

dgoodwin: Ah now I understand, thanks. Added

stbenjam: You're still missing the entry in `GetStringValue()`.  Sorry, this filtering thing is awkward but I think it's still used a few places.

dgoodwin: Sorry, added that one too.

PR #699, opened 2022-11-23
Fix or filtering

dgoodwin:
Wasn't working as we inject an AND for the release first, then we try to
add ORs but of course the AND already picks up all tests, so ORs had no
effect.

This makes it possible to OR on test names. However it still doesn't
work if you use any filtering on fields beyond the rawFilter, as we
split into raw and post filtering which comes after.

[TRT-694](https://issues.redhat.com//browse/TRT-694)

### DIFF
--- a/pkg/filter/filterable.go
+++ b/pkg/filter/filterable.go
@@ -61,86 +61,88 @@ type FilterItem struct {
 	Value    string   `json:"value"`
 }
 
-func (f FilterItem) orFilterToSQL(db *gorm.DB, filterable Filterable) *gorm.DB { //nolint
+func (f FilterItem) orFilterToSQL(db *gorm.DB, filterable Filterable) (orFilter string, orParams interface{}) { //nolint
+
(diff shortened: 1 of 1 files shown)


dgoodwin:
@stbenjam what do you think, this gets us or filtering on names at least, but if you add in any of the post-filter fields they get tacked on after and only take effect on what was in the raw filter, i.e.:

```sql
...FROM "prow_test_report_7d_matview" WHERE release = '4.12' AND ("name" LIKE '%openshift-tests should work%' or "name" LIKE '%pathological%') GROUP BY name,open_bugs,watchlist) as results WHERE current_runs > 0 or previous_runs > 0) as final_results WHERE "current_runs" > '0'
```

Or filtering is kind of busted conceptually in that regard. However this gets us something at least.

neisw:
Happy to lgtm based on my review, unless you want Stephen to give it a once over.  I'm presuming you have already done heavy local testing.

PR #703, opened 2022-11-25
Allow scaling to multiple pods

dgoodwin:
This is a fairly big change to how we run sippy in kube with the goal of being able to run multiple replicas, and deploy new versions without downtime. 

# What Changed

Updating database schema no longer happens by default when you start the server. If you specify the --init-database flag it will, which can be used when loading database or starting the server. In the official kube manifests, the new cronjob will include it, but the server will not. I didn't want multiple server pods coming up fighting over schema at the same time.

Subsequently refresh metrics is now broken out from refresh data. The /refresh API endpoint is removed, which is probably a good thing considering what could be done with it. Metrics are refreshed in a goroutine every 20 minutes.

The fetchdata container is broken out into a separate CronJob for both Kube and OpenShift. I also moved the command to do so directly into the CronJob itself, instead of in a script in the container, but only for OpenShift. (Kube script is still doing a little more) This helped development in that it avoided pushing new images as often, and is consistent with how we run the server. The cronjob is configured to not overlap, if an old instance is still running, the new will not start. A manual execution of the cronjob can be started at any time with `kubectl create job --from=cronjob/fetchdata fetchdata-m1`

The fetchdata cronjob will implicitly refresh materialized views after it finishes, instead of calling the /refresh API endpoint.

With the fetchdata breakout, sippy should now need very little memory in on-going use. We'll need to see in practice but I suspect far less than the 512MB we're requesting.

The server DeploymentConfig now runs two replicas, with a rolling update strategy, and a pre-rollout hook that will update the schema before we start rolling out the new pods. They also start immediately.

Sample manifests for running Sippy on OpenShift are updated and now a directory for Kube is included. Assuming you setup a secret for GCS credentials, running is about as simple as a kubectl apply on the directory. These manifests can be diffed against those we use on DPCR relatively cleanly. We may want to think about moving them to Kube native Deployments here however, in the future, but then no more diffing. It would let devs run against kind/minikube/k3s though.

# Caveats

It is possible that schema updates may impact still running pods from the last version, there isn't much we can do about this, and it should be extremely rare as we usually are not deleting columns, changing types, etc. The worst case scenario would be a breaking change to a matview perhaps. 

It is possible that we update schema while a fetchdata cronjob is running, things may break and restart and try again. This should be relatively rare.

Metrics will need a little look, but I suspect will function fine as it does not. We'll be hitting the service and routing to a random pod, that's ok for most of what we alert on as both pods will be reporting the same job/test stats. It won't be fine for monitoring any pod specific metrics like memory use etc. There is a way to do this but I'll need to look into it. I think the "up" metric we use should be ok as is as well.

# Plan to Roll Out

  1. Sync manifests with DPCR and open PR
  2. Hold both PRs
  3. When ready, scale sippy down and merge the sippy PR.
  4. Merge DPCR PR.
  5. Likely will need to cancel any builds/cronjobs that start too soon and run them again.

[TRT-704](https://issues.redhat.com//browse/TRT-704)

- Break postgres container out for scaling replicas
- Trim deploymentconfig for non-prod use
- Delete outdated historical/cluster specifics
- Remove some newlines in log msgs
- Fetchdata command in yaml, not a separate script
- Move postgres secret from example to yaml
- Major change to move refreshing matviews out to fetchdata
- Don't store postgresql revision history
- Rolling update strategy to update one pod at a time with a 10s pause as well
- Keep 5 revisions in history for sippy
- Init database schema before each deployment
- Don't allow overlapping fetchdata cronjobs
- Update process for building in cluster
- Remove the data volume no longer needed for prow + postgresql
- Move openshift manifests to sub-dir
- Add kube manifests copied from dpcr, needs edits
- Prep kube fetchdata script
- Update kube manifests for new way of deploying
- Full matview refresh on updates
- Add an error and exit if server is run when db has not been initialized
- Add note about --init-database to DEVELOPMENT.md


stbenjam:
> Updating database schema no longer happens by default when you start the server. If you specify the --init-database flag it will, which can be used when loading database or starting the server. In the official kube manifests, the new cronjob will include it, but the server will not. I didn't want multiple server pods coming up fighting over schema at the same time.

This sounds problematic - the schema could change, new front end pods would get triggered by the new build, but the schema wouldn't get updated until some time later. Is that handled somehow? I didn't look too much at the code, but will next week.

 I think we we talked about this before I had proposed using an [advisory lock](https://hashrocket.com/blog/posts/advisory-locks-in-postgres).  The first instance to start could grab an advisory lock, migrate the schema, and then release the lock.   The rest would wait for the lock to be released and then start.

dgoodwin:
> > Updating database schema no longer happens by default when you start the server. If you specify the --init-database flag it will, which can be used when loading database or starting the server. In the official kube manifests, the new cronjob will include it, but the server will not. I didn't want multiple server pods coming up fighting over schema at the same time.
> 
> This sounds problematic - the schema could change, new front end pods would get triggered by the new build, but the schema wouldn't get updated until some time later. Is that handled somehow? I didn't look too much at the code, but will next week.
> 
> I think when we talked about this before I had proposed using an [advisory lock](https://hashrocket.com/blog/posts/advisory-locks-in-postgres). The first instance to start could grab an advisory lock, migrate the schema, and then release the lock. The rest would wait for the lock to be released and then start.

I believe it's handled yes, there's a pre-deployment hook on the DeploymentConfig, when we roll out a new version of code, the first thing we do is update schema, before any new pods are started. It was a neat feature I discovered in OpenShift DCs, not available on Kube Deployments.

main.go
```diff
@@ -12,6 +12,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/openshift/sippy/pkg/db/models"
+	"github.com/openshift/sippy/pkg/sippyserver/metrics"
```
stbenjam, line 16: Nit, this belongs below in the other openshift/sippy imports

main.go
```diff
@@ -459,6 +479,32 @@ func (o *Options) runServerMode(pinnedDateTime *time.Time) error {
...
+				err := metrics.RefreshMetricsDB(dbc, util.GetReportEnd(pinnedDateTime))
+				if err != nil {
+					log.WithError(err).Error("error refreshing metrics")
+				}
+			case <-quit:
+				ticker.Stop()
+				return
+			}
+		}
+	}()
```
stbenjam, line 505: Will we end up in a weird state if the replicas are refreshing metrics at different times? Let's say we have 2 sippy containers that start 10 minutes offset each other.  Couldn't we end up in a situation where we flap on alerts briefly when one instance reached an alerting threshold, and another didn't because it didn't refresh it's metrics yet?

There's https://github.com/krayzpipes/cronticker but it hasn't been updated in years.

Maybe it's possible to make the metrics ingress endpoint sticky to only ever hit one of the endpoints?

resources/ingress.yaml
```diff
@@ -1,23 +0,0 @@
-apiVersion: extensions/v1beta1
```
stbenjam, line 1: Do we not want to keep the ingress configuration as an example in this repo too? https://github.com/openshift/continuous-release-jobs/blob/master/config/clusters/dpcr/services/sippy/ingress-sippy.yaml

main.go
dgoodwin: Hmm that's a very good question. 

I think we possibly best mitigate with the "for:" lines in the alert config, saying to only fire if the condition has been true for at least 10 minutes? We have some set to 1s, which we could change now that this would be in play. Maybe we set every pod to refresh metrics every 5 mins, and use at least 10 mins for firing alerts to give us a good buffer. I think metrics refreshes are pretty quick. Thoughts?

resources/ingress.yaml
dgoodwin: I wanted the directory to be usable with a straight kubectl apply -f dir/, and the ingress felt too deployment specific. I don't think devs or e2e needs it, we port forward, so thought it would be best to leave it out and cite it in the README as an implementation detail of whomever is running sippy.  

Is there a way to have a generic Ingress setup considering the TLS/domains/hosts in that example linked above?

main.go
stbenjam: I think that would work

dgoodwin: I will go bump those thresholds for the sister PR to this one so it merged together.

dgoodwin: Changed to 5m here.

dgoodwin:
/hold cancel

Here we go...

PR #705, opened 2022-11-28
Add sql queries for aggregating most common backstop issues

dgoodwin:
Adds an sql query for both firing alerts, and pathologically repeating events.

[TRT-723](https://issues.redhat.com//browse/TRT-723)

### DIFF
--- a/scripts/sql/find-job-runs-with-alert.sql
+++ b/scripts/sql/find-job-runs-with-alert.sql
@@ -1,0 +1,32 @@
+SELECT
+    r.timestamp,
+    md.metadata->>'alert' AS "alert",
+    md.metadata->>'namespace' AS "namespace",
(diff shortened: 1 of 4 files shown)


dgoodwin:
/hold

Need to factor in release, our top hits are 4.9 right now I suspect.

dgoodwin:
/hold

This is rapidly getting too slow the more data we accumulate.

dgoodwin:
/hold 

Again, found we can't really see alerts data properly, needed to look at flakes, but also we can't tell what's broken out and what isn't.

dgoodwin:
/hold cancel

I think we're good now, filtered out alerts we already know are acceptable, added scripts to find job runs where a particular problem is occurring. They can be slow when matviews are refreshing but normally are ok.

PR #707, opened 2022-11-29
Audit deps in lint target, rather than npm

dgoodwin:
This will allow us to only fail the lint job, and still be able to build / release if we choose to override.


Makefile
```diff
@@ -19,6 +19,7 @@ test: builddir npm
 	LANG=en_US.utf-8 LC_ALL=en_US.utf-8 cd sippy-ng; CI=true npm test -- --coverage
 
 lint: builddir npm
+	cd sippy-ng; npm audit --production
```
stbenjam, line 22: Can we move the comment on line 29 and 30 up here too?

stbenjam: Do we also need to `cd ..`? Or perhaps move this below go-lint.sh.

dgoodwin: Yup I was surprised too.

dgoodwin:
Comment has been relocated.

PR #708, opened 2022-11-29
TRT-697: validate incoming ProwJobRun

neisw:
Validate incoming ProwJobRun and return elevated risk if invalid

### DIFF
--- a/pkg/sippyserver/server.go
+++ b/pkg/sippyserver/server.go
@@ -796,6 +796,9 @@ func (s *Server) jsonJobRunRiskAnalysis(w http.ResponseWriter, req *http.Request
 				"code": http.StatusBadRequest, "message": res.Error.Error()})
 			return
 		}
+
+		// TODO: get the count of tests that were run overall
+
 	} else {
 		err := json.NewDecoder(req.Body).Decode(&jobRun)
 		if err != nil {
@@ -804,6 +807,28 @@ func (s *Server) jsonJobRunRiskAnalysis(w http.ResponseWriter, req *http.Request
 				"message": fmt.Sprintf("error decoding prow job run json in request body: %s", err)})
 			return
 		}
+
+		// validate the jobRun isn't empty
+		// valid case where test artifacts are not available
+		// we want to mark this as a high risk
+		if isValid, detailReason := isValidProwJobRun(jobRun); !isValid {
+
+			log.Warn("Invalid ProwJob provided for analysis, returning elevated risk")
+			result := apitype.ProwJobRunRiskAnalysis{
+				OverallRisk: apitype.FailureRisk{
+					Level:   apitype.FailureRiskLevelHigh,
+					Reasons: []string{fmt.Sprintf("Invalid ProwJob provided for analysis: %s", detailReason)},
+				},
+			}
+
+			// respond ok since we handle it
+			api.RespondWithJSON(http.StatusOK, w, result)
+			return
+		}
+
+		// TODO:
+		// check to see if prow_job_test_count was included in the query params
+
 		// We don't expect the caller to fully populate the ProwJob, just it's name,
 		// override the input by looking up the actual ProwJob so we have access to release and variants.
 		job := &models.ProwJob{}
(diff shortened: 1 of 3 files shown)


pkg/sippyserver/server.go
```diff
@@ -804,6 +807,27 @@ func (s *Server) jsonJobRunRiskAnalysis(w http.ResponseWriter, req *http.Request
...
+		// validate the jobRun isn't empty
+		// valid case where test artifacts are not available
+		// we want to mark this as a high risk
+		if (jobRun == &models.ProwJobRun{} || &jobRun.ProwJob == &models.ProwJob{} || jobRun.ProwJob.Name == "") {
+
+			log.Warn("Invalid ProwJob provided for analysis, returning elevated risk")
+			result := apitype.ProwJobRunRiskAnalysis{
+				OverallRisk: apitype.FailureRisk{
+					Level:   apitype.FailureRiskLevelHigh,
+					Reasons: []string{"Invalid ProwJob provided for analysis"},
```
dgoodwin, line 820: Could you just add in a little more detail on why it was invalid, specifically that no test results were provided.

neisw: Sure, in this case the complete ProwJobRun is empty so it is really the fact that we don't have a Name at this point.  I can add checks & specify that the ProwJobRun is empty or that the Name is missing.

dgoodwin: Ahh I see I was thinking it was just a lack of tests, this is a little more serious. I gotcha. Anything as crisp as possible is good though.

pkg/sippyserver/server.go
```diff
@@ -804,6 +807,28 @@ func (s *Server) jsonJobRunRiskAnalysis(w http.ResponseWriter, req *http.Request
 				"message": fmt.Sprintf("error decoding prow job run json in request body: %s", err)})
 			return
 		}
+
+		// validate the jobRun isn't empty
+		// valid case where test artifacts are not available
+		// we want to mark this as a high risk
+		if (jobRun == &models.ProwJobRun{} || &jobRun.ProwJob == &models.ProwJob{} || jobRun.ProwJob.Name == "") {
```
DennisPeriquet, line 814: I'm betting there's no test in CI to confirm this is working.

I'm new enough to golang that this line doesn't make much sense to me because on line 803, you decode the json to `&jobRun` (and in the case of no `test-failures-summary*.json` files it's an empty json value), and above, `jobRun` is defined as `&models.ProJobRun()` and you are comparing it to `&models.ProwJobRun{}` to check if it's empty -- won't those two things always be different?

If you already tested it and it works as expected, I guess I have to lookup something like "how to check if a json value is null".

neisw: That's a good catch really.  I added a test case to my latest push to replicate what we are seeing from the origin side.  What appears to happen is we pass a nil pointer that gets encoded to a string with the value "null" but no errors.. Then when we try to decode we again don't get any errors but the jobRun is set to nil.  I had to add that nil check...  I was anticipating that since there wasn't an error we were getting an empty prowJobRun, hence the &models.ProwJobRun{}.  Personally I would expect an error if I pass a nil pointer in to encode but that doesn't seem to be what happens either...

DennisPeriquet: I'm building openshift-tests and then doing this to try it out:
```
./openshift-tests risk-analysis --junit-dir ./junit/ --sippy-url http://localhost:8080/api/jobs/runs/risk_analysis
```

where the `./junit` dir is empty.  This is the resulting body:

```
{
  "ProwJobName": "",
  "ProwJobRunID": 0,
  "Release": "",
  "CompareRelease": "",
  "Tests": null,
  "OverallRisk": {
    "Level": {
      "Name": "High",
      "Level": 10
    },
    "Reasons": [
      "Invalid ProwJob provided for analysis: empty ProwJobRun"
    ]
  },
  "OpenBugs": null
}
```

neisw: Nice!

DennisPeriquet: I went into debugger and see `jobRun` is `nil` when using an empty `./junit` dir.

PR #714, opened 2022-12-01
Refactor db load error handling to tolerate problems and keep going

dgoodwin:
This is to hopefully allow more resilient syncing without abandoning
large portions because of a single area in any specific section of the
code.

All errors are re-logged at the end of the process, and if any were
present an overall error is returned causing the job to fail. This will
hopefully fire alerts, and the logs at the end of the pod should
summarize what went wrong.

[TRT-598](https://issues.redhat.com//browse/TRT-598)

### DIFF
--- a/main.go
+++ b/main.go
@@ -12,7 +12,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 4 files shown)


neisw:
lgtm, so just to summarize loadReleases, loadProw, loadBugs and updateWatchlist are independent and a failure in one doesn't impact the other.  Inside each of those you fail on critical errors but continue on non critical ones and report an error summary at the end.

I'm good tagging it unless you want additional review.

dgoodwin:
I think so yes, I cannot think of critical deps between them, obviously some things might be missing in later stages if the earlier fails, but it should all run I suspect. If not I can deal with it as it comes up, I'll keep an eye on job failures. Probably ok to tag, thank you!

PR #721, opened 2022-12-06
Skip ignored tests during risk analysis

dgoodwin:
This was causing Cluster should remain functional during upgrade to show up in the risk analysis with unknown state, because we don't import this in sippy.

[TRT-703](https://issues.redhat.com//browse/TRT-703)


stbenjam:
Makes sense

/lgtm

PR #723, opened 2022-12-06
TRT-662: elevate risk if missing large group of tests

neisw:
Analyze current job test counts against historical counts


pkg/api/job_runs.go
```diff
@@ -114,28 +115,60 @@ func JobRunRiskAnalysis(dbc *db.DB, jobRun *models.ProwJobRun) (apitype.ProwJobR
...
-	jobRun.ProwJob.Bugs = jobBugs
 
 	// Pre-load test bugs as well:
 	if len(jobRun.Tests) <= maxFailuresToFullyAnalyze {
 		for i, tr := range jobRun.Tests {
 			bugs, err := query.LoadBugsForTest(dbc, tr.Test.Name, true)
 			if err != nil {
-				return apitype.ProwJobRunRiskAnalysis{}, err
+				// SELECT * FROM "tests" WHERE name = '[sig-arch][Feature:ClusterUpgrade] Cluster should remain functional during upgrade [Disruptive] [Serial]' ORDER BY "tests"."id" LIMIT 1;
+				// causes an error here, don't think we need to fail due to missing bugs - error(*errors.errorString) *{s: "record not found"}
```
neisw, line 161: I can remove the commented queries but I'm curious about them and wanted to document / followup a little

pkg/db/query/job_queries.go
```diff
@@ -12,6 +12,56 @@ import (
...
+	var jobRuns []models.ProwJobRun
+	res := dbc.DB.Joins("ProwJob").
+		Preload("Tests").Order("updated_at desc").Limit(10).Find(&jobRuns, "prow_job_id = ? AND succeeded = true", prowJobID)
+	if res.Error != nil {
+		return 0, res.Error
+	}
+
+	if len(jobRuns) == 0 {
+		// select * from prow_job_runs where prow_job_id = 1805 and succeeded = true limit 10;
+		// has no 'succeeded = true' runs
```
neisw, line 41: Same here, interesting to have a job that never passes?

pkg/db/query/job_queries.go
```diff
@@ -12,6 +12,56 @@ import (
...
+
+	t := 0
+
+	if len(jobRuns) > 0 {
+
+		for _, jobRun := range jobRuns {
+			t += len(jobRun.Tests)
+		}
+
+		t /= len(jobRuns)
```
dgoodwin, line 59: /= ? TIL!

pkg/db/query/job_queries.go
```diff
@@ -12,6 +12,56 @@ import (
...
+
+	if len(jobRuns) == 0 {
+		// select * from prow_job_runs where prow_job_id = 1805 and succeeded = true limit 10;
+		// has no 'succeeded = true' runs
+		// if we didn't get any then open it up to just the most recent runs
+		res = dbc.DB.Joins("ProwJob").
+			Preload("Tests").Order("updated_at desc").Limit(10).Find(&jobRuns, "prow_job_id = ?", prowJobID)
+		if res.Error != nil {
+			// we didn't get an error on our first query so we can use that result (0) and continue
+			log.Warnf("Error during fallback historical count query for prowJobId %d", prowJobID)
```
dgoodwin, line 47:
```suggestion
			log.WithError(err).Warnf("Error during fallback historical count query for prowJobId %d", prowJobID)
```

pkg/db/query/job_queries.go
```diff
@@ -12,6 +12,56 @@ import (
...
+	prowJobRunTestCount = len(tests)
+
+	return prowJobRunTestCount, nil
+}
+
+func ProwJobHistoricalTestCounts(dbc *db.DB, prowJobID uint) (int, error) {
+
+	var jobRuns []models.ProwJobRun
+	res := dbc.DB.Joins("ProwJob").
+		Preload("Tests").Order("updated_at desc").Limit(10).Find(&jobRuns, "prow_job_id = ? AND succeeded = true", prowJobID)
```
dgoodwin, line 34: Nit: Suggest using "timestamp" instead of updated_at, which is the time the job ran rather than when we went into the db.

Should we filter here on timestamp > now - 2 weeks? Wondering if there are concerns with finding those successes like 3 months back.

pkg/sippyserver/server.go
```diff
@@ -793,6 +794,12 @@ func (s *Server) jsonJobRunRiskAnalysis(w http.ResponseWriter, req *http.Request
 		}
 
 		// TODO: get the count of tests that were run overall
+		// select id, count(*) from prow_job_run_tests where prow_job_run_id = jobRunID
+		jobRunTestCount, err = query.JobRunTestCount(s.db, jobRunID)
+		if err != nil {
+			log.Errorf("Error getting job run test count %v", err)
```
dgoodwin, line 800:
```suggestion
			log.WithError(err).Error("Error getting job run test count")
```

This one's a NIT if you've got the error in the message anyhow but structured logging is still good for us to keep in mind.

pkg/sippyserver/server.go
```diff
@@ -837,8 +843,22 @@ func (s *Server) jsonJobRunRiskAnalysis(w http.ResponseWriter, req *http.Request
 		jobRun.ProwJob = *job
 	}
 
+	historicalCount, err := query.ProwJobHistoricalTestCounts(s.db, jobRun.ProwJob.ID)
+
+	// if we had an error we will continue the risk analysis and not elevate based on test counts
+	if err != nil {
+		log.Errorf("Error comparing historical job run test count %v", err)
```
dgoodwin, line 850: Nit: WithError

pkg/sippyserver/server.go
```diff
@@ -837,8 +843,22 @@ func (s *Server) jsonJobRunRiskAnalysis(w http.ResponseWriter, req *http.Request
...
+		historicalCount = 0
+	}
+
+	if jobRunTestCount < 0 {
+		// if we were not able to determine
+		log.Error("Unable to determine job run test count, initializing to historical count")
+		jobRunTestCount = historicalCount
+	}
```
dgoodwin, line 858: Can this code move into the JobRunRiskAnalysis?

pkg/api/job_runs.go
```diff
@@ -171,7 +204,7 @@ func JobRunRiskAnalysis(dbc *db.DB, jobRun *models.ProwJobRun) (apitype.ProwJobR
 // testResultsFunc is used for injecting db responses in unit tests.
 type testResultsFunc func(testName string, release, suite string, variants []string) (*apitype.Test, error)
 
-func runJobRunAnalysis(jobRun *models.ProwJobRun, compareRelease string,
+func runJobRunAnalysis(jobRun *models.ProwJobRun, compareRelease string, jobRunTestCount int, historicalRunTestCount int,
 	testResultsFunc testResultsFunc) (apitype.ProwJobRunRiskAnalysis, error) {
 
 	logger := log.WithFields(log.Fields{
```
dgoodwin, line 210: Continuing with my theme of nitpicking logging, could you do us a favor and move this contextual logger up in the the parent function, then pass it to runJobRunAnalysis as an argument. (maybe drop or adjust the "func" in each per your preference) Then you can use this for all the logging in the function above and we'll consistently know what job run the problems are coming from. It could then be passed to any other function we log from to make sure we know what we're looking at when we see those messages.

I think I should have done it this way originally.

DEVELOPMENT.md
```diff
@@ -150,6 +155,16 @@ If you are *not* loading a backup for your data, you will need to initialize and
...
+If you *are* loading a backup created via pg_restore you need to point to the sippy_openshift db.
+
+```bash
+./sippy --server \
+  --release 4.11 \
+  --log-level=debug \
+  --database-dsn="postgresql://postgres:password@localhost:5432/sippy_openshift" \
+  --mode=ocp
+```
+
```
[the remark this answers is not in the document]
neisw: I used what Dennis has above

`pg_restore -h localhost -U postgres -p 5432 --verbose -Fc -C -d postgres ./sippy-backup-2022-10-20.dump`

It works without adding the users / roles but generates errors.  The DB goes to sippy_openshift so when my debug stopped working (before I updated the dsn)  I went chasing the errors I saw.  And then ultimately had to update the dsn.

pkg/db/query/job_queries.go
neisw: heh, I only learned since I got  a lint nag

pkg/api/job_runs.go
```diff
@@ -114,28 +115,60 @@ func JobRunRiskAnalysis(dbc *db.DB, jobRun *models.ProwJobRun) (apitype.ProwJobR
...
+
+			if err != nil {
+				log.Errorf("Error parsing minor version for %s: %v", compareRelease, err)
+			}
+
+			// ignoring the errors above after logging since we are in an edge case
+			if major < 4 || (major == 4 && minor < 13) {
+				jobRunTestCount = historicalCount
+			}
+		}
```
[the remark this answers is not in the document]
stbenjam: Sippy already vendors https://github.com/hashicorp/go-version, would that work for you? 

https://github.com/openshift/sippy/blob/7a8599a7dc28e003c4045c0a641c2b861b0f75a1/pkg/testidentification/ocp_variants.go#L385-L396

pkg/db/query/job_queries.go
stbenjam: Getting the avg number of tests would be faster to do in SQL

Example:

```sql
SELECT
    avg(count)
FROM (
    SELECT
        prow_job_run_id,
        count(*)
    FROM
        prow_job_run_tests
        INNER JOIN prow_job_runs ON prow_job_runs.id = prow_job_run_tests.prow_job_run_id
        INNER JOIN prow_jobs ON prow_jobs.id = prow_job_runs.prow_job_id
    WHERE
        prow_jobs.name = 'periodic-ci-openshift-release-master-ci-4.13-e2e-aws-ovn-upgrade'
        AND prow_job_runs.timestamp >= CURRENT_DATE - interval '7' day
    GROUP BY
        prow_job_run_id) t;
```

neisw: All that new knowledge lost with more efficient sql queries...  I was wondering about raw query performance vs gorm.  Thanks for the pointer.

DEVELOPMENT.md
```diff
@@ -40,6 +40,11 @@ psql -h localhost -U postgres -p 5432 postgres < sippy-backup-2022-05-02.sql
 
 or with a command like this if you are using the custom format file:
 
+Optional - If it's a new db then initialize roles (or just ignore the errors):
+```bash
+echo "CREATE USER sippyro;" | psql postgresql://postgres:password@localhost:5432/postgres
+echo "CREATE USER rdsadmin;" | psql postgresql://postgres:password@localhost:5432/postgres
+```
```
dgoodwin, line 47: The accounts we use probably should not be appearing in public docs esp if the errors are harmless.

DEVELOPMENT.md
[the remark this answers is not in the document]
dgoodwin: It's the -C causing that behavior, "When this option is used, the database named with -d is used only to issue the initial DROP DATABASE and CREATE DATABASE commands. All data is restored into the database name that appears in the archive.". 

Let's just drop the -C from above and then we don't need to doc a forked path.

PR #732, opened 2022-12-14
import from bigquery

dgoodwin:
- Proxy gorm logging through logrus and only show warnings and above
- Add option to import from OpenShift CI BigQuery


pkg/prowloader/bigqueryjobs.go
```diff
@@ -0,0 +1,111 @@
...
+	BuildID        string    `bigquery:"prowjob_build_id"`
+	Type           string    `bigquery:"prowjob_type"`
+	Cluster        string    `bigquery:"prowjob_cluster"`
+	StartTime      time.Time `bigquery:"prowjob_start_ts"`
+	CompletionTime time.Time `bigquery:"prowjob_completion_ts"`
+}
+
+// GetURL returns the guessed URL for the prow job using it's name and build ID.
+func (jr bigqueryProwJobRun) GetURL() string {
+	return fmt.Sprintf("https://prow.ci.openshift.org/view/gcs/origin-ci-test/logs/%s/%s", jr.JobName, jr.BuildID)
```
stbenjam, line 110: This won't work for presubmits, the URL is different.

https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-kube-apiserver-operator/1403/pull-ci-openshift-cluster-kube-apiserver-operator-master-e2e-aws-ovn/1598319526313725952/

Could we just get testplatform to store the URL in their table?

dgoodwin:
I know I promised 200 lines and 136,000 is a little big over that. Minus refactoring and vendoring we're close I think. :)

pkg/prowloader/bigqueryjobs.go
dgoodwin: Looked like it was working for some but I see now it's not consistent at the very least. Working on getting it into the bigquery db.

/hold

dgoodwin: I got the importer now setting prowjob_url, and updated this code to use it. Includes a filter for not null as that breaks serialization if we pickup any with out it set. I'll keep an eye on the db to make sure none are slipping in still, but by this time this rolls out, everything should have a URL set.

main.go
```diff
@@ -445,6 +439,51 @@ func (o *Options) Run() error { //nolint:gocyclo
...
+		log.WithError(err).Error("CRITICAL error getting GCS client which prevents importing prow jobs")
+		allErrs = append(allErrs, err)
+		return allErrs
+	}
+
+	var bigQueryClient *bigquery.Client
+	if o.LoadOpenShiftCIBigQuery {
+		bigQueryClient, err = bigquery.NewClient(context.Background(), "openshift-gce-devel",
+			option.WithCredentialsFile(o.GoogleServiceAccountCredentialFile))
+		// TODO: support GoogleOAuthClientCredentailFile like above for GCS?
```
dgoodwin, line 460: Anyone know if this is needed?

stbenjam: It's a second auth option but I don't think we've ever used it

pkg/db/db.go
```diff
@@ -29,9 +29,29 @@ type DB struct {
...
+func (w log2LogrusWriter) Printf(msg string, args ...interface{}) {
+	w.entry.Debugf(msg, args...)
+}
+
 func New(dsn string) (*DB, error) {
+	gormLogger := logger.New(
+		log2LogrusWriter{entry: log.WithField("source", "gorm")},
+		logger.Config{
+			SlowThreshold:             2 * time.Second,
+			LogLevel:                  logger.Warn, // Change to info if you want to see all sql being used, but it's a LOT
```
stbenjam, line 47: Could we get this set to `--log-level` CLI option? So I could use that instead of editing code to see the SQL queries.

pkg/prowloader/bigqueryjobs.go
```diff
@@ -0,0 +1,109 @@
...
+	row := pl.dbc.DB.Table("prow_job_runs").Select("max(timestamp)").Row()
+	err := row.Scan(&lastProwJobRun)
+	if err != nil || lastProwJobRun.IsZero() {
+		log.WithError(err).Warn("no last prow job run found (new database?), importing last two weeks")
+		lastProwJobRun = time.Now().Add(-14 * 24 * time.Hour)
+	} else {
+		// adjust the last job run time slightly, we're querying all jobs that have completed since our last recorded
+		// job START time, but we'll subtract another 10 minutes just in case there were upload delays to gcs/bigquery:
+		lastProwJobRun = lastProwJobRun.Add(-10 * time.Minute)
+	}
```
stbenjam, line 30: I don't think this is enough.  Timestamp is the job start time, so imagine this:

- Job A starts at 10am
- Job B starts at 10:15 am
- Job B completes in 15 minutes, maybe due to an infra error
- Fetchdata runs around 10:30, picks up job B

You'll never get job A from big query using this logic.

stbenjam: I think we're up to allowing 8h for aggregated jobs, that's probably the minimum lookback.  To be safe I'd say even 12h.

stbenjam:
Have we compared the two syncing mechanisms? When we went from testgrid->prow, we had 2 instances up for a while and compared the results.  It would be nice to know if the bigquery tables are reliable, and have all the same jobs we're getting when syncing from the authoritative source.

dgoodwin:
I could grab the oldest dev backup we have and restore locally, then import the last week from bigquery, and compare to prod sippy.

dgoodwin:
Code updated per review, and I have a 2 week import running for comparison I'll report back on tomorrow.

dgoodwin:
I've done some comparison with a freshly loaded 2 weeks of data from bigquery against prod sippy. Results look very good for the most part, a few things off by a little here and there which seems to depend on who imported last. The job histogram looked very different at one point (far more 100% success jobs in my local copy), but this appears to have been caused by single run jobs that were not yet picked up in the prod sippy. The charts stabilized after another refresh cycle.

To demonstrate, here's a comparison of two screenshots for the most commonly run jobs (>200):

Screenshot from 2023-01-10 13-20-24
Screenshot from 2023-01-10 13-20-32

This shows the two being virtually identical, and I think seeing this list is probably sufficient to have confidence we're picking up all the job runs normally.

stbenjam:
One other thing I meant to ask, does BigQuery have presubmit jobs too?

dgoodwin:
It does yes, prowjob_type = presubmit in the jobs table.

PR #761, opened 2023-01-10
Add test grid link to additional jobs not defined in release controller

xueqzhan:
[TRT-753](https://issues.redhat.com/browse/TRT-753)

### DIFF
--- a/pkg/prowloader/prow.go
+++ b/pkg/prowloader/prow.go
@@ -215,6 +215,18 @@ func (pl *ProwLoader) generateTestGridURL(release, jobName string) *url.URL {
 			jobType = "blocking"
 		} else if informingJobs.Has(jobName) {
 			jobType = "informing"
+		} else if strings.HasPrefix(jobName, "release-openshift-") ||
+			strings.HasPrefix(jobName, "promote-release-openshift-") ||
+			strings.HasPrefix(jobName, "periodic-ci-openshift-hypershift-main-periodics-") ||
+			strings.HasPrefix(jobName, "periodic-ci-openshift-multiarch") ||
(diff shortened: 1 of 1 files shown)


pkg/prowloader/prow.go
```diff
@@ -215,6 +215,18 @@ func (pl *ProwLoader) generateTestGridURL(release, jobName string) *url.URL {
...
+			strings.HasPrefix(jobName, "periodic-ci-openshift-hypershift-main-periodics-") ||
+			strings.HasPrefix(jobName, "periodic-ci-openshift-multiarch") ||
+			strings.HasPrefix(jobName, "periodic-ci-openshift-release-master-ci-") ||
+			strings.HasPrefix(jobName, "periodic-ci-openshift-release-master-okd-") ||
+			strings.HasPrefix(jobName, "periodic-ci-openshift-release-master-nightly-") ||
+			strings.HasPrefix(jobName, "periodic-ci-shiftstack-shiftstack-ci-main-periodic-") {
+			// This special case is here because of the special handling in test grid generator defined in
+			// https://github.com/openshift/ci-tools/blob/2492ee3127698a2e6db17f5a8e457c73fae26ddd/cmd/testgrid-config-generator/main.go#L219
+			// We should try to stay in sync.
+			jobType = "informing"
```
stbenjam, line 229: Could we not put this in the sippy-config-generator instead? At least that way it's in the same rpeo as the test-grid-generator.  And theoretically we could move them to shared code so there's no need to stay in sync.

xueqzhan: Good point!! Didn't come across my mind. :(

PR #769, opened 2023-01-16
TRT-778: add info logging to investigate missing prow_pull_request entries

neisw:
Want to see if we can get a better idea where we are stopping when calling findOrAddPullRequests


stbenjam:
AFAIK we moved away from prow.go, is there no equivalent in the bigquery loader?

PR #777, opened 2023-01-18
TRT-783: Add support for ChangeLogJSON

neisw:
Starting with a draft request to get feedback.  Brad reverted the ChangeLogJSON changes so we have time to review.

Main differences are noted in the test case

HTML Was:
`  href="https://github.com/openshift/oc/tree/983567b394f8d1efcc9c938b01af30d173c8fc5d"'`

JSON has a trailing ')' (we can remove but better handled in release controller if they can):
` "path": "https://github.com/openshift/oc/tree/983567b394f8d1efcc9c938b01af30d173c8fc5d)",`

We do not get the CurrentOSURL in JSON
In HTML we received:
`<li>Red Hat Enterprise Linux CoreOS <a target="_blank" href="https://releases-rhcos-art.apps.ocp-virt.prod.psi.redhat.com/?release=413.86.202301170928-0&amp;stream=releases%2Frhcos-4.13">413.86.202301170928-0</a></li>`

In JSON we don't get that URL, could try reconstructing it on our end if need be.
` {"name": "Red Hat Enterprise Linux CoreOS","version": "413.86.202301170928-0"}`

We don't necessarily need the new test to be pushed, but it is useful during the transition.

Are we ok with the new struct names or are there other preferences?


stbenjam:
Awesomesauce, thanks for digging into this.Ddid you share the missing data with Brad? Hopefully they can get this fixed so we can get ride of the terrible html parser I wrote

neisw:
I will share with Brad in the AM.  Didn't want to ping him late in the day.

neisw:
Discussion was to bring this in disabled and review again when the release-controller changes are reintroduced.

pkg/releasesync/releasesync.go
```diff
@@ -191,12 +191,30 @@ func releaseDetailsToDB(architecture string, tag ReleaseTag, details ReleaseDeta
...
+	// 	release.CurrentOSVersion = jsonChangeLog.CurrentOSVersion
+	// 	release.PreviousOSURL = jsonChangeLog.PreviousOSURL
+	// 	release.PreviousOSVersion = jsonChangeLog.PreviousOSVersion
+	// 	release.OSDiffURL = jsonChangeLog.OSDiffURL
+	//
+	// 	release.PreviousReleaseTag = jsonChangeLog.PreviousReleaseTag
+	// 	release.Repositories = jsonChangeLog.Repositories
+	// 	release.PullRequests = jsonChangeLog.PullRequests
+	//
+	// } else {
```
stbenjam, line 210: I would suggest writing this in a way that it will just work without needing to uncomment the code.  You should be able to check if ChangeLogJSON is populated, and if so use that.

pkg/releasesync/releasesync.go
```diff
@@ -220,6 +238,91 @@ func releaseDetailsToDB(architecture string, tag ReleaseTag, details ReleaseDeta
...
+
+	for _, c := range changeLogJSON.Components {
+		if c.Name == "Kubernetes" {
+			releaseChangeLogJSON.KubernetesVersion = c.Version
+		} else if c.Name == "Red Hat Enterprise Linux CoreOS" {
+			releaseChangeLogJSON.CurrentOSVersion = c.Version
+
+			// doesn't seem like we have all the right data to build this out
+			// https://releases-rhcos-art.apps.ocp-virt.prod.psi.redhat.com/?release=413.86.202301170928-0&stream=releases%2Frhcos-4.13
+			// releaseChangeLogJson.CurrentOSURL = fmt.Sprintf("https://releases-rhcos-art.apps.ocp-virt.prod.psi.redhat.com/?release=%s&stream=", releaseChangeLogJson.CurrentOSVersion, "???")
```
stbenjam, line 254: Is there a jira somewhere with CRT tracking adding the missing fields? These URL's are important for me when I debug suspected OS issues, it shows a diff between the last OS version and the current one

neisw:
It looked like the changelog_parser had the previous and current anchor extraction reversed when I was comparing with the JSON data.  Worth double checking.

pkg/releasesync/releasesync.go
neisw: The jira tracking the URL changes upstream from release-controller is https://issues.redhat.com/browse/OCPCRT-171.  Brads changes restoring the previously missing URLs from the JSON is https://github.com/openshift/release-controller/pull/491/files.  When comparing the output it looked as though we had our urls reversed for currentOS and previous os I updated the changelog_parser to reflect that.

pkg/releasesync/releasesync_test.go
```diff
@@ -166,3 +170,119 @@ func buildReleaseDetails(hasFailedBlockingJobs bool) ReleaseDetails {
...
+	releaseChangeLog.CurrentOSURL, releaseChangeLog.CurrentOSVersion, releaseChangeLog.PreviousOSURL, releaseChangeLog.PreviousOSVersion, releaseChangeLog.OSDiffURL = changelog.CoreOSVersion()
+	releaseChangeLog.PreviousReleaseTag = changelog.PreviousReleaseTag()
+	releaseChangeLog.Repositories = changelog.Repositories()
+	releaseChangeLog.PullRequests = changelog.PullRequests()
+
+	releaseChangeLogJSON := parseChangeLogJSON("test", releaseDetails.ChangeLogJSON)
+
+	if releaseChangeLogJSON.KubernetesVersion != releaseChangeLog.KubernetesVersion {
+		t.Fatalf("ReleaseChangeLog Kubernetes versions don't match.  ChangeLog: %s, ChangeLogJson: %s", releaseChangeLog.KubernetesVersion, releaseChangeLogJSON.KubernetesVersion)
+	}
```
stbenjam, line 204: Nit, but all these if's are challenging to read for me. If you can't directly use `reflect.DeepEqual` to just compare both structs entirely, I think this would look nicer to use helpers from testify (it's already vendored in sippy):

```go
assert.EqualValues(t, releaseChangeLogJSON.KubernetesVersion, releaseChangeLog.KubernetesVersion, "Kubernetes version")
```

It will produce a diff and a helpful error message and it's a bit more concise.

pkg/releasesync/releasesync_test.go
```diff
@@ -166,3 +170,119 @@ func buildReleaseDetails(hasFailedBlockingJobs bool) ReleaseDetails {
...
+				continue
+			}
+
+			found = true
+
+			// the quotes are different.. skip this test
+			// ReleaseChangeLog PullRequest Description don't match for console.  ChangeLog: display ‘Control plane is hosted’ alert only when isCl…, ChangeLogJson: display 'Control plane is hosted' alert only when isCl…
+			// if prBase.Description != prJSON.Description {
+			// 	t.Fatalf("ReleaseChangeLog PullRequest Description don't match for %s.  ChangeLog: %s, ChangeLogJson: %s", prJSON.Name, prBase.Description, prJSON.Description)
+			// }
```
stbenjam, line 272: What's going on here? I prefer not to merge commented out code blocks.  There's a difference between the two pull request descriptions in the changelog formats?

pkg/releasesync/releasesync_test.go
```diff
@@ -166,3 +170,119 @@ func buildReleaseDetails(hasFailedBlockingJobs bool) ReleaseDetails {
 	return releaseDetails
 
 }
+
+func TestChangeLog(t *testing.T) {
```
stbenjam, line 174: Thanks for the tests. I do find it a bit cumbersome to read, IMHO it should not be 100+ lines.  I'd maybe see about reducing some of the lines that were duplicated from releasesync.go into common helpers

stbenjam (commented):
This looks good to me, thanks for the unit tests.  I have some nitpicky things about the unit tests, but I won't hold up merging on it

stbenjam:
/lgtm
/hold

Release hold when ready

pkg/releasesync/releasesync_test.go
neisw: In this case there is a difference in the quotes between

pkg/releasesync/releasesync_test.go
neisw: Honestly I wasn't sure if we wanted the test going forward.  I used it get familiar with / build out the changes.  I don't mind putting some time into refactoring and cleaning it up.  I also didn't know if once the JSON was available we would remove the HTML parsing and not even need the test going forward.

stbenjam: Yea I guess that makes sense, I'm fine if you just want to `/hold cancel` and let this merge as is.  We can remove the HTML parsing code and the test once the new json thing is ready

neisw:
/hold cancel

Discussed in standup today.  Created https://issues.redhat.com/browse/TRT-814 to track the removal of the HTML parsing and refactoring of the final JSON parsing tests.

PR #802, opened 2023-02-02
Hide low frequency jobs by default

stbenjam:
TRT-824

This adjusts the Sippy UI to stop showing low frequency jobs by default. For jobs that run less than daily, especially once a week jobs, when one run fails it ends up showing up in various reporting showing the job significantly regressed.

### DIFF
--- a/pkg/api/health.go
+++ b/pkg/api/health.go
@@ -189,8 +189,8 @@ func calculateJobResultStatistics(results []apitype.Job) (currStats, prevStats s
 	prevStats.Histogram = make([]int, 10)
 
 	for _, result := range results {
-		// Skip jobs that have no runs
-		if result.CurrentRuns == 0 {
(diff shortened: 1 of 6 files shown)


config/e2e-openshift.yaml
```diff
@@ -0,0 +1,8 @@
+prow:
+  url: https://prow.ci.openshift.org/prowjobs.js
+releases:
+  "4.13":
+    jobs:
+      periodic-ci-openshift-release-master-nightly-4.13-e2e-aws-sdn-serial: true
+      periodic-ci-openshift-release-master-ci-4.13-e2e-azure-ovn-upgrade: true
+      periodic-ci-openshift-release-master-nightly-4.13-e2e-gcp-sdn: true
```
dgoodwin, line 8: Neat idea.

PR #824, opened 2023-02-12
TRT-845: Risk analysis pr commenting - Validate record, cap error att…

neisw:
Check the pending record before writing the comment

- Check when we last attempted to write a comment
- Skip if we are within wait period
- Track number of errors, increase wait based on that number
- Delete the record if too many errors
- Update the last attempted time when we are ready to try adding the comment
- Check the contents of existing comment if there is one and if it matches the pending update leave it and skip commenting
- Sort by job name when risk level matches so we have deterministic comment content

### DIFF
--- a/pkg/db/models/pr_commenting.go
+++ b/pkg/db/models/pr_commenting.go
@@ -11,12 +11,14 @@ const (
 )
 
 type PullRequestComment struct {
-	CreatedAt   time.Time
-	UpdatedAt   time.Time
-	PullNumber  int    `json:"pullNumber" gorm:"primaryKey"`
-	CommentType int8   `json:"commentType" gorm:"primaryKey"`
-	SHA         string `json:"sha" gorm:"primaryKey"`
-	Org         string `json:"org" gorm:"primaryKey"`
-	Repo        string `json:"repo" gorm:"primaryKey"`
-	ProwJobRoot string `json:"prowJobRoot"`
+	CreatedAt             time.Time
+	UpdatedAt             time.Time
+	PullNumber            int       `json:"pullNumber" gorm:"primaryKey"`
+	CommentType           int8      `json:"commentType" gorm:"primaryKey"`
+	SHA                   string    `json:"sha" gorm:"primaryKey"`
+	Org                   string    `json:"org" gorm:"primaryKey"`
+	Repo                  string    `json:"repo" gorm:"primaryKey"`
+	ProwJobRoot           string    `json:"prowJobRoot"`
+	LastCommentAttempt    time.Time `json:"lastCommentAttempt"`
+	FailedCommentAttempts int       `json:"failedCommentAttempts"`
 }
--- a/pkg/github/commenter/ghcommenter.go
+++ b/pkg/github/commenter/ghcommenter.go
@@ -97,7 +97,7 @@ func (ghc *GitHubCommenter) GetCurrentState(org, repo string, number int) (*gith
 	return ghc.githubClient.PRFetch(org, repo, number)
 }
 
-func (ghc *GitHubCommenter) FindExistingCommentID(org, repo string, number int, commentKey, commentID string) (*int64, error) {
+func (ghc *GitHubCommenter) FindExistingCommentID(org, repo string, number int, commentKey, commentID string) (*int64, *string, error) {
 	return ghc.githubClient.FindCommentID(org, repo, number, commentKey, commentID)
 }
 
@@ -190,7 +190,7 @@ func (ghc *GitHubCommenter) UpdatePendingCommentRecords(org, repo string, number
 
 			// otherwise we either aren't the latest sha or we have merged so clear
 			clearRecord := cmtupdt
-			ghc.ClearPendingRecord(clearRecord.Org, clearRecord.Repo, clearRecord.PullNumber, clearRecord.SHA, &clearRecord)
+			ghc.ClearPendingRecord(clearRecord.Org, clearRecord.Repo, clearRecord.PullNumber, clearRecord.SHA, commentType, &clearRecord)
 		}
 	}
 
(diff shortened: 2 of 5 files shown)


pkg/github/commenter/ghcommenter.go
```diff
@@ -213,25 +213,83 @@ func (ghc *GitHubCommenter) UpdatePendingCommentRecords(org, repo string, number
 	}
 }
 
-func (ghc *GitHubCommenter) ClearPendingRecord(org, repo string, number int, sha string, record *models.PullRequestComment) {
+func (ghc *GitHubCommenter) ValidatePendingRecordComment(org, repo string, number int, sha string, commentType models.CommentType) (bool, error) {
+	// check for this record in our table
```
DennisPeriquet, line 217: can you pls put the comment above the `func` to say "ValidatePendingRecordComment checks for this record in our table, ..."

pkg/github/commenter/ghcommenter.go
```diff
@@ -213,25 +213,83 @@ func (ghc *GitHubCommenter) UpdatePendingCommentRecords(org, repo string, number
...
+
+func (ghc *GitHubCommenter) UpdatePendingRecordErrorCount(org, repo string, number int, sha string, commentType models.CommentType) error {
+	pullRequestComment, err := ghc.getPendingRecord(org, repo, number, sha, commentType)
+
+	if err != nil {
+		return err
+	}
+
+	// this would be the tenth failure so just delete it
+	if pullRequestComment.FailedCommentAttempts > 8 {
```
DennisPeriquet, line 255: I don't understand this.  You say tenth, yet you are mentioning `8`

pkg/github/commenter/ghcommenter.go
```diff
@@ -213,25 +213,83 @@ func (ghc *GitHubCommenter) UpdatePendingCommentRecords(org, repo string, number
...
+		}
+		log.WithField("org", org).
+			WithField("repo", repo).
+			WithField("number", number).
+			WithField("sha", sha).Warn("Exceeded failed attempts, deleted comment record.")
+		return nil
+	}
+
+	res := ghc.dbc.DB.Model(&models.PullRequestComment{}).Where("org = ? AND repo = ? AND pull_number = ? AND sha = ? AND comment_type = ?  ", pullRequestComment.Org, pullRequestComment.Repo, pullRequestComment.PullNumber, pullRequestComment.SHA, pullRequestComment.CommentType).Update("failed_comment_attempts", pullRequestComment.FailedCommentAttempts+1)
+	if res.Error != nil {
```
DennisPeriquet, line 268: ah, maybe the `+1` here, makes it 10.

pkg/github/commenter/ghcommenter.go
```diff
@@ -213,25 +213,83 @@ func (ghc *GitHubCommenter) UpdatePendingCommentRecords(org, repo string, number
...
+			return res.Error
+		}
+		log.WithField("org", org).
+			WithField("repo", repo).
+			WithField("number", number).
+			WithField("sha", sha).Warn("Exceeded failed attempts, deleted comment record.")
+		return nil
+	}
+
+	res := ghc.dbc.DB.Model(&models.PullRequestComment{}).Where("org = ? AND repo = ? AND pull_number = ? AND sha = ? AND comment_type = ?  ", pullRequestComment.Org, pullRequestComment.Repo, pullRequestComment.PullNumber, pullRequestComment.SHA, pullRequestComment.CommentType).Update("failed_comment_attempts", pullRequestComment.FailedCommentAttempts+1)
```
DennisPeriquet, line 267: ah, maybe the `+1` here, makes it 10.

pkg/github/commenter/ghcommenter.go
DennisPeriquet: and please include what the bool return value means.  I think false means didn't do anything (either due to error or not enough time elapsed) and true means it updated the last comment.

pkg/sippyserver/pr_commenting_processor.go
```diff
@@ -127,6 +127,9 @@ type RiskAnalysisEntryList []RiskAnalysisEntry
 func (r RiskAnalysisEntryList) Len() int      { return len(r) }
 func (r RiskAnalysisEntryList) Swap(i, j int) { r[i], r[j] = r[j], r[i] }
 func (r RiskAnalysisEntryList) Less(i, j int) bool {
+	if r[i].Value.RiskLevel.Level == r[j].Value.RiskLevel.Level {
+		return r[i].Value.Name > r[j].Value.Name
```
DennisPeriquet, line 131: can you pls explain why this is necessary?  if the risk levels are identical, how is sorting by name a meaningful ordering?

pkg/github/commenter/ghcommenter.go
DennisPeriquet: Just a humble :-) opinion on the name `ValidatePendingRecordComment`, ... are you validating or updating the last attempt?

pkg/github/commenter/ghcommenter.go
```diff
@@ -213,25 +213,83 @@ func (ghc *GitHubCommenter) UpdatePendingCommentRecords(org, repo string, number
...
+	// this would be the tenth failure so just delete it
+	if pullRequestComment.FailedCommentAttempts > 8 {
+		res := ghc.dbc.DB.Delete(pullRequestComment)
+		if res.Error != nil {
+			return res.Error
+		}
+		log.WithField("org", org).
+			WithField("repo", repo).
+			WithField("number", number).
+			WithField("sha", sha).Warn("Exceeded failed attempts, deleted comment record.")
```
DennisPeriquet, line 263: optional: add some text about `Exceeded n failed attempts` for whatever we set for `n`.

pkg/github/commenter/ghcommenter.go
neisw: If the pending comment is valid the function will update the last attempt and return true.  If the pending comment is not valid it will not make any update and return false.  I can move the doc for that and change the name to ValidateAndUpdatePendingRecordComment if that is clearer.

pkg/github/commenter/ghcommenter.go
neisw: The incoming failure is the 10th.  It could be written as >= 9 but I went with > 8 (meaning it is likely 9) and the incoming attempt is the 10th so bail.

pkg/sippyserver/pr_commenting_processor.go
neisw: We want the order of the comment text to always be the same so we can compare the comment update with the existing comment and be sure we are writing a new comment (not a second processing of the existing run which can happen).  So the order that the failed jobs appears in the comment is important along with ranking based on the Level.

pkg/db/models/pr_commenting.go
```diff
@@ -11,12 +11,14 @@ const (
...
+	CreatedAt             time.Time
+	UpdatedAt             time.Time
+	PullNumber            int       `json:"pullNumber" gorm:"primaryKey"`
+	CommentType           int8      `json:"commentType" gorm:"primaryKey"`
+	SHA                   string    `json:"sha" gorm:"primaryKey"`
+	Org                   string    `json:"org" gorm:"primaryKey"`
+	Repo                  string    `json:"repo" gorm:"primaryKey"`
+	ProwJobRoot           string    `json:"prowJobRoot"`
+	LastCommentAttempt    time.Time `json:"lastCommentAttempt"`
+	FailedCommentAttempts int       `json:"failedCommentAttempts"`
```
DennisPeriquet, line 23: postgres/db-migration noob question: I see these are new.  How do those get added? is there some db-migration that just happens automatically?

neisw: Yep, I know you found it already but for documentation purposes..

https://github.com/openshift/sippy/blob/4fca373f93a9e0b89585744ce3f5cd61ac11eb73/pkg/db/db.go#L128

stbenjam: Yea there's some gorm magic that does it when we call AutoMigrate on the models

https://github.com/openshift/sippy/blob/master/pkg/db/db.go#L66

PR #826, opened 2023-02-14
Test analysis page should URI encode test names for bugs API

stbenjam:
TRT-843

Currently test names aren't being URI encoded for the bugs API, causing errors when there's characters like `#` in a test name.

Example: https://sippy.dptools.openshift.org/sippy-ng/tests/4.13/analysis?test=Create%20the%20different%20workloads%20from%20Add%20page.Create%20the%20different%20workloads%20from%20Add%20page%20Create%20Sample%20Application%20from%20Add%20page%3A%20GS-03-TC05%20(example%20%231)


stbenjam:
/override ci/prow/e2e

There's no UI tests in e2e, and this is a UI change only

PR #829, opened 2023-02-16
Add infrastructure success metrics

stbenjam:
[TRT-858](https://issues.redhat.com//browse/TRT-858)

This adds new metrics for per-platform infrastructure success rates. This is a global value not constrained by releases, but we could scope it by release if we thought it was useful.

This metric is the true success rate, so alerts based on this value will be taking into account both failures and flakes (i.e. retrying infra worked). This will let us make sure that we don't excessively start flaking on infra and not detect the problem.

### DIFF
--- a/pkg/testidentification/ocp_variants.go
+++ b/pkg/testidentification/ocp_variants.go
@@ -21,7 +21,8 @@ var openshiftJobsNeverStableRaw string
 var openshiftJobsNeverStable = strings.Split(openshiftJobsNeverStableRaw, "\n")
 
 var (
-	// variant regexes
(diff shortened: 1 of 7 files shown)


pkg/testidentification/ocp_variants.go
```diff
@@ -98,6 +99,20 @@ var (
 		"vsphere-ipi",
 		"vsphere-upi",
 	)
+
+	allPlatforms = sets.NewString(
```
deepsm007, line 103: Should we add libvirt?

stbenjam: Sure, we didn't have a libvirt variant but I just added one

PR #834, opened 2023-02-16
TRT-800: Missing variant data during Risk analysis

neisw:
Addresses [TRT-838](https://issues.redhat.com/browse/TRT-838) as well as [TRT-800](https://issues.redhat.com/browse/TRT-800)

- Missing variant data like presubmit / agnostic tests
- Missing data all together - Sets Risk Level MissingData
- Missing tests - Sets Risk Level IncompleteTests
- Missing variant match on TestRuns, secondary pass looking for the first super set of variants
- Accepts job variant data on incoming test-failures-summary data for risk analysis
- Performs secondary pass to match test run when we don't have complete variant data for a match
- Looks for new job-variants.json on import to fill in variant data from cluster info when available

### DIFF
--- a/pkg/api/health.go
+++ b/pkg/api/health.go
@@ -182,35 +182,39 @@ func calculateJobResultStatistics(results []apitype.Job) (currStats, prevStats s
 		prevPercentages = append(prevPercentages, result.PreviousPassPercentage)
 	}
 
-	data := stats.LoadRawData(currPercentages)
-	mean, _ := stats.Mean(data)
-	sd, _ := stats.StandardDeviation(data)
-	quartiles, _ := stats.Quartile(data)
-	p95, _ := stats.Percentile(data, 95)
+	if len(currPercentages) > 0 {
+		data := stats.LoadRawData(currPercentages)
+		mean, _ := stats.Mean(data)
+		sd, _ := stats.StandardDeviation(data)
+		quartiles, _ := stats.Quartile(data)
+		p95, _ := stats.Percentile(data, 95)
 
-	currStats.Mean = mean
-	currStats.StandardDeviation = sd
-	currStats.Quartiles = []float64{
-		testreportconversion.ConvertNaNToZero(quartiles.Q1),
-		testreportconversion.ConvertNaNToZero(quartiles.Q2),
-		testreportconversion.ConvertNaNToZero(quartiles.Q3),
+		currStats.Mean = mean
+		currStats.StandardDeviation = sd
+		currStats.Quartiles = []float64{
+			testreportconversion.ConvertNaNToZero(quartiles.Q1),
+			testreportconversion.ConvertNaNToZero(quartiles.Q2),
+			testreportconversion.ConvertNaNToZero(quartiles.Q3),
+		}
+		currStats.P95 = p95
 	}
-	currStats.P95 = p95
 
-	data = stats.LoadRawData(prevPercentages)
-	mean, _ = stats.Mean(data)
-	sd, _ = stats.StandardDeviation(data)
-	quartiles, _ = stats.Quartile(data)
-	p95, _ = stats.Percentile(data, 95)
+	if len(prevPercentages) > 0 {
+		data := stats.LoadRawData(prevPercentages)
+		mean, _ := stats.Mean(data)
+		sd, _ := stats.StandardDeviation(data)
+		quartiles, _ := stats.Quartile(data)
+		p95, _ := stats.Percentile(data, 95)
 
-	prevStats.Mean = mean
-	prevStats.StandardDeviation = sd
-	prevStats.Quartiles = []float64{
-		testreportconversion.ConvertNaNToZero(quartiles.Q1),
-		testreportconversion.ConvertNaNToZero(quartiles.Q2),
-		testreportconversion.ConvertNaNToZero(quartiles.Q3),
+		prevStats.Mean = mean
+		prevStats.StandardDeviation = sd
+		prevStats.Quartiles = []float64{
+			testreportconversion.ConvertNaNToZero(quartiles.Q1),
+			testreportconversion.ConvertNaNToZero(quartiles.Q2),
+			testreportconversion.ConvertNaNToZero(quartiles.Q3),
+		}
+		prevStats.P95 = p95
 	}
-	prevStats.P95 = p95
 
 	return currStats, prevStats
 }
--- a/pkg/api/job_runs.go
+++ b/pkg/api/job_runs.go
@@ -196,18 +196,29 @@ func JobRunRiskAnalysis(dbc *db.DB, jobRun *models.ProwJobRun, jobRunTestCount i
 				},
 				LinkOperator: "and",
 			}
-			trs, _, err := BuildTestsResults(dbc, release, "default", false, false,
+			testResults, _, err := BuildTestsResults(dbc, release, "default", false, false,
 				fil)
 			if err != nil {
 				return nil, err
 			}
 			gosort.Strings(variants)
-			for _, tr := range trs {
+			for _, testResult := range testResults {
 				// this is a weird way to get the variant we want, but it allows re-use
 				// of the existing code.
-				gosort.Strings(tr.Variants)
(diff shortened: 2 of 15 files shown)


pkg/api/job_runs.go
```diff
@@ -211,6 +211,17 @@ func JobRunRiskAnalysis(dbc *db.DB, jobRun *models.ProwJobRun, jobRunTestCount i
 				}
 			}
 
+			// otherwise, what is our best match...
+			// do something more expensive and check to see
+			// which tr contains all the variants we have currently
+			for _, tr := range trs {
```
stbenjam, line 217: trs = test results? Please say test result at least in the comment, or [change the variable names](https://blogg.itverket.no/dont-think-shortening-variable-names-2/)

pkg/api/job_runs.go
```diff
@@ -211,6 +211,17 @@ func JobRunRiskAnalysis(dbc *db.DB, jobRun *models.ProwJobRun, jobRunTestCount i
...
+				// we didn't find an exact variant match
+				// next best guess is the first variant list that contains all of our known variants
+				if stringSubSlicesEqual(variants, tr.Variants) && tr.SuiteName == suite {
+					return &tr, nil
+				}
```
stbenjam, line 222: - How is trs sorted? I think it's not deterministic which set of variants you'd pick, meaning different analyses could end up doing different comparisons.

- This seems to require we have at a test result that has *all* of the variants from the job run, but I think what we want is really a fuzzy best match kind of thing that works both ways.  If we get a job run that's "amd64, ha, fips" but we didn't have "fips" results, we'd be ok taking "amd64, ha" from the test results.

pkg/db/models/prow.go
```diff
@@ -175,3 +176,16 @@ type ProwPullRequest struct {
...
+	FromRelease           string
+	Platform              string
+	Architecture          string
+	Network               string
+	Topology              string
+	NetworkStack          string
+	CloudRegion           string
+	CloudZone             string
+	ClusterVersionHistory []string
+}
```
stbenjam, line 191: Just a thought, but I wonder if we shouldn't just actually store this information for a job run instead of throwing it away.  

We could address that once we get cluster-data.json written out and consumed by sippy

stbenjam (commented):
Looks like you're actively pushing changes so I'll review more later, but a few comments

pkg/api/job_runs.go
neisw: sure, just copied what was above but I can refactor both

pkg/api/job_runs.go
neisw: Yes, this is non deterministic but it is also meant to be a last resort / better than 'unknown'.  I added it before we pulled in the job variant data so we may not even want it.

pkg/db/models/prow.go
neisw: So this is getting pulled in on import now and feeding the variants.  We can look to build that out more for things like CloudRegion, zone & ClusterVersionHistory

pkg/api/job_runs.go
neisw: I did want the early feedback on this so thanks.  I think the original case was we were missing the platform for the agnostic case.  So I added this so that we could match a test run (anyone really) that had at least the same set of variants that the incoming test had.  Ultimately with the addition of the job variant data coming in we probably don't need this but we can discuss other use cases and how we want this to behave overall.

neisw: If you are saying I should add a third pass that flipped the superset check I can definitely do that.  I can also look to collect all the superset matches and the add some logic to which one gets returned maybe the fewest differences or a sorted list of the variant values.

pkg/testidentification/ocp_variants_test.go
```diff
@@ -50,7 +52,7 @@ func Test_openshiftVariants_IdentifyVariants(t *testing.T) {
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			v := openshiftVariants{}
-			if got := v.IdentifyVariants(tt.name, tt.release); !reflect.DeepEqual(got, tt.want) {
+			if got := v.IdentifyVariants(tt.name, tt.release, models.ClusterData{}); !reflect.DeepEqual(got, tt.want) {
```
stbenjam, line 55: Could we add a couple test cases for the cluster data? I think it'd be pretty straight forward to add a cluster data field to the tests struct above

neisw: Yeah, good point and nice catch.  

Wondering if the scan in gcs is making the load data time out in the e2e now... didn't see that before I renamed the file but can't imagine that would do it.

stbenjam:
https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_sippy/834/pull-ci-openshift-sippy-master-e2e/1628844017833742336/artifacts/e2e/sippy-e2e-sippy-e2e-setup/artifacts/sippy-load.log has the log

Looks like it's failing to import from the cluster-data file.  We're still deploying the one from the install step?  Maybe we need to remove it if it's not compatible

pkg/api/health.go
```diff
@@ -182,35 +182,39 @@ func calculateJobResultStatistics(results []apitype.Job) (currStats, prevStats s
...
-	p95, _ = stats.Percentile(data, 95)
-
-	prevStats.Mean = mean
-	prevStats.StandardDeviation = sd
-	prevStats.Quartiles = []float64{
-		testreportconversion.ConvertNaNToZero(quartiles.Q1),
-		testreportconversion.ConvertNaNToZero(quartiles.Q2),
-		testreportconversion.ConvertNaNToZero(quartiles.Q3),
-	}
-	prevStats.P95 = p95
```
neisw, line 213: Not related to the work in this PR but we are getting back empty percentages arrays that generate NaN mean, etc. which then cause JSON serialization errors in the e2e so added a check to make sure we have data.

neisw: Looks like https://github.com/openshift/sippy/pull/844/files#diff-1094b7bba124eec28f852d0ca61f46517f446298e4e4d119f5f6c771e1032c0bL157 

Removed some code that was handling the NaNs previously

`Percentage: testreportconversion.ConvertNaNToZero(testReport.CurrentPassPercentage),`

neisw:
/hold

hold to coordinate merge timing between origin and this pr

pkg/api/health.go
neisw: Not sure that code is in play, it is looking more like the currentRuns is < 7 so we continue in calculateJobResultStatistics loop and wind up with no entries in currPercentages or prevPercentages. 

[PR 854](https://github.com/openshift/sippy/pull/854) was used to validate the issue was external to this PR

pkg/apis/api/types.go
```diff
@@ -756,6 +756,8 @@ type RiskLevel struct {
 
 var FailureRiskLevelNone = RiskLevel{Name: "None", Level: 0}
 var FailureRiskLevelLow = RiskLevel{Name: "Low", Level: 1}
-var FailureRiskLevelMedium = RiskLevel{Name: "Medium", Level: 5}
-var FailureRiskLevelUnknown = RiskLevel{Name: "Unknown", Level: 7}
-var FailureRiskLevelHigh = RiskLevel{Name: "High", Level: 10}
+var FailureRiskLevelUnknown = RiskLevel{Name: "Unknown", Level: 25}
+var FailureRiskLevelMedium = RiskLevel{Name: "Medium", Level: 50}
+var FailureRiskLevelIncompleteTests = RiskLevel{Name: "IncompleteTests", Level: 75}
```
deepsm007, line 761: Should be have a bigger difference between incompleteTests (level: 70)and missingData (level: 80)??

neisw: My feeling is they were similar in that something was missing.  Overall I wanted to spread the ranges out more but I do think we should cluster the things that are common.  If we had another missing / incomplete we could make it 74 or 77, etc. but I definitely wanted to have room between High and MissingData for things that might come up that we consider a higher risk.

PR #838, opened 2023-02-20
Retrieve html url so that we can link to the actual PR in sippy UI

DennisPeriquet:
[TRT-859](https://issues.redhat.com//browse/TRT-859)

I believe we want the html url for the pull request info:

```
$ curl https://api.github.com/repos/openshift/cluster-resource-override-admission-operator/pulls/101 > x.json
$ cat x.json |jq .url
"https://api.github.com/repos/openshift/cluster-resource-override-admission-operator/pulls/101"
$ cat x.json |jq .html_url
"https://github.com/openshift/cluster-resource-override-admission-operator/pull/101"
```

The .html_url is the html link to the pull request which is what a user would want to click on vs. the other link in .url which is the raw api endpoint.

### DIFF
--- a/pkg/prowloader/github/github.go
+++ b/pkg/prowloader/github/github.go
@@ -285,7 +285,7 @@ func (c *Client) PRFetch(org, repo string, number int) (prEntry *PREntry, err er
 		prEntry = &PREntry{
 			MergedAt: pr.MergedAt,
 			Title:    pr.Title,
-			URL:      pr.URL,
+			URL:      pr.HTMLURL,
 			State:    pr.State,
 		}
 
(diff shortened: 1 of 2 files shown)


stbenjam:
Looks like the e2e failures are because there's been no recent job runs of the `nightly` jobs configured in https://github.com/openshift/sippy/blob/master/config/e2e-openshift.yaml#L6-L8, since we haven't had some nightlies in a while. I think it's OK to override

/lgtm
/override ci/prow/e2e

DennisPeriquet:
@stbenjam thanks.  I was trying to figure out why the test hasn't passed since since Feb 17. I did notice a difference in logs that pass vs. fail and this stuck out: `time="2023-02-20T21:16:43.664Z" level=warning msg="upgrade report for release did not include test: [sig-sippy] upgrade should work" release="{4.13}"` So that makes sense.

PR #840, opened 2023-02-21
Add http metrics for sippy api's

stbenjam:
[TRT-808](https://issues.redhat.com//browse/TRT-808)

This uses go-http-metrics to start measuring http response times so we can hopefully gather more data and see if things are slower during mat view refreshes, or other performance bottlenecks.


neisw:
/hold

Just to keep it from merging until you are around

stbenjam:
Failing again because lack of jobs

/override ci/prow/e2e

PR #841, opened 2023-02-21
Bump sippy to use ubi9 and go 1.18

stbenjam:
We're currently running go1.16, and for some things I want to do with collecting metrics on the http endpoints, I need go1.17 or newer. Seems like a good opportunity to go to ubi9 too.


stbenjam:
/override ci/prow/e2e

Also requires https://github.com/openshift/release/pull/36611 to merge concurrently.

A bit of a catch22, both PR's need to merge.

stbenjam:
/hold

I want release to merge first

PR #870, opened 2023-03-08
Change how aggregated variants are filtered out

stbenjam:
[TRT-837](https://issues.redhat.com//browse/TRT-837)

This moves the filtering of aggregation from test results from being hardcoded to being done in the UI.  By moving it to the UI, the filter can be removed and you can view tests for only aggregation. We need to do this because there are tests that *only* appear in aggregation (e.g. the new disruption tests).

Aggregated will be filtered out by default, but users wanting to see the aggregated results can add a filter for `variants contains aggregated`.

The reason for needing a default exclusion is because the jobs that make up an aggregated run are already tracked in Sippy. We are injecting unwanted hysteresis that will cause changes to pass rates to lag behind if you do not filter it out. The effect is probably small. 

Here's an example:

The follow test runs 10x and passes 7.

```
[sig-pizza] Pizza should be delivered within 30 minutes
```

It has a pass rate of 70%, in line with historical results.

If it is aggregated, we have an 11th entry in Sippy of a pass. Sippy actually thinks the test passed 8 times out of 11 -- a pass rate of 72.7%.

### DIFF
--- a/pkg/db/matviews.go
+++ b/pkg/db/matviews.go
@@ -258,7 +258,7 @@ FROM prow_job_run_tests
    LEFT JOIN suites on suites.id = prow_job_run_tests.suite_id
    JOIN prow_job_runs ON prow_job_runs.id = prow_job_run_tests.prow_job_run_id
    JOIN prow_jobs ON prow_job_runs.prow_job_id = prow_jobs.id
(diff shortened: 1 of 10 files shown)


stbenjam:
/hold

I want to think about this a little more to see if there's a better way

stbenjam:
/hold cancel

Tested and I think this approach works

PR #902, opened 2023-03-28
Add an API for fetching job run intervals

dgoodwin:
This will hopefully soon be used by a sippy UI similar to the current
spyglass charts.

For now we fetch from GCS. In future, we may be pulling from loki or
opensearch.

To work around the fact there are often two intervals files (upgrade and
conformance phases), for now we return both added together with a
filename attached to each.

[TRT-917](https://issues.redhat.com//browse/TRT-917)


pkg/api/jobrunintervals/gcsscanner.go
```diff
@@ -0,0 +1,216 @@
...
+	"regexp"
+	"strings"
+	"time"
+
+	"github.com/openshift/sippy/pkg/apis/api"
+	log "github.com/sirupsen/logrus"
+	"golang.org/x/net/html"
+)
+
+// Original code borrowed from https://github.com/vrutkovs/kaas/
```
stbenjam, line 18: Could we not consolidate our other GCS scanning code to make it work for both?

https://github.com/openshift/sippy/tree/master/pkg/prowloader/gcs

dgoodwin: Wasn't aware of that code, should be possible, I'll take a look.

pkg/api/jobrunintervals/gcsscanner.go
```diff
@@ -0,0 +1,216 @@
...
+	baseFile := tokens[len(tokens)-1]
+
+	resp, err := http.Get(url.String())
+	if err != nil {
+		return nil, err
+	}
+
+	defer resp.Body.Close()
+
+	body, err := ioutil.ReadAll(resp.Body)
```
stbenjam, line 199: How big are the intervals files? You could use Go's built in ReverseProxy to just stream the data https://pkg.go.dev/net/http/httputil#ReverseProxy

pkg/api/jobrunintervals/gcsscanner.go
dgoodwin: Ok I gave this a go, but it bubbles up into the server now needing gcs client, gcs credentials, possibly new manifest changes for the cli args. All of this would be for code I intend to fully delete once I can get these stored in some system like opensearch/loki. I don't think switching the current gcs code to use what I added is worthwhile as the current code is probably more correct.

My request here is to leave this be, I will delete it once I can get a better source for the intervals to live and query from.

dgoodwin: Actually hang on, we may still need this code. Unless we decide to stuff everything into opensearch/loki soon, we're going to need something to load individual jobs we're interested. I expected that to be sippy, if we don't have the data, pull it from gcs and load it up. In that case, gcs would still be needed, and doing this more properly may make sense.

pkg/api/jobrunintervals/gcsscanner.go
dgoodwin: They're around 15-30M today, I've switched the code over to the existing GCS stuff though so hopefully this isn't a concern, or one we can address if it becomes an issue.

pkg/sippyserver/server.go
```diff
@@ -892,6 +893,57 @@ func (s *Server) jsonJobRunRiskAnalysis(w http.ResponseWriter, req *http.Request
...
+
+	logger = logger.WithField("jobRunID", jobRunID)
+
+	jobRun, _, err = api.FetchJobRun(s.db, jobRunID, logger)
+
+	if err != nil {
+		api.RespondWithJSON(http.StatusBadRequest, w, map[string]interface{}{
+			"code": http.StatusBadRequest, "message": err.Error()})
+		return
+	}
```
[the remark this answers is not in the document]
dgoodwin: Pushed an update for this

pkg/api/jobrunintervals/gcsscanner.go
dgoodwin: Now switched over to use the gcs code you wrote for prow. Seems faster too.

main.go
```diff
@@ -657,6 +657,14 @@ func (o *Options) runServerMode(pinnedDateTime *time.Time, gormLogLevel gormlogg
 		return err
 	}
 
+	gcsClient, err := gcs.NewGCSClient(context.TODO(),
+		o.GoogleServiceAccountCredentialFile,
+		o.GoogleOAuthClientCredentialFile,
+	)
+	if err != nil {
+		log.WithError(err).Warn("unable to create GCS client, some APIs may not work")
```
stbenjam, line 665: If bigquery support will be optional going forward we should maybe use the `capabilities` feature in pkg/sippyserver/server.go, but I expect it won't be optional much longer

PR #918, opened 2023-04-05
Improve the DateTimePicker local vs. UTC display

DennisPeriquet:
This change makes the DateTimePicker stay in UTC time for header text (i.e., the hour displayed at the top).

PR-pic1

The DateTimePicker interactive arrows remain in local time (header text remains UTC).

Since `getHours(date)` method was adding more time to the time represented in the interactive arrow on the clock it was removed.  The `getMinutes(date)` date didn't look like it was doing anything.

`getHourText(date)` sets the header text to UTC; I ignore `ampm` to keep it in 24 hour format.

`format(date, formatString)` affects the time in the Date slot for the date that was chosen.

Set a few other dates to UTC format. 

TODO:

- [ ] Try a few more tests before review.

### DIFF
--- a/sippy-ng/src/datagrid/GridToolbarFilterDateUtils.js
+++ b/sippy-ng/src/datagrid/GridToolbarFilterDateUtils.js
@@ -5,18 +5,16 @@ export class GridToolbarFilterDateUtils extends DateFnsUtils {
     super()
   }
 
-  getHours(date) {
-    return date.getUTCHours()
-  }
-
-  getMinutes(date) {
-    return date.getUTCMinutes()
-  }
-
   format(date, formatString) {
     return `${format(utcToZonedTime(date, 'UTC'), formatString, {
       timeZone: 'Etc/UTC',
       locale: this.locale,
     })}`
+  }
+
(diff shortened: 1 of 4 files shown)


stbenjam:
Thanks for looking at this, I had invested some time before and could never figure it out.

Anyway, this looks like an improvement but at least for me the hour picker shows local time, with header in UTC:

image

DennisPeriquet:
Yes, it's an improvement but not quite there.  I made the DateTimePicker header showing time in UTC (instead of suddenly changing after clicking on the day).

For now, the interactive arrow on the clock remains as local time.  That will require a more involved effort because I think it's the DateTimePicker that has to move to UTC which is probably something that can be fixed using a theme on the DateTimePicker or a global theme that uses UTC time.

stbenjam:
The other option is picking a different library like https://github.com/wojtekmaj/react-datetime-picker, although it doesn't match the style of material-ui.

We could also split the date and time into two inputs with https://v4.mui.com/components/pickers/, and use two input boxes for date and time and combine them together.

PR #921, opened 2023-04-07
Make Jobs table match query for build clusters

DennisPeriquet:
[TRT-955](https://issues.redhat.com//browse/TRT-955)

For the build cluster, the query is for [periodic jobs](https://github.com/openshift/sippy/blob/d6f82be99bbb7c170bf580a80febf6cedbc36644/pkg/db/query/build_clusters.go#L66) only.  This makes the query for the JobsTable match.

### DIFF
--- a/sippy-ng/src/build_clusters/BuildClusterDetails.js
+++ b/sippy-ng/src/build_clusters/BuildClusterDetails.js
@@ -1,5 +1,5 @@
 import { Card, Container, Grid, Typography } from '@material-ui/core'
-import { filterFor, getReportStartDate } from '../helpers'
+import { filterFor, getReportStartDate, not } from '../helpers'
 import { JobStackedChart } from '../jobs/JobStackedChart'
 import { ReportEndContext } from '../App'
 import JobRunsTable from '../jobs/JobRunsTable'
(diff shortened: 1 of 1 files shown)


sippy-ng/src/build_clusters/BuildClusterDetails.js
```diff
@@ -13,13 +13,16 @@ export default function BuildClusterDetails(props) {
...
+            <Typography variant="h6">
+              Periodic Job Runs On {props.cluster}
+            </Typography>
             <JobRunsTable
               pageSize={10}
               hideControls={true}
               filterModel={{
                 items: [
                   filterFor('cluster', 'equals', props.cluster),
+                  filterFor('name', 'starts with', 'periodic'),
```
stbenjam, line 25: There are some periodics that start with `release-` or `promote-` unfortunately.  I think you could do ` not(filterFor('name', 'starts with', 'pull'))` to be more accurate

deepsm007: Will a regex match be possible for `*periodic*`?

stbenjam: The periodics that start with release or promote don't contain periodic anywhere in the name, but I believe all presubmits begin with `pull-`

DennisPeriquet: @deepsm007 we probably could, but if I understand correctly, we'd have to plumb a new regex-style `Operator` and logic to implement it [here](https://github.com/openshift/sippy/blob/cebacf96c35f9c540cea8348c8f94cb662383b8b/pkg/filter/filterable.go#L64).

PR #937, opened 2023-04-19
TRT-974: Log error but continue with VariantTestReport

neisw:


pkg/api/install.go
```diff
@@ -107,9 +107,11 @@ func VariantTestsReport(dbc *db.DB, release string, reportType v1.ReportType,
 	for testName := range tests {
 		allReport, err := query.TestReportExcludeVariants(dbc, release, testName, excludedVariants)
 		if err != nil {
-			return variantColumns, tests, err
+			// log the error and keep going
+			log.WithError(err).Errorf("Failed to query test report for: %s", testName)
+		} else {
+			tests[testName]["All"] = allReport
```
stbenjam, line 113: What does allReport contain if it returns an error?

neisw: Looks empty in the debugger and based on https://github.com/openshift/sippy/blob/c343a00e15c79c9ee6ce27aed51d3eb7a9147869/pkg/db/query/test_queries.go#L151

PR #940, opened 2023-04-21
Port Prow Job Intervals UI to Sippy

dgoodwin:
TRT-918

Today we store fixed html artifacts after every job run which we use to debug jobs. However once published this html can't be easily updated or tweaked with improvements. It's also not possible to share links directly to things we're looking at.

This work ports the existing intervals chart over to sippy giving us a central location we can link to from prow jobs. We can rollout improvements to the UI in sippy and have them immediately available for past jobs. All filtering includes support for query parameters so we can link directly to what we're looking at.

Uses a pre-existing API that pulls all intervals files from GCS and merges them together, but the UI has support for filtering down to just one, or both files if present. By default you're shown the first after sorting the filenames. (timestamped)

For now, the prow job ID you specify must be in your DB, this is because I can't find the bucket for the job without more info. Once we move to loki, a prow job ID should be enough and we can then view intervals from a sippy that has no jobs actually imported. This is important so there are no delays to using this tool, while waiting for sippy to actually import the job on the hour.

I have not ported David's latest work on the pre-defined 10 rows you can filter with using lodash, this is just a simpler single regex specification, plus some buttons that allow you to include/exclude categories or files. 

A default selection of categories is provided if none are specified.


sippy-ng/src/App.js
```diff
@@ -436,6 +437,13 @@ export default function App(props) {
                     )}
                   />
 
+                  <Route
+                    path="/prow_job_runs/:jobrunid"
```
dgoodwin, line 441: Was thinking this would be a generic page for details of a job run, perhaps it should be more intervals specific. I can't imagine we'd want anything else on the screen.

stbenjam:
I think you're missing the timelines-chart in packages.json

sippy-ng/src/App.js
stbenjam: I could see us having more info about a job run (risk anaylsis maybe?) or eventually even some testgrid like view.

WDYT about `/jobs_runs/:jobrunid/intervals` ?  Just so if we don't break urls later if we need to refactor things. Also do we need prow in the url name?

sippy-ng/src/components/TimelineChart.js
```diff
@@ -0,0 +1,130 @@
...
+// in origin previously.
+export default function TimelineChart({ eventIntervals, data }) {
+  const ref = useRef(null)
+
+  useEffect(() => {
+    let chart = null
+    const node = ref.current
+    if (node) {
+      //const el = document.querySelector('#chart')
+      let ordinalScale = d3scale
```
stbenjam, line 16: I think this could be a `const` outside of useEffect, it won't change if data changes.

sippy-ng/src/prow_job_runs/ProwJobRun.js
```diff
@@ -0,0 +1,926 @@
...
+      ['pod_logs', 'Pod Logs'],
+      ['interesting_events', 'Interesting Events'],
+      ['alerts', 'Alerts'],
+      ['node_state', 'Node State'],
+      ['e2e_test_failed', 'E2E Failed'],
+      ['e2e_test_flaked', 'E2E Flaked'],
+      ['e2e_test_passed', 'E2E Passed'],
+      ['disruption', 'Disruption'],
+    ])
+  )
```
stbenjam, line 35:
```js
const allCategories = {
  operator_unavailable: 'Operator Unavailable',
  operator_progressing: 'Operator Progressing',
  operator_degraded: 'Operator Degraded',
  pods: 'Pods (careful)',
  pod_logs: 'Pod Logs',
  interesting_events: 'Interesting Events',
  alerts: 'Alerts',
  node_state: 'Node State',
  e2e_test_failed: 'E2E Failed',
  e2e_test_flaked: 'E2E Flaked',
  e2e_test_passed: 'E2E Passed',
  disruption: 'Disruption'
};
```

didn't work?

sippy-ng/src/prow_job_runs/ProwJobRun.js
```diff
@@ -0,0 +1,926 @@
...
+
+  const [categoryButtonState, setCategoryButtonState] = useState(() => {
+    const params = new URLSearchParams(window.location.search)
+
+    let categories = []
+    let categoriesParam = params.get('categories')
+    if (categoriesParam) {
+      categories = categoriesParam.split(',')
+    }
```
stbenjam, line 44: Can you use `useQueryParams` instead? There are string array params built-in, and it manages the URLs for you.

sippy-ng/src/prow_job_runs/ProwJobRun.js
```diff
@@ -0,0 +1,926 @@
...
+      initialState['operator_unavailable'] = true
+      initialState['operator_progressing'] = true
+      initialState['operator_degraded'] = true
+      initialState['pod_logs'] = true
+      initialState['interesting_events'] = true
+      initialState['alerts'] = true
+      initialState['node_state'] = true
+      initialState['e2e_test_failed'] = true
+      initialState['disruption'] = true
+    }
```
stbenjam, line 66: If you used useQueryParams, you wouldn't need any of this.  I'd expect this to be a default prop on the component like this:

```js
ProwJobRun.defaultProps = {
  categories:  {
    operator_unavailable: true,
    operator_progressing: true,
    operator_degraded: true,
    pod_logs: true,
    interesting_events: true,
    alerts: true,
    node_state: true,
    e2e_test_failed: true,
    disruption: true,
  },
}

ProwJobRun.propTypes = {
  categories: PropTypes.object,
}
```

And then for useQueryParams, the default value would be the prop:

```
 const [names = props.categories, setCategories] = useQueryParam('category', ArrayParam) 
```

stbenjam: And then if someone clicked the button, you'd just call `setCategories(<new categories map>)` and the URL would be updated, the categories variable would get updated, and the intervals table should automatically rerender

sippy-ng/src/prow_job_runs/ProwJobRun.js
```diff
@@ -0,0 +1,926 @@
...
+    return ''
+  })
+
+  const fetchData = () => {
+    if (isLoaded) {
+      console.log(
+        'fetchData called due to UI update but intervals already loaded, re-using intervals'
+      )
+      return
+    }
```
stbenjam, line 100: You could avoid this but having multiple `useEffect` hooks, so you could have one with no conditions so fetchdata gets called only one time, and then one that calls updatefiltering when  [categoryButtonState, history, selectedIntervalFiles]) gets  updated

sippy-ng/src/prow_job_runs/ProwJobRun.js
dgoodwin: I don't know what I did, but this works, thanks!

sippy-ng/src/prow_job_runs/ProwJobRun.js
dgoodwin: Gave it a shot, let me know what you think.

dgoodwin:
I think I got everything, and it seems to be working.

stbenjam:
Need to run `npm install timelines-chart` so it gets pulled into packages.json.  That's why the tests are failing here.

Once I do that it looks like it works well - a few comments you could fix later after merging this

- Any idea why in firefox I get these big text headers? 

image

- Jobs that don't have intervals the page goes blank http://localhost:3000/sippy-ng/job_runs/1651642012622917632/intervals - the api probably needs to return a 404 instead of `{items:null}`

dgoodwin:
Doh I did the npm command in top level sippy, it's fixed now. 

Those headers, I saw those as well at one point, and then they went away. I've developed this on Firefox so I'm really unclear why that's showing up and going away.

PR #948, opened 2023-04-26
Fix the logs to get better idea of search.ci calls

DennisPeriquet:
This helps us get better logging re: TRT-970.

Log the number of requests to tell us how many batches (of 50) were successfully processed.

Use log messages specific to tests vs. jobs so we can tell where the errors actually came from.


stbenjam:
Nice.  There's some failures on this in the e2e run, it fails immediately :(

/lgtm
/override ci/prow/e2e

DennisPeriquet:
thanks.  gonna take a look at that e2e test https://issues.redhat.com/browse/TRT-990

PR #955, opened 2023-04-30
TRT-892: Match jobnames, not variants

neisw:
Review changes switching from variant matches to jobname matches

### DIFF
--- a/pkg/api/job_runs.go
+++ b/pkg/api/job_runs.go
@@ -178,28 +178,34 @@ func findReleaseMatchJobNames(dbc *db.DB, jobRun *models.ProwJobRun, compareRele
 					// how do we handle never-stable
 					if len(job.Variants) == 1 && job.Variants[0] == testidentification.NeverStable {
 						hasNeverStableJob = true
-					} else {
-						gosort.Strings(job.Variants)
-						if stringSlicesEqual(variants, job.Variants) {
+					}
 
-							jobIds, err := query.ProwJobRunIds(dbc, job.ID)
+					gosort.Strings(job.Variants)
+					if stringSlicesEqual(variants, job.Variants) {
 
-							if err != nil {
-								logger.WithError(err).Errorf("Failed to query job run ids for %d", job.ID)
-								continue
-							}
+						jobIds, err := query.ProwJobRunIds(dbc, job.ID)
 
-							totalJobRunsCount += len(jobIds)
-							allJobNames = append(allJobNames, "'"+job.Name+"'")
+						if err != nil {
+							logger.WithError(err).Errorf("Failed to query job run ids for %d", job.ID)
+							continue
 						}
+
+						totalJobRunsCount += len(jobIds)
+						allJobNames = append(allJobNames, "'"+job.Name+"'")
 					}
 				}
 
-				if len(allJobNames) == 0 && hasNeverStableJob {
-					return nil, 0, errors.New(testidentification.NeverStable)
+				// logging at info for now so we can monitor, can dial down to debug if / when preferred
+				if len(allJobNames) > 0 {
+					logger.Infof("Matched job name: %s to %v", jobRun.ProwJob.Name, allJobNames)
 				}
 
-				return allJobNames, totalJobRunsCount, nil
+				var err error
+				if hasNeverStableJob {
+					err = errors.New(testidentification.NeverStable)
+				}
+
+				return allJobNames, totalJobRunsCount, err
 			}
 
 		}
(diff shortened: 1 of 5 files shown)


pkg/api/job_runs.go
```diff
@@ -116,6 +118,78 @@ func FetchJobRun(dbc *db.DB, jobRunID int64, logger *log.Entry) (*models.ProwJob
 	return jobRun, jobRunTestCount, nil
 }
 
+func findReleaseMatchJobNames(dbc *db.DB, jobRun *models.ProwJobRun, compareRelease string, logger *log.Entry) ([]string, int, error) {
```
dgoodwin, line 121: I think I can see what this is doing but I am not immediately clear why, could you add godoc or PR description to outline the change being made and the algorithm to do so, and maybe provide some small real world examples. I'm unclear about the segment matching as well as how variants continue to play into it.

pkg/api/job_runs.go
```diff
@@ -116,6 +121,99 @@ func FetchJobRun(dbc *db.DB, jobRunID int64, logger *log.Entry) (*models.ProwJob
...
+
+							jobIds, err := query.ProwJobRunIds(dbc, job.ID)
+
+							if err != nil {
+								logger.WithError(err).Errorf("Failed to query job run ids for %d", job.ID)
+								continue
+							}
+
+							totalJobRunsCount += len(jobIds)
+							allJobNames = append(allJobNames, "'"+job.Name+"'")
```
dgoodwin, line 193: We should log when we find a matching job name, perhaps at debug level, either here or when we get the final list back out of the func.

pkg/api/job_runs.go
```diff
@@ -272,6 +436,13 @@ func runJobRunAnalysis(jobRun *models.ProwJobRun, compareRelease string, jobRunT
 		response.OverallRisk.Reasons = append(response.OverallRisk.Reasons,
 			fmt.Sprintf("%d tests failed in this run: High", len(jobRun.Tests)))
 		return response, nil
+
+	// Return early is this is for a never-stable job
+	case neverStableJob:
+		response.OverallRisk.Level = apitype.FailureRiskLevelLow
+		response.OverallRisk.Reasons = append(response.OverallRisk.Reasons,
+			"Job is marked as never-stable")
+		return response, nil
```
dgoodwin, line 445: Is it feasible to proceed with test analysis on these so we could be watching risk of jobs we're hoping to get OUT of never-stable, or is that largely impossible? With variant matching probably not feasible, but with job name we might get something useful?

pkg/api/job_runs.go
```diff
@@ -336,6 +527,64 @@ func runJobRunAnalysis(jobRun *models.ProwJobRun, compareRelease string, jobRunT
...
+	}
+
+	// we had the case where etcd-scaling was reporting
+	// high risk using the variant methodology due
+	// to the analysis including more stable non scaling job results
+	// so, we don't want to just return the highest risk analysis
+	// bias is towards jobname analysis at this point
+	// we could check for the number of runs but
+	// that introduces flakiness on which analysis we are using
+	// for now we have a deterministic result based on the jobnames analsysis so return it
```
dgoodwin, line 575: nit: typo in analysis

pkg/api/job_runs_test.go
```diff
@@ -267,3 +367,78 @@ func TestSubSliceEqual(t *testing.T) {
...
+		},
+		{
+			name:              "variants risk level",
+			variantsRiskLevel: &apitype.Test{CurrentPassPercentage: 100, CurrentRuns: 100, Variants: []string{"aws", "ovn"}},
+			expectedResponse:  []string{"This test has passed 100.00% of 100 runs on release 4.14 [aws ovn] in the last week."},
+			expectedRiskLevel: apitype.FailureRiskLevelHigh.Level,
+			compareRelease:    "4.14",
+			jobNames:          []string{"jobname-4.14", "jobname-4.13"},
+		},
+	}
```
dgoodwin, line 434: Nice tests!

dgoodwin (commented):
This looks pretty good to me just a few nits, it seems we'll be able to easily tell which version got used so that should do the tricky nicely.

pkg/api/job_runs.go
dgoodwin: Disregard I saw later it was in the result message.

pkg/api/job_runs.go
neisw: We can try.  If it looks to be at all flaky I say dump it and get value from the other 99% though.

PR #970, opened 2023-05-08
Use regex for metadata extractor function for pathological tests

DennisPeriquet:
[TRT-1015](https://issues.redhat.com//browse/TRT-1015)

This allows us to process metadata for all pathological tests.  There used to be only two of them; there are now 126:

```
sippy_openshift=> select count(*) from tests where name  LIKE '%pathological%';
 count 
-------
   126
```

Using this regex will allow us to get the 124 new ones.

### DIFF
--- a/pkg/prowloader/testoutputmetadata.go
+++ b/pkg/prowloader/testoutputmetadata.go
@@ -1,6 +1,7 @@
 package prowloader
 
 import (
+	"fmt"
 	"reflect"
 	"regexp"
 	"strings"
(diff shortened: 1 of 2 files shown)


pkg/prowloader/testoutputmetadata.go
```diff
@@ -20,12 +21,26 @@ func GetTestOutputMetadataExtractors() map[string]TestOutputMetadataExtractorFun
...
+	extractorFunc, ok := GetTestOutputMetadataExtractors()[testName]
+	if ok {
+		return extractorFunc, nil
+	}
+	if pathologicalTestNameMatch.MatchString(testName) {
+		return pathologicalEventsMetadataExtractor, nil
+	}
+	return nil, fmt.Errorf("extractor function not found for %s", testName)
+}
+
```
dgoodwin, line 43: This looks ok, but could you add a test to the testoutputmetadata_test.go file and make sure we're getting what we expect? Make sure the new test output is the same as the old so we can be confident once deployed, that we're getting what we expect.

DennisPeriquet: I changed the test output to have a real log (mostly so the diff captures my new test and not part of the existing one).

PR #975, opened 2023-05-10
Bump golangci-lint and fix new issues

stbenjam:
golangci-lint is too old for go1.19, so we need to bump it, and fix all the new issues. This needs to merge first since buildroot changes aren't tested in presubmits.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -5,7 +5,6 @@ run:
 linters:
   disable-all: true
   enable:
-    - deadcode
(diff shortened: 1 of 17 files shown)


stbenjam:
/override ci/prow/lint

Docker.buildroot changes don't get tested in presubmits. It works locally for me though.

neisw:
/lgtm

Ran make lint locally as well to double check

PR #977, opened 2023-05-10
Allow test analysis page to show tests that belong to aggregated jobs

xueqzhan:
[TRT-1018](https://issues.redhat.com/browse/TRT-1018)

### DIFF
--- a/sippy-ng/src/tests/TestAnalysis.js
+++ b/sippy-ng/src/tests/TestAnalysis.js
@@ -51,7 +51,6 @@ export function TestAnalysis(props) {
     filterModel = {
       items: [
         filterFor('name', 'equals', testName),
-        not(filterFor('variants', 'contains', 'aggregated')),
         not(filterFor('variants', 'contains', 'never-stable')),
       ],
     },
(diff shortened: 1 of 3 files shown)


sippy-ng/src/tests/TestAnalysis.js
```diff
@@ -51,7 +51,6 @@ export function TestAnalysis(props) {
     filterModel = {
       items: [
         filterFor('name', 'equals', testName),
-        not(filterFor('variants', 'contains', 'aggregated')),
```
stbenjam, line 54: This reintroduces the aggregated variant for ALL tests, meaning you'll be throwing off the overall numbers for every test.  See https://github.com/openshift/sippy/pull/404#issue-1217311618 for an explanation of why this adds unwanted hysteresis.

One solution is from the tests table is to link to the analysis page with the variant filters it has currently set.

stbenjam: aggregated is currently also a terminal variant, like never-stable.  It's worthwhile to fix that, so aggregated test and job analysis will show the per-nurp performance, but we need to check that aggregated gets filtered out everywhere else by default (do our metrics filter that out?) too.

stbenjam: Once a job gets assigned a terminal variant it gets no others (never-stable and aggregated), there's a card to fix it https://issues.redhat.com/browse/TRT-1004.

stbenjam:
This PR looks great, thanks

/lgtm

PR #991, opened 2023-05-23
PowerVS: Add PowerVS to sippy variant detection

hamzy:
There are now two PowerPC CIs: one using libvirt and one using PowerVS.

### DIFF
--- a/pkg/testidentification/ocp_variants.go
+++ b/pkg/testidentification/ocp_variants.go
@@ -49,6 +49,7 @@ var (
 	ovirtRegex     = regexp.MustCompile(`(?i)-ovirt`)
 	ovnRegex       = regexp.MustCompile(`(?i)-ovn`)
 	// proxy jobs do not have a trailing -version segment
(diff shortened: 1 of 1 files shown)


pkg/testidentification/ocp_variants.go
```diff
@@ -49,7 +49,8 @@ var (
 	ovirtRegex     = regexp.MustCompile(`(?i)-ovirt`)
 	ovnRegex       = regexp.MustCompile(`(?i)-ovn`)
 	// proxy jobs do not have a trailing -version segment
-	ppc64leRegex      = regexp.MustCompile(`(?i)-ppc64le`)
```
deepsm007, line 52: Its better to only add a new variant for powervs rather than updating the existing to `libvirt-ppc64le`.

hamzy: I would have thought making the other one more precise would be preferable, but I will go back to the more confusing way.

PR #1002, opened 2023-06-01
Ignore disruption by default an match api parameter names in ComponentReadiness UI

DennisPeriquet:
[TRT-1068](https://issues.redhat.com//browse/TRT-1068)

This is dependent on https://github.com/openshift/sippy/pull/1003
I tested this against PR1003 and traversed all 5 pages and got no errors from the API.

### DIFF
--- a/sippy-ng/src/component_readiness/CompReadyUtils.js
+++ b/sippy-ng/src/component_readiness/CompReadyUtils.js
@@ -282,8 +282,8 @@ export function getUpdatedUrlParts(
     confidence: confidence,
     pity: pity,
     minFail: minFail,
-    ignoreDisruption: ignoreDisruption,
-    ignoreMissing: ignoreMissing,
+    ignore_disruption: ignoreDisruption,
+    ignore_missing: ignoreMissing,
     //component: component,
   }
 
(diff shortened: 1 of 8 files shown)


sippy-ng/src/component_readiness/CompReadyUtils.js
```diff
@@ -282,8 +282,8 @@ export function getUpdatedUrlParts(
     confidence: confidence,
     pity: pity,
     minFail: minFail,
-    ignoreDisruption: ignoreDisruption,
-    ignoreMissing: ignoreMissing,
+    ignore_disruption: ignoreDisruption,
+    ignore_missing: ignoreMissing,
```
stbenjam, line 286: Why the mismatch between camel and snake case?

DennisPeriquet: the left side is used for the api call; the right side is the value I'm using.

DennisPeriquet: the left side is used for the api call; the right side is the value I'm using.

stbenjam: Why? The rest of the values here are camel case above these two lines

PR #1007, opened 2023-06-02
Add search in rows and columns for page 1

DennisPeriquet:

* Add some TextField UI elements for the regexes
* Re-factor code that forms columns (so it can be re-used to make it clear what the comparison is against)
* applying the filters to component and column names (and column results). 

NOTE:  The changes look bigger due to the indentations.

### DIFF
--- a/sippy-ng/src/component_readiness/ComponentReadiness.js
+++ b/sippy-ng/src/component_readiness/ComponentReadiness.js
@@ -11,6 +11,7 @@ import {
   cancelledDataTable,
   formatLongDate,
   formatLongEndDate,
+  formColumnName,
   getAPIUrl,
   getColumns,
   getUpdatedUrlParts,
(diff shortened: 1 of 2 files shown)


sippy-ng/src/component_readiness/ComponentReadiness.js
```diff
@@ -145,6 +147,17 @@ export default function ComponentReadiness(props) {
   const classes = useStyles()
   const theme = useTheme()
 
+  const [searchComponentRegex, setSearchComponentRegex] = useState('')
```
stbenjam, line 150:
```suggestion
  const [searchComponentRegex, setSearchComponentRegex] = useQueryParam(
    'searchComponent',
    SafeStringParam
  )
```

stbenjam:
Disregard my comment, the query parameter makes me lose characters when I type too fast, we can figure it out later if we want these searches shared in the URL

/lgtm

stbenjam:
/override ci/prow/e2e

It's search.ci again

PR #1009, opened 2023-06-02
e2e interval charts: limit tooltip width to first word width

vrutkovs:
Two quality-of-life improvements for e2e charts:

*  wrap tooltips in e2e chart to make sure that tooltips don't overflow X when a long label is used.
*  copy segment label contents to clipboard when its clicked

Ref: https://github.com/openshift/origin/pull/27965

### DIFF
--- a/sippy-ng/src/components/TimelineChart.js
+++ b/sippy-ng/src/components/TimelineChart.js
@@ -106,6 +106,20 @@ export default function TimelineChart({ eventIntervals, data }) {
         .maxLineHeight(20)
         .maxHeight(10000)
         .zColorScale(ordinalScale)
+        .segmentTooltipContent((d) => {
+          return (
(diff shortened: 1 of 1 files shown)


dgoodwin:
Probably was not fair of me to suggest you add this here Vadim, something looks broken and it would require a little knowledge on standing up sippy db, api, and node UI. We can pickup this PR from here, thank you for trying, it will help us get it in regardless.

vrutkovs:
Sure, I don't mind leaving it under review for some time

stbenjam:
I think the e2e test was just hitting search.ci problems, I tested this locally and works for me, but I'm curious if @dgoodwin saw some other problem

/test e2e

dgoodwin:
No I didn't dig into the failures, mistakenly assumed it was the PR.

PR #1011, opened 2023-06-02
Add Assessment to ComponentReadiness TestReport

DennisPeriquet:
The "Probability of ..." text has also been revamped based on the assessment.


DennisPeriquet:
need to debug a Nan value

PR #1020, opened 2023-06-05
Show test suite and name in component readiness

stbenjam:
If you look at this page:

https://sippy.dptools.openshift.org/sippy-ng/component_readiness/capability?baseEndTime=2023-05-16%2023%3A59%3A59&baseRelease=4.13&baseStartTime=2023-04-18%2000%3A00%3A00&capability=Other&component=Networking%20%2F%20cluster-network-operator&confidence=95&groupBy=cloud%2Carch%2Cnetwork&ignoreDisruption=true&ignoreMissing=false&minFail=3&pity=5&sampleEndTime=2023-06-05%2023%3A59%3A59&sampleRelease=4.14&sampleStartTime=2023-05-08%2000%3A00%3A00

You'll notice a lot of duplicate tests. It's because we consider same named tests from different suites as different tests (it's possible two suites could define a generic test like "TestClusterCreate" and we want to make sure we're not combining results from different suites).

This adds the suite name to the API, and shows it in the UI.

### DIFF
--- a/pkg/api/component_report.go
+++ b/pkg/api/component_report.go
@@ -11,11 +11,12 @@ import (
 
 	"cloud.google.com/go/bigquery"
 	fischer "github.com/glycerine/golang-fisher-exact"
-	apitype "github.com/openshift/sippy/pkg/apis/api"
-	"github.com/openshift/sippy/pkg/util/sets"
 	"github.com/pkg/errors"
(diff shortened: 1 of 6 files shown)


DennisPeriquet:
I pulled this down and visited all the pages that changed.  I also see the testSuite on the front of the test names.  Looks good.

PR #1022, opened 2023-06-05
Allow selecting only red components

DennisPeriquet:

More page 1 features to allow users to focus on what they care about.

I focus only on the components/rows to avoid too many filters.  If we really want to filter on red columns, we can do that separately.

### DIFF
--- a/sippy-ng/src/component_readiness/ComponentReadiness.js
+++ b/sippy-ng/src/component_readiness/ComponentReadiness.js
@@ -22,6 +22,7 @@ import {
   noDataTable,
 } from './CompReadyUtils'
 import {
+  Checkbox,
   Drawer,
   Grid,
   TableContainer,
@@ -156,6 +157,11 @@ export default function ComponentReadiness(props) {
   const handleSearchColumnRegexChange = (event) => {
     const searchValue = event.target.value
     setSearchColumnRegex(searchValue)
+  }
+
+  const [redOnlyChecked, setRedOnlyChecked] = React.useState(false)
+  const handleRedOnlyCheckboxChange = (event) => {
+    setRedOnlyChecked(event.target.checked)
   }
 
   const [drawerOpen, setDrawerOpen] = React.useState(true)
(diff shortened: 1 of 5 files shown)


stbenjam:
I like the idea of having the controls all in the same area, what do you think about moving this next to the search filters? And how hard would it be to only show the red squares, i.e. make the rest of the green blank / gray? You could maybe [make the green square images greyscale](https://www.w3schools.com/howto/howto_css_image_bw.asp)? That'd make it easy to see the red

image

DennisPeriquet:
cool.  I didn't know you could do that grayscale thing - thx.
Checkbox on top does look cleaner -- will do.

sippy-ng/src/component_readiness/ComponentReadiness.js
```diff
@@ -801,6 +807,22 @@ export default function ComponentReadiness(props) {
...
+                        />
+                        <label
+                          htmlFor="redOnlyCheckbox"
+                          style={{
+                            textAlign: 'left',
+                            marginTop: 15,
+                          }}
+                        >
+                          Red Only
+                        </label>
```
stbenjam, line 825:
```suggestion
                        <FormControlLabel
                          control={
                            <Checkbox
                              checked={redOnlyChecked}
                              onChange={handleRedOnlyCheckboxChange}
                              color="primary"
                              size="small"
                              style={{ borderRadius: 1 }}
                            />
                          }
                          htmlFor="redOnlyCheckbox"
                          style={{
                            textAlign: 'left',
                            marginTop: 15,
                          }}
                          label="Red Only"
                        ></FormControlLabel>
```

Using the material UI label makes the alignment of the checkbox a little better

Old:
image


New:
image

DennisPeriquet: Ah, definitely nicer.  Styling changes I tried could not get that text closer.

stbenjam:
I think you'll need to add the import too:

```
[eslint] 
src/component_readiness/ComponentReadiness.js
  Line 810:26:  'FormControlLabel' is not defined  react/jsx-no-undef
```

Unfortunately github didn't let me make a suggestion on the imports

sippy-ng/src/component_readiness/ComponentReadiness.js
DennisPeriquet: will add import

PR #1024, opened 2023-06-06
Include column filtering when selecting Red Only

DennisPeriquet:

This applies the "Red Only" checkbox so we also filter out columns that have no red.


stbenjam:
Neat

/lgtm

PR #1031, opened 2023-06-08
Revert "Include column filtering when selecting Red Only"

DennisPeriquet:
This reverts commit 22e0928893e00bba4e292dd5c053aba0da0232f3 ([TRT-1082](https://issues.redhat.com//browse/TRT-1082)).

Looks like this broke the part about getting the column/environment.

Holding because a simple solution is under test ...


DennisPeriquet:
I'm convinced https://github.com/openshift/sippy/pull/1032 addresses the problem; the mistake seems pretty obvious.

PR #1032, opened 2023-06-08
fix misplaced parenthesis on column match

DennisPeriquet:
A fix that might supersede the revert in https://github.com/openshift/sippy/pull/1031.

The second test in my check is not part of the match (which mangled up the `searchColumnRegex` test).  This seems to work on preliminary testing - but I'm going to test more thoroughly.


DennisPeriquet:
Interesting error; the optional `{}` and optional `return` on arrow function made this non-obvious.  I'm curious if this is one of the javascript "classic" bugs.

PR #1057, opened 2023-06-26
Add to/from dates to test report page plus human readable message

DennisPeriquet:
[TRT-1098](https://issues.redhat.com//browse/TRT-1098)

Add the dates in the test report for easy reference.  Attempt to print out a human interpretable summary related to the GA dates we know about.

yada

### DIFF
--- a/sippy-ng/src/component_readiness/CompReadyTestReport.js
+++ b/sippy-ng/src/component_readiness/CompReadyTestReport.js
@@ -35,20 +35,6 @@ const cancelFetch = () => {
   abortController.abort()
 }
 
-const printStats = (statsLabel, stats) => {
-  return (
-    <Fragment>
-      {statsLabel} Release: {stats.release}
-      <ul>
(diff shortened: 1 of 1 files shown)


sippy-ng/src/component_readiness/CompReadyTestReport.js
```diff
@@ -125,8 +122,28 @@ export default function CompReadyTestReport(props) {
...
+          })
+          .forEach((r) => {
+            tmpRelease[r] = data.ga_dates[r]
+          })
+        setVersions(tmpRelease)
+      })
+      .catch((error) => console.error(error))
+      .finally(() => {
+        isVersionFetchDone = true
+        checkFetchesDone()
```
stbenjam, line 146: We're duplicating this in a few places, I think it'd be better to have a context for the release API.  It's the solution I'm working on as an alternative to #1058.

stbenjam:
/lgtm

Can replace the /api/releases call with a context later

PR #1058, opened 2023-06-26
Change dates only on click of 4week, 2week, 4WeeksBeforeGA buttons

DennisPeriquet:
This makes it so that the dates on the DatePicker are changed only when the user clicks one of the buttons.  This avoids the dates being reset when the first page refreshes.


stbenjam:
This prevents the GA date being set for the historical release though, I'm working on a fix for this already

DennisPeriquet:
ha. I needed it to test the https://github.com/openshift/sippy/pull/1057 :-) glad you're already on it, I'll wait for your fix -- thanks!

PR #1064, opened 2023-06-27
Refactor release selector

stbenjam:
I think this works correctly now.

### DIFF
--- a/sippy-ng/src/App.js
+++ b/sippy-ng/src/App.js
@@ -190,7 +190,7 @@ export default function App(props) {
   let landingPage = ''
   if (fetchError !== '') {
     landingPage = <Alert severity="error">{fetchError}</Alert>
-  } else if (releases.releases.length > 0) {
+  } else if (releases?.releases?.length > 0) {
     landingPage = (
       <ReleaseOverview
         key={releases.releases[0]}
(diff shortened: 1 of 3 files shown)


stbenjam:
/hold

Not quite, I've confused something

sippy-ng/src/App.js
```diff
@@ -189,282 +190,291 @@ export default function App(props) {
   let landingPage = ''
   if (fetchError !== '') {
     landingPage = <Alert severity="error">{fetchError}</Alert>
-  } else if (releases.length > 0) {
-    landingPage = <ReleaseOverview key={releases[0]} release={releases[0]} />
+  } else if (releases.releases.length > 0) {
```
DennisPeriquet, line 193: Not required, but to avoid stack traces, we can protect against null `releases` or null `releases.releases`, like: `if (releases && releases.releases && releases.releases.length > 0)` or `if (releases?.releases?.length) {`

stbenjam: Yea that's a fair comment. Will fix

stbenjam: Fixed

DennisPeriquet:
I tested this and I see that the date is no longer reset after clicking the GenerateReport button.

PR #1088, opened 2023-07-10
Dedupe junit table results in component readiness

stbenjam:
[TRT-1139](https://issues.redhat.com//browse/TRT-1139)

Using the simpler fix for DISTINCT(file_path) doesn't seem to work for
all pages. We have the  unfortunate scenario where we're getting duplicated test results in many places, which is caused by both retried failures and a bug in openshift-tests that causes duplicates (https://issues.redhat.com/browse/OCPBUGS-16039).

### DIFF
--- a/pkg/api/component_report.go
+++ b/pkg/api/component_report.go
@@ -20,37 +20,27 @@ import (
 )
 
 const (
-	// Some test suites in OpenShift retry, resulting in potentially multiple
-	// failures for the same test in a job.  Component Readiness is currently
-	// counting these as separate failures, resulting in an outsized impact on
-	// our statistical analysis.
+	// This query de-dupes the test results. There are two issues:
 	//
-	// Counting them only once with the schema we have is challenging, this
-	// creates a couple of CTE to split the table into failed test results,
-	// and non-failed (success and flake).  We partition the failed test
-	// results by name and prow job ID, and only select the first row from
-	// each partition and then union it all with the non-failed table.
-	singleFailureJunitTable = `
-		WITH failed_testcases AS (
+	// 1. Some test suites in OpenShift retry, resulting in potentially multiple
+	//    failures for the same test in a job.  Component Readiness is currently
+	//    counting these as separate failures, resulting in an outsized impact on
+	//    our statistical analysis.
+	//
+	// 2. There is a second bug where successful test cases are sometimes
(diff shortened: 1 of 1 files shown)


pkg/api/component_report.go
```diff
@@ -19,6 +19,30 @@ import (
...
+	//    our statistical analysis.
+	//
+	// 2. There is a second bug where successful test cases are sometimes
+	//    recorded by openshift-tests more than once, it's tracked by
+	//    https://issues.redhat.com/browse/OCPBUGS-16039
+	//
+	dedupedJunitTable = `
+		WITH deduped_testcases AS (
+			SELECT  *,
+				ROW_NUMBER() OVER(PARTITION BY file_path, test_name, testsuite ORDER BY success_val, flake_count) AS row_num
```
xueqzhan, line 37: Do I understand this correctly that, if there are multiple rows with different success_val (0 and 1 in retry case), you prefer 0?

stbenjam: Yea that's the way it works, but maybe I should prefer success or flake instead when we have multiple rows

xueqzhan: It seems more reasonable to me that way. If the retry succeeds, I would think we would like that to be a pass.

stbenjam: Moving to DESC broke a lootttt. Something is off with this still

image

xueqzhan: This is weird. My little experiment seems to indicate it is fine. It is puzzling.

stbenjam: Yea DESC won't work because flake rows keep their failures around - we need ASC sort so we get successes and flakes first.

From Justin:

```
Sure. DOM is too heavy, so the XML is streamed in. There is no global reasoning about all elements up front.
https://github.com/jupierce/prowjobs-to-bigquery/blob/375954cec27f4a38bc85507326ba3c3781b3ea2c/gcs_finalize_event.py#L562-L579
So the second row is a success but includes a 'flake_count' of 1 informing you that a preceding failure row is ultimately to be considered a flake.
True failure count is determined by total-sum(flake)-sum(success). Embodied in the fisher's exact POC in code like: https://github.com/jupierce/fishers_exact/blob/7078ff7fc6eff18f133221c608bd742e9876b07f/readiness/main/views.py#L562 .
```

PR #1089, opened 2023-07-10
remove jobs without release impact from sippy main page

deads2k:
Per the PPC team, regressions in the libvirt ppc jobs do not impact our release readiness.  Because of this, we don't want them to disruption the signal of other ppc or other libvirt jobs in our standard summarization or regression analysis.  The team can still inspect their results by checking a job filter in sippy, but the rest of the org will have a cleaner ship/no-ship analysis to watch and focus their efforts on.


jaypoulz:
It would have been good to include the *libvirt-s390x jobs as part of the patch. I'll see if I can file a follow-up tomorrow so we can keep these in sync.

jaypoulz:
After continued conversation today, I am awaiting guidance from leadership before taking further action.

PR #1092, opened 2023-07-11
TRT-1125: add check for test with suite name when missing analysis re…

neisw:
As documented in [TRT-1125](https://issues.redhat.com/browse/TRT-1125) found a case where we failed to find test results due to the name being prepened with the suite name in the db.

When we don't find results this makes a second pass with the suite name prepended to the test name.

[sig-imageregistry] Image registry remains available using new connections -> Cluster upgrade.[sig-imageregistry] Image registry remains available using new connections

Open question, should we limit the suite names we do this for?  Currently it will try for all suites any time we don't get valid results back.

### DIFF
--- a/pkg/api/job_runs.go
+++ b/pkg/api/job_runs.go
@@ -467,6 +467,14 @@ func runJobRunAnalysis(jobRun *models.ProwJobRun, compareRelease string, jobRunT
 		if testResultsJobNameFunc != nil {
 			if len(jobNames) < 5 && len(jobNames) > 0 {
 				testResultsJobNames, errJobNames = testResultsJobNameFunc(ft.Test.Name, jobNames)
+
+				if errJobNames == nil && testResultsJobNames != nil {
+					if testResultsJobNames.CurrentRuns == 0 {
+						// do we need to prepend the suite name to the test?
+						testResultsJobNames, errJobNames = testResultsJobNameFunc(fmt.Sprintf("%s.%s", ft.Suite.Name, ft.Test.Name), jobNames)
+					}
+				}
+
 			} else {
 				loggerFields.Warningf("Skipping job names test analysis due to jobNames length: %d", len(jobNames))
 			}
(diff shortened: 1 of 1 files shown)


pkg/api/job_runs.go
```diff
@@ -467,6 +467,14 @@ func runJobRunAnalysis(jobRun *models.ProwJobRun, compareRelease string, jobRunT
 		if testResultsJobNameFunc != nil {
 			if len(jobNames) < 5 && len(jobNames) > 0 {
 				testResultsJobNames, errJobNames = testResultsJobNameFunc(ft.Test.Name, jobNames)
+
+				if errJobNames == nil && testResultsJobNames != nil {
+					if testResultsJobNames.CurrentRuns == 0 {
+						// do we need to prepend the suite name to the test?
+						testResultsJobNames, errJobNames = testResultsJobNameFunc(fmt.Sprintf("%s.%s", ft.Suite.Name, ft.Test.Name), jobNames)
```
DennisPeriquet, line 474: is that a suite name?

There are 3 suites:

```sql
sippy_openshift=> select name from suites ;
          name           
-------------------------
 openshift-tests
 openshift-tests-upgrade
 sippy
(3 rows)
```

But looking for that pattern above, there are a lot of other things that come before the ".""

```sql
sippy_openshift=> SELECT DISTINCT
    SUBSTRING(name FROM 1 FOR POSITION('.' IN name) - 1) AS name_prefix
FROM
    tests
WHERE
    name LIKE '%.[%'
ORDER BY
    name_prefix;
                                name_prefix                                 
----------------------------------------------------------------------------
 Backup from one node and recover on another
 Cluster API Suite
 Cluster Controller Suite
 Cluster upgrade
 CNF Features e2e integration tests
 CNF Features e2e setup
 CNF Features e2e validation
 Controller Suite
 E2E Suite
 End-to-end
 Environment Setup Suites
 github
 Integration Suite
 Kubeconfig Controller Suite
 Kubernetes e2e suite
 Machine Suite
 metal infra
 Metallb Operator E2E Suite
 Metallb Operator Validation Suite
 OADP E2E Suite
 odo e2e scenarios
 OSD e2e suite
 Performance Addon Operator configuration
 Performance Addon Operator latency e2e tests
 Performance Test
 PlatformOperators Suite
 PTP e2e integration tests
 PTP e2e tests : Parallel
 Quorum Loss and Restore
 Repos
 [sig-storage] Ensure that all storage objects are deleted during the tests
 SRIOV Operator conformance tests
 Subsystem Suite
 Tests Suite
 Test Suites
(35 rows)
```

Does `ft.Suite.Name` get you what you want?

dgoodwin: This is precisely the trouble I got into when trying to split the suite names out, possibly related to some legacy compatability we may no longer care about. When we parse the xml, we should explicitly know the suite name. It's just a question of migrating data perhaps.

dgoodwin:
Also I can't think of a reason you'd need to filter on suite names, in this location it looks like we explicitly know it, unlike other places where we have to guess at supported list of suites. Change still makes sense to me.

/lgtm

PR #1093, opened 2023-07-11
Sync and show jira incidents on the sippy calendar

stbenjam:
[TRT-1067](https://issues.redhat.com//browse/TRT-1067)

image

### DIFF
--- a/pkg/api/jira.go
+++ b/pkg/api/jira.go
@@ -1,11 +1,13 @@
 package api
 
 import (
+	"time"
+
 	apitype "github.com/openshift/sippy/pkg/apis/api"
 	"github.com/openshift/sippy/pkg/db"
 )
 
-func GetJIRAIncidentsFromDB(dbClient *db.DB) ([]apitype.CalendarEvent, error) {
+func GetJIRAIncidentsFromDB(dbClient *db.DB, start, end *time.Time) ([]apitype.CalendarEvent, error) {
 	// Get JIRA Incidents for display in calendar
 	var incidents []apitype.CalendarEvent
 	res := dbClient.DB.Table("jira_incidents").Select(`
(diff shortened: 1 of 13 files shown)


pkg/jiraloader/jiraloader.go
```diff
@@ -0,0 +1,114 @@
...
+	   us public jira cards which is probably what we want, I don't think we're ever doing non-public
+	   incidents. That way we don't leak anything embargoed. If at some point we do need a token, you
+	   can do it with the below, but make sure it has limited privileges and can only see public cards.
+
+		token := os.Getenv("JIRA_TOKEN")
+		req.Header.Add("Authorization", "Bearer "+token)
+	*/
+
+	client := &http.Client{}
+	req, err := http.NewRequest("GET", "https://issues.redhat.com/rest/api/2/search?jql=labels%20in%20(trt-incident,trt-resolved)", nil)
```
neisw, line 42: I know we had discussed trt-resolved but I believe we decided to just clone incidents and keep one for our board, closing when done and the other wouldn't have trt-incident on it.  Just mentioning in case we don't want to include trt-resolved at this time.

pkg/api/jira.go
```diff
@@ -0,0 +1,20 @@
...
+func GetJIRAIncidentsFromDB(dbClient *db.DB) ([]apitype.CalendarEvent, error) {
+	// Get JIRA Incidents for display in calendar
+	var incidents []apitype.CalendarEvent
+	res := dbClient.DB.Table("jira_incidents").Select(`
+		start_time AS start,
+		resolution_time AS end,
+		key as jira,
+		key || ': ' || summary AS title,
+		'incident' AS phase,
+		'TRUE' as all_day`).Scan(&incidents)
```
neisw, line 17: Any concerns about this growing over time?  I guess (hope) it would take some time for the incidents to get so big we wouldn't want them all in one query.  Didn't know if we could get the year or something from the calendar view.

stbenjam: Makes sense, fixed.  The calendar sends the start/end ranges anyway for what the user is looking at.

pkg/jiraloader/jiraloader.go
stbenjam: Fixed

PR #1098, opened 2023-07-12
Embed the URL for the API call for convenient access for debugging

DennisPeriquet:
[TRT-1144](https://issues.redhat.com//browse/TRT-1144)

If there's a problem in ComponentReadiness, we often need the URL for the api call to verify the call and resulting data.  We can set `debugMode` to display these on the page while developing; but for those who just need the URL, this would entail some setup time.  This will allow you to retrieve the URL by way of the javascript console which is more convenient.

I added an `id="callStr"` to facility easy search manually or programmatically.


stbenjam:
I don't really have an objection to an extra div, but showing the API call this way is not really something most apps do. I wish we didn't show this at all, average users don't need it.  And someone developing sippy should have browser tools open.

You can view all the network requests for a page

image

And copy the API URL in various formats, including directly as an executable curl command

image

DennisPeriquet:
The debug tools are the way to go.

PR #1117, opened 2023-07-26
DO NOT MERGE: Always exclude alibaba but avoid displaying it as an option

DennisPeriquet:
[TRT-1151](https://issues.redhat.com//browse/TRT-1151)

This is one way to avoid having alibaba the ComponentReadiness UI -- we exclude it by default without giving the option to not exclude it.

You can probably also do this in the backend.

Seeking opinions.


stbenjam:
I think it'd be easier to maintain a list of excluded things in the backend

DennisPeriquet:
Sounds good to me -- thanks.

PR #1118, opened 2023-07-26
Add architecture filter to rejected-payloads.py script

stbenjam:


scripts/rejected-payloads.py
```diff
@@ -140,13 +145,15 @@ def verifyArgs(args):
     list_parser.set_defaults(action='list')
     list_parser.add_argument('-r', '--release', help='Specifies a release, like 4.11', default=None)
     list_parser.add_argument('-s', '--stream', help='Specifies a stream, like nightly or ci', default=None)
+    list_parser.add_argument('-e', '--architecture', help='Specifies an architecture, like amd64', default='amd64')
```
dgoodwin, line 148: WDYT about --arch over --architecture, and could we steal -a here given this is largely unused now, just leaving --all below?

PR #1126, opened 2023-08-01
[wip] Derive releases from config, not command line

stbenjam:
This makes OKD releases start showing up in Sippy.

The Sippy config already contains the release/job mapping, use it and not the command line flags. We should eventually remove the command line flags for `--release` and just read it from the config, but it's a little messy until we rip out TestGrid legacy code.


stbenjam:
TRT-1125

PR #1129, opened 2023-08-02
AGENT-688: Add agent to the sippy variant

pawanpinjarkar:


pkg/testidentification/ocp_variants_test.go
```diff
@@ -85,6 +85,10 @@ func Test_openshiftVariants_IdentifyVariants(t *testing.T) {
 			clusterData: models.ClusterData{Release: "4.13", Network: "sdn", Platform: "azure", Architecture: "arm64", Topology: "single"},
 			want:        []string{"azure", "arm64", "sdn", "ha"},
 		},
+		{
+			name:        "periodic-ci-openshift-release-master-nightly-4.14-e2e-agent-ha-dualstack",
```
andfasano, line 89: Probably the compact would be a better choice for that?

pawanpinjarkar: This test mainly just checks the variant name `agent` so I think this is enough, others are default.

andfasano: Isn't fetching the results from that kind of job?

dgoodwin: No it's just a small unit test to make sure we saw "agent" in the job name, no bearing on what we fetch.

dgoodwin:
I don't think this works until https://github.com/openshift/release/blob/master/ci-operator/step-registry/agent/test/agent-test-chain.yaml is resolved, you're not running the tests that are used to build the table for install and upgrade success by operator.

pkg/testidentification/ocp_variants_test.go
pawanpinjarkar: @andfasano ok, so I see there is a bug in the existing code which encounters when we use the job name  `periodic-ci-openshift-release-master-nightly-4.14-e2e-agent-compact-ipv4` in the test case which results adding the topology twice in the variant array.

dgoodwin: Lets keep the release check too.

pkg/testidentification/ocp_variants.go
```diff
@@ -357,3 +364,17 @@ func (openshiftVariants) IsJobNeverStable(jobName string) bool {
...
+
+	for _, val := range variants {
+		if _, ok := uniqueMap[val]; !ok {
+			uniqueMap[val] = true
+			uniqueArr = append(uniqueArr, val)
+		}
+	}
+
+	return uniqueArr
+}
```
stbenjam, line 380: I'd suggest using a Set implementation instead, https://pkg.go.dev/k8s.io/apimachinery/pkg/util/sets

pkg/testidentification/ocp_variants_test.go
```diff
@@ -17,78 +17,78 @@ func Test_openshiftVariants_IdentifyVariants(t *testing.T) {
 		{
 			name:    "periodic-ci-openshift-hypershift-main-periodics-conformance-aws-ovn-4-12",
 			release: "4.12",
-			want:    []string{"aws", "amd64", "ovn", "ha", "hypershift"},
+			want:    []string{"amd64", "aws", "ha", "hypershift", "ovn"},
```
stbenjam, line 20: I think we have testify vendored here, https://pkg.go.dev/github.com/stretchr/testify/assert#ElementsMatch instead of reflection probably works more reliably.  The new version of Sets from apimachinery doesn't guarantee order anymore IIRC

stbenjam: Thanks for fixing all this btw

pawanpinjarkar: Thanks for the review. I am keeping this on hold as we have to get a confirmation about adding `agent` as a new variant. This is because there are some tests that we are not running in our workflow in CI. Once that is resolved, I will address the suggestions on this PR.

pawanpinjarkar:
This depends on https://github.com/openshift/release/pull/43917
/hold

PR #1130, opened 2023-08-02
TRT-839: Store cluster-data in sippy db

deepsm007:


neisw:
The story mentions   _It would be nice if additional data could be added to the job run in columns, like architecture, network stack, cloud region_ 

Original story discusses storing a json blob,  let's discuss our long term plans and usage for this data.

Also we might want to revisit TRT-880 to validate what data we are collecting in that file.

PR #1144, opened 2023-08-11
Add a new view for each test that failed in all jobs on a payload

dgoodwin:
New Tab on the page to view a particular payload, that lists all tests that failed across all jobs in the payload, how many times, and which jobs it failed in. 

Helpful just to quickly show the tests that may be failing in multiple
jobs.

### DIFF
--- a/pkg/sippyserver/server.go
+++ b/pkg/sippyserver/server.go
@@ -423,6 +423,35 @@ func (s *Server) jsonGetPayloadAnalysis(w http.ResponseWriter, req *http.Request
 	api.RespondWithJSON(http.StatusOK, w, result)
 }
 
+// jsonGetPayloadTestFailures is an api to fetch information about what tests failed across all jobs in a specific
+// payload.
+func (s *Server) jsonGetPayloadTestFailures(w http.ResponseWriter, req *http.Request) {
+	log.Info("hello")
+	payload := req.URL.Query().Get("payload")
+	if payload == "" {
+		api.RespondWithJSON(http.StatusBadRequest, w, map[string]interface{}{
+			"code":    http.StatusBadRequest,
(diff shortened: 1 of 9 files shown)


pkg/sippyserver/server.go
```diff
@@ -423,6 +423,35 @@ func (s *Server) jsonGetPayloadAnalysis(w http.ResponseWriter, req *http.Request
 	api.RespondWithJSON(http.StatusOK, w, result)
 }
 
+// jsonGetPayloadTestFailures is an api to fetch information about what tests failed across all jobs in a specific
+// payload.
+func (s *Server) jsonGetPayloadTestFailures(w http.ResponseWriter, req *http.Request) {
+	log.Info("hello")
```
DennisPeriquet, line 429: nit: not sure you need it -- but harmless.

sippy-ng/src/releases/PayloadTestFailures.js
```diff
@@ -0,0 +1,281 @@
...
+
+  if (!isLoaded) {
+    return <p>Loading...</p>
+  }
+
+  if (rows.length === 0) {
+    return <Typography variant="h5">Analysis not found.</Typography>
+  }
+
+  console.log('hello world')
```
DennisPeriquet, line 190: nit: not sure you need it -- but harmless.

DennisPeriquet (approved):
/lgtm
You can remove the console logs now or later

DennisPeriquet:
and I look forward to use this!

sippy-ng/src/releases/PayloadTestFailures.js
dgoodwin: My bad, thanks!

https://github.com/openshift/sippy/pull/1148

PR #1160, opened 2023-08-23
Ensure stale issues get updated for incidents

stbenjam:
[TRT-1194](https://issues.redhat.com//browse/TRT-1194)

If we removed the incident label from a jira, we want to make sure we update the database row still. So, we keep track of jira incidents we didn't see yet and then query jira to figure out what happened. If the jira got the trt-incident label removed, we found out when that happened and set it as the incident resolution date.

I tested this change by:
- Syncing incidents to a local DB
- Removing the incident label from [TRT-1202](https://issues.redhat.com//browse/TRT-1202)
- Resyncing, and confirm the resolution date was set to when I removed the label
- Added label back, resynced, and confirmed issue was marked unresolved again

### DIFF
--- a/pkg/jiraloader/jiraloader.go
+++ b/pkg/jiraloader/jiraloader.go
@@ -13,6 +13,7 @@ import (
 	v1jira "github.com/openshift/sippy/pkg/apis/jira/v1"
 	"github.com/openshift/sippy/pkg/db"
 	"github.com/openshift/sippy/pkg/db/models"
+	"github.com/openshift/sippy/pkg/util/sets"
 )
 
 type JIRALoader struct {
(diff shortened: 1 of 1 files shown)


pkg/jiraloader/jiraloader.go
```diff
@@ -29,6 +30,16 @@ const jiraTimeLayout = "2006-01-02T15:04:05.000Z0700"
 
 func (jl *JIRALoader) LoadJIRAIncidents() error {
 	start := time.Now()
+	log.Infof("populating jira incident cache...")
+	var dbIssues []string
+	jl.dbc.DB.Table("jira_incidents").Pluck("key", &dbIssues)
```
neisw, line 35: If I understand it correctly this will get all of the keys from the table each time, I guess when I was thinking about it I was thinking we would only want entries without resolution_time.  Just wondering about the 5 years from now case or if we move to another system and don't have historical context, etc.

pkg/jiraloader/jiraloader.go
```diff
@@ -69,32 +80,32 @@ func (jl *JIRALoader) LoadJIRAIncidents() error {
...
+		if err != nil {
+			log.WithError(err).Errorf("couldn't query details for %+v", issue)
+			continue
 		}
 
+		model, err := issueToDB(issue)
+		if err != nil {
+			log.WithError(err).Errorf("couldn't convert jira issue to db model")
+			continue
+		}
```
neisw, line 108: I guess we do a query for all issues with label trt-incident over all time and then save them all, maybe something in the future for us to consider only those modified in the last 30 days or something.  Doesn't have to be this PR just walking through it

pkg/jiraloader/jiraloader.go
stbenjam: 5 years from now even if you had 5 incidents a day, 7 days a week you'd have under 10K rows.  Even at 100,000 incidents this cache would be at most 4MB, it's not really worth optimizing.

stbenjam: For resolution_time, what if you unresolve an incident which I think we've done in the past?

pkg/jiraloader/jiraloader.go
stbenjam: Yea I can limit the jira query, will update

stbenjam:
@neisw I think I've addressed your comments, have a look

PR #1165, opened 2023-08-25
TRT-1122: Get sippy ready for 4.15

deepsm007:
Adding 4.15 release to sippy and start collecting data from testgrid
/cc @stbenjam


stbenjam:
Note: we don't collect anything from testgrid anymore. It's fine to update fetch-testgrid.sh until we remove it though.

stbenjam:
Can you also update sippy-ng/src/component_readiness/ComponentReadiness.js to replace 4.14 with 4.15 and 4.13 with 4.14?

PR #1175, opened 2023-09-06
Allow passing in a jobName for sippy intervals

DennisPeriquet:
[TRT-1215](https://issues.redhat.com//browse/TRT-1215)

Today, when you click on the Intervals button on a prow job under the Debug section, only the jobID is passed.  From the jobID, we can get the GCS bucket url via the sippy database, but only after that prow job is added to the sippyDB (which happens hourly).

This change allows us to make use of a prow job name (if passed in as a parameter).  This obviates the need to wait for the sippyDB to be populated with the prow job; as such, we can get the intervals immediately.

https://github.com/openshift/release/pull/43064 will make use of this functionality.

Test:

- [x] Test links for new rehearse and presubmit jobs (both of these are of type "presubmit") (i.e., where jobName, repoInfo and pullNumber is passed)
- [x] Test links for new periodic jobs (i.e., where jobName is passed)
- [x] Ensure backward compatibility for rehearse, presubmit, and periodic jobs (i.e., when just jobID is passed)
- [x] Ensure `curl 'http://localhost:8080/api/jobs/runs/intervals?prow_job_run_id=1697965504431394816&prow_job_name=periodic-ci-openshift-release-master-ci-4.14-e2e-azure-ovn-upgrade'` gets data
- [x] Ensure `http://localhost:3000/sippy-ng/job_runs/1697965504431394816/periodic-ci-openshift-release-master-ci-4.14-e2e-azure-ovn-upgrade/intervals?categories=operator_unavailable&categories=operator_progressing&categories=operator_degraded&categories=pod_logs&categories=interesting_events&categories=alerts&categories=node_state&categories=e2e_test_failed&categories=disruption&filterText=&intervalFiles=e2e-events_20230902-143224.json` results in rendered intervals.

### DIFF
--- a/pkg/api/jobrunintervals/job_run_intervals.go
+++ b/pkg/api/jobrunintervals/job_run_intervals.go
@@ -14,26 +14,35 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
-// JobRunIntervals fetches intervals for a given job run.
-func JobRunIntervals(gcsClient *storage.Client, dbc *db.DB, jobRunID int64, logger *log.Entry) (*apitype.EventIntervalList, error) {
+// JobRunIntervals fetches intervals for a given job run by fetching from the prow job's GCS bucket path
+// contructed using the prow job name and jobID using one of these methods:
+// 1) using a GCS path that was calculated and passed in (we can retrieve intervals immediately)
+// 2) looking up the url given the jobRunID and extracting the prow job name (we need to wait until the sippyDB is populated)
+// If the GCS path could not be calculated, it will be empty.
+func JobRunIntervals(gcsClient *storage.Client, dbc *db.DB, jobRunID int64, gcsPath string, logger *log.Entry) (*apitype.EventIntervalList, error) {
 
-	// Right now, we need the job run in our DB to fetch its URL, so we can find the GCS
-	// bucket. This means until sippy imports the job, you will not be able to fetch
-	// its intervals.
-	// However, long term we expect these to live in an external system, and the prow job ID
-	// should be all we need to look it up. At that point the limitation should be removed.
-	jobRun, _, err := api.FetchJobRun(dbc, jobRunID, logger)
-	if err != nil {
-		logger.WithError(err).Error("error querying job run")
-		return nil, err
-	}
-
-	parts := strings.Split(jobRun.URL, gcs.OpenshiftGCSBucket)
-	path := parts[1][1:]
-	log.WithField("path", path).Debug("calculated gcs path")
 	bkt := gcsClient.Bucket(gcs.OpenshiftGCSBucket)
-	gcsJobRun := gcs.NewGCSJobRun(bkt, path)
 
+	var gcsJobRun *gcs.GCSJobRun
+
+	if len(gcsPath) > 0 {
+		log.WithField("gcsPath", gcsPath).Debug("calculated gcs path from job attributes")
+		gcsJobRun = gcs.NewGCSJobRun(bkt, gcsPath)
+	} else {
+		// Fall back to looking up the job run ID in the DB and extracting the URL that way.
+		// This is here to support older prow jobs where only the jobID was passed.  Eventually,
+		// we will not have to fallback because we will expect all jobs to pass in enough
+		// information to construct a full GCS bucket path.
+		jobRun, _, err := api.FetchJobRun(dbc, jobRunID, logger)
+		if err != nil {
+			logger.WithError(err).Error("error querying job run")
+			return nil, err
+		}
+		parts := strings.Split(jobRun.URL, gcs.OpenshiftGCSBucket)
+		path := parts[1][1:]
+		log.WithField("path", path).Debug("calculated gcs path")
+		gcsJobRun = gcs.NewGCSJobRun(bkt, path)
+	}
 	intervalFiles := gcsJobRun.FindAllMatches([]*regexp.Regexp{gcs.GetIntervalFile()})
 
 	// We will often match multiple files here, one for upgrade phase, one for conformance
--- a/sippy-ng/src/App.js
+++ b/sippy-ng/src/App.js
@@ -460,9 +460,14 @@ export default function App(props) {
                     />
 
                     <Route
-                      path="/job_runs/:jobrunid/intervals"
+                      path="/job_runs/:jobrunid/:jobname?/:repoinfo?/:pullnumber?/intervals"
                       render={(props) => (
-                        <ProwJobRun jobRunID={props.match.params.jobrunid} />
+                        <ProwJobRun
+                          jobRunID={props.match.params.jobrunid}
+                          jobName={props.match.params.jobname}
+                          repoInfo={props.match.params.repoinfo}
+                          pullNumber={props.match.params.pullnumber}
+                        />
                       )}
                     />
 
--- a/sippy-ng/src/prow_job_runs/ProwJobRun.js
+++ b/sippy-ng/src/prow_job_runs/ProwJobRun.js
@@ -60,12 +60,24 @@ export default function ProwJobRun(props) {
 
   const fetchData = () => {
     let queryString = ''
-    console.log('hello world we got the prow job run id of ' + props.jobRunID)
+    console.log(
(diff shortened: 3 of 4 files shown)


pkg/api/jobrunintervals/job_run_intervals.go
```diff
@@ -15,26 +15,46 @@ import (
 )
 
 // JobRunIntervals fetches intervals for a given job run.
-func JobRunIntervals(gcsClient *storage.Client, dbc *db.DB, jobRunID int64, logger *log.Entry) (*apitype.EventIntervalList, error) {
+// It fetches from the prow job's GCS bucket path contructed using the prow job name and jobID
```
dgoodwin, line 18: Godoc standard requires the function name to be the first word in the comment, i.e. this should remain "JobRunIntervals fetches", not "It fetches".

pkg/api/jobrunintervals/job_run_intervals.go
```diff
@@ -15,26 +15,46 @@ import (
...
+// using one of these methods:
+// 1) looking up the url given the jobRunID and extracting the prow job name (we need to wait until the sippyDB is populated)
+// 2) using a GCS path that was calculated and passed in (we can retrieve intervals immediately)
```
dgoodwin, line 21: Flip the ordering here, 2) is the one we try first.

pkg/api/jobrunintervals/job_run_intervals.go
```diff
@@ -15,26 +15,46 @@ import (
...
-	parts := strings.Split(jobRun.URL, gcs.OpenshiftGCSBucket)
-	path := parts[1][1:]
-	log.WithField("path", path).Debug("calculated gcs path")
-	bkt := gcsClient.Bucket(gcs.OpenshiftGCSBucket)
-	gcsJobRun := gcs.NewGCSJobRun(bkt, path)
+	// If there we couldn't calculate the gcsPath, then use the database if we can.
+	if len(gcsPath) == 0 ||
+		// If there is no match, there will still be one empty item.
+		// This means either the GCS bucket path is wrong or there are no interval files.
+		(len(intervalFiles) == 1 && len(intervalFiles[0]) == 0) {
```
dgoodwin, line 38: I don't think we should lump the "no gcsPath provided" case together with gcsPath provided but wrong and fall back to looking up from the db. If you gave us a path, we should try to use it and fail if it's wrong, not silently try to do something else.

Please refactor to an if len(gcsPath) > 0 and an else.

pkg/api/jobrunintervals/job_run_intervals.go
```diff
@@ -15,26 +15,46 @@ import (
...
 
-	intervalFiles := gcsJobRun.FindAllMatches([]*regexp.Regexp{gcs.GetIntervalFile()})
+		// Fall back to looking up the job run ID in the DB and extracting the URL that way.
```
dgoodwin, line 40: Remove the rest of the comment below this line. Maybe clarify that this is for legacy support and we expect callers should pass in job name.

sippy-ng/src/App.js
```diff
@@ -460,9 +460,14 @@ export default function App(props) {
                     />
 
                     <Route
-                      path="/job_runs/:jobrunid/intervals"
+                      path="/job_runs/:jobrunid/:jobname/:repoinfo/:pullnumber/intervals"
```
dgoodwin, line 463: This looks busted for non-presubmits where you only specify jobname? Does it still work if the others are missing? It makes me nervous either way unless the behavior being relied on is documented. Could we do query params here as well? Curious what @stbenjam would like, we're trying to keep backward compatibility, but also accept and use new info.

sippy-ng/src/prow_job_runs/ProwJobRun.js
```diff
@@ -60,12 +60,24 @@ export default function ProwJobRun(props) {
...
+    console.log(
+      'We got the prow job run id of ' + props.jobRunID,
+      props.jobName
+        ? `jobName of ${props.jobName}`
+        : '' + props.pullNumber
+        ? `pullNumber of ${props.pullNumber}`
+        : '' + props.repoInfo
+        ? `repoInfo of ${props.repoInfo}`
+        : ''
+    )
```
dgoodwin, line 72: Please find a  simpler way to log this, it's fine if we print "jobName=" if there was no jobName for example.

pkg/api/jobrunintervals/job_run_intervals.go
```diff
@@ -15,26 +15,46 @@ import (
...
-	if err != nil {
-		logger.WithError(err).Error("error querying job run")
-		return nil, err
+	var gcsJobRun *gcs.GCSJobRun
+	var intervalFiles [][]string
+	if len(gcsPath) > 0 {
+		log.WithField("gcsPath", gcsPath).Debug("calculated gcs path from job attributes")
+		bkt := gcsClient.Bucket(gcs.OpenshiftGCSBucket)
+		gcsJobRun = gcs.NewGCSJobRun(bkt, gcsPath)
+		intervalFiles = gcsJobRun.FindAllMatches([]*regexp.Regexp{gcs.GetIntervalFile()})
```
dgoodwin, line 31: You can break this line out of the if/else and put it at the end I think, and remove the var declaration above.

pkg/api/jobrunintervals/job_run_intervals.go
```diff
@@ -15,26 +15,46 @@ import (
...
-	// should be all we need to look it up. At that point the limitation should be removed.
-	jobRun, _, err := api.FetchJobRun(dbc, jobRunID, logger)
-	if err != nil {
-		logger.WithError(err).Error("error querying job run")
-		return nil, err
+	var gcsJobRun *gcs.GCSJobRun
+	var intervalFiles [][]string
+	if len(gcsPath) > 0 {
+		log.WithField("gcsPath", gcsPath).Debug("calculated gcs path from job attributes")
+		bkt := gcsClient.Bucket(gcs.OpenshiftGCSBucket)
```
dgoodwin, line 29: This line can be moved out of the if and shared.

dgoodwin (commented):
Good job handling the presubmits case, I did not see that coming. 

Excellent PR description too, rest of us need to do more of that.

sippy-ng/src/App.js
DennisPeriquet: Making them optional allows us to not pass them and still use the route.

sippy-ng/src/prow_job_runs/ProwJobRun.js
DennisPeriquet: The pretty printer insists on making it look ugly.

This:

```js
      props.jobName
        ? `jobName of ${props.jobName}`
        : '' + props.pullNumber
```

should be this:

```js
      props.jobName ? `jobName of ${props.jobName}` : '' + props.pullNumber
```

But, I like your suggestion to simplify with "jobName=" style.

sippy-ng/src/App.js
```diff
@@ -460,9 +460,14 @@ export default function App(props) {
                     />
 
                     <Route
-                      path="/job_runs/:jobrunid/intervals"
+                      path="/job_runs/:jobrunid/:jobname?/:repoinfo?/:pullnumber?/intervals"
```
dgoodwin, line 463: Cool!

sippy-ng/src/prow_job_runs/ProwJobRun.js
```diff
@@ -102,7 +114,16 @@ export default function ProwJobRun(props) {
...
+          'Could not retrieve intervals for ' +
+            'jobRunID=' +
+            props.jobRunID +
+            'jobName=' +
+            props.jobName +
+            'pullNumber=' +
+            props.pullNumber +
+            'repoInfo=' +
+            ', ' +
+            error
```
dgoodwin, line 126: i think you're lacking spaces between your params.

dgoodwin (commented):
Otherwise lgtm.

sippy-ng/src/prow_job_runs/ProwJobRun.js
DennisPeriquet: fixed and squashed -- thx

PR #1181, opened 2023-09-11
Narrow scope of Component Readiness defaults

DennisPeriquet:
[TRT-1230](https://issues.redhat.com//browse/TRT-1230)

This will focus the Component Readiness defaults (by excluding several items) so we can reduce the query search space and possibly make better use of the cache.

* Release to evaluate = 4.14 (from/to relative to now)
* Historical release = 4.13 (from/to set to 4 weeks before GA)


stbenjam:
Looks reasonable to me

/lgtm

PR #1191, opened 2023-09-13
Parallelize job gcs bucket processing

stbenjam:
[TRT-1190](https://issues.redhat.com//browse/TRT-1190)

We're spending a lot of time processing gcs buckets and importing jobs, such that when combined with the matview refreshes we're regularly approaching or exceeding the 60 minute window. This changes the gcs importer code to import up to 10 jobs in parallel.  Gorm is (mostly) thread safe, and I think I've protected all the places in our code that isn't.

### DIFF
--- a/pkg/prowloader/prow.go
+++ b/pkg/prowloader/prow.go
@@ -776,6 +776,7 @@ func (pl *ProwLoader) prowJobRunTestsFromGCS(ctx context.Context, pj *prow.ProwJ
 
 	syntheticSuite, jobResult := testconversion.ConvertProwJobRunToSyntheticTests(*pj, testCases, pl.syntheticTestManager)
 	pl.extractTestCases(syntheticSuite, testCases)
+	log.Infof("synthetic suite had %d tests", syntheticSuite.NumTests)
 
 	results := make([]*models.ProwJobRunTest, 0)
 	for k, v := range testCases {
--- a/pkg/prowloader/prow.go
+++ b/pkg/prowloader/prow.go
@@ -189,10 +189,10 @@ func (pl *ProwLoader) LoadProwJobsToDB() []error {
 				}
 				if err := pl.processProwJob(ctx, job); err != nil {
 					errsCh <- err
-				} else {
-					pl.jobsImportedCount.Add(1)
-					log.Infof("%d of %d job runs completed", pl.jobsImportedCount.Load(), total)
+					log.WithError(err).Warningf("couldn't import job %s/%s, continuing", job.Spec.Job, job.Status.BuildID)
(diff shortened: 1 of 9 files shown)


pkg/prowloader/prow.go
```diff
@@ -153,46 +168,103 @@ func (pl *ProwLoader) LoadProwJobsToDB() []error {
...
+	// Start pl.maxConcurrency consumers
+	var wg sync.WaitGroup
+	for i := 0; i < pl.maxConcurrency; i++ {
+		wg.Add(1)
+		go func(ctx context.Context) {
+			defer wg.Done()
+			for job := range queue {
+				if ctx.Err() != nil {
+					errsCh <- ctx.Err()
+					fmt.Println("consumer exiting, got error")
```
dgoodwin, line 187: Lets flip to logrus logging WithError

pkg/prowloader/prow.go
```diff
@@ -484,15 +561,24 @@ func (pl *ProwLoader) prowJobToJobRun(pj prow.ProwJob, release string, newJobRun
 			}
 		}
 		if saveDB {
-			pl.dbc.DB.Save(&dbProwJob)
+			fmt.Println("update")
```
dgoodwin, line 564: logrus or remove

pkg/prowloader/prow.go
```diff
@@ -623,30 +711,44 @@ func (pl *ProwLoader) findOrAddPullRequests(refs *prow.Refs, pjPath string) []mo
...
 
+	pl.suiteCacheLock.RLock()
 	if id, ok := pl.suiteCache[name]; ok {
+		pl.suiteCacheLock.RUnlock()
 		return id
 	}
+	pl.suiteCacheLock.RUnlock()
 
+	pl.suiteCacheLock.Lock()
+	defer pl.suiteCacheLock.Unlock()
```
dgoodwin, line 751: Looks like an unnecessary lock/unlock here.

dgoodwin (commented):
It looks ok to me, just a couple nits. I'm going to hold just for a bit until we settle things down in the db world.

/hold

pkg/prowloader/prow.go
stbenjam: It's needed since we're manipulating the cache a little bit further down

stbenjam:
> It looks ok to me, just a couple nits. I'm going to hold just for a bit until we settle things down in the db world.

Yea sounds good, I also need to dig into why e2e is failing. Sync completes fine, but tests are failing

stbenjam:
/hold cancel

Working now

PR #1192, opened 2023-09-13
Add metric for prow job loading time

stbenjam:
Let's collect some data on how long it takes us to load our prow jobs. This will help us see the impact of https://github.com/openshift/sippy/pull/1191.


neisw:
/lgtm

/hold
just in case you want to be around when it merges. unhold when ready

PR #1227, opened 2023-10-04
Refactor fetch data loaders into single interface

stbenjam:

We have a variety of data sources: jira, github, prow, etc. They all look slightly different. This refactors all of them to a standard loader interface. We also now gather metrics and push them to Prometheus for each stage of loading.


stbenjam:
/hold

Let my last PR bake for a while.

PR #1236, opened 2023-10-09
Sync ci-test-mapping from bigquery

stbenjam:

- Sync the list of OCPBUGS components and their ID's to Sippy (from Jira)
- Sync the ci-test-mapping test ownerships information from bigquery
- Wire up that data together
- Adds the jira component information to the test API, and adds file-a-jira links to the test analysis page

Getting the file-a-bug link into component readiness is more complicated, will tackle separately.


stbenjam:
> /hold
> 
> I need to link suite id

https://github.com/openshift/sippy/pull/1255 needed first

pkg/db/tableswapper/tableswapper.go
```diff
@@ -0,0 +1,63 @@
...
+	// Swap tables and drop the old one
+	return ts.dbc.DB.Transaction(func(tx *gorm.DB) error {
+		if err := tx.Exec(fmt.Sprintf("ALTER TABLE %s RENAME TO %s", ts.tableName, ts.oldTable)).Error; err != nil {
+			return err
+		}
+		if err := tx.Exec(fmt.Sprintf("ALTER TABLE %s RENAME TO %s", ts.stagingTable, ts.tableName)).Error; err != nil {
+			return err
+		}
+		return tx.Exec(fmt.Sprintf("DROP TABLE %s", ts.oldTable)).Error
+	})
```
dgoodwin, line 54: How does this play out with indicies and cascades in play? Any risk of someone doing the wrong thing there and wiping out data we didn't intend?

pkg/db/tableswapper/tableswapper.go
```diff
@@ -0,0 +1,63 @@
...
+}
+
+// Cleanup removes any leftover tables.  Good idea to defer this after calling CreateStagingTable.
+func (ts *TableSwapper) Cleanup() error {
+	if err := ts.dbc.DB.Exec(fmt.Sprintf("DROP TABLE IF EXISTS %s", ts.stagingTable)).Error; err != nil {
```
dgoodwin, line 59: Can't recall if these Exec statements log nicely still, I recall shutting something off, but if not we'd want logging on all major operations in this module. Disregard if this already appears in the log cleanly.

pkg/dataloader/componentloader/componentloader.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+
+	// Link up the ci-test-mapping records to Sippy's test_ids
+	for _, m := range mappings {
+		// Find the test or suite.test in Sippy DB:
+		var test models.Test
+		res := cl.db.DB.Table("tests").First(&test, "name = ?", m.Name)
+		if res.Error == gorm.ErrRecordNotFound {
+			// Search for test with suite name, because of how sippy handles unknown suites
+			res = cl.db.DB.Table("tests").First(&test, "name = ?", fmt.Sprintf("%s.%s", m.Suite, m.Name))
```
dgoodwin, line 71: This will go after today I assume.

pkg/db/tableswapper/tableswapper.go
stbenjam: Oh that's a REALLY good question.  I'll have a think about this, it was a shortcut to easily mirror a table from bigquery.  It might be better to do it the hard way.

stbenjam: Ok, I redid this for both ways I sync data.  I upsert this data into the main table and insert it into a staging table, then any records in the main table not present in the staging table are deleted to keep both sources in sync. Then the staging table gets deleted.

pkg/dataloader/jiraloader/jiraloader.go
```diff
@@ -0,0 +1,354 @@
...
+		jl.errors = append(jl.errors, oldRecords.Error)
+	}
+
+	log.WithFields(log.Fields{
+		"component_count": len(components),
+		"obsolete":        oldRecords.RowsAffected,
+	}).Infof("component load complete in %+v", time.Since(start))
+}
+
+func (jl *JiraLoader) incidentLoader() {
```
stbenjam, line 199: This incident loading code is unchanged, just moved to it's own function

pkg/dataloader/jiraloader/jiraloader.go
```diff
@@ -0,0 +1,354 @@
...
+
+	// Load incidents
+	jl.incidentLoader()
+}
+
+func (jl *JiraLoader) Errors() []error {
+	return jl.errors
+}
+
+func findResolutionTime(issue *v1jira.Issue) *time.Time {
```
stbenjam, line 53: This one is also a straight move

pkg/dataloader/jiraloader/jiraloader.go
```diff
@@ -0,0 +1,354 @@
...
+			jl.errors = append(jl.errors, res.Error)
+			log.WithError(err).Warningf("failed to save component %q", c.Name)
+			return
+		}
+
+		fmt.Println(mc.ID)
+	}
+
+	log.Infof("deleting old records...")
+	oldRecords := jl.dbc.DB.Exec(fmt.Sprintf(`DELETE FROM jira_components WHERE id NOT IN (SELECT id FROM %s)`, tmpTable))
```
dgoodwin, line 187: How many records in tmpTable today? This will break at a certain point with postgres but hopefully we're far away. 64k items IIRC?

pkg/dataloader/jiraloader/jiraloader.go
```diff
@@ -0,0 +1,354 @@
...
+
+	var components []v1jira.Component
+	err = json.Unmarshal(body, &components)
+	if err != nil {
+		jl.errors = append(jl.errors, err)
+		return
+	}
+
+	tblUpdater := stagingtablemanager.New(jl.dbc, "jira_components")
+	tmpTable, err := tblUpdater.CreateStagingTable()
```
dgoodwin, line 137: Should you do a Cleanup prior to each Create? (in case process got killed or something)

pkg/dataloader/jiraloader/jiraloader.go
dgoodwin: What are the advantages to the staging table vs maintaining a list of what you expect in memory and using it in the query?

stbenjam: > How many records in tmpTable today? This will break at a certain point with postgres but hopefully we're far away.

What would break with this particular query? This is designed to not give a list of thousands of ID's directly.

> What are the advantages to the staging table vs maintaining a list of what you expect in memory and using it in the query?

Yea it's around 64K items.  That itself isn't a big deal, I couldn't find good information about what the maximum query size was since it's dependent on a bunch of factors.  I was just thinking in general about our future needs to sync data from bigquery.  Most of the docs online recommend using temporary tables like this and doing upserts for large syncs.

I've gone back and forth on this a bunch.  Maybe it's worth worrying about when we have more data to sync, I pushed up the id-based version if you prefer that for now

pkg/dataloader/jiraloader/jiraloader.go
stbenjam: The tables get a timestamp appended,  so another downside to this approach is potentially need to clean-up that if it happens

pkg/dataloader/jiraloader/jiraloader.go
```diff
@@ -0,0 +1,331 @@
...
+		}
+		ids = append(ids, mc.ID)
+	}
+
+	log.Infof("deleting old records...")
+	oldRecords := jl.dbc.DB.Where("id NOT IN ?", ids).Unscoped().Delete(&models.JiraComponent{})
+	if oldRecords.Error != nil {
+		log.WithError(oldRecords.Error).Warningf("couldn't delete old records")
+		jl.errors = append(jl.errors, oldRecords.Error)
+	}
```
dgoodwin, line 168: Totally wasn't necessary to switch to another variant I just wanted to learn the benefits as it's not an approach I've used before. I'm ok with either you want to proceed with.

dgoodwin:
/lgtm
/hold

Held so you can decide If you want to proceed with the temp table, that's fine with me just let me know and I'll re-lgtm.

pkg/dataloader/jiraloader/jiraloader.go
stbenjam: No this is fine, I've gone back and forth on this a bunch and AFAIK we're not close to any limits for the tables we're doing now.

PR #1255, opened 2023-10-16
Change handling of suites to stop prepending some test suite names to tests

stbenjam:

This fixes Sippy's handling of suites.  It makes it more consistent with how we're doing things in bigquery and component readiness, meaning we'll be able to bigquery consume data more easily: for example, we'll be able to sync from the junit table instead of directly from prow, or use the component mapping table in Sippy.

This PR:
 - limits the suites we import to the ones in our DB (this list is in sippy's code, but will probably eventually move to bigquery)
 - stop prepending suite names to the front of tests
 - has a script to remove the prefixes and update existing records

Merging suggestion:
- disable fetchdata and db backups, scale down sippy temporarily
- merge this PR
- run `go run ./scripts/updatesuites.go` (takes about ~4 hours)
- do a vacuum full (~2.5 hours)
- enable fetchdata and scale sippy back up

### DIFF
--- a/pkg/db/migrations.go
+++ b/pkg/db/migrations.go
@@ -1,0 +1,79 @@
+package db
+
+import (
+	"fmt"
+	"strings"
+
+	log "github.com/sirupsen/logrus"
+	"gorm.io/gorm"
+
+	"github.com/openshift/sippy/pkg/db/models"
+)
+
+func testNameWithoutSuite(dbc *DB) error {
+	// Get list of suites
+	var knownSuites []models.Suite
+	if res := dbc.DB.Model(&models.Suite{}).Find(&knownSuites); res.Error != nil {
+		return res.Error
+	}
+	for _, suite := range knownSuites {
+		log.Infof("processing suite %q", suite.Name)
+		suiteID := suite.ID
+
+		var testsWithPrefix []models.Test
+		if err := dbc.DB.Table("tests").
+			Where(fmt.Sprintf("name like '%s.%%'", suite.Name)).
+			Scan(&testsWithPrefix).Error; err != nil {
+			return err
+		}
+
+		for _, oldTest := range testsWithPrefix {
+			log.Infof("processing test %q", oldTest.Name)
+			var newTest models.Test
+			newTestName := strings.TrimPrefix(oldTest.Name, fmt.Sprintf("%s.", suite.Name))
+
+			if err := dbc.DB.Where("name = ?", newTestName).First(&newTest).Error; err != nil {
+				log.Infof("no existing test found, renaming and adding suite to prow job run tests...")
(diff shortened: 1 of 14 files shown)


pkg/db/migrations.go
```diff
@@ -0,0 +1,90 @@
...
+	"gorm.io/gorm"
+
+	"github.com/openshift/sippy/pkg/db/models"
+)
+
+// testNameWithoutSuite removes test suite prefixes from tests in the database
+// and assigns the suite to all the prow_job_run_tests. When the test can't be
+// renamed because the unprefixed version exists in the DB, use that one and remove
+// the prefixed version.
+func testNameWithoutSuite(dbc *DB) error {
```
stbenjam, line 17: Maybe this should just be a one-time script we run against the production DB?

stbenjam:
/hold

Not ready to merge yet

scripts/updatesuites.go
```diff
@@ -0,0 +1,144 @@
...
+					return err
+				}
+			} else { //nolint
+				log.Infof("existing test found, making it the default and removing the old one...")
+				err := dbc.Transaction(func(tx *gorm.DB) error {
+					// Update rows in the prow_job_run_tests table and then delete the old oldTest row.
+					if err := tx.Model(&models.ProwJobRunTest{}).Where("test_id = ?", oldTest.ID).Updates(models.ProwJobRunTest{TestID: newTest.ID, SuiteID: &suiteID}).Error; err != nil {
+						log.WithError(err).Warningf("error updating prow_job_run_tests for oldTest ID %d", oldTest.ID)
+						return err
+					}
```
dgoodwin, line 124: Can you log the number of rows affected easily here?

dgoodwin (commented):
Really nice work on a very delicate operation. Could you adjust the logging a little per this review and then post a log of migrating the dev backup? Would love to eye ball what it's going to do.

scripts/updatesuites.go
stbenjam: Yup, added.  14,838,319 was what I got for the snapshot restore

pkg/db/suites.go
```diff
@@ -1,27 +1,52 @@
...
+	"Operator results",
+	"Symptom detection",
+	"Tests Suite",
+	"cluster install",
+	"cluster nodes ready",
+	"cluster nodes",
+	"gather core dump",
+	"hypershift-e2e",
+	"metal infra",
+	"step graph",
```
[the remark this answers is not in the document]
stbenjam: I added a note to the development.md doc but I think this is transitional anyway.  I want to move this to bigquery and sync it down to sippy so we have a canonical list of suites TRT cares about

PR #1267, opened 2023-10-23
Remove overlapping Error material-ui import

DennisPeriquet:
[TRT-1294](https://issues.redhat.com//browse/TRT-1294)

Only the change in `sippy-ng/src/tests/TestDurationChart.js` fixes the problem mentioned in the Jira.  But, the same problem exists in the other 5 files.

The `import { Error } from '@material-ui/icons'` line overlaps with the `Error` built-in which we use for `throw/catch` so we can propagate error strings to users during error conditions.  This overlap causes React to render the `bundle.js` in a way that we didn't actually intend and hence causes mysterious errors like `TypeError: cx is not a constructor` when the code handles an error condition.  Removal of that import allows React to render the bundle.js as we intended.

I was able to reproduce this by returning a 500 status in a local version of sippy.

TODO:

- [ ] Test the other 5 files with their error conditions.

### DIFF
--- a/sippy-ng/src/prow_job_runs/ProwJobRun.js
+++ b/sippy-ng/src/prow_job_runs/ProwJobRun.js
@@ -6,7 +6,6 @@ import {
   useQueryParam,
 } from 'use-query-params'
 import { Button, ButtonGroup, TextField } from '@material-ui/core'
-import { Error } from '@material-ui/icons'
 import { stringify } from 'query-string'
 import { useHistory } from 'react-router-dom'
 import Alert from '@material-ui/lab/Alert'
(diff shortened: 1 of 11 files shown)


stbenjam:
Looks like a few more places need fixing, because they use both the Error icon and call "throw New Error"

```
$ find . -type f -exec grep -l '<Error' {} \; | while read -r file; do
    grep -q 'throw new Error' "$file" && echo "$file"
done
./pull_requests/PullRequestsTable.js
./releases/PayloadStreamsTable.js
./releases/ReleasePayloadAcceptance.js
./releases/ReleasePayloadTable.js
./tests/TestTable.js
```

The solution there is rename the Error icon import, `import { Error as ErrorIcon } from '@material-ui/icons';` should work

stbenjam:
It's grumpy about the js imports formatting.  `npx eslint --fix .` should fix it.

DennisPeriquet:
thanks for the tip.

PR #1273, opened 2023-10-25
WIP: introduced a shadowing condition for linter

DennisPeriquet:
[TRT-1343](https://issues.redhat.com//browse/TRT-1343)

Turn on variable shadowing detection using the [no-shadow ESlint setting](https://eslint.org/docs/latest/rules/no-shadow) to try to catch more shadowing problems during compile-time.

`Error`, `History`, `event`, `fetch`, `history`, `location`, `name`, `open`, `status`, `test`, show up as "already a global variable" so I renamed all usages of these to something else.

### DIFF
--- a/sippy-ng/src/App.js
+++ b/sippy-ng/src/App.js
@@ -137,30 +137,30 @@ export default function App(props) {
       fetch(process.env.REACT_APP_API_URL + '/api/capabilities'),
       fetch(process.env.REACT_APP_API_URL + '/api/report_date'),
     ])
-      .then(([releases, capabilities, reportDate]) => {
-        if (releases.status !== 200) {
-          throw new Error('server returned ' + releases.status)
+      .then(([fetchedReleases, fetchedCapabilities, fetchedReportDate]) => {
+        if (fetchedReleases.status !== 200) {
+          throw new Error('server returned ' + fetchedReleases.status)
         }
 
-        if (capabilities.status !== 200) {
-          throw new Error('server returned ' + capabilities.status)
+        if (fetchedCapabilities.status !== 200) {
+          throw new Error('server returned ' + fetchedCapabilities.status)
         }
 
-        if (reportDate.status !== 200) {
-          throw new Error('server returned ' + reportDate.status)
+        if (fetchedReportDate.status !== 200) {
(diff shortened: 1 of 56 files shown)


DennisPeriquet:
With `builtinGlobals: true` option added, now I see:

```
$ npx eslint ./src/prow_job_runs/ProwJobRun.js 

/Users/dperique/mygit/dperique/NG/Dennis/Redhat/PR/sippy/sippy-ng/src/prow_job_runs/ProwJobRun.js
    9:10  error  'Error' is already a global variable     <-- wanted to catch this one                                        no-shadow
   18:9   error  'history' is already a global variable                                           no-shadow
  167:9   error  'filteredIntervals' is already declared in the upper scope on line 23 column 10  no-shadow
  226:31  error  'event' is already a global variable                                             no-shadow
  926:13  error  'data' is already declared in the upper scope on line 873 column 9               no-shadow

✖ 5 problems (5 errors, 0 warnings)
```

sippy-ng/src/App.js
```diff
@@ -288,116 +288,122 @@ export default function App(props) {
 
                     <Route
                       path="/release/:release/tags/:tag"
-                      render={(props) => (
+                      render={(theProps) => (
```
stbenjam, line 291: To be honest I don't like this, js has scopes for a reason and shadowing variables outside your scope you're not going to use doesn't seem problematic to me.  Is it possible to just turn on the global shadowing? I think that's golang's default behavior and I find it not too intrusive.

DennisPeriquet: I hear ya.  I started out by doing a bunch of `/* eslint-{en|dis}able no-shadow */` around those and then maybe changing the name of `props` at the top (one line change) vs. `theProps` everywhere which I've never seen on any example code.  I didn't see much in the way of making this more forgiving in [the doc](https://eslint.org/docs/latest/rules/no-shadow#rule-details).  I also noticed that doc is for v8.52.0 and we're running v7.32.0 so some options don't work.  I might try upgrading eslint and see what happens.

I definitely would like to eliminate the changes regarding the local scope stuff esp. for things like in the `then` clauses in promises and functions with `event`.

DennisPeriquet:
closed in favor of https://github.com/openshift/sippy/pull/1293

PR #1279, opened 2023-10-27
Add link to file an issue about sippy to sidebar

stbenjam:
TRT-1295

And remove deprecated sidebar items


sippy-ng/src/components/Sidebar.js
```diff
@@ -62,18 +62,20 @@ export default function Sidebar(props) {
...
     setOpen((prevState) => ({ ...prevState, [id]: !prevState[id] }))
   }
 
+  function reportAnIssueURI() {
+    const description = `Describe your feature request or bug:\n\n
+    
+    Relevant Sippy URL:\n
+    ${window.location.href}\n\n`
+    return `https://issues.redhat.com/secure/CreateIssueDetails!init.jspa?pid=12323832&issuetype=17&description=${safeEncodeURIComponent(
+      description
```
DennisPeriquet, line 75: default priority is shown as `Blocker`.  I tried these and could not find `Normal`.  Maybe `Minor` is the way to go with `&priority=4`?

We can also say default priority is set to whatever we pick so they can choose it consciously.

```
  4 = minor
  3 = major
  2 = critical
  1 = blocker
  0 = invalid
```

stbenjam: image

It's 10200 lol

stbenjam:
@DennisPeriquet Thanks, nice find. Updated

PR #807, opened 2023-02-03
[WIP] Migrate Sippy to cobra subcommands

stbenjam:
The handling of options in main.go has gotten a little bit out of control. This moves everything to individualized sub-commands for separation of concerns, easier usability/discoverability, and better maintainability as Sippy gains new features.

Todo's:
- [ ] Update documentation
- [ ] Update deployment manifests
- [ ] Fix e2e test script
- [ ] Make sure kube/testgrid configurations are fully ported

Example output:

```
Sippy reports on job and test statistics, sliced by various filters
including name, suite, or NURP+ variants (network, upgrade, release,
platform, etc).

Usage:
  sippy [command]

Available Commands:
  completion  Generate the autocompletion script for the specified shell
  help        Help about any command
  load        Load data from various import sources
  migrate     Migrates the PostgreSQL database to the latest schema.
  refresh     Refresh data in database such as materialized views
  serve       Run the sippy server
  snapshot    Create snapshots using current sippy overview API json and store in the database

Flags:
  -h, --help               help for sippy
      --log-level string   Log level (trace,debug,info,warn,error) (default info) (default "info")

Use "sippy [command] --help" for more information about a command.
```

```
./sippy load --help                                                                                                                   ST 286   command-structure 
Load data from various import sources

Usage:
  sippy load [command]

Available Commands:
  bugs        Load bugs from search.ci.openshift.org
  prow        Load job runs and test data from prow
  releases    Load releases from the OpenShift release controllers

Flags:
  -h, --help   help for load

Global Flags:
      --log-level string   Log level (trace,debug,info,warn,error) (default info) (default "info")

Use "sippy load [command] --help" for more information about a command.
```

Example usages:
```
$ ./sippy migrate --database-dsn postgresql://postgres:password@localhost:5432/postgres
$ ./sippy load prow --config ./config/openshift.yaml --google-service-account-credential-file ~/.bigquery.json --load-github --mode ocp --releases 4.13
$ ./sippy --log-level trace snapshot --name asdf --release 4.13
$ ./sippy serve
```

PR #1288, opened 2023-10-31
[poc] [wip] Qe branch

deads2k:
playing with how to wire against QE tests.

Looks like the next thing is to create component_mapping and component_mapping_latest variants.

1. when nothing comes back from queries, it's hard to debug on the UI which just fail
2. links to jobs broken
3. not much history in the five weeks to GA.  Lots of different jobs.  What do they all mean?  Can we run more?  20 per week each in the lead up release and then 20 each week for ECs?

### DIFF
--- a/pkg/api/component_report.go
+++ b/pkg/api/component_report.go
@@ -231,10 +231,10 @@ func (c *componentReportGenerator) getJobRunTestStatusFromBigQuery() (
 	errs := []error{}
 	queryString := fmt.Sprintf(`WITH latest_component_mapping AS (
 						SELECT *
-						FROM ci_analysis_us.component_mapping cm
+						FROM ci_analysis_us.qe_component_mapping cm
 						WHERE created_at = (
 								SELECT MAX(created_at)
-								FROM openshift-gce-devel.ci_analysis_us.component_mapping))
+								FROM openshift-gce-devel.ci_analysis_us.qe_component_mapping))
 					SELECT
 						ANY_VALUE(test_name) AS test_name,
 						ANY_VALUE(testsuite) AS test_suite,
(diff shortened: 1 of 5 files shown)


pkg/api/component_report.go
```diff
@@ -51,7 +51,7 @@ const (
 						ELSE 2
 					END) AS row_num
 			FROM
-				ci_analysis_us.junit
+				ci_analysis_us.qe_junit
```
stbenjam, line 54: Sippy has a config file, I've been thinking if we're going to productize sippy maybe these kinds of values can go there? The dataset and project should also probably be customizable.  It could also be command-line flags but feels like sippy's already got too many.

https://github.com/openshift/sippy/blob/master/pkg/apis/config/v1/types.go

And the configs are in the config/ dir.

This would also let them setup a postgres and pull in their job data if they were interested

deads2k: I think I'd suggesting forcing the usage of a different `ci_analysis_us` and keeping the table names the same.  This PR is just looking at what it takes to rejigger enough to see data.

IOW, for this I'd redo the import

deads2k: > It could also be command-line flags but feels like sippy's already got too many.

Having read through this briefly, I think the first cut I'd make is to separate out the various modes into differnet binaries.

stbenjam: I started to break things out into subcommands here, not sure if it's useful if you're intending to do something similar.  I never got around to finishing it: https://github.com/openshift/sippy/pull/807

deads2k: > I started to break things out into subcommands here, not sure if it's useful if you're intending to do something similar. I never got around to finishing it: #807

Yeah, I'd probably start somewhere like that.

deads2k:
dead

PR #1303, opened 2023-11-06
Don't generate component readiness metrics until we have a cache

stbenjam:
If we don't have a cache, then we don't want component readiness metrics, because you'll end up hitting bigquery every 5 minutes ($$$).


stbenjam:
/hold

Needs a better mechanism

PR #1304, opened 2023-11-06
[wip] Remember the last update for expensive metrics

stbenjam:
Don't update component readiness more than every 6 hours.

### DIFF
--- a/pkg/db/models/metrics.go
+++ b/pkg/db/models/metrics.go
@@ -1,0 +1,6 @@
+package models
+
+type MetricsRefresh struct {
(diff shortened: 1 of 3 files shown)


pkg/db/models/metrics.go
```diff
@@ -0,0 +1,6 @@
+package models
+
+type MetricsRefresh struct {
```
dgoodwin, line 3: How about a more generic LastUpdate, as there could be other uses for parts of sippy that want to track when something was last refreshed/updated.

PR #1308, opened 2023-11-10
trt-1167: check if updated axios is the OOM issue

neisw:


stbenjam:
npm lint succeeds fine.   It looks like golangci-lint is OOM'ing and that's happened in the past when there's a mismatch between the golang verison and the version golangci-lint was built with, maybe we got a new go somehow?

stbenjam:
We also don't seem to remove axios, we can just remove it entirely: https://github.com/openshift/sippy/pull/1310

stbenjam:
/close

https://github.com/openshift/sippy/pull/1310 fixed the oom

PR #1309, opened 2023-11-13
Upgrade to Material UI v5

stbenjam:
We're currently on v4, which will go EOL when v6 comes out (soon).  This updates Sippy to material UI v5, working through https://mui.com/material-ui/migration/migration-v4/ and addressing other bugs as I find them.

Collecting issues here:
- [X] Update deps to mui v5
- [X] Run codemod fixes
- [X] Theme colors have changed 
- [X] Sidebar doesn't work anymore
- [x] Fix component readiness side bar - it's totally wrong on the first page load, and then looks fine on refresh
- [x] Job analysis page doesn't load at all
- [x] Component readiness tables are off to the side
- [x] Release overview page doesn't get centered
- [x] Test table is off the borders of the page
- [x] Tests-by-variage page (i.e. install) has vertical column headers?? http://localhost:3000/sippy-ng/install/4.15/operators
- [x] Margins on in-page tabs are way off e.g. [this page](http://localhost:3000/sippy-ng/jobs/4.15?filters=%7B%22items%22%3A%5B%7B%22columnField%22%3A%22current_runs%22%2C%22operatorValue%22%3A%22%3E%3D%22%2C%22value%22%3A%227%22%7D%2C%7B%22columnField%22%3A%22variants%22%2C%22operatorValue%22%3A%22contains%22%2C%22value%22%3A%22never-stable%22%2C%22not%22%3Atrue%7D%5D%7D&sortField=net_improvement&sort=asc)
- [x] Jobs page icons have blue hover (underlined from codemods?)
- [x] date pickers are broken (and component readiness won't load)

PR #1314, opened 2023-11-16
Add a dark mode

stbenjam:
Fun little addition that's more possible after fixing up some of our CSS mess in https://github.com/openshift/sippy/pull/1309.  I think things are all usable, but the theme colors might need some tweaking.

cc: @neisw Since you asked about dark mode :) 

Screenshot 2023-11-15 at 22 20 15


stbenjam:
I don't think I broke anything.  Dark mode colors might need some tweaks but I think it's good enough for now

PR #1320, opened 2023-11-20
Fix date picker for filters

stbenjam:
Bug from the v5 upgrade. Job analysis page date filters aren't working, it needed to be wrapped with the LocalizationProvider like I did for component readiness.


DennisPeriquet:
tested the broken one in prod and tested this locally and it's fixed.
/lgtm

PR #1330, opened 2023-11-24
TRT-1357: Publish disruption delta metrics

dgoodwin:
Query new views in BigQuery which contain the complex logic, whatever is
in those views, we publish a metric for. One of them will need to be
updated as part of the GA date tasks.

Metrics published for the P50/75/95 delta now compared to both 30 days
prior to previous GA, and two weeks prior to two weeks ago.

We'll alert on these to notify ourselves of regressions in either.

### DIFF
--- a/pkg/api/disruption_report.go
+++ b/pkg/api/disruption_report.go
@@ -1,0 +1,79 @@
+package api
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"cloud.google.com/go/bigquery"
+	apitype "github.com/openshift/sippy/pkg/apis/api"
+	bqcachedclient "github.com/openshift/sippy/pkg/bigquery"
+	log "github.com/sirupsen/logrus"
+	"google.golang.org/api/iterator"
+)
+
+func GetDisruptionVsPrevGAReportFromBigQuery(client *bqcachedclient.Client) (apitype.DisruptionReport, []error) {
+	generator := disruptionReportGenerator{
+		client:   client.BQ,
+		ViewName: "BackendDisruptionPercentilesDeltaCurrentVsPrevGA",
+	}
+
+	return getReportFromCacheOrGenerate[apitype.DisruptionReport](client.Cache, generator, generator.GenerateReport, apitype.DisruptionReport{})
+}
+
+func GetDisruptionVsTwoWeeksAgoReportFromBigQuery(client *bqcachedclient.Client) (apitype.DisruptionReport, []error) {
+	generator := disruptionReportGenerator{
+		client:   client.BQ,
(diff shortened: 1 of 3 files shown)


pkg/sippyserver/metrics/metrics.go
```diff
@@ -130,9 +138,17 @@ func RefreshMetricsDB(dbc *db.DB, bqc *bqclient.Client, variantManager testident
 	refreshPayloadMetrics(dbc, reportEnd)
 
 	if bqc != nil {
+		/* TODO: don't commit this
+
 		if err := refreshComponentReadinessMetrics(bqc); err != nil {
 			log.WithError(err).Error("error refreshing component readiness metrics")
 		}
+
+		*/
```
stbenjam, line 147: Looks like you committed this :)

dgoodwin: I know myself. :) Thanks!

pkg/api/disruption_report.go
```diff
@@ -0,0 +1,79 @@
...
+	"google.golang.org/api/iterator"
+)
+
+func GetDisruptionVsPrevGAReportFromBigQuery(client *bqcachedclient.Client) (apitype.DisruptionReport, []error) {
+	generator := disruptionReportGenerator{
+		client:   client.BQ,
+		viewName: "BackendDisruptionPercentilesDeltaCurrentVsPrevGA",
+	}
+
+	return getReportFromCacheOrGenerate[apitype.DisruptionReport](client.Cache, generator, generator.GenerateReport, apitype.DisruptionReport{})
```
stbenjam, line 21: I don't think the caching will work, getReportFromCacheOrGenerate will use the JSON marshalled `generator` as the cache key, but there's no public fields in generator so it'll probably just marshal to `{}` (getReportFromCacheORGenerate should probably check for that).

dgoodwin: Ok added, and the cache will error if it sees an empty json struct

pkg/api/component_report.go
```diff
@@ -1183,6 +1184,9 @@ func getReportFromCacheOrGenerate[T any](c cache.Cache, cacheKey interface{}, ge
 		if err != nil {
 			return defaultVal, []error{err}
 		}
+		if bytes.Equal(cacheKey, []byte("{}")) {
+			return defaultVal, []error{fmt.Errorf("cache key serialized to an empty json struct, must have public fields")}
+		}
```
stbenjam, line 1189: I'm not sure this would get noticed, what do you think about panicking like this? https://github.com/openshift/sippy/pull/1333

dgoodwin: Updated to remove my version.

PR #1342, opened 2023-11-29
Add more info about the IncompleteTests message in Risk Analsys

DennisPeriquet:
I saw this message in a [PR](https://github.com/openshift/origin/pull/28423#issuecomment-1831034152) but didn't understand what it meant.

My hope is that this extra verbiage will make it more obvious.


pkg/api/job_runs.go
```diff
@@ -435,7 +435,8 @@ func runJobRunAnalysis(jobRun *models.ProwJobRun, compareRelease string, jobRunT
 	case jobRunTestCount < (int(float64(historicalRunTestCount) * .75)):
 		response.OverallRisk.Level = apitype.FailureRiskLevelIncompleteTests
 		response.OverallRisk.Reasons = append(response.OverallRisk.Reasons,
-			fmt.Sprintf("Tests for this run (%d) are below the historical average (%d): IncompleteTests", jobRunTestCount, historicalRunTestCount))
+			fmt.Sprintf("Tests for this run (%d) are below the historical average (%d): IncompleteTests", jobRunTestCount, historicalRunTestCount),
+			"Not enough tests ran to make a reasonable risk analysis; this could be due to infra, installation, or upgrade problems")
```
dgoodwin, line 439: Adding another "reason" here is probably going to work but a little misleading and could get something confused. You want this text in with the reason it's applicable to, please move into the Sprintf.

Anything counting reasons would think two things were wrong when it was one.

DennisPeriquet: change pushed

PR #1344, opened 2023-11-29
Show revert count in repositories API

stbenjam:
TRT-1288

This adds a revert count into the repositories API based on data in the Sippy DB.

*Note*:: this also changes the behavior of partial SQL filters "contains", "starts with", "ends with" to use case-insensitive matching.  That seems like a better behavior anyway.

### DIFF
--- a/pkg/apis/api/types.go
+++ b/pkg/apis/api/types.go
@@ -48,10 +48,11 @@ type Pagination struct {
 }
 
 type Repository struct {
-	ID       int    `json:"id"`
-	Org      string `json:"org"`
-	Repo     string `json:"repo"`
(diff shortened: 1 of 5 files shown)


neisw:
/hold
release when ready

dgoodwin:
Excellent!

DennisPeriquet:
using production DB, and local UI, at https://github.com/openshift/sippy/pull/1344/files, `Revert count` column is empty.

stbenjam:
Production DB or Production API? There's an API change to the golang code.

It works for me fine locally

image

DennisPeriquet:
ah, you're right, I need a local sippy api

PR #1350, opened 2023-12-04
TRT-1249: Show regressed tests on component readiness

xueqzhan:


sippy-ng/src/component_readiness/RegressedTestsModal.js
```diff
@@ -0,0 +1,207 @@
...
+export default function RegressedTestsModal(props) {
+  // define table columns
+  const columns = [
+    {
+      field: 'test_name',
+      headerName: 'Test Name',
+      flex: 40,
+      renderCell: (param) => <div className="test-name">{param.value}</div>,
+      sortable: false,
+    },
```
stbenjam, line 70: I'd suggest making the column order Component, Capability, then Test Name (and sorted by Component first)

sippy-ng/src/component_readiness/ComponentReadiness.js
```diff
@@ -401,6 +404,11 @@ export default function ComponentReadiness(props) {
   const [isLoaded, setIsLoaded] = React.useState(false)
   const [data, setData] = React.useState({})
 
+  const [regressedTestDialog, setRegressedTestDialog] = React.useState(false)
```
stbenjam, line 407: Suggest making this a query parameter so the modal is open when you hit the back button (or share a link)

```suggestion
  const [regressedTestDialog = false, setRegressedTestDialog] = useQueryParam(
    'regressedModal',
    BooleanParam
  )```

In order to make this work nicely you'll have to yoink this parameter out of the URL in the `key` since it doesn't require a new API call

Example here:

https://github.com/openshift/sippy/commit/d7291880f92d29c9fb72725cb3c5add6b00f019d#diff-0dd37ad68b4331fc273d39705d19e402ed1eba635da3b349ff622c8dd09f3c3eR486

**Comment by stbenjam** [comment]
You'll probably want to rebase to pull in recent changes before the rebase gets too difficult

File: sippy-ng/src/component_readiness/RegressedTestsModal.js [hunk 3]
```diff
@@ -0,0 +1,207 @@
...
+    {
+      field: 'test_suite',
+      headerName: 'Test Suite',
+      flex: 15,
+      renderCell: (param) => <div className="test-name">{param.value}</div>,
+      sortable: false,
```
stbenjam, line 76: Can we make all of these columns sortable?

sippy-ng/src/component_readiness/RegressedTestsModal.js
```diff
@@ -0,0 +1,207 @@
...
+        onClose={props.close}
+      >
+        <Grid className="regressed-tests-dialog">
+          <Typography
+            variant="h6"
+            style={{ marginTop: 20, marginBottom: 20, marginLeft: 20 }}
+          >
+            Regressed Tests
+          </Typography>
+          <DataGrid
```
stbenjam, line 168: Can we add a quick filter to this table?

https://mui.com/x/react-data-grid/filtering/quick-filter/

dgoodwin:
I'm only seeing two results when I run locally, likely due to a comment Stephen has in the review above, but testing locally this looks awesome and will be very useful, thanks Ken!

sippy-ng/src/component_readiness/ComponentReadiness.js
```diff
@@ -502,6 +504,15 @@ export default function ComponentReadiness(props) {
     setRedOnlyChecked(false)
   }
 
+  const regressedTests = mergeRegressedTests(data)
+  const [regressedTestDialog = false, setRegressedTestDialog] = useQueryParam(
+    'regressedModal',
+    BooleanParam
+  )
+  const closeRegressedTestsDialog = () => {
+    setRegressedTestDialog(false)
```
stbenjam, line 513:
```suggestion
    setRegressedTestDialog(false, 'replace')
```

This should fix the history problems you were having

sippy-ng/src/component_readiness/ComponentReadiness.js
```diff
@@ -897,290 +908,289 @@ export default function ComponentReadiness(props) {
...
+                                textAlign: 'left',
+                                marginTop: 15,
+                              }}
+                              label="Red Only"
+                            ></FormControlLabel>
+                            <Button
+                              style={{ marginTop: 20 }}
+                              variant="contained"
+                              color="secondary"
+                              onClick={() => setRegressedTestDialog(true)}
```
stbenjam, line 1079: Here too

```suggestion
                              onClick={() => setRegressedTestDialog(true, 'replace')}
```

stbenjam (commented):
Looks really good

sippy-ng/src/component_readiness/ComponentReadiness.js
```diff
@@ -897,290 +908,289 @@ export default function ComponentReadiness(props) {
...
-                          transformOrigin={{
-                            vertical: 'top',
-                            horizontal: 'center',
-                          }}
-                        >
-                          Link copied!
-                        </Popover>
-                        <Tooltip title="Clear searches">
-                          <IconButton
-                            color="primary"
```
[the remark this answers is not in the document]
xueqzhan: This is weird, I am still seeing the buttons.

sippy-ng/src/component_readiness/ComponentReadiness.js
xueqzhan: Turned out it is 'replaceIn'. Using 'replace', I was seeing all kinds of weird behaviors. I think it replaced all parameters.

sippy-ng/src/component_readiness/ComponentReadiness.js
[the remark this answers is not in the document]
stbenjam: I don't, I'm running your latest branch. I don't see where the link copy or clear are in your code either.  The image I shared has the 2 buttons I'm talking about.  The search boxes are still there.

image

sippy-ng/src/component_readiness/RegressedTestsModal.js
```diff
@@ -0,0 +1,245 @@
...
+export default function RegressedTestsModal(props) {
+  // define table columns
+  const columns = [
+    {
+      field: 'component',
+      headerName: 'Component',
+      flex: 20,
+      renderCell: (param) => (
+        <div className="test-name">
+          <Tooltip title={'Component'}>{param.value}</Tooltip>
```
stbenjam, line 70: Sorry I meant to have the tooltip show the value.  In the screenshot I shared before all the values were truncated with ellipsis.  Like this:

```
<Tooltip title={param.value}>{param.value}</Tooltip>
```

But now you also wrapped these in a `<div className="test-name">` which turns off the truncate.  You can probably drop the tooltip now, depending which view you prefer.

stbenjam:
It's an existing problem but this toolbar has weird spacing, mostly from the buttons I added.  I'll try to fix it, but otherwise this is incredible

image



/lgtm

PR #1379, opened 2023-12-18
TRT-1259: Fix time filter to use ISO format that is required by API

xueqzhan:


DennisPeriquet:
pulled this down and added a filter and it looks like this:

```
  "columnField": "release_time",
  "operatorValue": "<=",
  "value": "2023-12-11T05:00:00.000Z"
```

Looks good.

PR #1382, opened 2023-12-20
Update to new GCS bucket

stbenjam:


pkg/dataloader/prowloader/prow.go
```diff
@@ -515,12 +516,9 @@ func (pl *ProwLoader) prowJobToJobRun(ctx context.Context, pj *prow.ProwJob, rel
 		return err
 	}
 
-	parts := strings.Split(pjURL.Path, pl.bktName)
```
stbenjam, line 518: This code was splitting on the bucket name, but at the moment the bucket name is test-platform-results but the path still has origin-ci-test.  I moved to a regexp instead.

pkg/dataloader/prowloader/prow.go
```diff
@@ -39,6 +38,8 @@ import (
 	"github.com/openshift/sippy/pkg/util/sets"
 )
 
+var gcsPathStrip = regexp.MustCompile(`.*/gs/[^/]+/`)
```
xueqzhan, line 41: Maybe document a simple example of what this is trying to match?

pkg/dataloader/prowloader/prow.go
```diff
@@ -515,12 +516,9 @@ func (pl *ProwLoader) prowJobToJobRun(ctx context.Context, pj *prow.ProwJob, rel
...
 	}
 
-	parts := strings.Split(pjURL.Path, pl.bktName)
-	pjLog.WithField("path", pjURL.Path).WithField("bucket", pl.bktName).WithField("parts", parts).Info("building path")
-	if len(parts) < 2 {
-		return fmt.Errorf("bucket name '%s' does not match prowjob url path '%s'", pl.bktName, pjURL.Path)
-	}
-	path := parts[1][1:]
+	// Get the path in the gcs bucket, strip out the bucket name and anything before it
+	path := gcsPathStrip.ReplaceAllString(pjURL.Path, "")
```
DennisPeriquet, line 520: the original code on line 520-522 handles the case when the parse via `Split` isn't as expected by returning an `error`.  In the new code here, if the regex doesn't match, the path is the same string but is not handled.

DennisPeriquet: I'm assuming you want to carry forward the error handling that the original code did.

stbenjam: Hm, I didn't think it was worthwhile originally but I guess if the path ends up being blank we could do something silly like traverse the whole bucket which would be expensive.

PR #1399, opened 2024-01-05
Qe branch 3

deads2k:
I think I pushed.

### DIFF
--- a/cmd/sippy/component_readiness.go
+++ b/cmd/sippy/component_readiness.go
@@ -1,0 +1,170 @@
+package main
+
+import (
+	"context"
+	"io/fs"
+	"net/http"
+	"os"
+	"time"
+
+	"cloud.google.com/go/storage"
+	resources "github.com/openshift/sippy"
+	v1 "github.com/openshift/sippy/pkg/apis/config/v1"
+	"github.com/openshift/sippy/pkg/bigquery"
+	"github.com/openshift/sippy/pkg/dataloader/prowloader/gcs"
+	"github.com/openshift/sippy/pkg/flags"
+	"github.com/openshift/sippy/pkg/sippyserver"
+	"github.com/pkg/errors"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+	log "github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+	"gopkg.in/yaml.v3"
+)
+
+type ComponentReadinessFlags struct {
+	GoogleCloudFlags *flags.GoogleCloudFlags
+	BigQueryFlags    *flags.BigQueryFlags
+	CacheFlags       *flags.CacheFlags
+
+	Config      string
+	LogLevel    string
+	ListenAddr  string
+	MetricsAddr string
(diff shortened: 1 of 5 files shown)


pkg/api/component_report.go
```diff
@@ -742,6 +742,10 @@ func (c *componentReportGenerator) normalizeProwJobName(prowName string) string
 	if prev, err := previousRelease(c.SampleRelease.Release); err == nil {
 		name = strings.ReplaceAll(name, prev, "X.X")
 	}
+	name = strings.ReplaceAll(name, "f28", "fXX")
+	name = strings.ReplaceAll(name, "f14", "fXX")
+	name = strings.ReplaceAll(name, "f7", "fXX")
+	name = strings.ReplaceAll(name, "f6", "fXX")
 	return name
```
stbenjam, line 749: Will they have other frequencies?  I know you hate regex but this isn't too bad:

	re := regexp.MustCompile(`-f\d+`)
	output := re.ReplaceAllString(input, "-fXX")

cmd/sippy/component_readiness.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+
+	cacheClient, err := f.CacheFlags.GetCacheClient()
+	if err != nil {
+		return errors.WithMessage(err, "couldn't get cache client")
+	}
+
+	var bigQueryClient *bigquery.Client
+	var gcsClient *storage.Client
+	if f.GoogleCloudFlags.ServiceAccountCredentialFile != "" {
```
deads2k, line 127: I'd like to push this logic into the GoogleCloudFlags type, so we can be consistent with how to get output from the input.

pkg/sippyserver/install_details.go
```diff
@@ -7,12 +7,22 @@ import (
 )
 
 func (s *Server) jsonUpgradeReportFromDB(w http.ResponseWriter, req *http.Request) {
```
deads2k, line 9: Overall, would we try to refactor Server to have a componentREadinessServer?

pkg/sippyserver/install_details.go
```diff
@@ -7,12 +7,22 @@ import (
...
+		return
+	}
+
 	release := req.URL.Query().Get("release")
 
 	api.PrintUpgradeJSONReportFromDB(w, req, s.db, release)
 }
 
 func (s *Server) jsonInstallReportFromDB(w http.ResponseWriter, req *http.Request) {
+	if s.db == nil {
```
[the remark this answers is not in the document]
stbenjam: There's an if block in the server section that checks for s.db.  Leftover from when we were partially on testgrid disk data and partially on postgres.  You could move all these under there.

stbenjam:
I don't think this has to be perfect to merge, it's making progress and would keep you from having to maintain a separate branch.  I think we should just merge this once lint passes.

pkg/sippyserver/install_details.go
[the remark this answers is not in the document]
stbenjam: I refactored this quite a bit in https://github.com/openshift/sippy/pull/1559, it includes most of this PR

PR #1406, opened 2024-01-09
TRT-1331:  Use precomputed tables for component readiness  [2]

stbenjam:
Merge after https://github.com/openshift/sippy/pull/1405.

This moves to using views, and scheduled queries that precompute some data for component readiness.  This is an improvement in code readability by moving the complex SQL queries out as well.

I'll have another PR later that makes Sippy create/update these views

There are some new views/queries:
- component_mapping_latest [View]: the latest mappings for components from the table
- junit_deduped [View]: the flattened junit table in the format CR expects (one row per-test result-per-job)
- junit_precomputed_by_test_and_day [Scheduled query]:  runs every 8 hours to sum the results by day (the most granular time period CR cares about), performance and cost savings

### DIFF
--- a/pkg/api/component_report.go
+++ b/pkg/api/component_report.go
@@ -52,7 +52,7 @@ const (
 						ELSE 2
 					END) AS row_num
 			FROM
-				ci_analysis_us.junit
+				%s.junit
 			WHERE modified_time >= DATETIME(@From)
 			AND modified_time < DATETIME(@To)
 			AND skipped = false
@@ -99,7 +99,7 @@ func getSingleColumnResultToSlice(query *bigquery.Query) ([]string, error) {
 
 func GetComponentTestVariantsFromBigQuery(client *bqcachedclient.Client, gcsBucket string) (apitype.ComponentReportTestVariants, []error) {
 	generator := componentReportGenerator{
-		client:    client.BQ,
+		client:    client,
 		gcsBucket: gcsBucket,
 	}
 
@@ -113,7 +113,7 @@ func GetComponentReportFromBigQuery(client *bqcachedclient.Client, gcsBucket str
 	excludeOption apitype.ComponentReportRequestExcludeOptions,
 	advancedOption apitype.ComponentReportRequestAdvancedOptions) (apitype.ComponentReport, []error) {
 	generator := componentReportGenerator{
-		client:        client.BQ,
+		client:        client,
 		gcsBucket:     gcsBucket,
 		BaseRelease:   baseRelease,
 		SampleRelease: sampleRelease,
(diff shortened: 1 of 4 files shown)


pkg/api/component_report.go
```diff
@@ -253,10 +215,6 @@ func (c *componentReportGenerator) getJobRunTestStatusFromBigQuery() (
 						AND @Variant IN UNNEST(variants)
 						AND cm.id = @TestId `
 	commonParams := []bigquery.QueryParameter{
-		{
-			Name:  "IgnoredJobs",
-			Value: ignoredJobsRegexp,
-		},
```
dgoodwin, line 259: In the context of a productized sippy, it might be best to keep the ignored jobs configurable. I'd expect we'll want view schema managed automatically and common for any deployment of sippy, but the jobs to ignore certain jobs something configured in the app itself and applied dynamically. WDYT.

pkg/api/component_report.go
```diff
@@ -348,31 +306,27 @@ func (c *componentReportGenerator) getTestStatusFromBigQuery() (
...
+            SUM(total_count) AS total_count,
+            SUM(success_count) AS success_count,
+            SUM(flake_count) AS flake_count,
+            ANY_VALUE(component) AS component,
+            ANY_VALUE(capabilities) AS capabilities,
+            ANY_VALUE(jira_component) AS jira_component,
+            ANY_VALUE(jira_component_id) AS jira_component_id
+          FROM
+            %s.junit_precomputed_by_test_and_day
+          WHERE date >= @From AND date <= @To`, c.client.Dataset)
```
dgoodwin, line 329: Can you clarify the use of flat_variants and variants (repeated) in the schema for this table. Just plotting for the externalized job variants work coming.

stbenjam: Both have the same data, variants is the array, flat_variants is the comma-separated string

pkg/api/component_report.go
stbenjam: It doesn't quite work with the materialized view (scheduled query) to do it that way, because I'm precomputing test results for whole days grouped by network/upgrade/arch/platform/variants. I had planned to make the ignored jobs configurable in Sippy when it creates the views.

If I do it at the job level, it'll be less useful for performance/cost, but I'm not sure how much, I'll look into it

pkg/api/component_report.go
dgoodwin: Sorry I was wondering why both were present.

stbenjam: I'm not sure, it's the existing SQL Ken wrote.  I'm guessing it's to do with bigquery limitations on working with the repeated  column (arrays)

stbenjam: Actually it's coming from the junit table that Justin created, I don't know why there's both.  Probably for flexibility?

stbenjam:
I don't intend to tackle this one now, details about how we use views is a little unclear in light of trying to productize things.  I'll defer this until we get that sorted.

PR #1415, opened 2024-01-11
TRT-1405: Add context to each CR page to show releases and dates

xueqzhan:


xueqzhan:
/hold

I have some progress with enabling sidebar for subsequent pages. Maybe we don't need this at all.

PR #1420, opened 2024-01-12
TRT-1435: Add a FAQ for component readiness

stbenjam:
image

image


neisw:
/lgtm
/hold
Holding for review on is we should force the external links to open in a new tab.  But ok to have it merge as is.  
Very cool adding the FAQ and links to test mapping repository, etc.!

PR #1425, opened 2024-01-16
trt-622: use build_clusters capabilities value to drive tools panel i…

neisw:
…tems

image

### DIFF
--- a/sippy-ng/src/components/Sidebar.js
+++ b/sippy-ng/src/components/Sidebar.js
@@ -96,38 +96,6 @@ export default function Sidebar(props) {
       </List>
       <CapabilitiesContext.Consumer>
         {(value) => {
-          if (value.includes('build_clusters')) {
-            return (
-              <Fragment>
-                <Divider />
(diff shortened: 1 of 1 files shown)


sippy-ng/src/components/Sidebar.js
```diff
@@ -97,38 +97,6 @@ export default function Sidebar(props) {
       <CapabilitiesContext.Consumer>
         {(value) => {
           if (value.includes('build_clusters')) {
```
stbenjam, line 99: This doesn't look right to me, it ends up putting the entire tools menu under the build_clusters capability.   You can move this whole block under Tools, starting with the CapabilitiesContext.Consumer, but instead of returning the Divider and List headers, just return the ListItem

neisw: Yeah, what was unclear to me was if we wanted that nested inside of 
`  if (value.includes('openshift_releases')) {`

But let me play around with it more

PR #1433, opened 2024-01-19
Add mechanism to indicate that a regression is allowed in sippy

deads2k:
The core part of the change that allows regressions to be introduced into the product.

/assign @stbenjam

### DIFF
--- a/pkg/api/component_report.go
+++ b/pkg/api/component_report.go
@@ -17,6 +17,7 @@ import (
 
 	apitype "github.com/openshift/sippy/pkg/apis/api"
 	bqcachedclient "github.com/openshift/sippy/pkg/bigquery"
+	"github.com/openshift/sippy/pkg/regressionallowances"
 	"github.com/openshift/sippy/pkg/util/sets"
 )
 
@@ -909,7 +910,8 @@ func (c *componentReportGenerator) generateComponentTestReport(baseStatus map[ap
 		if !ok {
 			reportStatus = apitype.MissingSample
 		} else {
-			reportStatus, _ = c.assessComponentStatus(sampleStats.TotalCount, sampleStats.SuccessCount, sampleStats.FlakeCount, baseStats.TotalCount, baseStats.SuccessCount, baseStats.FlakeCount)
+			approvedRegression := regressionallowances.IntentionalRegressionFor(c.SampleRelease.Release, testID.ComponentReportColumnIdentification, testID.TestID)
+			reportStatus, _ = c.assessComponentStatus(sampleStats.TotalCount, sampleStats.SuccessCount, sampleStats.FlakeCount, baseStats.TotalCount, baseStats.SuccessCount, baseStats.FlakeCount, approvedRegression)
 		}
 		delete(sampleStatus, testIdentification)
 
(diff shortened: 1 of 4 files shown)


stbenjam:
This adds OCP-specific logic to component readiness which we're trying to avoid adding more of. It is fine if we're using this to make a shipping decision for 4.15 and want everything green but I'd prefer if the PR description/commit message had an explanation that this is temporary and what the ideas are for a productized version of this (at least a link to a jira card so it's not dropped).

pkg/api/component_report.go
```diff
@@ -1188,18 +1194,27 @@ func (c *componentReportGenerator) assessComponentStatus(sampleTotal, sampleSucc
 			if c.MinimumFailure != 0 && (sampleTotal-sampleSuccess-sampleFlake) < c.MinimumFailure {
 				return apitype.NotSignificant, fischerExact
 			}
+
 			basisPassPercentage := float64(baseSuccess+baseFlake) / float64(baseTotal)
 			samplePassPercentage := float64(sampleSuccess+sampleFlake) / float64(sampleTotal)
+			effectivePityFactor := c.PityFactor
+			if approvedRegression != nil && approvedRegression.RegressedPassPercentage < int(basisPassPercentage) {
```
stbenjam, line 1201: Looks like basisPassPercentage is actually a ratio, needs to be `* 100`.  Same below on line 1204.  Once that is fixed, the squares get greened

pkg/api/component_report.go
```diff
@@ -1188,18 +1194,27 @@ func (c *componentReportGenerator) assessComponentStatus(sampleTotal, sampleSucc
...
+				// product owner chose a required pass percentage, so we all pity to cover that approved pass percent
+				// plus the existing pity factor to limit, "well, it's just *barely* lower" arguments.
+				effectivePityFactor = int(basisPassPercentage) - approvedRegression.RegressedPassPercentage + c.PityFactor
```
stbenjam, line 1204: Maybe also a sanity check to make suer effectivePityFactor is never below 0 or c.PityFactor.  Could just log and reset, or panic since it probably means the approvedRegression has messed up values

pkg/regressionallowances/types.go
```diff
@@ -0,0 +1,123 @@
...
+		return fmt.Errorf("Upgrade must be specified.")
+	}
+
+	var targetMap map[regressionKey]IntentionalRegression
+
+	switch release {
+	case release415:
+		targetMap = regressions_415
+	default:
+		fmt.Errorf("unknown release: %q", release)
```
stbenjam, line 112: missing return

pkg/regressionallowances/types.go
```diff
@@ -0,0 +1,123 @@
...
+	switch release(releaseString) {
+	case release415:
+		targetMap = regressions_415
+	default:
+		return nil
+	}
+
+	inKey := keyFor(testID, variant)
+	t := targetMap[inKey]
+	return &t
```
stbenjam, line 32:
```suggestion
if t, ok := targetMap[inKey]; ok {
    log.Debugf("found approved regression: %+v", t)
    return &t
}
return nil
```

deads2k:
> It is fine if we're using this to make a shipping decision for 4.15 and want everything green but I'd prefer if the PR description/commit message had an explanation that this is temporary and what the ideas are for a productized version of this (at least a link to a jira card so it's not dropped).

I suspect this will continue until productization is more thoroughly completed.  I have opened https://issues.redhat.com/browse/TRT-1448 and will add details in the commit history as well.

PR #1436, opened 2024-01-19
allow exception for SDN installation failure

deads2k:
This PR demonstrates how to reduce the acceptable reliability for a particular capability in OCP.  Once merged, this change will allow installations on SDNs to go from 100% reliable to 50% reliable.  The format clearly exposes
1. What is changing
2. What the previous reliability was
3. What the new reality is
4. Why we are allowing the regression

for ease of review by the product owner (Steve Cuppett)  who can decide to regress the capability and reliability of the platform.

/hold

Publishing this as an example of a current 4.15 failure.

### DIFF
--- a/pkg/regressionallowances/intentional_4.15_regressions.go
+++ b/pkg/regressionallowances/intentional_4.15_regressions.go
@@ -1,5 +1,7 @@
 package regressionallowances
 
+import "github.com/openshift/sippy/pkg/apis/api"
+
 //nolint:all
 func init() {
 	/*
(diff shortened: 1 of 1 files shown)


pkg/regressionallowances/intentional_4.15_regressions.go
```diff
@@ -22,4 +24,23 @@ func init() {
...
+			Variant: api.ComponentReportColumnIdentification{ // this indicates the selectivity of the choice
+				Network:  "sdn",
+				Upgrade:  "upgrade-micro",
+				Arch:     "amd64",
+				Platform: "metal-ipi",
+			},
+			PreviousPassPercentage:    100,
+			PreviousSampleSize:        61, // number of runs used for the percentage
+			RegressedPassPercentage:   50,
+			RegressedSampleSize:       12, // number of runs used for the percentage
```
nstielau, line 43: Are teams supposed to come up with these?  If so, we'll probably need to give guidance.

pkg/regressionallowances/intentional_4.15_regressions.go
```diff
@@ -22,4 +24,23 @@ func init() {
...
+			ReasonToAllowInsteadOfFix: "I can't explain this, but Ben Bennett maybe?",
```
wking, line 44: Is this property expected to link work to fix the regression, when such work is planned?  I could see:

* We regress, and we want to block 4.y patch releases until we fix it.  There's a tracking OCPBUGS-# pointed at 4.y.z.
* We regress, and decide that's just how the world is now.  We land an intentional regression exception and do not remove it until we stop running Sippy on that old 4.y, long after its end-of-life.
* We regress, and decide that we want to fix it, but that it's not bad enough to block further 4.y patch releases.  There's a tracking OCPBUGS-# pointed at 4.y.z and an intentional regression exception that gets removed once we're done touching builds that lack the fix.

It's that last case where I'd expect an OCPBUGS-# link somewhere in the regression declaration, and I'm wondering if that is structured somewhere, or unstructured in this property, or unstructured in the commit message, or...?

PR #1444, opened 2024-01-22
add ability to indicate an issue was resolved and impacted some job runs

deads2k:
This allows rapid re-greening, while remaining sensitive to "the problem is still happening after it was supposed to be fixed".

The one that I added should turn the red cell on [azure](https://sippy.dptools.openshift.org/sippy-ng/component_readiness/test_details?arch=amd64&baseEndTime=2023-10-31%2023%3A59%3A59&baseRelease=4.14&baseStartTime=2023-10-04%2000%3A00%3A00&capability=Other&component=Installer%20%2F%20openshift-installer&confidence=95&environment=ovn%20upgrade-micro%20amd64%20azure%20standard&excludeArches=arm64%2Cheterogeneous%2Cppc64le%2Cs390x&excludeClouds=openstack%2Cibmcloud%2Clibvirt%2Covirt%2Cunknown&excludeVariants=hypershift%2Cosd%2Cmicroshift%2Ctechpreview%2Csingle-node%2Cassisted%2Ccompact&groupBy=cloud%2Carch%2Cnetwork&ignoreDisruption=true&ignoreMissing=false&minFail=3&network=ovn&pity=5&platform=azure&sampleEndTime=2024-01-22%2023%3A59%3A59&sampleRelease=4.15&sampleStartTime=2024-01-16%2000%3A00%3A00&testId=cluster%20install%3A0cb1bb27e418491b1ffdacab58c5c8c0&testName=install%20should%20succeed%3A%20overall&upgrade=upgrade-micro&variant=standard) back to green.

/assign @stbenjam


pkg/componentreadiness/resolvedissues/types.go
```diff
@@ -0,0 +1,240 @@
...
+	// TODO consider whether jira URL should be required and get the resolution date from there
+	ResolutionDate time.Time
+}
+
+type PayloadIssue struct {
+	// required
+	PullRequestURL string
+
+	// TODO switch to detecting this from the payload
+	// required
```
stbenjam, line 42: The payload approach is tricky, you only get the time it was created IIRC.  We probably need to start storing the time it was accepted.  Until then you'll have to figure the payload time + some number of hours.  There's a delay from payload creation to job start, and then the duration of the running jobs.  8 hours?

There are cron jobs running as well that get pulled into CR, which could start after the nightly was created but be running the old payload.

deads2k: > but we only have the creation time; add job start delay and duration (around 8 hours?).

I don't understand what action I should take in this PR based on this.

stbenjam: I'm commenting on a TODO so nothing if you don't plan to do this now

deads2k: >I'm commenting on a TODO so nothing if you don't plan to do this now

ok, thanks.  I don't plan to do anything about this now.

neisw:
Yes, thanks for the heads up.  And the explanation regarding how reducing sampleTotal increases [samplePassPercentage](https://github.com/openshift/sippy/pull/1444/files#diff-4731a371f11b299e4a1203ccd7c38a9578c9cece373ec82736aea8732260a977R1211)

/lgtm

stbenjam:
/retitle TRT-1442: add ability to indicate an issue was resolved and impacted some job runs

stbenjam:
Wrong jira

/retitle TRT-1448: add ability to indicate an issue was resolved and impacted some job runs

deads2k:
> This greens the square and code looks ok to me other than lint.
> 
> @neisw do you want to take a look since you're working on the triage tool?

still red on https://sippy.dptools.openshift.org/sippy-ng/component_readiness/main?regressedModal=0 .  I wonder what I missed

neisw:
Is this the thing Ken is chasing?  Looks green when you go down one level...  I can try running locally to see what results I get.

image

neisw:
When I run it locally I see it calculate numberOfIgnoredSampleJobRuns and it looks right to me

image

PR #1448, opened 2024-01-23
TRT-1225: Allow user to force refresh the cache, add time/date to the API and pages

stbenjam:
This adds the time the report was generated, and ability to force a reload of the data in the cache.

image


stbenjam:
Will fix the build issues tomorrow, IDE removed some imports for some reason

/hold

stbenjam:
/hold

It may not be needed if we get https://github.com/openshift/sippy/pull/1452 to work

stbenjam:
Actually I think it's useful even if #1452 goes in, in cases where we're doing things like adding known issues (e.g. #1454)

/hold cancel

PR #1451, opened 2024-01-24
require jira bug for a regression allowance

deads2k:
Consistent suggestion for allowances

/assign @jupierce @stbenjam

### DIFF
--- a/pkg/regressionallowances/types.go
+++ b/pkg/regressionallowances/types.go
@@ -2,10 +2,10 @@ package regressionallowances
 
 import (
 	"fmt"
-
-	log "github.com/sirupsen/logrus"
+	"net/url"
 
 	"github.com/openshift/sippy/pkg/apis/api"
+	log "github.com/sirupsen/logrus"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/regressionallowances/types.go
```diff
@@ -95,6 +96,9 @@ func addIntentionalRegression(release release, in IntentionalRegression) error {
 	if len(in.ReasonToAllowInsteadOfFix) == 0 {
 		return fmt.Errorf("reasonToAllowInsteadOfFix must be specified")
 	}
+	if len(in.JiraBug) == 0 {
+		return fmt.Errorf("jiraBug must be specified")
+	}
```
stbenjam, line 101: If we always expect a link WDYT about?

```suggestion
if _, err := url.ParseRequestURI(in.JiraBug); err != nil {
	return fmt.Errorf("jiraBug must be a valid URL")
}
```

deads2k: good idea.  Changed.

PR #1452, opened 2024-01-24
TRT-1232: Add time rounding factor for component readiness release time

xueqzhan:


stbenjam:
The cache times need to be adjusted as well, so we're always expiring entries at the edge of the next rounding boundary

pkg/sippyserver/server.go
```diff
@@ -656,6 +659,18 @@ func (s *Server) jsonComponentReportTestDetailsFromBigQuery(w http.ResponseWrite
 	api.RespondWithJSON(http.StatusOK, w, outputs)
 }
 
+func (s *Server) parseReleaseTime(timeStr string) (time.Time, error) {
+	releaseTime, err := time.Parse(time.RFC3339, timeStr)
+	if err != nil {
+		return releaseTime, err
+	}
+	now := time.Now().UTC()
+	if s.crTimeRoundingFactor != 0 && releaseTime.Year() == now.Year() && releaseTime.YearDay() == now.YearDay() {
```
stbenjam, line 668: `now().Format("2006-01-02") == releaseTime.Format("2006-01-02")` is a little more readable to me, I always forget YearDay is not the day of the month

main.go
```diff
@@ -295,6 +302,13 @@ func (o *Options) Validate() error {
 		return fmt.Errorf("--db-only-mode cannot be set to false (deprecated flag soon to be removed, feature now mandatory)")
 	}
 
+	if o.CRTimeRoundingFactor > maxCRTimeRoundingFactor {
+		return fmt.Errorf("please provide a value smaller than %v for --component-readiness-time-rounding-factor", maxCRTimeRoundingFactor)
+	}
+	if o.CRTimeRoundingFactor > 0 {
+		api.CacheDuration = o.CRTimeRoundingFactor
```
stbenjam, line 309: This isn't enough, you have to make the cache expire at the boundary itself.  If I hit a URL at 12:23 you'll be caching the data until 13:23.  You instead want it to expire at 13:00.

The caching will need to be adjusted for each request, the Cache interface likely needs changing to support this kind of per-request TTL.

main.go
```diff
@@ -54,6 +55,8 @@ const (
 	defaultLogLevel                = "info"
 	defaultDBLogLevel              = "warn"
 	commentProcessingDryRunDefault = true
+	defaultCRTimeRoundingFactor    = 1 * time.Hour
```
stbenjam, line 58: WDYT of starting with 4 hours? I'm not sure an hour has much benefit, and it will increase costs

stbenjam:
Sorry forgot about the metrics code: https://github.com/openshift/sippy/blob/master/pkg/sippyserver/metrics/metrics.go

Can we make sure the CR metrics are aligned the same way?

pkg/sippyserver/metrics/metrics.go
```diff
@@ -30,6 +30,11 @@ const (
 	blockerScoreToAlertOn = 70
 )
 
+var (
+	// CRTimeRoundingFactor will be synced with CRTimeRoundingFactor from Command Option
+	CRTimeRoundingFactor = 0 * time.Hour
```
stbenjam, line 35: Can you clarify this comment? I think setting this to zero makes metrics hit component readiness every 5 minutes 💸 💸 💸

xueqzhan: This parameter is always synced with the command argument --component-readiness-time-rounding-factor. People can intentional disable rounding by setting --component-readiness-time-rounding-factor=0. When they do, this variable stays the default value of 0s and the old logic (24 hours one) will kick in.

main.go
```diff
@@ -295,6 +302,14 @@ func (o *Options) Validate() error {
 		return fmt.Errorf("--db-only-mode cannot be set to false (deprecated flag soon to be removed, feature now mandatory)")
 	}
 
+	if o.CRTimeRoundingFactor > maxCRTimeRoundingFactor {
+		return fmt.Errorf("please provide a value smaller than %v for --component-readiness-time-rounding-factor", maxCRTimeRoundingFactor)
+	}
+	if o.CRTimeRoundingFactor > 0 {
+		api.CacheRoundingDuration = o.CRTimeRoundingFactor
+		metrics.CRTimeRoundingFactor = o.CRTimeRoundingFactor
+	}
```
stbenjam, line 311: Ah this is the part [I was missing](https://github.com/openshift/sippy/pull/1452/files#r1468042657), the main is modifying the values in the other packages.  Personally I'm not a big fan of global state like this

stbenjam: There's a CacheOption struct in the BQ Client, or it could just be passed into the functions.  Does that make it too complicated?

pkg/sippyserver/metrics/metrics.go
stbenjam: Ah I see, can you add a comment that 0 = disabled. I see the `>` check below now

PR #1469, opened 2024-02-05
TRT-1232: Use flat_variants in place of variant to generate consistent data bet…

xueqzhan:
…ween CR pages

We used to use variants to create column grouping and test details query parameter. But since test could belong to multiple variants, like ['serial'] and ['serial', 'techpreview'], we grouped those tests in the test details page when asked for variant=serial. 

For the other pages of component readiness, the status of a cell represents those of a SINGLE regressed test case underneath it. In the example above, it is either the one with variant [‘serial’] or the one with variant [‘serial’, ‘techpreview’].  

That resulted in inconsistent data between test details page and other pages of CR. 

This change uses flat_variants in place of variants when grouping columns and sending query parameters.

### DIFF
--- a/pkg/api/component_report_test.go
+++ b/pkg/api/component_report_test.go
@@ -47,7 +47,7 @@ var (
 	}
 	groupByVariantComponentReportGenerator = componentReportGenerator{
 		gcsBucket:                             "test-platform-results",
-		ComponentReportRequestVariantOptions:  apitype.ComponentReportRequestVariantOptions{GroupBy: "cloud,arch,network,variant"},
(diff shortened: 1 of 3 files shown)


pkg/api/component_report_test.go
```diff
@@ -220,17 +230,11 @@ func TestGenerateComponentReport(t *testing.T) {
...
 		Arch:     "amd64",
 		Network:  "ovn",
-		Variant:  "standard",
-	}
-	columnAWSAMD64OVNVariantFips := apitype.ComponentReportColumnIdentification{
-		Platform: "aws",
-		Arch:     "amd64",
-		Network:  "ovn",
-		Variant:  "fips",
+		Variant:  "standard, fips",
```
dgoodwin, line 237: It looks like this should not have a space after the comma if it's intended to match the db flat_variants col.

PR #1479, opened 2024-02-12
trt-1465: cache query data

neisw:
Refactor to allow the report to be regenerated reusing cached bigquery results.

### DIFF
--- a/pkg/api/component_report.go
+++ b/pkg/api/component_report.go
@@ -103,8 +103,9 @@ func getSingleColumnResultToSlice(query *bigquery.Query) ([]string, error) {
 
 func GetComponentTestVariantsFromBigQuery(client *bqcachedclient.Client, gcsBucket string) (apitype.ComponentReportTestVariants, []error) {
 	generator := componentReportGenerator{
-		client:    client,
-		gcsBucket: gcsBucket,
+		GeneratorVersion: 1,
+		client:           client,
+		gcsBucket:        gcsBucket,
 	}
 
 	return getDataFromCacheOrGenerate[apitype.ComponentReportTestVariants](client.Cache, cache.RequestOptions{}, "component_readiness_variants", generator.GenerateVariants, apitype.ComponentReportTestVariants{})
@@ -119,12 +120,12 @@ func GetComponentReportFromBigQuery(client *bqcachedclient.Client, gcsBucket str
 	cacheOption cache.RequestOptions,
 ) (apitype.ComponentReport, []error) {
 	generator := componentReportGenerator{
-		CacheType:     "data",
-		client:        client,
-		gcsBucket:     gcsBucket,
-		cacheOption:   cacheOption,
-		BaseRelease:   baseRelease,
-		SampleRelease: sampleRelease,
+		GeneratorVersion: 1,
+		client:           client,
+		gcsBucket:        gcsBucket,
+		cacheOption:      cacheOption,
+		BaseRelease:      baseRelease,
+		SampleRelease:    sampleRelease,
 		ComponentReportRequestTestIdentificationOptions: testIDOption,
 		ComponentReportRequestVariantOptions:            variantOption,
 		ComponentReportRequestExcludeOptions:            excludeOption,
@@ -142,12 +143,12 @@ func GetComponentReportTestDetailsFromBigQuery(client *bqcachedclient.Client, gc
 	advancedOption apitype.ComponentReportRequestAdvancedOptions,
 	cacheOption cache.RequestOptions) (apitype.ComponentReportTestDetails, []error) {
 	generator := componentReportGenerator{
-		CacheType:     "data",
-		client:        client,
-		gcsBucket:     gcsBucket,
-		cacheOption:   cacheOption,
-		BaseRelease:   baseRelease,
-		SampleRelease: sampleRelease,
+		GeneratorVersion: 1,
+		client:           client,
+		gcsBucket:        gcsBucket,
+		cacheOption:      cacheOption,
+		BaseRelease:      baseRelease,
+		SampleRelease:    sampleRelease,
 		ComponentReportRequestTestIdentificationOptions: testIDOption,
 		ComponentReportRequestVariantOptions:            variantOption,
 		ComponentReportRequestExcludeOptions:            excludeOption,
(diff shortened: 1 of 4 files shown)


pkg/apis/api/types.go
```diff
@@ -817,23 +818,40 @@ type ComponentReportRequestAdvancedOptions struct {
...
+// implement encoding.TextMarshaler for json map key marshalling support
+func (s ComponentTestIdentification) MarshalText() (text []byte, err error) {
+	type t ComponentTestIdentification
+	return json.Marshal(t(s))
+}
+
+func (s *ComponentTestIdentification) UnmarshalText(text []byte) error {
+	type t ComponentTestIdentification
+	return json.Unmarshal(text, (*t)(s))
 }
```
stbenjam, line 855: Are these needed? The struct should just be serializable on it's own.

pkg/api/component_report.go
```diff
@@ -119,6 +119,7 @@ func GetComponentReportFromBigQuery(client *bqcachedclient.Client, gcsBucket str
 	cacheOption cache.RequestOptions,
 ) (apitype.ComponentReport, []error) {
 	generator := componentReportGenerator{
+		CacheType:     "data",
```
stbenjam, line 122: What does CacheType do? I don't see it used anywhere

pkg/api/utils.go
```diff
@@ -40,8 +40,12 @@ func getReportFromCacheOrGenerate[T any](c cache.Cache, cacheOptions cache.Reque
 				}).Debugf("cache hit")
 				var cr T
 				if err := json.Unmarshal(res, &cr); err != nil {
+					// we got bad data back from the cache,
+					// should we log error then fall through and call generateFn?
 					return defaultVal, []error{err}
```
stbenjam, line 45: I think this behavior's right, we shouldn't get bad data in the cache and if we do we need to not hide it.  I'd probably wrap the error though, `errors.WithMessagef(err, "got bad data back from the cache.  cacheKey=%+v", cacheKey)` or similar.  Otherwise we'll just get a JSON parse error without much context.

pkg/api/component_report.go
```diff
@@ -192,21 +195,29 @@ func (c *componentReportGenerator) GenerateVariants() (apitype.ComponentReportTe
...
+	log.Infof("getTestStatusFromBigQuery completed in %s with %d sample results and %d base results from db", time.Since(before), len(componentReportTestStatus.SampleStatus), len(componentReportTestStatus.BaseStatus))
 	now := time.Now()
-	report.GeneratedAt = &now
-	return report, nil
+	componentReportTestStatus.GeneratedAt = &now
+	return componentReportTestStatus, nil
 }
 
-func (c *componentReportGenerator) GenerateTestDetailsReport() (apitype.ComponentReportTestDetails, []error) {
+func (c *componentReportGenerator) GenerateTestDetailsReport(cc cache.Cache, cacheKey interface{}) (apitype.ComponentReportTestDetails, []error) {
```
stbenjam, line 220: cache should probably just go into `componentReportGenerator` struct as a private field.  And don't we know the type of cacheKey for this function (instead of `interface{}`)?

stbenjam (commented):
A couple of quick comments, haven't tested yet

pkg/api/component_report.go
neisw: It is used in the cacheKey, added a comment [here](https://github.com/openshift/sippy/blob/9dfe15eea256c6a54d72bca9d0a56b0c532e319c/pkg/api/utils.go#L47) but worth noting in the struct as well

pkg/apis/api/types.go
neisw: Yes I was getting errors [here](https://github.com/openshift/sippy/blob/9dfe15eea256c6a54d72bca9d0a56b0c532e319c/pkg/api/utils.go#L55) found this [issue](https://stackoverflow.com/questions/55335296/problem-with-marshal-unmarshal-when-key-of-map-is-a-struct) added a brief comment [here](https://github.com/openshift/sippy/blob/9dfe15eea256c6a54d72bca9d0a56b0c532e319c/pkg/apis/api/types.go#L846)

pkg/api/component_report.go
stbenjam: What's it's purpose though? I'm not seeing anything read from it

stbenjam: I see - `CacheType: data` doesn't really mean anything then, it's just used to force a different cache key.  What about a version field? Maybe an integer `GeneratorVersion` that can bumped as needed if we make incompatible changes again

neisw: Thought was if the same generator cached different types of objects, data vs. report that there was meaning.  But I don't mind switching it to version either, that does fit better for the current usage.

stbenjam:
Thanks looks good now. One last concern now that I've tested it: this is quite a bit slower now since we have to redo the calculations -- about 250x slower (3.2s vs 12.5ms). 3 seconds is on the cusp of feeling sluggish to me.  Is the trade-off worth it?

This version:

INFO[2024-02-14T13:21:41.545-05:00] responded to request                          elapsed=3.193243291s method=GET uri="/api/component_readiness?baseEndTime=2023-10-31T23:59:59Z&baseRelease=4.14&baseStartTime=2023-10-04T00:00:00Z&confidence=95&excludeArches=arm64,heterogeneous,ppc64le,s390x&excludeClouds=openstack,ibmcloud,libvirt,ovirt,unknown&excludeVariants=hypershift,osd,microshift,techpreview,single-node,assisted,compact&groupBy=cloud,arch,network&ignoreDisruption=true&ignoreMissing=false&minFail=3&pity=5&sampleEndTime=2024-02-14T23:59:59Z&sampleRelease=4.15&sampleStartTime=2024-02-08T00:00:00Z"

Previous:

INFO[2024-02-14T13:22:50.992-05:00] responded to request                          elapsed=12.458583ms method=GET uri="/api/component_readiness?baseEndTime=2023-10-31T23:59:59Z&baseRelease=4.14&baseStartTime=2023-10-04T00:00:00Z&confidence=95&excludeArches=arm64,heterogeneous,ppc64le,s390x&excludeClouds=openstack,ibmcloud,libvirt,ovirt,unknown&excludeVariants=hypershift,osd,microshift,techpreview,single-node,assisted,compact&groupBy=cloud,arch,network&ignoreDisruption=true&ignoreMissing=false&minFail=3&pity=5&sampleEndTime=2024-02-14T23:59:59Z&sampleRelease=4.15&sampleStartTime=2024-02-08T00:00:00Z"

neisw:
/close

superseded by https://github.com/openshift/sippy/pull/1589

PR #1483, opened 2024-02-14
TRT-1237: First phase of Save Views for Component Readiness

DennisPeriquet:
On the left bar, you will now see a "View" selector (inspired by `GridToolbarViewSelector` as used in sippy's Jobs table)

Think of a "view" as a collection of configuration settings in Component Readiness.  This allows selection between some commonly used views (or configuration settings) that different groups of users care about.

I arbitrarily chose some views.  We can tweak them as needed.  This first phase gives us an ability to store and apply views.

Screenshot 2024-02-19 at 4 41 58 PM

Saved views are persisted to browser local storage.  You can view the storage by "inspecting the page" (F12 on some browsers), and going to the Application tab:

Screenshot 2024-02-14 at 4 35 05 PM


DennisPeriquet:
close for now because I'm making headway on save/delete

DennisPeriquet:
/hold

Before merge, I'd like folks to think about and give feedback on the pre-configured views I added.

sippy-ng/src/component_readiness/CompReadyVars.js
```diff
@@ -240,6 +264,141 @@ export const CompReadyVarsProvider = ({ children }) => {
...
+        Confidence: defaultConfidenceParam,
+        Pity: defaultPityParam,
+        'Min Fail': defaultMinFailParam,
+        'Ignore Missing': defaultIgnoreMissingParam,
+        'Ignore Disruption': defaultIgnoreDisruptionParam,
+      },
+    },
+    Strict: {
+      config: {
+        help: 'Default view except Pity and MinFail set to 0',
```
stbenjam, line 286: This is probably not to useful IMHO, you'll trigger statistically significant results on pass rate changes on the order of 0.01%. It'll turn most of the board red.

sippy-ng/src/component_readiness/SavedViews.js
```diff
@@ -0,0 +1,248 @@
...
+        onClick={handleOpenSaveDialog}
+      >
+        Save Current View
+      </Button>
+
+      <Dialog open={saveDialogIsOpen} onClose={handleCloseSaveDialog}>
+        <DialogTitle>Save Current View</DialogTitle>
+        <DialogContent>
+          <DialogContentText>
+            Name for new view (max {maxSavedViewLength} characters)
```
stbenjam, line 182: Why is the max 15 characters?

Can you also add some text explaining this is in the local browser storage only.

It might be nice if there was some kind of share button that let the user save the view in their own storage, but not required.

stbenjam:
If I go to a page like this: http://localhost:3000/sippy-ng/component_readiness/env_capabilities?arch=amd64&baseEndTime=2024-02-28%2023%3A59%3A59&baseRelease=4.15&baseStartTime=2024-02-01%2000%3A00%3A00&component=Cluster%20Version%20Operator&confidence=95&environment=ovn%20amd64%20gcp&excludeArches=arm64%2Cheterogeneous%2Cppc64le%2Cs390x&excludeClouds=openstack%2Cibmcloud%2Clibvirt%2Covirt%2Cunknown&excludeVariants=hypershift%2Cosd%2Cmicroshift%2Ctechpreview%2Csingle-node%2Cassisted%2Ccompact&groupBy=cloud%2Carch%2Cnetwork&ignoreDisruption=true&ignoreMissing=false&minFail=3&network=ovn&pity=5&platform=gcp&sampleEndTime=2024-03-07%2023%3A59%3A59&sampleRelease=4.16&sampleStartTime=2024-03-01%2000%3A00%3A00

And then try to select a view like Heterogeneous - it breaks the page.

DennisPeriquet:
> If I go to a page like this: http://localhost:3000/sippy-ng/component_readiness/env_capabilities?arch=amd64&baseEndTime=2024-02-28%2023%3A59%3A59&baseRelease=4.15&baseStartTime=2024-02-01%2000%3A00%3A00&component=Cluster%20Version%20Operator&confidence=95&environment=ovn%20amd64%20gcp&excludeArches=arm64%2Cheterogeneous%2Cppc64le%2Cs390x&excludeClouds=openstack%2Cibmcloud%2Clibvirt%2Covirt%2Cunknown&excludeVariants=hypershift%2Cosd%2Cmicroshift%2Ctechpreview%2Csingle-node%2Cassisted%2Ccompact&groupBy=cloud%2Carch%2Cnetwork&ignoreDisruption=true&ignoreMissing=false&minFail=3&network=ovn&pity=5&platform=gcp&sampleEndTime=2024-03-07%2023%3A59%3A59&sampleRelease=4.16&sampleStartTime=2024-03-01%2000%3A00%3A00
> 
> And then try to select a view like Heterogeneous - it breaks the page.

Good find.  That was an interesting one; the test for empty table has been updated.

sippy-ng/src/component_readiness/SavedViews.js
DennisPeriquet: I originally set it to 15 thinking it would make the left sidebar too wide.  Turns out it renders/wraps ok for longer names so I made the max 100 characters.

For that last part, do you mean save the views to the user's disk?

dgoodwin:
Does saved views using browser local storage imply that these are per user, you create your desired view, but nobody else will see it? 

Is there a mechanism for us to define views centrally that anyone can see in that dropdown and use?

DennisPeriquet:
> Does saved views using browser local storage imply that these are per user, you create your desired view, but nobody else will see it?

Yes.

> 
> Is there a mechanism for us to define views centrally that anyone can see in that dropdown and use?

We define the globally visible views in [CompReadyVars.js](https://github.com/openshift/sippy/pull/1483/files#diff-7f9164f6776cee7e39226f6d4e0d4501b09d2cae48ba2c7283903dd7c2cbd56eR267-R358). If anyone wanted a new view, they'd have to make a PR.

sippy-ng/src/component_readiness/CompReadyVars.js
```diff
@@ -240,6 +264,127 @@ export const CompReadyVarsProvider = ({ children }) => {
...
+        'Exclude Upgrades': defaultExcludeUpgradesCheckedItems,
+        'Exclude Variants': defaultExcludeVariantsCheckedItems,
+        Confidence: defaultConfidenceParam,
+        Pity: defaultPityParam,
+        'Min Fail': defaultMinFailParam,
+        'Ignore Missing': defaultIgnoreMissingParam,
+        'Ignore Disruption': defaultIgnoreDisruptionParam,
+      },
+    },
+    AWS_GCP_Azure: {
```
stbenjam, line 332: I think you can drop this one

sippy-ng/src/component_readiness/CompReadyVars.js
```diff
@@ -240,6 +264,127 @@ export const CompReadyVarsProvider = ({ children }) => {
...
+        'Min Fail': defaultMinFailParam,
+        'Ignore Missing': defaultIgnoreMissingParam,
+        'Ignore Disruption': defaultIgnoreDisruptionParam,
+      },
+    },
+    ARM64: {
+      config: {
+        help: 'Default view but only for amd64',
+        'Group By': defaultGroupByCheckedItems,
+        'Exclude Arches': ['amd64', 'ppc64le', 's390x', 'heterogeneous'],
```
stbenjam, line 304: Hm, it'll probably be better once we've got server-side views, since we'll be able to do some kind of "include" version of the filter.  If we add a new arch now then it'll show up in this view along with ARM since filters are excluded.

sippy-ng/src/component_readiness/CompReadyVars.js
```diff
@@ -240,6 +264,101 @@ export const CompReadyVarsProvider = ({ children }) => {
...
+    // If there is data in the local storage and it has keys, then merge it with the
+    // initial views.  Otherwise, just use the initial views.
+    const savedViews = localStorage.getItem('savedViews')
+    if (savedViews && Object.keys(JSON.parse(savedViews)).length > 0) {
+      const mergedViews = { ...initialViews, ...JSON.parse(savedViews) }
+      return mergedViews
+    } else {
+      return initialViews
+    }
+  }
```
dgoodwin, line 358: Was just using bookmarks discussed? Why would we save to local storage when we could fully qualify a URL? I forsee problems with what people have locally in the future, whereas bookmarks gives us a path to reporting a problem with a URL, that might be more tricky if we hit some bug due to local storage.

DennisPeriquet: ah, bookmark is a browser bookmark and not the bookmark widget in the sippy UI.

DennisPeriquet:
we'll close this for now but let's keep the branch around as we'll still need some kind of UI to allow users to pick a server side view; we will back off on user saved views since they may become problematic in the future.

PR #1515, opened 2024-02-27
TRT-1437: Automate alert update when new releases are introduced

xueqzhan:
This PR adds a mechanism to silence an alert for older releases. It uses a map to map a metric to a count. The count defines how many recent releases for which the alert should be generated for that metric. The metric for releases older than the count will be label as silenced. 

Alert definitions will be updated once this merges. We will remove specific releases in the alert definition. Instead, the current release definitions are translated into counts in this PR for any specific metrics. 

As new releases are introduced, a rollover will happen so that older releases will fall off the sliding window and have their alerts silenced automatically.

EDIT:
The above implementation is changed to add just "releaseStatus" label to all the metrics with label "release". The value of releaseStatus is obtained from the bigquery Releases table. This will leave the alert definition to filter by interested releaseStatus.

### DIFF
--- a/pkg/sippyserver/metrics/metrics.go
+++ b/pkg/sippyserver/metrics/metrics.go
@@ -31,15 +31,38 @@ const (
 	blockerScoreToAlertOn = 70
 )
 
+const (
+	jobPassRatioMetricName                  = "sippy_job_pass_ratio" // #nosec G101
+	disruptionVsPrevGAMetricName            = "sippy_disruption_vs_prev_ga"
+	payloadHoursSinceLastAcceptedMetricName = "sippy_payloads_hours_since_last_accepted"
+	payloadConsecutiveRejectionsMetricName  = "sippy_payloads_consecutively_rejected"
+	payloadPossibleTestBlockersMetricName   = "sippy_payloads_possible_test_blockers"
+	installSuccessMetricName                = "sippy_install_success_last"
+	upgradeSuccessMetricName                = "sippy_upgrade_success_last"
+	installSuccessDeltaToPrevWeekMetricName = "sippy_install_success_delta_last"
+	upgradeSuccessDeltaToPrevWeekMetricName = "sippy_upgrade_success_delta_last"
+	payloadHoursSinceLastOSUpgradeName      = "sippy_payloads_hours_since_last_os_upgrade"
+)
+
+const (
+	silencedTrue  = "true"
+	silencedFalse = "false"
+)
+
+const (
+	releasePresubmits = "Presubmits"
+	releaseStatusEOL  = "End of life"
+)
+
 var (
 	buildClusterHealthMetric = promauto.NewGaugeVec(prometheus.GaugeOpts{
 		Name: "sippy_build_cluster_pass_ratio",
 		Help: "Ratio of passed job runs for a build cluster in a period (2 day, 7 day, etc)",
 	}, []string{"cluster", "period"})
 	jobPassRatioMetric = promauto.NewGaugeVec(prometheus.GaugeOpts{
-		Name: "sippy_job_pass_ratio",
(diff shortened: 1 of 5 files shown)


neisw:
I suppose I'm curious about the sliding window approach.  For the TRT monitoring alerts like `payloadHoursSinceLastAcceptedMetricName` we will still have to toggle to 2 at branch cut then back to 1 at GA I presume.  And for the periodic alerts we haven't automatically been removing older releases.  Not saying this is a bad idea, just curious what folks think pros / cons.

neisw:
Can we just check for missing GA date for things like `payloadHoursSinceLastAcceptedMetricName` then when we set that date it stops alerting.

For the periodics if you do a start release version then you probably don't need to update since the frequency of those jobs roll off over time.

If you start with something for 4.15 now and set 2 then at 4.17 you have to set 3, etc. until you get the full cycle of releases, maybe 5, that you want to support.

dgoodwin:
Ken and I just talked through an idea to decouple the decision of what is silenced from sippy itself. Goal is to allow three scenarios:

- TRT wants to alert on unreleased, in devel releases. (such as 4.15 (prior to yesterday) and 4.16).
- Staff eng wants to alert on the past 3 GA'd releases.
- Component team wants to alert on their jobs for every release since 4.12.

To allow all of this to happen in the alerts definitions, without requiring anyone to ever need to update sippy as well, we are thinking of adding 4 new labels, naming is up for discussion:

- devel or ga = true|false - this would be so TRT could monitor for any branches that are not yet GA.
- ga_offset - an int representing how far back this ga'd release is from the latest. 4.15 would have ga_offset=0, as that just ga'd. 4.14 would be ga_offset=1, and so forth. Then alerts could alert on the past X GA releases with ga_offset < X.
- release_major=4 and release_minor=16, alerts could then say fire on my job (with a regex for the job name) for every release since 4.12, I believe some alerts already do this by continuously adding to the 4.12|4.13|4.14 ...

stbenjam:
> devel or ga = true|false - this would be so TRT could monitor for any branches that are not yet GA.

I would suggest not using a boolean but using a string value in case you need a third value (inevitably when I make something a boolean, I very soon need a third value)

stbenjam:
> Ah darn, Stephen's comment jogged my memory, the whole premise may be flawed, all prom label values are strings, the numeric comparisons may not work at all.

What if you had a "release_owner" label?  Sippy knows enough to make it trt, staff-engineer, eol, etc depending whether it's GA'd, or how old it is.  Then one label replaces devel and ga_offset

dgoodwin:
Yeah that could work, perhaps state="devel|ga|eol" so we don't have to encode who owners are in sippy. eol could be reflected in the Releases table I think.

xueqzhan:
I updated the description to reflect the new change.

pkg/sippyserver/metrics/metrics.go
```diff
@@ -73,33 +96,58 @@ var (
...
+		return releases, nil
+	}
+	return query.ReleasesFromDB(dbc)
+}
+
+func getReleaseStatus(releases []query.Release, release string) string {
+	releaseStatus := releaseStatusEOL
+	for _, r := range releases {
+		if r.Release == release && len(r.Status) != 0 {
+			releaseStatus = r.Status
```
DennisPeriquet, line 136: minor and optional suggestion: it would be nice to do a `break` here since you found the release.  And later, you can tell if the release was not found which is probably an erroneous condition.  But if we can guarantee the new `Releases` table will always contain the release, maybe it doesn't matter.

But, if you account for this by defaulting to `releaseStatusEOL` in line 133, if the release is not found, maybe it's fine.

DennisPeriquet (commented):
minor comment

PR #1531, opened 2024-03-07
TRT-1304: Job Variant Registry

dgoodwin:
This adds the first revision of the job variants registry.

A load-ocp-job-variants command is added for our specific job variant calculations. The output is a json file containing the mapping of job names to variants, you can see the current results [here](https://gist.github.com/dgoodwin/ae189286e338b17dfdc3736c7eb37f71) for review of how variants are being calculated. The output file is intended to be a mechanism third party users of sippy can use to hook into use of the next command that was added, they just need to generate a json file mapping their job names to variants via whatever program/script they wish.

```
$ sippy load-ocp-job-variants --google-service-account-credential-file ~/creds/creds.json
```

A sync-job-variants command is added to reconciling the json file of desired/expected state out to bigquery. Inserts are fast, deleting a job that disappeared is sort of slow, and updating a variant on lots of jobs is very slow. (2-3s per query) We expect about 3600 jobs if looking for just those that ran in the last 3 months.

```
$ sippy sync-job-variants --google-service-account-credential-file ~/creds/creds.json ocp-expected-job-variants.json --bigquery-project openshift-ci-data-analysis
```

There is support for loading a cluster-data.json file and comparing values against those we calculated by the job name, with one or the other side winning depending on what variant we're talking about. By default, the job name calculation wins.

Next steps will be to:
- auto-generate and manage a view where the variants become columns. (per design document)
- get component readiness using these variants rather than those in the junit table


stbenjam:
```
{
  "aggregator-periodic-ci-openshift-hypershift-release-4.16-periodics-e2e-aws-ovn-conformance": {
    "Architecture": "amd64",
    "FeatureSet": "default",
    "FromRelease": "4.16",
    "FromReleaseMajor": "4",
    "FromReleaseMinor": "16",
    "Installer": "ipi",
    "Network": "ovn",
    "NetworkAccess": "default",
    "Owner": "eng",
    "Platform": "aws",
    "Release": "4.16",
    "ReleaseMajor": "4",
    "ReleaseMinor": "16",
    "Scheduler": "default",
    "SecurityMode": "default",
    "Suite": "unknown",
    "Topology": "external",
    "Upgrade": "none"
  },
```

Is there any `aggregated` variant planned?

cmd/sippy/loadocpvariants.go
```diff
@@ -0,0 +1,93 @@
...
+	f.BigQueryFlags.BindFlags(fs)
+	f.GoogleCloudFlags.BindFlags(fs)
+	fs.StringVar(&f.OutputFile, "o", "ocp-expected-job-variants.json", "Output json file for job variant data")
+}
+
+func NewLoadOCPJobVariantsCommand() *cobra.Command {
+	f := NewLoadJobVariantsFlags()
+
+	cmd := &cobra.Command{
+		Use:   "load-ocp-job-variants",
```
stbenjam, line 43: Does this command have to be OCP specific? Could it be `load-job-variants` and take a mode flag? Similar to how the variant registry works https://github.com/openshift/sippy/blob/master/pkg/testidentification/variants.go#L9

Future us would probably appreciate leaving the design more open if we productize this part

pkg/variantregistry/ocp.go
```diff
@@ -0,0 +1,448 @@
...
+
+const (
+	VariantAggregation   = "Aggregation" // aggregated or none
+	VariantArch          = "Architecture"
+	VariantFeatureSet    = "FeatureSet" // techpreview / standard
+	VariantInstaller     = "Installer"  // ipi / upi / assisted
+	VariantNetwork       = "Network"
+	VariantNetworkAccess = "NetworkAccess" // disconnected / proxy / standard
+	// TODO
+	VariantNetworkStack     = "NetworkStack" // ipv4 / ipv6 / dual
```
stbenjam, line 228: I don't see dualstack jobs being identified https://gist.github.com/dgoodwin/ae189286e338b17dfdc3736c7eb37f71#file-gistfile1-txt-L8621

pkg/variantregistry/ocp.go
```diff
@@ -0,0 +1,448 @@
...
+	start := time.Now()
+
+	// TODO: pull presubmits for sippy as well
+
+	// For the primary list of all job names, we will query everything that's run in the last 3 months:
+	// TODO: for component readiness queries to work in the past, we may need far more than jobs that ran in 3 months
+	// since start of 4.14 perhaps?
+	// TODO: unfortunate use of another project here
+	query := v.BigQueryClient.Query(`SELECT prowjob_job_name, MAX(prowjob_url) AS prowjob_url, MAX(prowjob_build_id) AS prowjob_build_id FROM ` +
+		"`openshift-gce-devel.ci_analysis_us.jobs` " +
```
stbenjam, line 62: Maybe this could be a flag? Like "jobs-table"?

pkg/variantregistry/ocp.go
```diff
@@ -0,0 +1,448 @@
...
+				"variant":  k,
+				"fromJob":  jnv,
+				"fromFile": v,
+			})
+
+			switch k {
+			case VariantArch:
+				// Job name identification wins for arch, heterogenous jobs can show cluster data with
+				// amd64 as it's read from a single node.
+				jLog.Infof("variant mismatch: using %s from job name", k)
```
stbenjam, line 169:
```suggestion
				jLog.Warningf("variant mismatch: using %s from job name", k)
```

pkg/variantregistry/ocp.go
```diff
@@ -0,0 +1,448 @@
...
+				continue
+			default:
+				jLog.Infof("variant mismatch: using %s from job run variants file", k)
```
stbenjam, line 172:
```suggestion
				jLog.Warningf("variant mismatch: using %s from job run variants file", k)
```

pkg/variantregistry/ocp.go
```diff
@@ -0,0 +1,448 @@
...
+
+type prowJobLastRun struct {
+	JobName  string              `bigquery:"prowjob_job_name"`
+	JobRunID string              `bigquery:"prowjob_build_id"`
+	URL      bigquery.NullString `bigquery:"prowjob_url"`
+}
+
+// LoadExpectedJobVariants queries all known jobs from the gce-devel "jobs" table (actually contains job runs).
+// This effectively is every job that actually ran in the last several years.
+func (v *OCPVariantLoader) LoadExpectedJobVariants(ctx context.Context) (map[string]map[string]string, error) {
```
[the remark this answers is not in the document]
dgoodwin: This may cause problems for our planned view representing these variants as columns for integration and ease of use, I don't think that's going to work with multi-valued variants, and would require a more involved integration into the three apps using this thing. Maybe that's worthwhile but we need to make sure.

My hope was that if it's important enough we need to slice on it, but not important enough to get a variant on all jobs ,that it would collapse into a pre-existing variant with hyphenation, i.e. virtualmedia could be something like: Installer=ipi-virtualmedia. 

Need to think on this and maybe discuss, I'm not sure which way to go.

cmd/sippy/loadocpvariants.go
dgoodwin: My intention was to flag it clearly as ocp specific so third parties get a clear signal this is not for them. The intention was that the hook point is the file this generates, a non-ocp sippy user can do whatever they like to generate that file, and then the second command is reusable to get it all synced to bigquery on any cadence they wish.

QE might have very different rules, we probably would take that code in this repo, so perhaps --mode is a better idea. I'll get it switched.

pkg/variantregistry/ocp.go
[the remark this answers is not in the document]
dgoodwin: Team discussed, we think there's a good chance we can get by with the current implementation with a few levers. Some of the desired "tags" we discussed fit into existing variants such as suite, others are probably worthy of getting a variant on all jobs, and those that alter an existing variant significantly we can hyphenate. We'll see how it goes.

pkg/variantregistry/ocp_test.go
```diff
@@ -0,0 +1,333 @@
...
+			job: "periodic-ci-openshift-hypershift-release-4.16-periodics-e2e-aws-ovn-conformance",
+			expected: map[string]string{
+				VariantRelease:          "4.16",
+				VariantFromRelease:      "4.16",
+				VariantReleaseMajor:     "4",
+				VariantReleaseMinor:     "16",
+				VariantFromReleaseMajor: "4",
+				VariantFromReleaseMinor: "16",
+				VariantArch:             "amd64",
+				VariantInstaller:        "ipi", // Is this ok for hypershift?
```
stbenjam, line 56: Is it IPI? I don't think it uses the openshift-installer does it?

I'm not sure how to show hypershift, but we probably do need to differentiate it somehow.  For example, we'll probably want to show hypershift on aws and hypershift on metal. Not sure if that's a case for hyphenation (hypershift-aws) or a new category

dgoodwin: I'll find out, iirc they use a custom installer binary.

dgoodwin: With regards to platform I think it would be ControlPlane=external, Platform=aws

stbenjam: Does that always mean hypershift? I'm just trying to imagine how we would show a view in component readiness where we could look at just hypershift for example

dgoodwin: ControlPlane external means hypershift i think yes, but per the installer comment we'll also have Installer=Hypershift.

dgoodwin:
I think I got everything, only thing I'm pondering now is possibly future proofing that file format a little, rather than a fix list of maps. I'll do that soon and update the gist.

stbenjam:
@dgoodwin Lint needs fixing

cmd/sippy/syncvariants.go
```diff
@@ -0,0 +1,89 @@
...
+		GoogleCloudFlags: flags.NewGoogleCloudFlags(),
+	}
+}
+
+func (f *SyncJobVariantsFlags) BindFlags(fs *pflag.FlagSet) {
+	f.BigQueryFlags.BindFlags(fs)
+	f.GoogleCloudFlags.BindFlags(fs)
+}
+
+func NewSyncJobVariantsCommand() *cobra.Command {
```
stbenjam, line 37: One thing that crossed my mind is that this is basically a loader, I wonder if the variant registry shouldn't just be a dataloader in pkg/dataloader.

Then there'd be no need for another command, you'd just run `./sippy load --loader variants`

I wouldn't hold up merging on this, but you do get some benefits like this gets merged up to metrics on grafana, and you get plotting of how long it takes, error counts, etc.

dgoodwin: So I have two new commands in this pr, load variants, which generates a json file, and sync variants, which syncs it to the db and takes an input filename param. I think sync is the better fit for load command, but input file is an issue. Two commands as the latter was intended for third party users, whereas the current "load variants" command was just for our logic, and maybe QEs.

I think I'll rename my load-job-variants to generate-job-variants to output the json file, and then I can move sync over to be a loader, and that will use an assumed filename? I can't really add a --input param as load takes multiple loaders and we wouldn't know which it was intended for. Seem reasonable?

pkg/variantregistry/syncer.go
```diff
@@ -0,0 +1,230 @@
...
+		if _, ok := expectedVariants[currJobName]; !ok {
+			deleteJobs = append(deleteJobs, currJobName)
+		}
+	}
+
+	return insertVariants, updateVariants, deleteVariants, deleteJobs
+}
+
+func loadCurrentJobVariants(bqClient *bigquery.Client) (map[string]map[string]string, error) {
+	query := bqClient.Query(`SELECT * FROM openshift-ci-data-analysis.sippy.JobVariants ORDER BY JobName, VariantName`)
```
dgoodwin, line 139: @stbenjam  What did we decide on that meeting I missed the end of, sippy or ci-data-analysis or gce-devel.something?

cmd/sippy/syncvariants.go
stbenjam: >  then I can move sync over to be a loader, and that will use an assumed filename? I can't really add a --input param as load takes multiple loaders and we wouldn't know which it was intended for. Seem reasonable?

Ah I didn't think about that - it's probably OK to keep this as you have it then. If you want the data loader interface, I don't think it's a problem to add a parameter that's only used for one loader.  Not all loaders use all all arguments

dgoodwin:
openshift-gce-devel.ci_analysis_us.job_variants is live with the full 24k jobs and their variants.

xueqzhan:
/lgtm

@dgoodwin This looks great to me. I am looking forward to using it in the apps.

dgoodwin:
Updated for lint fixes both in the new code and some new things being picked up in the project, probably from some new version of something in lint.

Issue #1537, opened 2024-03-13
Missing the dep info when running make


jianzhangbjz:
As follows, 
```console
jiazha-mac:sippy jiazha$ make
Makefile:5: *** "Missing  in PATH".  Stop.
```

PR #1538, opened 2024-03-13
display the right dep info when running make

jianzhangbjz:
Address #1537, after this fix, 
```console
jiazha-mac:sippy jiazha$ make
Makefile:5: *** "Missing npm in PATH".  Stop.
```


stbenjam:
This is great, thanks for the fix

/lgtm

PR #1540, opened 2024-03-13
TRT-1484: Migrate bugloader to use bigquery data sources

stbenjam:
This migrates from using search.ci to fetching bug data from the
bigquery tables.  We link bugs to tests by joining on the component
mapping table, which is our canonical source of test names.  We link
jobs to tests by using the jobs table from bigquery.

Linking variants is deferred, as there's no single bug that currently
uses this feature, but it could be restored by using the new variant
registry.

Todo's:

- [X] Link tests to bugs
- [X] Link jobs to bugs
- [x] Missing data we have today needs to go to bigquery (target/affect version info, bug components)
- [x] ~Support variant linking~ Deferred after discussing with team


pkg/dataloader/bugloader/bugloader_test.go
stbenjam, line 1: Just for local testing for now

PR #1557, opened 2024-03-22
TRT-1404: Port Sippy Intervals UI to Use Structured Data and Various Improvements

dgoodwin:
The move to structured intervals is underway, the fields in json remain tempStructuredLocator and tempStructuredMessage. Sippy now uses these fields instead of the legacy locator/message which will soon be removed, and the structured version will be moved to those names. When that happens, I will update this code in the sippy API to detect both and massage the data so old jobs can still be loaded fine.

In addition:
- we now load a specific intervals file, defaulting to the first spyglass intervals file we find.
- a dropdown allows loading other files.

More changes coming.

### DIFF
--- a/sippy-ng/src/prow_job_runs/ProwJobRun.js
+++ b/sippy-ng/src/prow_job_runs/ProwJobRun.js
@@ -358,20 +358,6 @@ function filterIntervals(
 function mutateIntervals(eventIntervals) {
   // Structure the locator data and then categorize the event
   lodash.forEach(eventIntervals, function (eventInterval) {
-    // break the locator apart into an object for better filtering
-    eventInterval.locatorObj = {}
-    if (eventInterval.locator.startsWith('e2e-test/')) {
-      eventInterval.locatorObj.e2e_test = eventInterval.locator.slice(
-        eventInterval.locator.indexOf('/') + 1
-      )
-    } else {
-      let locatorChunks = eventInterval.locator.split(' ')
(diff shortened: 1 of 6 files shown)


sippy-ng/src/prow_job_runs/ProwJobRun.js
```diff
@@ -547,151 +556,101 @@ function groupIntervals(filteredIntervals) {
...
 function isPodLog(eventInterval) {
-  if (isEtcdLeadership(eventInterval)) {
-    return false
-  }
-  if (eventInterval.locator.includes('src/podLog')) {
+  if (eventInterval.tempSource === 'PodLog') {
     return true
   }
-  return false
+  return eventInterval.tempSource === 'EtcdLog'
```
DennisPeriquet, line 595: Just a sanity check: this says that you want to include etcdLog intervals in the `pod-logs` section.  Is that what you want?

dgoodwin: yes that was intended for now. If we get to the end game every Source gets it's own section, so that would change then I think.

DennisPeriquet:
just one small observation: I get the errors below on the dev console when reloading the intervals page.  If I click one of the squares (e.g., NodeState), it goes away.  Seems like something didn't get initialized. This can be fixed in another PR.

Screenshot 2024-04-05 at 8 55 50 AM

PR #1560, opened 2024-03-25
TRT-1555: Scripted Incident Attribution

neisw:
Need to finalize the location of the triaged_incidents table and update.

- updates gen-resolved-issue.py to write records to bigquery table
- updates component_report to read resolved / triaged records from bigquery
- separates sample and base queries and caches results independently

### DIFF
--- a/pkg/api/component_report.go
+++ b/pkg/api/component_report.go
@@ -24,17 +24,9 @@ import (
 	"github.com/openshift/sippy/pkg/util/sets"
 )
 
-// move this out
-const TriagedIncidentsProjectID = "openshift-ci-data-analysis"
-const TriagedIncidentsDataSetID = "fsbabcock_test"
-const TriagedIncidentsTableID = "api_resolved_incidents_b"
-
-// long term moves to
-// const TriagedIncidentsProjectID="openshift-gce-devel"
-// const TriagedIncidentsDataSetID="ci_analysis_us"
-// const TriagedIncidentsTableID="triaged_incidents"
-
 const (
+	triagedIncidentsTableID = "triaged_incidents"
+
 	ignoredJobsRegexp = `-okd|-recovery|aggregator-|alibaba|-disruptive|-rollback|-out-of-change|-sno-fips-recert`
 
 	// This query de-dupes the test results. There are multiple issues present in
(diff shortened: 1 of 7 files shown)


dgoodwin (commented):
Apologies I got pulled away before I could fully finish, but here's a little feedback.

pkg/api/component_report.go
```diff
@@ -64,6 +74,19 @@ const (
...
+
+var (
+	ComponentReport        GeneratorType = "ComponentReport"
+	SampleTestStatus       GeneratorType = "SampleTestStatus"
+	BaseTestStatus         GeneratorType = "BaseTestStatus"
+	SampleJobRunTestStatus GeneratorType = "SampleJobRunTestStatus"
+	BaseJobRunTestStatus   GeneratorType = "BaseJobRunTestStatus"
+	TriagedIncidents       GeneratorType = "TriagedIncidents"
+	TestDetailsReport      GeneratorType = "TestDetailsReport"
+	TestVariants           GeneratorType = "TestVariants"
```
stbenjam, line 87: I'm not a huge fan of this.  This was the thing I mentioned I was thinking about.  I can see why the GeneratorType is useful to store, but it's really more a product of some of the downsides of the current implementation.

I wish each of the generators was a different type instead of overloading componentReportGenerator to do many different things. Then each one could implement a new Generator interface, which could simplify the caching layer (anything implementing Generator could be cached, without having to pass so much to  getDataFromCacheOrGenerate)

And that would facilitate getting the generators out of the API layer entirely.  We've coupled the API and data model layers a bit too tightly IMHO.

stbenjam (commented):
A couple of comments, I think you can move this out of draft and run it through CI.

pkg/api/component_report.go
neisw: I'm open to working through this with you either in this PR or a latter one.  I have started to create new generators for the base / sample query and triaged incidents.  Would you be open to the Generator having an interface that provided the cache key which it could optionally add a prefix to guarantee uniqueness for it's generator type?

stbenjam: > Would you be open to the Generator having an interface that provided the cache key which it could optionally add a prefix to guarantee uniqueness for it's generator type?

Yea, exactly - I was thinking the interface impl would manage it's cache key.

PR #1569, opened 2024-03-27
Revert "TRT-1484: Migrate bugloader to use bigquery data sources"

stbenjam:
Reverts openshift/sippy#1540

Affects versions and last changed are coming up blank in sippy production even though the table has the values.  They were populated when I tested locally.   Maybe I broke something on a late change to the PR.  Go back to the old bug loader while I sort this out.


neisw:
/lgtm
/hold
release when ready

PR #1572, opened 2024-03-28
Fix install success metrics releaseStatus label

xueqzhan:


xueqzhan:
/hold 

Let me go through all metrics to make sure they are fine.

PR #1584, opened 2024-04-03
Revert "TRT-1588: Avoid NULL for prowjob_start_ts"

stbenjam:
Reverts openshift/sippy#1581

Sippy not syncing:

```
time="2024-04-03T19:28:20.509Z" level=error msg="loader \"prow\" returned error: googleapi: Error 400: Unrecognized name: prowjob_start_ts; Did you mean prowjob_start? at [16:20], invalidQuery"
971
Error: errors were encountered while loading database, see logs for details
```


pkg/dataloader/prowloader/bigqueryjobs.go
```diff
@@ -52,7 +52,6 @@ func (pl *ProwLoader) fetchProwJobsFromOpenShiftBigQuery() ([]prow.ProwJob, []er
 		"FROM `ci_analysis_us.jobs` " +
 		`WHERE TIMESTAMP(prowjob_completion) > @queryFrom
 	       AND prowjob_url IS NOT NULL
-	       AND prowjob_start_ts IS NOT NULL
```
stbenjam, line 55: The column is actually called `prowjob_start`, we cast it to a timestamp above

PR #1585, opened 2024-04-03
fix wrong name in SELECT ... AND clause

DennisPeriquet:
fix oversight on the AND clause


DennisPeriquet:
I can put on top of the revert if it's more clean

PR #1589, opened 2024-04-03
Trt 1555 redux

neisw:
Reverts the revert and fixes the wrong common query being loaded for getJobRunTestStatusFromBigQuery:

https://github.com/openshift/sippy/commit/c32e1fcf0bfab7b2c7254a468509d6c1aa464c8a


stbenjam:
/retitle TRT-1555 Scripted Incident Attribution redux 
/lgtm
/hold

Let's merge in the morning

PR #1597, opened 2024-04-05
Remove bparees from approvers

bparees:
Been a long time since it's made sense for me to be an approver here.  Sippy has been in good hands for quite a while.


dgoodwin:
/lgtm

yes-sir-yes-boss

PR #1606, opened 2024-04-09
Just test sippy e2e in CI

DennisPeriquet:
This is just a test


DennisPeriquet:
there are other PRs that can test e2e so we don't need this one.

PR #1607, opened 2024-04-11
Include total/success/flake counts in the regressed tests

dgoodwin:
The main API for component readiness did not include the current counts
for the regressed test pass rates, only a status indicating how severe
of a regression was detected.

This PR attempts to add the sample counts to every regressed test in the
response.


pkg/apis/api/types.go
```diff
@@ -914,7 +914,10 @@ type ComponentReportTestIdentification struct {
 
 type ComponentReportTestSummary struct {
 	ComponentReportTestIdentification
-	Status ComponentReportStatus `json:"status"`
+	Status       ComponentReportStatus `json:"status"`
+	TotalCount   int                   `json:"total_count" bigquery:"total_count"`
```
neisw, line 918: Are the bigquery notations needed?

pkg/api/component_report.go
```diff
@@ -1382,12 +1393,12 @@ func (c *componentReportGenerator) generateComponentTestReport(baseStatus map[ap
 		delete(sampleStatus, testIdentification)
 
 		rowIdentifications, columnIdentifications := c.getRowColumnIdentifications(testIdentification, baseStats)
-		updateCellStatus(rowIdentifications, columnIdentifications, testID, reportStatus, aggregatedStatus, allRows, allColumns)
+		updateCellStatus(rowIdentifications, columnIdentifications, testID, reportStatus, sampleStats, aggregatedStatus, allRows, allColumns)
```
neisw, line 1396: The report status gets modified by adjusting the sample stats in assessComponentStatus.  Those adjustments won't be reflected here but I think that will be ok if we are ok with my comment @ line 1018

pkg/api/component_report.go
```diff
@@ -1007,7 +1010,9 @@ type cellStatus struct {
 }
 
 func getNewCellStatus(testID apitype.ComponentReportTestIdentification,
-	reportStatus apitype.ComponentReportStatus, existingCellStatus *cellStatus) cellStatus {
+	reportStatus apitype.ComponentReportStatus, sampleStats apitype.ComponentTestStatus,
+	existingCellStatus *cellStatus) cellStatus {
+
 	var newCellStatus cellStatus
 	if existingCellStatus != nil {
 		if (reportStatus < apitype.NotSignificant && reportStatus < existingCellStatus.status) ||
```
neisw, line 1018: With the current logic to change reportStatus based on tiraged incidents we won't add regressed tests if the reportStatus is modified due to those triaged incidents.  This includes the move to using the triagedregression statuses.  Presume we are ok here, but mentioning.

pkg/api/component_report.go
xueqzhan: Good catch. We might need to add the compensation amount to the sampleStats struct to have proper numbers reflected in the regressedTests. 

While looking at the code, I wonder if there is a potential miscount involving the compensation calculation. It seems to me that the triaged compensation amount is based on jobRuns involved. But the sample stats is test based. That means, if a test appears multiple times in a job run, it is only counted once for compensation. This should be common in an upgrade job.

neisw: My other comment was that we won't be storing the sampleStatus unless the reportStatus is still considered a regression.  But granted if we made an adjustment and were still considered regressed that adjustment would not be reflected in the sampleStats.  Still think we are ok for our current needs and keeping the 'real' values may in fact be desirable.

Curious about your second comment.  We might need to walk through that more though I'm thinking potential exposure is low.  You are saying that the same testID could fail twice in a single job run and only be counted once for triage.  Might be something to review with Stephen, I know he has some deduping logic and it might cover this case so a test failure is only counted once per job run.  Not sure though.

pkg/apis/api/types.go
dgoodwin: No I suppose not, removing!

pkg/api/component_report.go
xueqzhan: If a triaged issue is associated with a job that have multiple potential regressions, do we count the compensation for all of them even if only one is being worked on? If so, the other regressions should still show up in the regressedTests and maybe we do like the current handling.

Let's sync in the office hour in case I am throwing out nonsense here. :)

neisw: Triaged Regressions are TestID specific.  So if there were 3 TestIDs that were being triaged there would be 3 entries (each with the list of jobruns that were considered triaged for that TestID).

What I don't know is if the same TestID would be reused for conformance & parallel.  I wouldn't think so but can't state definitively without more digging.

Sorry Devan, we hijacked your PR...

dgoodwin:
WIP commit pushed for what I'm facing. 

I've never logged a testID I can find in the json report output, and vice versa. Running the code now I'll see dupe tests in the result, then run it again and that test is gone and new ones appear multiple times.

pkg/api/component_report.go
```diff
@@ -1056,16 +1068,16 @@ func updateCellStatus(rowIdentifications []apitype.ComponentReportRowIdentificat
...
 			}
 		} else {
 			for _, columnIdentification := range columnIdentifications {
 				existing, ok := row[columnIdentification]
 				if !ok {
-					row[columnIdentification] = getNewCellStatus(testID, reportStatus, nil)
+					row[columnIdentification] = getNewCellStatus(testID, reportStatus, sampleStats, nil)
 				} else {
-					row[columnIdentification] = getNewCellStatus(testID, reportStatus, &existing)
+					row[columnIdentification] = getNewCellStatus(testID, reportStatus, sampleStats, &existing)
```
dgoodwin, line 1080: Traced to this line being called 8x for the same test ID + row id + column id.

pkg/api/component_report.go
```diff
@@ -1022,9 +1027,13 @@ func getNewCellStatus(testID apitype.ComponentReportTestIdentification,
 		newCellStatus.status = reportStatus
 	}
 	if reportStatus < apitype.MissingSample {
+		logger.Infof("adding regressed test")
```
dgoodwin, line 1030: This shows the 8 calls happening at various times:

❯ cat allout.txt| grep openshift-tests:fefc532f9a1e9928f7b22d188a7d4799  | grep "adding regressed" | grep Build
time="2024-04-12T13:52:44.351-03:00" level=info msg="adding regressed test" columnID="{ovn  amd64 vsphere }" row="{Build    }" testID="openshift-tests:fefc532f9a1e9928f7b22d188a7d4799"
time="2024-04-12T13:52:44.373-03:00" level=info msg="adding regressed test" columnID="{ovn  amd64 vsphere }" row="{Build    }" testID="openshift-tests:fefc532f9a1e9928f7b22d188a7d4799"
time="2024-04-12T13:52:44.601-03:00" level=info msg="adding regressed test" columnID="{ovn  amd64 vsphere }" row="{Build    }" testID="openshift-tests:fefc532f9a1e9928f7b22d188a7d4799"
time="2024-04-12T13:52:44.806-03:00" level=info msg="adding regressed test" columnID="{ovn  amd64 vsphere }" row="{Build    }" testID="openshift-tests:fefc532f9a1e9928f7b22d188a7d4799"
time="2024-04-12T13:52:44.815-03:00" level=info msg="adding regressed test" columnID="{ovn  amd64 vsphere }" row="{Build    }" testID="openshift-tests:fefc532f9a1e9928f7b22d188a7d4799"
time="2024-04-12T13:52:45.139-03:00" level=info msg="adding regressed test" columnID="{ovn  amd64 vsphere }" row="{Build    }" testID="openshift-tests:fefc532f9a1e9928f7b22d188a7d4799"
time="2024-04-12T13:52:45.177-03:00" level=info msg="adding regressed test" columnID="{ovn  amd64 vsphere }" row="{Build    }" testID="openshift-tests:fefc532f9a1e9928f7b22d188a7d4799"
time="2024-04-12T13:52:45.249-03:00" level=info msg="adding regressed test" columnID="{ovn  amd64 vsphere }" row="{Build    }" testID="openshift-tests:fefc532f9a1e9928f7b22d188a7d4799"

PR #1608, opened 2024-04-15
Fix a lurking bug with testID not being initialized in sampleStats loop

dgoodwin:


dgoodwin:
Credit to @xueqzhan for finding the fix after the bug was identified.

xueqzhan:
/lgtm

Thanks!

PR #1609, opened 2024-04-15
Trt 1594 regression icons

neisw:
Shows 'triaged' icons for regressions that are cleared via triaged incidents.

Screenshot showing the watch count triaged regression under Unknown

image

Small amount of cleanup for sampleQueryGenerator and baseQueryGenerator since we pulled in ComponentReportGenerator for cache key making the separate storage of SampleRelease and BaseRelease redundant.

### DIFF
--- a/pkg/api/component_report.go
+++ b/pkg/api/component_report.go
@@ -1017,7 +1017,9 @@ func getNewCellStatus(testID apitype.ComponentReportTestIdentification,
 	} else {
 		newCellStatus.status = reportStatus
 	}
-	if reportStatus < apitype.MissingSample {
+	// don't show triaged regressions in the regressed tests
+	// need a new UI to show active triaged incidents
+	if reportStatus < apitype.ExtremeTriagedRegression {
 		newCellStatus.regressedTests = append(newCellStatus.regressedTests, apitype.ComponentReportTestSummary{
 			ComponentReportTestIdentification: testID,
 			Status:                            reportStatus,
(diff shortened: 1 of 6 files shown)


pkg/api/component_report.go
```diff
@@ -1660,21 +1661,73 @@ func (c *componentReportGenerator) assessComponentStatus(sampleTotal, sampleSucc
...
 			if c.IgnoreMissing {
 				status = apitype.NotSignificant
 
 			} else {
 				status = apitype.MissingSample
 			}
 		} else {
-			if c.MinimumFailure != 0 && (sampleTotal-sampleSuccess-sampleFlake) < c.MinimumFailure {
-				return apitype.NotSignificant, fischerExact
+			// see if we had a significant regression prior to adjusting
```
xueqzhan, line 1673: I wonder if it is more readable if we create a common function to do the calculation. The function will be called twice, one for preAdjustment and one for postAdjustment, and then we can compare the result to decide on the report value.

neisw: I started out thinking I would.  But then each case seemed to have it's own prechecks and the common logic seemed to come down to

```
fischer.FisherExactTest(baseTotal-baseSuccess-baseFlake,
					baseSuccess+baseFlake,
					sampleTotal-sampleSuccess-sampleFlake,
					sampleSuccess+sampleFlake)
				significant = r < 1-float64(c.Confidence)/100
				fischerExact = r
```
Didn't seem worth the additional refactor but I could if you think I should.  Maybe more would fall out too...

PR #1614, opened 2024-04-17
TRT-1574: Server Side Views

dgoodwin:
Server now accepts a CLI param for `--component-readiness-views views.yaml`.

```yaml
---
- name: main
  variant_options:
    group_by: ''
    platform: ''
    upgrade: ''
    arch: ''
    network: ''
    variant: ''
  exclude_options:
    exclude_platforms: ''
    exclude_arches: ''
    exclude_networks: ''
    exclude_upgrades: ''
    exclude_variants: ''
  advance_options:
    minimum_failure: 0
    confidence: 0
    pity_factor: 0
    ignore_missing: false
    ignore_disruption: false
```

Adds a new /api/component_readiness/views API call to list these views, and their params. This is intended for future use in the UI to list params, and allow a user to stop using the view but tweak it's settings to get their own custom URL.

/api/component_readiness now accepts a ?view=main param. If specified, all the params above should conflict with the fact you specified a view. 

Thus far, start/end times are not part of the view, this will be up for discussion with the team. base and sample release are both still required, we may make base optional and assume to use sample-1 if sample is not GA yet, or sampleGA if it is.

### DIFF
--- a/cmd/sippy/component_readiness.go
+++ b/cmd/sippy/component_readiness.go
@@ -26,9 +26,10 @@ import (
 )
 
 type ComponentReadinessFlags struct {
-	GoogleCloudFlags *flags.GoogleCloudFlags
-	BigQueryFlags    *flags.BigQueryFlags
-	CacheFlags       *flags.CacheFlags
+	GoogleCloudFlags        *flags.GoogleCloudFlags
+	BigQueryFlags           *flags.BigQueryFlags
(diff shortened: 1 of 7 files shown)


dgoodwin:
/hold

We're working out how to use the variant registry in component readiness, which will impact the structure of these views. This can be reviewed but we'll hold off on merging until we get a final structure.

PR #1619, opened 2024-04-18
TRT-1304: Unrevert Job Variant Registry

dgoodwin:
Reverts openshift/sippy#1615


dgoodwin:
This should fix the error in fetchdata job, I thought it was a list of all possible loaders, it's actually the default set, and the fetchdata job runs with the default set.

PR #1628, opened 2024-04-22
WIP: Dynamic variants in Component Readiness

dgoodwin:
This is a PoC to explore how we could get component readiness working with dynamic variants instead of a fixed set as we do today.

The general approach outlined thus far:
- we control the query for component readiness in code, so programatically inject columns for each variant requested in the API call into the select, join, and group by portions of the query. No view is required for this, chatgpt revealed a way to do it with standard joins.
- switch to a more dynamic method of reading the column values back out of the bigquery results, as we don't have fixed columns anymore that will map to a static go struct.
- variants move to a map[string]string on the ComponentTestIdentification.
- because we use ComponentTestIdentification as a map key, and structs containing map fields cannot be used as map keys, switch to using a string representation (json serialized right now). Will need to watch computation cost but the implication of doing a few thousand serializations/deserializations of small json structs is likely insignificant.

### DIFF
--- a/pkg/api/component_report.go
+++ b/pkg/api/component_report.go
@@ -472,6 +472,21 @@ func (c *componentReportGenerator) getJobRunTestStatusFromBigQuery() (apitype.Co
 }
 
 func (c *componentReportGenerator) getCommonTestStatusQuery() (string, string, []bigquery.QueryParameter) {
+
+	variants := []string{"Network", "Upgrade", "Architecture", "Platform"}
+	selectVariants := ""
+	joinVariants := ""
+	groupByVariants := ""
+	for _, v := range variants {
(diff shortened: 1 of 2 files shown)


pkg/api/component_report.go
```diff
@@ -472,6 +477,29 @@ func (c *componentReportGenerator) getJobRunTestStatusFromBigQuery() (apitype.Co
 }
 
 func (c *componentReportGenerator) getCommonTestStatusQuery() (string, string, []bigquery.QueryParameter) {
+
+	// Simulate a list of varaints that were requested, this would be specified in the API request / UI in the real world...
+	variants := []string{"Network", "Upgrade", "Architecture", "Platform"}
```
dgoodwin, line 482: groupBy query param becomes the above array

Additional includeVariant query param comes in with Installer:hypershift for example. When includes are specified we inject a join and a where, but not a groupby for that variant name.

PR #1642, opened 2024-04-28
Fix cmd format

liangxia:


DEVELOPMENT.md
```diff
@@ -124,7 +124,7 @@ If you'd like to launch just Component Readiness, you can run:
 
 ```
 ./sippy component-readiness
```
stbenjam, line 126: Missing one here too

liangxia: Good catch. Fixed now.

stbenjam:
Thank you :) 

/lgtm

PR #1659, opened 2024-05-03
trt-1604: handle missing job bugs

neisw:
Passes in URL for job bugs


sippy-ng/src/jobs/JobAnalysis.js
```diff
@@ -89,18 +89,19 @@ export function JobAnalysis(props) {
...
-        return Promise.all([analysis.json(), bugs.json()])
+        return Promise.all([analysis.json()])
       })
-      .then(([analysis, bugs]) => {
+      .then(([analysis]) => {
         setAnalysis(analysis)
-        setBugs(bugs)
+        setBugsURL(
+          `${process.env.REACT_APP_API_URL}/api/jobs/bugs?${queryParams}`
+        )
```
neisw, line 104: Didn't want to copy the queryParams logic to BugTable.  Could have set that but figured passing in URL was more generic.

stbenjam:
Thanks for fixing this

/lgtm

PR #1661, opened 2024-05-03
TRT-1647: Allow overriding the URL used for Prow

stbenjam:


pkg/flags/prow.go
```diff
@@ -0,0 +1,30 @@
...
+		URL: "https://prow.ci.openshift.org",
+	}
+}
+
+func (f *ProwFlags) BindFlags(fs *pflag.FlagSet) {
+	fs.StringVar(&f.URL, "prow-url", f.URL, "URL for Prow")
+}
+
+func (f *ProwFlags) Validate() error {
+	_, err := url.ParseRequestURI(f.URL)
```
neisw, line 25: Looks like the code expects the URL to not have a trailing '/'.  Only question I have is if we input 'https://prow.ci.openshift.org/' does everything work or do we need to protect against that?   Looks like we are ok when I manually tested with 'https://prow.ci.openshift.org//view/gs/' so probably ok to ignore.

neisw:
/lgtm
/hold
remove hold presuming no concern on my validation question

pkg/flags/prow.go
stbenjam: Extra slashes are OK, if unsightly.

PR #1666, opened 2024-05-05
trt-1555: Support keeping only common job runs for all test ids

neisw:
* Removes 'OutputType' from JSON output since JSON is the default and persisting it requires manual step to remove when writing to DB.
* Updates [gen-resolved-issue.md](https://github.com/neisw/sippy/blob/0231b877cdd42b4165390e838785dcdf256bb8ed/hack/gen-resolved-issue.md) to document the input argument options
* Adds --match-all-job-runs to drop job runs from the incident records that don't contain all of the  specified test ids
** Likely needs more work to not drop everything when a testid / variant doesn't have any common job runs which will currently drop all job runs.

### DIFF
--- a/hack/gen-resolved-issue.py
+++ b/hack/gen-resolved-issue.py
@@ -501,6 +501,37 @@ def issue_details(record, issue_type, issue_description, issue_url):
 
     return issue_type, issue_description, issue_url
 
+def find_matching_job_run_ids(incidents):
+    # take the first test entry and get the set of job runs
+    # enumerate the remaining test entries and job runs and remove any that don't match from the first set
+    # when done if there are valid job runs set them for all of the test ids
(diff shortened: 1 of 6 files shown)


hack/gen-resolved-issue.py
```diff
@@ -620,6 +655,9 @@ def issue_details(record, issue_type, issue_description, issue_url):
         if None == os.environ.get('GOOGLE_APPLICATION_CREDENTIALS'):
```
stbenjam, line 655: Not related to to this change, but other tools including Google's use `GOOGLE_SA_CREDENTIAL_FILE` for this (including Sippy).  Might be nice to be consistent

neisw:
/close
replaced by https://github.com/openshift/sippy/pull/1674

PR #1678, opened 2024-05-08
trt-1469: Update CellStatus with TriagedIncidents

neisw:
* Adds triagedIncidents to cellStatus for gen-resolved-issue.py use and future UI
* Adds triagedIncidents support to gen-resovled-issue.py for updating existing incidents
* Moves some TriagedIncident types to api/types.go
* Adds doc for triaging regressions from JIRA requests
*  Updates `--match-all-job-runs` to be variant match specific


neisw:
/hold
looks like column  headers might be impacted by the new data

PR #1681, opened 2024-05-09
TRT-1661: Add link to title pages to help triage

DennisPeriquet:
This will adds a link in the lower right page to page API URL. This will (hopefully) obviate the need to use the debugging tools to retrieve the link used for the triage workflow.

### DIFF
--- a/sippy-ng/src/component_readiness/CopyPageURL.js
+++ b/sippy-ng/src/component_readiness/CopyPageURL.js
@@ -1,19 +1,18 @@
 import { Box, Button, Tooltip, Typography } from '@mui/material'
+import Link from '@mui/material/Link'
 import PropTypes from 'prop-types'
 import React from 'react'
 
 // This is used as a convenient way to capture the page URL for use with
 // the triaging workflow.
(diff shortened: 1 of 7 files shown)


stbenjam:
Missing a popover on successful copy, for consistency with the rest of them. You could also just make this a simple link horizontally next to the Generated at info and let folks use the right click copy link. I’m not sure I like the button.

sippy-ng/src/component_readiness/CopyPageURL.js
```diff
@@ -0,0 +1,25 @@
...
+export default function CopyPageURL(props) {
+  const handleCopyClick = () => {
+    navigator.clipboard.writeText(props.apiCallStr)
+  }
+
+  return (
+    <Box align="right">
+      <Typography variant="caption">
+        <Tooltip title={props.apiCallStr}>
+          <Button onClick={handleCopyClick}>copy page URL</Button>
```
dgoodwin, line 16: COPY API URL

With that it looks good to me. Page sounds like the current page URL. (i.e. sippy-ng)

PR #1695, opened 2024-05-14
TRT-1667: Track component readiness test regression start/end dates

dgoodwin:
Sippy's metrics loop which queries the default component readiness report already, now uses this data to maintain a new test_regressions table in bigquery tracking when a test first went regressed, and when it resolved. The opened date is now returned in a new sub-struct on the test summary for regressed/triaged tests, and displayed in the test modal.

Because params can change, the goal here is to specifically track when a test regressed *in the default report*. For simplicity now however, regardless of your parameters, we'll include that opened date in the response provided the test was still regressed in that report. In future, we may only include this value if you're using the "default" server side view.

The code attempts to re-use recent regressions, within the last two days, to limit flapping on regressions that are appearing/disappearing depending on borderline results available.

The opened/closed dates are approximate, depending on when the metrics loop runs and picks up the newest data.

### DIFF
--- a/pkg/apis/api/types.go
+++ b/pkg/apis/api/types.go
@@ -915,12 +915,6 @@ type ComponentReportTestIdentification struct {
 
 type ComponentReportTestSummary struct {
 	ComponentReportTestIdentification
-	// Status is an integer status we're trying to get away from in favor of the regression status struct. (TODO)
-	Status           ComponentReportStatus           `json:"status"`
-	RegressionStatus ComponentReportRegressionStatus `json:"regression_status"`
-}
-
-type ComponentReportRegressionStatus struct {
 	// Status is an integer representing the severity of the regression.
 	Status ComponentReportStatus `json:"status"`
 
@@ -929,7 +923,7 @@ type ComponentReportRegressionStatus struct {
 	// the regression with it's *default view* query. However we always include it in the response (if that test
 	// is regressed per the query params used). Eventually we should only include these details if the default view
 	// is being used, without overriding the start/end dates.
-	Opened *time.Time `json:"regressed_since"`
+	Opened *time.Time `json:"opened"`
 }
 
 type ComponentReportTestDetails struct {
--- a/pkg/apis/api/types.go
+++ b/pkg/apis/api/types.go
@@ -1107,7 +1107,7 @@ type ComponentReportTriageIncidentSummary struct {
 type TestRegression struct {
 	Release      string                 `bigquery:"release" json:"release"`
 	TestID       string                 `bigquery:"test_id" json:"test_id"`
-	TestName     string                 `bigquery:"test_name" json:"test_name"`
+	TestName     bigquery.NullString    `bigquery:"test_name" json:"test_name"`
 	RegressionID string                 `bigquery:"regression_id" json:"regression_id"`
 	Opened       time.Time              `bigquery:"opened" json:"opened"`
 	Closed       bigquery.NullTimestamp `bigquery:"closed" json:"closed"`
--- a/pkg/apis/api/types.go
+++ b/pkg/apis/api/types.go
@@ -1102,9 +1102,22 @@ type ComponentReportTriageIncidentSummary struct {
 	TriagedIncidents []TriagedIncident `json:"incidents"`
 }
 
+// TestRegression is used for rows in the test_regressions table and is used to track when we detect test
+// regressions opening and closing.
+type TestRegression struct {
+	Release      string                 `bigquery:"release" json:"release"`
+	TestID       string                 `bigquery:"test_id" json:"test_id"`
+	TestName     string                 `bigquery:"test_name" json:"test_name"`
+	RegressionID string                 `bigquery:"regression_id" json:"regression_id"`
+	Opened       time.Time              `bigquery:"opened" json:"opened"`
+	Closed       bigquery.NullTimestamp `bigquery:"closed" json:"closed"`
+	Variants     []TriagedVariant       `bigquery:"variants" json:"variants"`
+}
+
 type TriagedIncident struct {
-	Release      string                       `bigquery:"release" json:"release"`
-	TestID       string                       `bigquery:"test_id" json:"test_id"`
+	Release string `bigquery:"release" json:"release"`
+	TestID  string `bigquery:"test_id" json:"test_id"`
+	// TODO: should this be joined in instead of recording? test_name can change for a given test_id
 	TestName     string                       `bigquery:"test_name" json:"test_name"`
 	IncidentID   string                       `bigquery:"incident_id" json:"incident_id"`
 	ModifiedTime time.Time                    `bigquery:"modified_time" json:"modified_time"`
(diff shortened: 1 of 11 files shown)


dgoodwin:
/hold

Still doing some testing, the 90 minute buffer restrictions before you can delete/update new rows make it very slow to test all the edge cases.

pkg/componentreadiness/tracker/tracker.go
```diff
@@ -0,0 +1,284 @@
...
+
+func NewBigQueryStorage(client *bigquery.Client) Storage {
+	return &BigQueryStorage{client: client}
+}
+
+func (bqs *BigQueryStorage) ListCurrentRegressions(release string) ([]api.TestRegression, error) {
+	// List open regressions (no closed date), or those that closed within the last two days. This is to prevent flapping
+	// and return more accurate opened dates when a test is falling in / out of the report.
+	queryString := fmt.Sprintf("SELECT * FROM %s.%s WHERE release = @SampleRelease AND (closed IS NULL or closed > TIMESTAMP_SUB(CURRENT_TIMESTAMP(), INTERVAL 2 DAY))",
+		testRegressionsDataSet, testRegressionsTable)
```
neisw, line 43: Stephen has told me (in the past) to tell you it being better to reference bqs.client.DataSet instead of hardcoding is so we can configure the dataset via command options.

pkg/componentreadiness/tracker/tracker.go
```diff
@@ -0,0 +1,280 @@
...
+
+// BigQueryStorage is the primary implementation for real world usage, tracking when we detect test
+// regressions opening/closing in the Bigquery test_regressions table.
+type BigQueryStorage struct {
+	client *bigquery.Client
+}
+
+func NewBigQueryStorage(client *bigquery.Client) Storage {
+	return &BigQueryStorage{client: client}
+}
```
stbenjam, line 37: Naming things is hard. Big Query Storage is the name of a bigquery API, I think we should pick a different name to avoid overloading. This is in a directory called tracker.go, I'd expect see things related to the word "Tracking" in here.

Maybe RegressionTracker, and the impl BigQueryRegressionTracker?

pkg/componentreadiness/tracker/tracker.go
```diff
@@ -0,0 +1,280 @@
...
+	id := uuid.New()
+	newRegression := &api.TestRegression{
+		Release:      release,
+		TestID:       newRegressedTest.TestID,
+		TestName:     bigquery.NullString{StringVal: newRegressedTest.TestName, Valid: true},
+		RegressionID: id.String(),
+		Opened:       time.Now(),
+		Variants: []api.TriagedVariant{
+			{
+				Key:   "Network",
```
neisw, line 91: Feel free to take it or leave it, I created some [variant constants](https://github.com/openshift/sippy/blob/f41dff9ef1f23735c8ac2e1b5939caa1bcc0baeb/pkg/componentreadiness/resolvedissues/types.go#L13) with a comment to sync up with your [variantregistry constants](https://github.com/openshift/sippy/blob/1025904eb603f8cab6173c498e491ecc7bd5272f/pkg/variantregistry/ocp.go#L241).  Looks like yours merged, should we use them here and I can update mine in another PR?

pkg/componentreadiness/tracker/tracker.go
dgoodwin: Good point I'll adjust to something like that.

dgoodwin: I wanted to keep an underlying interface for abstracting away bigquery (for testing purposes... someday), so that RegressionTracker was already claimed, but I moved this to RegressionStore + BigQueryRegressionStore. Hopefully that helps.

pkg/componentreadiness/tracker/tracker.go
dgoodwin: Ahaaa that's why we have a wrapper client. On it!

dgoodwin:
Feedback addressed and fixed several bugs relating to re-using recently closed regressions.

sippy-ng/src/component_readiness/RegressedTestsModal.js
```diff
@@ -131,6 +131,34 @@ export default function RegressedTestsModal(props) {
       flex: 10,
       renderCell: (param) => <div className="test-name">{param.value}</div>,
     },
+    {
+      field: 'opened',
+      headerName: 'Regressed Since',
+      flex: 15,
+      valueGetter: (params) => params.row.regression_status.regressed_since,
+      valueGetter: (params) => {
```
neisw, line 139: Is it intentional you have `valueGetter` defined twice here?

dgoodwin: Nope! TY, fixing now.

pkg/apis/api/types.go
```diff
@@ -1102,9 +1116,22 @@ type ComponentReportTriageIncidentSummary struct {
 	TriagedIncidents []TriagedIncident `json:"incidents"`
 }
 
+// TestRegression is used for rows in the test_regressions table and is used to track when we detect test
+// regressions opening and closing.
+type TestRegression struct {
+	Release      string                 `bigquery:"release" json:"release"`
+	TestID       string                 `bigquery:"test_id" json:"test_id"`
+	TestName     bigquery.NullString    `bigquery:"test_name" json:"test_name"`
```
stbenjam, line 1124: Can the test name really be null?

pkg/apis/api/types.go
```diff
@@ -1102,9 +1116,22 @@ type ComponentReportTriageIncidentSummary struct {
...
+	RegressionID string                 `bigquery:"regression_id" json:"regression_id"`
+	Opened       time.Time              `bigquery:"opened" json:"opened"`
+	Closed       bigquery.NullTimestamp `bigquery:"closed" json:"closed"`
+	Variants     []TriagedVariant       `bigquery:"variants" json:"variants"`
```
stbenjam, line 1128: Nit, since this has become more than just Triage, I'd probably rename this to Variant

sippy-ng/src/component_readiness/RegressedTestsModal.js
stbenjam: I think most places in Sippy we're consistent about showing the relativeTime like "about 15 days ago" (there's a helper function), and use a `<Tooltip>` to show the raw value

sippy-ng/src/component_readiness/RegressedTestsModal.js
```diff
@@ -131,6 +131,33 @@ export default function RegressedTestsModal(props) {
       flex: 10,
       renderCell: (param) => <div className="test-name">{param.value}</div>,
     },
+    {
+      field: 'opened',
```
stbenjam, line 135: There's a hidden field here that could be dependent on whether we're in the default value or not.  We'd need to write something to check if query is the default, it could be done as a follow-up

pkg/apis/api/types.go
dgoodwin: Only because I added it after making the table. :) I'll go recreate it.

pkg/apis/api/types.go
dgoodwin: Sadly already claimed by sippy classic, but I made it ComponentReportVariant.

dgoodwin:
Feedback addressed.

I also noticed I was using Opened but regressed_since in json, that is now just 'opened'. Also flattened the new struct, just moved the Opened timestamp onto the TestSummary, I couldn't see a reason to keep in a sub-struct.

PR #1698, opened 2024-05-15
trt-1668: Add triage start time buffer

neisw:
Sample query uses modified time.  Triage job runs track start time.  Short term fix is to add a 6 hour buffer for triage job runs that are about to fall off but we still want to consider.


neisw:
/hold cancel

We don't have start or end times in the junit table currently.  For now just add in a buffer time when we are close to the sample start time.

PR #1703, opened 2024-05-16
TRT-1307: Component Readiness use new variant registry

xueqzhan:
/hold for team to experiment.


pkg/componentreadiness/resolvedissues/types.go
```diff
@@ -17,9 +17,8 @@ const variantArchitecture = "Architecture"
 const variantNetwork = "Network"
 const variantPlatform = "Platform"
 const variantUpgrade = "Upgrade"
-const variantVariant = "Variant"
 
-var triageMatchVariants = buildTriageMatchVariants([]string{variantArchitecture, variantNetwork, variantPlatform, variantUpgrade, variantVariant})
+var triageMatchVariants = buildTriageMatchVariants([]string{variantArchitecture, variantNetwork, variantPlatform, variantUpgrade})
```
neisw, line 21: We have to figure out how we transition things like `rt`, `serial`, etc. that are currently stored as 'variant' this change would break backwards compatibility with triaged incidents (we can live with that or migrate) but how do we keep the data that was previously there and considered relevant?

sippy-ng/src/component_readiness/CompReadyUtils.js
```diff
@@ -214,9 +218,9 @@ export function getStatusAndIcon(status, grayFactor = 0) {
...
 // concatenated to form a column name
 export function formColumnName(column) {
-  return Object.keys(column)
-    .filter((key) => key != 'status' && key != 'regressed_tests')
-    .map((key) => column[key])
+  let variants = column['variants']
+  return Object.keys(variants)
+    .map((key) => variants[key])
     .join(' ')
 }
```
stbenjam, line 225: There's a usage of this in the regressed tests modal that doesn't match the new style.  That modal looks broken actually

https://github.com/openshift/sippy/blob/960454694ecb62ef79ac1a3f1d807ed81f528152/sippy-ng/src/component_readiness/RegressedTestsModal.js#L37

sippy-ng/src/component_readiness/CompReadyMainInputs.js
```diff
@@ -33,6 +35,33 @@ const useStyles = makeStyles((theme) => ({
...
+    'NetworkStack',
+    'Owner',
+    'Release',
+    'ReleaseMajor',
+    'ReleaseMinor',
+    'Scheduler',
+    'SecurityMode',
+    'Suite',
+    'Topology',
+  ])
```
stbenjam, line 64: Not required here unless you think it's easy doable, but I think this should be driven by the server as much as possible.  Keeping productization in mind, someone else may have an entirely different set of variants

stbenjam:
In the last drill down, we lose the break apart - couldn't you force a new groupby on the very last drill down? 

Prod sippy:
image

This PR:
image

Forced new grouping looks a lot more like current version (but we seem to lose the triage status):
image

pkg/apis/api/types.go
```diff
@@ -831,31 +820,41 @@ type ComponentTestStatus struct {
...
-	TestID       string `json:"test_id"`
-	Network      string `json:"network"`
-	Upgrade      string `json:"upgrade"`
-	Arch         string `json:"arch"`
-	Platform     string `json:"platform"`
-	FlatVariants string `json:"flat_variants"`
+	TestID string `json:"test_id"`
+
+	// Proposed, need to serialize to use as map key
+	Variants map[string]string `json:"variants"`
```
dgoodwin, line 838: Only other option here would be to use a List of Key/Value structs we always make sure is sorted when created. But if the json marshal to string key is working for us hopefully we can just stick with it.

pkg/componentreadiness/resolvedissues/types.go
dgoodwin: Does anyone know what else is floating around out there using generic "variant"? traige data should be pretty easy to migrate, same for the new start/end date tracking. I can't immediately think what else might be around.

dgoodwin:
Triage status probably getting lost because of the mismatch in how it's stored today. Forrest I think planned for this well with the key/value variants in there, but today one of those keys is Variant=standard which would probably trip this up. A small db migration may be required to remove variant=Standard and move variant=rt to the Scheduler=realtime etc.

dgoodwin (commented):
Seems to be working very well locally.

sippy-ng/src/component_readiness/IncludeVariantCheckboxList.js
```diff
@@ -0,0 +1,28 @@
...
+          : []
+      }
+      setCheckedItems={setCheckedItems}
+    />
+  )
+}
+
+IncludeVariantCheckBoxList.propTypes = {
+  variantName: PropTypes.string,
+}
```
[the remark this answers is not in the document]
xueqzhan: This should be addressed.

pkg/sippyserver/metrics/metrics.go
```diff
@@ -330,7 +342,19 @@ func refreshComponentReadinessMetrics(client *bqclient.Client, prowURL, gcsBucke
...
+			}
+			archLabel, ok := col.Variants["Architecture"]
+			if !ok {
+				archLabel = ""
+			}
+			platLabel, ok := col.Variants["Platform"]
+			if !ok {
+				platLabel = ""
+			}
+			componentReadinessMetric.WithLabelValues(row.Component, networkLabel, archLabel, platLabel).Set(float64(col.Status))
```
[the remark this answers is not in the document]
xueqzhan: This should be addressed.

PR #1751, opened 2024-06-05
Guard the regression table maint with an opt-in flag.

dgoodwin:
This prevents conflicts where devs working on sippy are actually
modifying the regression tables, in the case of the recent variant
registry work, this is catastrophically wrong.


stbenjam:
/lgtm

But boolean flags :-( Maybe this could eventually be a task in the sippy-daemon instead since it's kind of a constantly running task

dgoodwin:
I like that idea re the daemon, provided we hit the cache, no extra comp readiness querying.

PR #1753, opened 2024-06-05
DO NOT MERGE: Just test e2e

DennisPeriquet:
The `technical-release-team/sippy-bigquery-job-importer` credential was modified to use the same one as the friday disruption data PR (which is read-only).


DennisPeriquet:
e2e passed

PR #1755, opened 2024-06-07
TRT-1218: Display intervals groups automatically by source

dgoodwin:
This change allows the intervals UI to display whatever origin publishes in the intervals files, grouped automatically on source.

By default we show only those with display=true as specified by origin, but an option is added in the UI to override this, though you may get strange results.

Our core sources are ordered and selected by default, and additional sources can be shown with buttons at the top. As before it should be possible to send someone a link to see exactly what you're looking at. Most main sources are colored as before but with a refined mechanism requiring less code in less places.

It's not perfect in current form, but I think it's good enough for a first pass and can be improved in future. 

Weak points: 
- I think some intervals are overlapping when viewing PodState section, these should probably broken out with row keys in the locator when published in origin.
- Two Sources had to be further broken down, ClusterOperator -> available/degraded/progressing, and E2ETest -> passed/failed/skipped. Arguably this too should happen in origin.
- Still not as fast as I'd like.
- Still not as flexible with search as I'd like.
- UI seems a little overly refreshy.


xueqzhan:
/lgtm

I like that you use source for category.

PR #1766, opened 2024-06-11
TRT-1550: enable date changes on CR test details

sosiouxme:


sippy-ng/src/component_readiness/CompReadyMainInputs.js
```diff
@@ -156,3 +156,63 @@ export default function CompReadyMainInputs(props) {
 
 // component and environment may be null so they are not required
 CompReadyMainInputs.propTypes = {}
+
+export function CompReadyTestDetailInputs(props) {
```
sosiouxme, line 160: might belong in a different file?

sippy-ng/src/component_readiness/CompReadyTestReport.js
```diff
@@ -157,6 +158,19 @@ export default function CompReadyTestReport(props) {
...
+  const datesEnv = useContext(CompReadyVarsContext)
+  useEffect(() => {
+    let tmpDates = {
+      baseStartTime: datesEnv.baseStartTime,
+      baseEndTime: datesEnv.baseEndTime,
+      sampleStartTime: datesEnv.sampleStartTime,
+      sampleEndTime: datesEnv.sampleEndTime,
+    }
+    setDates(tmpDates)
+  }, [])
```
stbenjam, line 172: You should have datesEnv in the list there `}, [datesEnv])`

sosiouxme: if I do that, I get in the js console `Error: Objects are not valid as a React child [...]. If you meant to render a collection of children, use an array instead.`

i don't really understand that yet, but the whole point of this is to keep these from updating when the context changes so... maybe leaving it out is fine? i modeled it after the example two above

stbenjam: If it works then that's fine, usually useEffect needs to know what states it's dependent on to execute, otherwise it should only ever run once

sippy-ng/src/component_readiness/CompReadyMainInputs.js
stbenjam: Yea if it's big enough usually it should go in a file by itself - do you share any code between the two that means they need to stay together?

stbenjam:
Looks like this works.  The UI isn't super consistent with the other pages; if you were interested in writing more react you could look about having a sidebar on the test details page with only the release date controls.  That'd be better from the UX angle IMHO

I won't hold up merging on that though,  unhold if you're ready.

/lgtm
/hold

sosiouxme:
I think it simplifies it further to just vary Sidebar rather than doing something separate. If I had realized that wasn't so hard I'd probably have done it that way to begin with.

Along the way, seems unnecessary to be plumbing `theme` down into Sidebar when it can just get that from context.

stbenjam:
Looks great!

/lgtm
/hold cancel

stbenjam:
Oh if I use the week selectors buttons, and then click generate report I get a blank page with some errors in the javascript console

image

PR #1723, opened 2024-05-23
TRT-1305: Move Sippy to using variant registry

stbenjam:
Note: tests will fail until we get bigquery working in tests.  This change also means we always require bigquery.

- The OCP variant manager now queries bigquery
- Adds a new "sync-variants" command that will sync all jobs variants instead of waiting until we see them again (test this locally against a dev backup with `./sippy load --loader sync-variants --google-service-account-credential-file ~/tokens/asdf.json`)
- UI for tests shows the top 8 variants (PANTFUSI) with the rest in a tooltip

Example:
```
----------------------
name     | periodic-ci-openshift-multiarch-master-nightly-4.16-ocp-heavy-build-ovn-remote-libvirt-s390x
variants | {Platform:libvirt,Architecture:s390x,Network:ovn,NetworkStack:ipv4,Topology:ha,FeatureSet:default,Upgrade:none,SecurityMode:default,Installer:ipi,Aggregation:none,NetworkAccess:default,Owner:eng,Release:4.16,ReleaseMajor:4,ReleaseMinor:16,Scheduler:default,Suite:unknown}
-[ RECORD 2 ]---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
name     | periodic-ci-openshift-multiarch-master-nightly-4.16-upgrade-from-nightly-4.15-ocp-e2e-upgrade-gcp-ovn-heterogeneous
variants | {Platform:gcp,Architecture:heterogeneous,Network:ovn,NetworkStack:ipv4,Topology:ha,FeatureSet:default,Upgrade:minor,SecurityMode:default,Installer:ipi,Aggregation:none,FromRelease:4.15,FromReleaseMajor:4,FromReleaseMinor:15,NetworkAccess:default,Owner:eng,Release:4.16,ReleaseMajor:4,ReleaseMinor:16,Scheduler:default,Suite:unknown}
-[ RECORD 3 ]---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
name     | periodic-ci-openshift-assisted-installer-agent-master-edge-subsystem-test-periodic
variants | {Architecture:amd64,NetworkStack:ipv4,Topology:ha,FeatureSet:default,Upgrade:none,SecurityMode:default,Installer:assisted,Aggregation:none,NetworkAccess:default,Owner:eng,Scheduler:default,Suite:unknown}
```

image

PR #1774, opened 2024-06-12
Revert "TRT-1305: Move Sippy to using variant registry"

stbenjam:
Reverts openshift/sippy#1723


stbenjam:
/hold

Will merge later

PR #1811, opened 2024-06-28
updated to handle new information in variant names

eggfoobar:
Updated display to split up the variant name to make it easier to see for overview grid and test grid

/assign @stbenjam 

Ex:
image

image


stbenjam:
Beautiful, thank you!!

/lgtm

PR #1826, opened 2024-07-03
trt-1712: Target build cluster

neisw:
Adds support to target a specific build cluster for job run / test failures

### DIFF
--- a/hack/gen-resolved-issue.py
+++ b/hack/gen-resolved-issue.py
@@ -754,6 +754,12 @@ def triage_regressions(regressed_tests, triaged_incidents, issue_url):
                 if "completionTime" in prow_job_json["status"]:
                     completion_time = prow_job_json["status"]["completionTime"]
 
+                build_cluster =  prow_job_json["spec"]["cluster"]
+
+                if args.target_build_cluster != None and len(args.target_build_cluster) > 0:
+                    if build_cluster != args.target_build_cluster:
+                        continue  
+
                 if job_matches(record, prow_url, start_time, issue_resolution_date):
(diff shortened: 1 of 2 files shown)


hack/gen-resolved-issue.py
```diff
@@ -754,6 +754,12 @@ def triage_regressions(regressed_tests, triaged_incidents, issue_url):
                 if "completionTime" in prow_job_json["status"]:
                     completion_time = prow_job_json["status"]["completionTime"]
 
+                build_cluster =  prow_job_json["spec"]["cluster"]
+
+                if args.target_build_cluster != None and len(args.target_build_cluster) > 0:
```
dgoodwin, line 759: just a little python optimization for future, you can just do "if args.target_build_cluster:", both None and "" eval to false in that context.

```
❯ python3
Python 3.12.4 (main, Jun  6 2024, 18:26:44) [Clang 15.0.0 (clang-1500.3.9.4)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> n = None
>>> n = ""
>>> if n:
...     print("Hi")
...
>>> n = None
>>> if n:
...     print("1")
...
>>>
>>> n = "asdadasd"
>>> if n:
...     print("1")
...
1

```

dgoodwin (approved):
Nice and easy! That's a good improvement. 
/lgtm

hack/gen-resolved-issue.py
neisw: Thanks, I had tried if len(args.target_build_cluster): but that failed for me.  Good to know.

PR #1839, opened 2024-07-08
Use new variant registry for component readiness

xueqzhan:
This branch has been running along the master for some time. We think it is time to merge to master.


neisw:
/hold

Need to review https://github.com/openshift/sippy/blob/c0048af6f2d8077567ffcf82ea8a9e73da9fb3a0/pkg/api/component_report.go#L64

dgoodwin:
> /hold
> 
> Need to review
> 
> https://github.com/openshift/sippy/blob/c0048af6f2d8077567ffcf82ea8a9e73da9fb3a0/pkg/api/component_report.go#L64

That was from my work on PR Component Readiness, /payload jobs get weird names that won't be in the job registry and thus can't be compared. The *real* underlying job name is in a prowjob annotation, which fortunately is in our db. The dynamic column is chatgpt special, let me know if you find a problem in there.

PR #1852, opened 2024-07-11
TRT-1702: Risk Analysis: ensure Test ID is populated

sosiouxme:
At the time that  `openshift-e2e-tests` requests it, test results haven't yet been associated with test IDs (we just have junits). So it is posting a `ProwJobRun` with partial `Test` records to the API, sans IDs. This PR changes the past-results queries by test name to also return the test ID, and uses that to populate RA results.

Messing with the queries worries me a little; every instance of a test name I saw only had a single test ID across multiple jobs, but I'm not certain that's reliable, so check me.

### DIFF
--- a/pkg/api/tests.go
+++ b/pkg/api/tests.go
@@ -195,7 +195,7 @@ func BuildTestsResults(dbc *db.DB, release, period string, collapse, includeOver
 	// Collapse groups the test results together -- otherwise we return the test results per-variant combo (NURP+)
 	variantSelect := ""
 	if collapse {
-		rawQuery = rawQuery.Select(`name,watchlist,jira_component,jira_component_id,` + query.QueryTestSummer).Group("name,watchlist,jira_component,jira_component_id")
+		rawQuery = rawQuery.Select(`id,name,watchlist,jira_component,jira_component_id,` + query.QueryTestSummer).Group("id,name,watchlist,jira_component,jira_component_id")
 	} else {
 		rawQuery = query.TestsByNURPAndStandardDeviation(dbc, release, table)
 		variantSelect = "suite_name, variants," +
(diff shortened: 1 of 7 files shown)


pkg/api/tests.go
```diff
@@ -210,9 +210,8 @@ func BuildTestsResults(dbc *db.DB, release, period string, collapse, includeOver
 	}
 
 	testReports := make([]apitype.Test, 0)
-	// FIXME: Add test id to matview, for now generate with ROW_NUMBER OVER
 	processedResults := dbc.DB.Table("(?) as results", rawQuery).
-		Select(`ROW_NUMBER() OVER() as id, watchlist, name, jira_component, jira_component_id,` + variantSelect + query.QueryTestSummarizer).
+		Select(`id, watchlist, name, jira_component, jira_component_id,` + variantSelect + query.QueryTestSummarizer).
```
neisw, line 214: This is likely worth additional review.  I'm wondering if there was a reason id wasn't used before and if this impacts other parts of the UI like the tests report.

neisw: Could select id as test_id and then keep the existing ROW_NUMBER() OVER() statement.  Don't know that we need it but if we wanted to keep the query as it was and just add in the test id.

sosiouxme: from the deleted FIXME comment and the state of the matview, i gathered this was written at a point where the test id didn't exist in the view, and so ROW_NUMBER was just a stand-in so there would be _something_, with the intent to go back and update the matview and the query. well, the matview got updated with a test id, but the query never did until this.

neisw: Looks like apitype.Test expects ID to be a float64: https://github.com/openshift/sippy/blob/cc0ba0efaad03249186db2b1e0b9fbfac41ef202/pkg/apis/api/types.go#L535

sosiouxme: /hold
the tests report is using the "synthetic" id and it gets the table wrong if results repeat the same id. that seems like an abuse of `Test.ID` to me but i suppose we can work around it.

no idea what the float64 cast is for but we can leave Test.ID to its current purpose and look up the real test id in a new field.

pkg/apis/api/types.go
```diff
@@ -757,7 +758,7 @@ type ProwJobRunRiskAnalysis struct {
 
 type ProwJobRunTestRiskAnalysis struct {
 	Name     string
-	TestID   uint
+	TestID   string
```
neisw, line 761: Curious how this will play out with the autodl table.  We currently have TestID as an int.  We can delete those tables and start over but if you think we will keep an int value and add the 'unique' string id maybe we should add a new string field here instead?  I wonder if TestUniqueId or similar is a good identifier, like what you have in:

 `town.unique_id as test_str_id,`

PR #1861, opened 2024-07-15
cr hysteresis

dgoodwin:
- **Return certainty in CR report and display in regressed tests table**
- **Store openRegressions on componentReportGenerator for easier access**
- **Introduce lag for a regression to clear**


stbenjam:
Does this fix the flip flopping we saw?

There's two values at play - how sure fisher's is there's a regression (confidence), which we require to be at least 95% certain, and pity which we set to 5%.

Fisher's can still be sure there's a regression between very small pass rates, e.g. a test that regresses between 98% and 97.8% could report a 95% confidence in a regression.

However, pity is the minimum difference between the basis and sample for reporting a regression.  I thought the problem (and I could be mistaken)  was we would often briefly cross the 5% pity and turn red, and then a single pass would drop us just below 5% so the square would go green.

stbenjam:
I think you could flip flop in both cases, where confidence is dipping under 95% and then above, but I think you could still have high confidence and flip flop on pity.  I guess it depends on what we were seeing in the real world

dgoodwin:
Discussed on slack but a summary here, I haven't noticed pity being involved yet, but I also haven't paid too much attention to it. Gut feel is that most of the examples I've looked at we're far outside our pity range by the time we're 95% sure of a regression. We decided we're going to track a little data in the regressions table to record the pass rate and fishers confidence when we open the regression, and when we close. This should provide more insight into things. This will come in as a separate PR.

This one should be good to go, lint and tests fixed now.

PR #1869, opened 2024-07-19
trt-1740: use valueGetter for variants column

neisw:
Use valueGetter for variants formColumnName  call as noted in the [doc](https://mui.com/x/react-data-grid/filtering/customization/).


dgoodwin:
/lgtm

Thanks!

PR #1881, opened 2024-07-24
TRT-1748: WIP: Track more stats when opening/closing regression records in bigquery

dgoodwin:
This PR is the result of a shiftweek convo where Stephen and I discussed tracking more info in the test regression records so we can better understand how they're behaving. 

To do this, I needed stats for when we close the regression. The old implementation can't do this because it just generated a report and then closed out any open regressions no longer in the report. Obviously we don't have stats at that point. Instead I had to refactor the regression tracking straight into component_report.go. 

I'm not thrilled with the implementation and worry it's more complexity than it's worth. However in this state it's working, and you can see the results in test_regressions_dgoodwin_temp. Fisher value for closing doesn't seem to work yet, not sure if it always is coming out 0 or just in the cases I've seen so far. 

So questions for reviewer:
- is the data useful
- is the implementation contributing too much to the complexity in component_report
- would inserting a row into a table every time we scan the report and see a regressed test + open regression be better? this would allow charting the values over time until it disappears. we could go back to old implementation i think.

### DIFF
--- a/pkg/api/component_report.go
+++ b/pkg/api/component_report.go
@@ -181,15 +181,17 @@ func GetComponentReportFromBigQuery(client *bqcachedclient.Client, prowURL, gcsB
 	variantOption apitype.ComponentReportRequestVariantOptions,
 	advancedOption apitype.ComponentReportRequestAdvancedOptions,
 	cacheOption cache.RequestOptions,
+	trackRegressions bool,
 ) (apitype.ComponentReport, []error) {
 	generator := componentReportGenerator{
-		client:        client,
-		prowURL:       prowURL,
-		gcsBucket:     gcsBucket,
-		cacheOption:   cacheOption,
-		BaseRelease:   baseRelease,
-		SampleRelease: sampleRelease,
-		triagedIssues: nil,
+		client:           client,
+		prowURL:          prowURL,
+		gcsBucket:        gcsBucket,
+		cacheOption:      cacheOption,
+		BaseRelease:      baseRelease,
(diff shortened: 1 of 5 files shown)


pkg/api/component_report.go
```diff
@@ -1647,6 +1671,13 @@ func (c *componentReportGenerator) generateComponentTestReport(baseStatus map[st
 		updateCellStatus(rowIdentifications, columnIdentification, testID, testStats, aggregatedStatus, allRows, allColumns, nil, c.openRegressions)
 	}
 
+	// Fallback for any open regressions in the db, that did not appear in this report.
+	// This would likely imply they no longer have sufficient sample, and this should be rare.
+	err := c.closeRegressionsForMissingTests(rLog, matchedOpenRegressions, now)
```
neisw, line 1676: Wondering how the Base sample plays into regression tracking.  If I change the base sample window to something small or less stable time period would that cause regressions to close for the main GA base sample time?  Should we consider differentiating open regressions based on the base sample release, start and end dates?

dgoodwin: Regression tracking should only be turned on in a very specific case which is the metrics loop, there should be no other way to turn it on. We should be able to assume that the base will remain fixed, and once server side views land that will be encoded in the view and should never change.

dgoodwin:
I'm still not sure if this is worth it, the old implementation was much cleaner. I almost think this would be better served by looking up test pass rates for given variants on the dates in question.

PR #1882, opened 2024-07-24
TRT-1743: Remove flake from success count for test cases

xueqzhan:
In openshift-tests, we create two junits for the same test to indicate a flake, one success one failure. But depending on the order in which the two junits appear in the file, we get different entries to the junit BQ table. In one case the success count includes the flake while in the other it does not. This has caused math errors in some status calculation.

### DIFF
--- a/pkg/api/component_report.go
+++ b/pkg/api/component_report.go
@@ -72,7 +72,11 @@ const (
 				CASE
 					WHEN flake_count > 0 THEN 0
 					ELSE success_val
-				END AS adjusted_success_val
+				END AS adjusted_success_val,
+				CASE
+					WHEN flake_count > 0 THEN 1
+					ELSE 0
+				END AS adjusted_flake_count
 			FROM
 				%s.junit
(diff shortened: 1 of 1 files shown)


pkg/api/component_report.go
```diff
@@ -431,7 +435,7 @@ func (c *componentReportGenerator) getTestDetailsQuery(allJobVariants apitype.Jo
 						ANY_VALUE(cm.jira_component_id) AS jira_component_id,
 						COUNT(*) AS total_count,
 						ANY_VALUE(cm.capabilities) as capabilities,
-						SUM(success_val) AS success_count,
+						SUM(adjusted_success_val) AS success_count,
 						SUM(flake_count) AS flake_count,
```
stbenjam, line 439: The other problem we found is flake_count isn't a 0 or 1, it can be higher.

You could replace these with `SUM(CASE WHEN flake_count > 0 THEN 1 ELSE 0 END) AS flake_count` I think to get the right answer

PR #1899, opened 2024-07-31
Revert "Update filtering to allow for owners"

neisw:
Reverts openshift/sippy#1898

Getting error `invalid variant value from includeVariant Owner:qe`


neisw:
/close

Ken added a dummy job to work around it

PR #1905, opened 2024-08-02
Server Side Views

dgoodwin:
This PR adds a first draft of server side views. No UI work is included, these are only accessible if you curl the API manually right now.

Adds /api/component_readiness/views to list all views configured on the server.
Adds support for relative start/end times of the format now, now-7d, ga, ga-30d, etc.
Removed some transient fields on API types that weren't really used.
Broke some query param parsing out to a component readiness package as it was getting unwieldy in server.go. Added unit testing.
Added unit testing for util that parses time strings and applies rounding.

A sample views file can be provided to the sippy component-readiness command with --component-readiness-views views.yaml.

```yaml
---
- name: 4.17-main
  base_release:
      release: "4.16"
      start: ga-30d
      end: ga
  sample_release:
      release: "4.17"
      start: now-7d
      end: now
  variant_options:
      column_group_by:
          Architecture: {}
          Network: {}
          Platform: {}
      db_group_by:
          Architecture: {}
          FeatureSet: {}
          Installer: {}
          Network: {}
          Platform: {}
          Suite: {}
          Topology: {}
          Upgrade: {}
      include_variants:
          Architecture:
              - amd64
          FeatureSet:
              - default
          Installer:
              - ipi
              - upi
          Owner:
              - eng
          Platform:
              - aws
              - azure
              - gcp
              - metal
              - vsphere
          Topology:
              - ha
  advanced_options:
      minimum_failure: 3
      confidence: 95
      pity_factor: 5
      ignore_missing: false
      ignore_disruption: true
```


pkg/apis/api/types.go
```diff
@@ -800,10 +800,19 @@ type PullRequestOptions struct {
...
+	End                time.Time           `json:"end,omitempty" yaml:"end,omitempty"`
+}
+
+// ComponentReportRequestRelativeReleaseOptions is an unfortunate necessity for views where we do not have
+// a fixed time, rather a relative time to now/ga. It is translated to the above normal struct before use.
+type ComponentReportRequestRelativeReleaseOptions struct {
+	Release            string              `json:"release" yaml:"release"`
+	PullRequestOptions *PullRequestOptions `json:"pull_request_options,omitempty" yaml:"pull_request_options,omitempty"`
+	RelativeStart      string              `json:"start,omitempty" yaml:"start,omitempty"`
+	RelativeEnd        string              `json:"end,omitempty" yaml:"end,omitempty"`
```
dgoodwin, line 815: This made me sad but I tried it several different ways, it seemed the cleanest to layer in the relative time parsing for a view and yet always have a time.Time available when we get into component_report.go.

pkg/util/utils.go
```diff
@@ -79,15 +83,71 @@ func URLForJob(dashboard, jobName string) *gourl.URL {
...
 }
 
-// ParseCRReleaseTime parses the time for component readiness. It adjusts the time based on a rounding factor
-func ParseCRReleaseTime(timeStr string, crTimeRoundingFactor time.Duration) (time.Time, error) {
-	releaseTime, err := time.Parse(time.RFC3339, timeStr)
-	if err != nil {
-		return releaseTime, err
+// releaseRelativeRE is a custom format we allow for times relative to now, or a releases ga date
+// (i.e. now-7d, ga-30d, ga, etc
+var releaseRelativeRE = regexp.MustCompile(`^(now|ga)(?:-([0-9]+)([d]))?$`)
```
stbenjam, line 88: Nice

stbenjam:
I really like the YAML format.  Are these maps to preserve ordering?

```
  variant_options:
      column_group_by:
          Architecture: {}
          Network: {}
          Platform: {}
      db_group_by:
          Architecture: {}
          FeatureSet: {}
          Installer: {}
          Network: {}
          Platform: {}
          Suite: {}
          Topology: {}
          Upgrade: {}
```

dgoodwin:
> I really like the YAML format. Are these maps to preserve ordering?
> 
> ```
>   variant_options:
>       column_group_by:
>           Architecture: {}
>           Network: {}
>           Platform: {}
>       db_group_by:
>           Architecture: {}
>           FeatureSet: {}
>           Installer: {}
>           Network: {}
>           Platform: {}
>           Suite: {}
>           Topology: {}
>           Upgrade: {}
> ```

Those are a side effect of the pre-existing use of sets on those types, which under the covers are implemented as a map with no values. We'd have to fork that type as well to avoid them or get into some kind of custom serialization. Let me know if you think thats worthwhile.

dgoodwin:
> Could releases also be relative? Maybe similar to dates, or you could just use an index? 0 for 4.17, -1 for 4.16, -2 for 4.15, etc.
> 
> e.g. compare 4.17 vs 4.16:
> 
> ```
> ---
> - name: "Release Comparison"
>   base_release:
>       release: dev-1
>       start: ga-30d
>       end: ga
>   sample_release:
>       release: dev
>       start: now-7d
>       end: now
> ```
> 
> Compare 4.17 vs 4.15:
> 
> ```
> ---
> - name: "2 Release Comparison"
>   base_release:
>       release: dev-2
>       start: ga-30d
>       end: ga
>   sample_release:
>       release: dev
>       start: now-7d
>       end: now
> ```
> 
> 4.16 vs 4.15:
> 
> ```
> ---
> - name: "Last Release"
>   base_release:
>       release: dev-1
>       start: ga-30d
>       end: ga
>   sample_release:
>       release: dev-2
>       start: now-7d
>       end: now
> ```

Good question. We definitely could but I think with the EAT team coming online and interested in many past releases, and considering promotion of a variant into the default views, it was feeling to me like we need views to be quite specific with a branching task to create the next view. Both are going to be problematic honestly. This might need more discussion and a followup PR.

dgoodwin:
Lint fixed. If it looks safe it would be great to slip this in, it will help avoid conflicts, and I can address any followups once back from PTO.

stbenjam:
Problem was on my side, looks like it doesn't break anything.

/lgtm

PR #1912, opened 2024-08-09
TRT-1619: cross-variant comparison UI

sosiouxme:
# step 1: UI inputs

Enable the UI for specifying variants for cross-comparison, both in the
UI parameters and API parameters. This introduces the "Compare" button  
in the variant selection sidebar, and plumbs the results through. This  
adds parameters for "variantCrossCompare" and "compareVariant" which    
specify (respectively) the variant groups that should query for 
different values in sample vs basis, and what those actual values should
be.

There was also a fair amount of refactoring and renaming along the way.
I tried to distinguish variantGroup ("Architecture") from individual    
variants ("Architecture:amd64"). Hopefully the result is clearer.

# step 2: API support

Since this could take a while, I added a second commit that removes the "Compare" button in the UI so that if we want, we can merge this PR without exposing the functionality.

### DIFF
--- a/sippy-ng/src/component_readiness/CompReadyVars.js
+++ b/sippy-ng/src/component_readiness/CompReadyVars.js
@@ -8,6 +8,8 @@ import {
   useQueryParam,
 } from 'use-query-params'
 import {
+  convertParamToVariantItems,
+  convertVariantItemsToParam,
   dateEndFormat,
   dateFormat,
   formatLongDate,
@@ -96,22 +98,6 @@ export const CompReadyVarsProvider = ({ children }) => {
     columnGroupByCheckedItemsParam = ['Platform', 'Architecture', 'Network'],
     setColumnGroupByCheckedItemsParam,
   ] = useQueryParam('columnGroupBy', ArrayParam)
-  const [
-    includeVariantsCheckedItemsParam = [
-      'Architecture:amd64',
-      'FeatureSet:default',
-      'Installer:ipi',
-      'Installer:upi',
-      'Owner:eng',
-      'Platform:aws',
-      'Platform:azure',
-      'Platform:gcp',
-      'Platform:metal',
-      'Platform:vsphere',
-      'Topology:ha',
-    ],
-    setIncludeVariantsCheckedItemsParam,
-  ] = useQueryParam('includeVariant', ArrayParam)
 
   const [confidenceParam = 95, setConfidenceParam] = useQueryParam(
     'confidence',
@@ -129,7 +115,7 @@ export const CompReadyVarsProvider = ({ children }) => {
   const [ignoreDisruptionParam = false, setIgnoreDisruptionParam] =
     useQueryParam('ignoreDisruption', BooleanParam)
 
-  // Create the variables to be used for api calls; these are initilized to the
+  // Create the variables to be used for api calls; these are initialized to the
   // value of the variables that got their values from the URL.
   const [columnGroupByCheckedItems, setColumnGroupByCheckedItems] =
     React.useState(columnGroupByCheckedItemsParam)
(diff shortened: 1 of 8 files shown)


sippy-ng/src/component_readiness/IncludeVariantCheckboxList.js
```diff
@@ -1,31 +1,190 @@
...
+            />
+          }
+          label={item}
+        />
+      ))}
+    </FormGroup>
+  )
+}
+
+function CompareVariantGroup(params) {
```
dgoodwin, line 152: This may depend on what you come up with for server side, but is there a plan yet for single vs multi-select? I suspect if you enable this compare mode you must be restricted to one value on each side for us to know what to compare to what.

sippy-ng/src/component_readiness/IncludeVariantCheckboxList.js
```diff
@@ -1,31 +1,190 @@
...
+          expandIcon={<ExpandMore />}
+        >
+          <Typography className="checkboxlist-label">
+            Include {variantGroupName}
+          </Typography>
+        </AccordionSummary>
+        <AccordionDetails>
+          {isCompareMode
+            ? CompareVariantGroup(params)
+            : IncludeVariantGroup(params)}
```
dgoodwin, line 125: TIL, this is cool.

sippy-ng/src/component_readiness/CompReadyVars.js
```diff
@@ -192,38 +178,69 @@ export const CompReadyVarsProvider = ({ children }) => {
...
+  const compareVariantsCheckedItems = convertParamToVariantItems(
+    compareVariantsCheckedItemsParam
+  )
+  const replaceCompareVariantsCheckedItems = (variant, checkedItems) => {
+    compareVariantsCheckedItems[variant] = checkedItems
+  }
+  // This is the list of variant groups (e.g. "Architecture") that have been selected for cross-variant comparison
+  const [variantCrossCompare, setVariantCrossCompare] = useState(
+    variantCrossCompareParam
+  )
```
dgoodwin, line 230: My js is weak, but why is this required when variantCrossCompareParam already exists?

dgoodwin (commented):
I think this looks fine to go in to me and iterate as work progresses, I doubt my ability to spot many js problems but I've given it my best shot. If you'd like someone else on the team to have a go, I'll add a hold just in case, but if you're happy with it feel free to hold cancel. Everything seems to work fine for me locally with this, and when i back out the commit that disabled the button.

/lgtm
/hold

sippy-ng/src/component_readiness/CompReadyVars.js
sosiouxme: This was what we mentioned yesterday, which I found out the hard way when I had the same reaction... have to keep app state and url state separate. When I tried using `variantCrossCompareParam` for app state, every time I checked a box it would update the url params and thus the page. So instead, in-app state keeps track of the UI, and then when you `Generate Report` the Param gets updated.

sippy-ng/src/component_readiness/IncludeVariantCheckboxList.js
sosiouxme: what, the ternary?

mainly just trying to make it readable... very messy to mix logic and elements

sippy-ng/src/component_readiness/IncludeVariantCheckboxList.js
sosiouxme: I am hoping that there will be no need to restrict it to one per side. You'll simply have basis variants and sample variants (e.g. IBM vs non-IBM arches). Tracing through the logic with Forrest yesterday, the only place where this seemed to break down is when referencing triage data, which could specify any variants, while the basis/sample would come back with data not grouped by the cross-compared variants, so we can't always correlate the triage data (while we could if sample/basis only had one variant each). But we concluded as a first approximation that for cross-compare scenarios, triage data may not be important.

sosiouxme (commented):
thanks for reviewing!

sippy-ng/src/component_readiness/IncludeVariantCheckboxList.js
sosiouxme: the side-by-side job comparison on test details will also be messy, since the job names could vary fairly wildly across variants. but that would really be the case whether there were single or multiple variants on each side, and again as a first pass i think we'll still have something useful.

stbenjam:
/lgtm

I tested it locally and everything seems to work.  This JS looks good to me too.

sippy-ng/src/component_readiness/IncludeVariantCheckboxList.js
dgoodwin: No the functions injecting more jsx, didn't realize we could do that.

sippy-ng/src/component_readiness/IncludeVariantCheckboxList.js
dgoodwin: On test details I think that's fine if virtually no jobs are matched, it's really about the overall data and links to job runs.

If you can pull this off with multi comparing to multi that sounds awesome.

PR #1926, opened 2024-08-15
TRT-1619: cross-variant comparison - pre-refactoring API

sosiouxme:
for the UI I refactored as I went along, which left the PR a bit messy. this time I think it's wise to do the obvious refactoring up front before updating API logic.

these are in separate commits for reviewing simplicity (also recommend ignoring whitespace):
* separate out component-readiness types into a `crtype` package and trim the names for legibility.
* separate out test_details logic out from component_report logic (same package though)
* refactor parameter parsing and passage

all the tests run. the net result should be no change to logic at all.

I would have liked to refactor the `componentReportGenerator` type too but that's a cache key and has 48 direct usages... it would be a tar pit. speaking of caches, though, would be good to verify that I haven't already broken caching with all the name changes.


dgoodwin:
first two bullet points are precisely what I was working on in shift week when i asked if anyone had substantial changes underway. I burned out and gave up on it, so this is absolutely awesome to see. I'd like to land is asap.

Caching still seems fine in local testing, subsequent requests come back immediately. 

/lgtm

PR #1927, opened 2024-08-15
TRT-1574: Server side views UI

dgoodwin:
This change adds the UI work for server side views. If a views file is provided to the server with `--component-readiness-views views.yaml`, a dropdown will appear at the top of the component readiness inputs. Bu default we will pre-select the first view matching our default sample release using the same logic we do today. Selecting a view pre-populates all of the inputs with the values from that view, including concrete versions of the relative timestamps in the view. (these are returned from the UI when views are listed, so the logic remains in one place) This allows users to easily customize a report from the base values of a view. Doing so will unselect any view in the dropdown, and selecting another view will take you directly back to that view.

As before, we start with short top level URLs for the top level report, ?view=foo. Drilling down results in fully qualified links as before.

Metrics and regression tracking are not yet using views, this will be completed in a followup.

If no views file is provided, the dropdown should not be present and the UI should function as before, allowing this to roll out and then we can enable with a PR for views.yaml ConfigMap in DPCR repo.

The views.yaml we'll deploy is added in this PR, allowing teams to self-service views they want in a PR and for us to validate the views config by hooking it into our e2e test jobs on this repo. As a result I've also added a very basic sanity check that a component readiness request is working in the e2e suite.


e2e-scripts/sippy-e2e-sippy-e2e-setup-commands.sh
```diff
@@ -185,12 +185,12 @@ spec:
         imagePullPolicy: ${SIPPY_IMAGE_PULL_POLICY:-Always}
         resources:
           limits:
-            memory: 1G
+            memory: 3G
```
dgoodwin, line 188: There was evidence we were OOMing on the first attempt and having to retry to get through the import.

sippy-ng/src/component_readiness/CompReadyMainInputs.js
```diff
@@ -80,8 +83,44 @@ export default function CompReadyMainInputs(props) {
       ></AdvancedOptions>
     </div>
   )
+
+  // eslint-disable-next-line react/prop-types
+  function ViewPicker({ enabled }) {
```
stbenjam, line 88: This should probably be in it's own file, IMHO it makes it a little easier to navigate the code base without an IDE telling you where this is

sippy-ng/src/component_readiness/CompReadyMainInputs.js
```diff
@@ -80,8 +83,44 @@ export default function CompReadyMainInputs(props) {
...
+    if (!enabled) return null
+    return (
+      <div className={classes.crRelease}>
+        <FormControl variant="standard">
```
stbenjam, line 92: View gets very small when it's empty, this should make it look a little better

```suggestion
        <FormControl variant="standard" sx={{ minWidth: 100 }}>
```

stbenjam (commented):
This is awesome, I can't wait to create one for QE.

This works really well in my testing, a few small comments to think about

sippy-ng/src/component_readiness/CompReadyMainInputs.js
```diff
@@ -93,10 +132,10 @@ export default function CompReadyMainInputs(props) {
           <Tooltip
             title={
               'Click here to generate a report that compares the release you wish to evaluate\
-               against a historical (previous) release'
+                                         against a historical (previous) release using all the specific parameters specified'
             }
           >
-            <Fragment>Generate Report</Fragment>
+            <Fragment>Generate Custom Report</Fragment>
```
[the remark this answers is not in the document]
dgoodwin: Lets discuss on team sync, I'm a little hesitant on the complexity there when we could make it smaller in much simpler ways, and the views selection behavior is kind of what I'd expect.

e2e-scripts/sippy-e2e-sippy-e2e-setup-commands.sh
```diff
@@ -185,12 +185,12 @@ spec:
...
         terminationMessagePath: /dev/termination-log
         terminationMessagePolicy: File
         command:  ["/bin/sh", "-c"]
         args:
-          - /bin/sippy load --init-database --load-openshift-ci-bigquery    --log-level=debug --release 4.14 --database-dsn=postgresql://postgres:password@postgres.sippy-e2e.svc.cluster.local:5432/postgres --mode=ocp --config ./config/e2e-openshift.yaml --google-service-account-credential-file /tmp/secrets/gcs-cred
+          - /bin/sippy load --init-database --load-openshift-ci-bigquery    --log-level=debug --release 4.16 --database-dsn=postgresql://postgres:password@postgres.sippy-e2e.svc.cluster.local:5432/postgres --mode=ocp --config ./config/e2e-openshift.yaml --google-service-account-credential-file /tmp/secrets/gcs-cred --component-readiness-views ./config/views.yaml
```
stbenjam, line 193: Could this option just be called `--views`? I could see us implementing views for other parts of sippy eventually.

config/views.yaml
```diff
@@ -0,0 +1,150 @@
+---
```
stbenjam, line 1: WDYT about adding a top-level key,

```
---
component_readiness:
  <views>
```

sippy-ng/src/component_readiness/CompReadyMainInputs.js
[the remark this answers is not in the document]
stbenjam: I was really making 3 separate comments, hiding/showing the button wasn't really the fix for the size.  I have concerns about usability where I think people are going to change the view and then immediately click the button unnecessarily.

Leaving the view selector populated is probably fine.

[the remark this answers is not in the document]
stbenjam: Alternative: maybe move the button below the release options? It's now no longer the most important thing on the page, so de-emphasizing it and moving it away from the view selector also probably solves it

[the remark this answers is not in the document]
dgoodwin: We decided on team meeting to do the last suggestion to keep the code a little simpler.

sippy-ng/src/component_readiness/CompReadyMainInputs.js
dgoodwin: Done

sippy-ng/src/component_readiness/CompReadyMainInputs.js
dgoodwin: Done

e2e-scripts/sippy-e2e-sippy-e2e-setup-commands.sh
dgoodwin: Done

config/views.yaml
dgoodwin: Done

dgoodwin:
Feedback addressed, and cross compare merged in. 

I discovered included variants were not syncing to the controls (or perhaps this broke with cross compare), so I hooked up clearing those URL params when view selected and populating controls with the view settings. I've also added a ha-vs-single view.

Still trying to get e2e to pass, moving to 4.16 seems too slow to import even with doubled timeouts. Took it back to 4.14, so we'll see if the test actually runs now.

dgoodwin:
/hold

The cross compare view may have exposed why we're cache missing, I was getting very different results, investigating.

sippy-ng/src/component_readiness/CompReadyUtils.js
```diff
@@ -291,7 +295,7 @@ export function formatLongDate(aLongDate, aDateFormat) {
     console.log('Error: unknown date format: ', typeof aLongDate)
     dateObj = new Date(aLongDate)
   }
-  return format(dateObj, aDateFormat)
+  return formatInTimeZone(dateObj, 'UTC', aDateFormat)
```
dgoodwin, line 298: This is a little dicey, but essentially we'd give the function a string like 2024-08-23T00:00:00Z

new Date() would parse that and if you printed it you'd get a date in your local time, i.e. on the 22nd but -3 for my timezone. 

then format would trim that to 2024-08-22T00:00:00 resulting in a control using a day before we told it to.

This should fix it and I *think* the UI is now behaving as you'd expect?

dgoodwin: Naturally caused problems. Seems the default date when you load in initially and get the default view is (a) not coming from the view, and (b) now today + 1 day somehow. Possible no controls are populated from default view on initial load.

dgoodwin:
Ok I believe I've fixed all known issues, the view is now properly populating controls from the default view on initial load, and pressing generate report will now hit the cache, the dates were off and that is probably why we were missing.

This should be ready for review again.

stbenjam:
On the test details page, the view dropdown appears -  can we hide it? We only want the date selectors on that page

I think you can do something similar to https://github.com/openshift/sippy/blob/37ec67dfaf9ebca488c356f63e95c5264955d7d9/sippy-ng/src/component_readiness/CompReadyMainInputs.js#L135

stbenjam:
If you click the link above, it loads -- but when I refresh the page I get a 500. The error makes me think it's something to do with cross-variant compare:

```
Error: Return code = 500 (error querying component test details from big query: [all dbGroupBy variants have to be defined for test details: Topology is missing])
```

config/views.yaml
```diff
@@ -0,0 +1,202 @@
...
+      db_group_by:
+        Architecture: {}
+        FeatureSet: {}
+        Installer: {}
+        Network: {}
+        Platform: {}
+        Suite: {}
+        Topology: {}
+        Upgrade: {}
+      include_variants:
```
stbenjam, line 26: We can deal with this later, but can 4.17+ drop `sdn` from the include networking variants?

dgoodwin:
Added a condition for display of the views dropdown.

Good find on the refresh, I think it's the assumption that when loading views we must be fresh entering the page and thus can syncView, but we may have been given fully qualified params that then wipes out. I'll try to figure out how to tell if we need to set a view on load or not.

dgoodwin:
Nightmare sequence of problems from that can of worms but I'm hopeful it's now stable. 

Found a bug we actually have in master where if you drill down into a component, then refresh, the column group by control loses it's defaults because we claimed it's an ArrayParam when it's actually a String with commas.

Fought with all kinds of date and timezone problems. I think we still have bugs here and should be using ISO8601 from the server view anyhow, all the time, but for now it works and cache hits are working too.

Refactored variable params vs variables such that the params should be undefined if for example, baseRelease= is not in the url query params. (baseReleaseParam thus should be undefined)

sippy-ng/src/component_readiness/CompReadyUtils.test.js
stbenjam, line 1: 💚

stbenjam:
I found one weird bug that we can probably handle separately.  I suspect it's somehow related to isLoaded  Otherwise this seems to work well for me.

1. Visit http://localhost:3000/sippy-ng/component_readiness/test_details?Aggregation=none&Architecture=amd64&Architecture=amd64&FeatureSet=default&FeatureSet=default&Installer=upi&Installer=upi&Network=ovn&Network=ovn&NetworkAccess=default&Platform=vsphere&Platform=vsphere&Scheduler=default&SecurityMode=default&Suite=unknown&Suite=unknown&Topology=ha&Topology=ha&Upgrade=none&Upgrade=none&baseEndTime=2024-06-27%2023%3A59%3A59&baseRelease=4.16&baseStartTime=2024-05-28%2000%3A00%3A00&capability=Operator&columnGroupBy=Architecture%2CNetwork%2CPlatform&compareVariant=Topology%3Asingle&component=Management%20Console&confidence=95&dbGroupBy=Platform%2CArchitecture%2CNetwork%2CFeatureSet%2CUpgrade%2CSuite%2CInstaller&environment=amd64%20default%20upi%20ovn%20vsphere%20unknown%20ha%20none&ignoreDisruption=true&ignoreMissing=false&includeVariant=Architecture%3Aamd64&includeVariant=FeatureSet%3Adefault&includeVariant=Installer%3Aipi&includeVariant=Installer%3Aupi&includeVariant=Owner%3Aeng&includeVariant=Platform%3Aaws&includeVariant=Platform%3Aazure&includeVariant=Platform%3Agcp&includeVariant=Platform%3Ametal&includeVariant=Platform%3Avsphere&includeVariant=Topology%3Aha&minFail=3&pity=5&sampleEndTime=2024-09-05%2023%3A59%3A59&sampleRelease=4.17&sampleStartTime=2024-08-29%2000%3A00%3A00&testId=openshift-tests%3Af9552d2b4b817d19bdcd666a7b8d1ece&testName=%5Bbz-Management%20Console%5D%20clusteroperator%2Fconsole%20should%20not%20change%20condition%2FAvailable&variantCrossCompare=Topology
2. Close the CR sidebar
3. Click "Component Readiness" (to go back to main page)
4. It says it's loading, but it never actually makes the API call



https://github.com/user-attachments/assets/5bb4b5ac-f527-48db-9555-b39aa8d4ff70

dgoodwin:
/hold cancel

I'll draft PR for dpcr to enable the flag and use them shortly.

PR #1949, opened 2024-08-28
api bugfix: race condition in sample/base queries

sosiouxme:
base and sample queries run in goroutines, based on the same parameter variables that were updated in between; sometimes this would result in both queries being the same. by separating the variables this is prevented.

(this was for test_details only; regular CR pages already came from correctly separated queries)


stbenjam:
Nice find

/lgtm

PR #1955, opened 2024-08-29
trt-1788: Ensure IncludeVariants not otherwise referenced are part of test_details query

neisw:
- Owner was not part of the TestDetails query and we were picking up non eng jobs.  
- Triage is based on TestDetails and perfscale job was added added to the triage job runs
- ComponentReadiness detected more triaged job runs than failures and skipped triage
- Fixed TD query to honor IncludeVariants not referenced elsewhere
- Found periodic-ci-openshift-cluster-control-plane-machine-set-operator-release-4.17-periodics-e2e-aws-arm showing up as amd64, updated variant parsing

### DIFF
--- a/pkg/variantregistry/ocp.go
+++ b/pkg/variantregistry/ocp.go
@@ -245,7 +245,7 @@ var (
 	// We're not sure what these aggregator jobs are but they exist as of right now:
 	aggregatorRegex = regexp.MustCompile(`(?i)aggregator-`)
 	alibabaRegex    = regexp.MustCompile(`(?i)-alibaba`)
-	arm64Regex      = regexp.MustCompile(`(?i)-arm64|-multi-a-a`)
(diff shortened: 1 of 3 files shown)


pkg/variantregistry/ocp.go
```diff
@@ -245,7 +245,7 @@ var (
 	// We're not sure what these aggregator jobs are but they exist as of right now:
 	aggregatorRegex = regexp.MustCompile(`(?i)aggregator-`)
 	alibabaRegex    = regexp.MustCompile(`(?i)-alibaba`)
-	arm64Regex      = regexp.MustCompile(`(?i)-arm64|-multi-a-a`)
+	arm64Regex      = regexp.MustCompile(`(?i)-arm64|-multi-a-a|-arm`)
```
neisw, line 248: Technically we could just remove `-arm64` in place of `-arm` but seems like the `-arm` job might be better off to be renamed at some point.

PR #1959, opened 2024-08-30
TRT-1784: Add variants for container runtime and cgroupsversion

xueqzhan:


pkg/variantregistry/ocp.go
```diff
@@ -436,6 +440,18 @@ func (v *OCPVariantLoader) IdentifyVariants(jLog logrus.FieldLogger, jobName str
...
+		variants[VariantContainerRuntime] = "crun"
+	} else {
+		variants[VariantContainerRuntime] = "runc"
+	}
+
+	if cgroupsv1Regex.MatchString(jobName) {
+		variants[VariantCGroupsVersion] = "v1"
+	} else {
+		variants[VariantCGroupsVersion] = "v2"
+	}
```
dgoodwin, line 453: Can you track down what the value is that gets set in a cluster to turn on v1 or v2? Should be visible in their job configuration, or you could ask the requestor. It might end up helpful at some point to have the variant value match the actual cluster value in case anyone ever is reading it from the cluster. (a future cluster-data.json enhancement perhaps)

PR #1968, opened 2024-09-03
trt-1792: Add all regressions view to RegressedTestsModal.js

neisw:
- Moved RegressedTests to its own panel
- Consolidated code creating regression data arrays
- Added new tab with all regressed tests
- Refactored Triaged Incidents Test Failures columns to better align with Regressed Tests

image

image


sosiouxme (approved):
/lgtm
more removed than added :)

PR #1976, opened 2024-09-06
TRT-1799:  Support previous values for allowed base regressions

neisw:
- Refactor to record IntentionalRegression success, flake and failure for previous and regressed values
- Support loading IntentionalRegressions via JSON
- - Part of [TRT-1448](https://issues.redhat.com/browse/TRT-1448) 
- - See https://github.com/openshift/sippy/commit/2ddb15cc01dee6ee43a351565b11856e0d019d92 for example
- Support overriding base total, success, flake and failure values with previous values for allowed regressions in the base release


pkg/regressionallowances/types.go
```diff
@@ -128,3 +147,26 @@ func addIntentionalRegression(release release, in IntentionalRegression) error {
 
 	return nil
 }
+
```
neisw, line 150: Groundwork for [TRT-1448](https://issues.redhat.com/browse/TRT-1448) adds support for json array of intentional regressions

neisw:
/retitle TRT-1799:  Support previous values for allowed base regressions

neisw:
linter error is not related, needs to be addressed outside of this pr

`path-to-regexp  0.2.0 - 7.1.0
Severity: high
path-to-regexp outputs backtracking regular expressions - https://github.com/advisories/GHSA-9wv6-86v2-598j
fix available via `npm audit fix --force`
Will install react-router-dom@6.26.2, which is a breaking change`

stbenjam:
Looks good to me, do you think we could include one of the examples from https://github.com/openshift/sippy/commit/2ddb15cc01dee6ee43a351565b11856e0d019d92 commented out in the file so there's a readily accessible example?

neisw:
> Looks good to me, do you think we could include one of the examples from [2ddb15c](https://github.com/openshift/sippy/commit/2ddb15cc01dee6ee43a351565b11856e0d019d92) commented out in the file so there's a readily accessible example?

Added some commented out JSON.  Hope is that goes away once we have example PR / real regressions added.

pkg/api/componentreadiness/component_report.go
```diff
@@ -1345,16 +1345,20 @@ func (c *componentReportGenerator) generateComponentTestReport(baseStatus map[st
...
+			var approvedRegression, baseRegression *regressionallowances.IntentionalRegression
 			if len(c.VariantCrossCompare) == 0 { // only really makes sense when not cross-comparing variants:
 				// look for corresponding regressions we can account for in the analysis
 				approvedRegression = regressionallowances.IntentionalRegressionFor(c.SampleRelease.Release, testID.ColumnIdentification, testID.TestID)
-				resolvedIssueCompensation, triagedIncidents = c.triagedIncidentsFor(testID)
+				baseRegression = regressionallowances.IntentionalRegressionFor(c.BaseRelease.Release, testID.ColumnIdentification, testID.TestID)
+				// ignore triage if we have an intentional regression
+				if approvedRegression == nil {
+					resolvedIssueCompensation, triagedIncidents = c.triagedIncidentsFor(testID)
+				}
```
stbenjam, line 1356: I don't notice any time window used here - how does it work if I were to select a different time value for 4.17 where the regression wasn't present?

Don't we want this to only work in the GA window case?

neisw: The IntentionalRegressions are expected to be added at the end of the release during the GA timeframe and lower the bar for the entire release.  That's not new with this PR.  We could build your case out if we needed but I think that is outside of what this PR is trying to do.

When adding the [base regression override](https://github.com/openshift/sippy/blob/f1363e049ed106202f2a99432af68f69dc9a79d4/pkg/api/componentreadiness/component_report.go#L1588) it should only kick in when the computed base values are less than the previous values.  So if we see the base improve we won't then lower to the previous values.

stbenjam: I don't think that's great protection and going to result in confusion.  We won't really be able to answer questions like "what was 4.17 like in the middle of the development cycle? At the beginning?" Or even slice the last month of GA because this is such a big hammer.

deads2k:
/approve

for the shape change

stbenjam:
/hold cancel

Was just checking it'll still build if lint was overriden

PR #1977, opened 2024-09-06
TRT-1784: Add CGroupMode and ContainerRuntime variants to CR

xueqzhan:


pkg/api/componentreadiness/component_report.go
```diff
@@ -135,6 +135,8 @@ var (
 		"Platform:metal",
 		"Platform:vsphere",
 		"Topology:ha",
+		"CGroupMode:v2",
+		"ContainerRuntime:runc",
```
dgoodwin, line 139: I think you now need to plug them into the config/views.yaml for this to actually take effect, this change will only work if no views are configured on the sippy deployment which is not the case for us.

dgoodwin: Do they also need to go into DBGroupBy?

xueqzhan: I do not think we want to add those to DBGroupBy for the following reasons. 

1. Whatever is added to DBGroupBy will end up in the column names in the last page before the test details page. We already have many variants there. 
2. Typical use case (like default) will only pick one of the valid case (e.g. CGroupMode:v2). If somebody really cares about the difference between the two, it is better to use the cross comparison mode. 
3. Those two variants are more transitional than permanent. For example, once crun is mature, most of the jobs probably will use that. It is only during transitional period when we are intersted in the differences between crun and runc.

PR #1984, opened 2024-09-09
Convert metrics and regression tracking to be view based

dgoodwin:
This adds consistency to our reporting and ensures we hit the cache more
often, leading to faster response times.

Metrics can now be enabled for multiple releases, or even auxiliary
views if a team is interested in getting alerts for their regressions vs
mainline.

Regression tracking can only be enabled for one view for each release,
-main is assumed. This allows us to always show the mainline regression
tracking data regardless what custom report you might be viewing.


dgoodwin:
/hold

Just in case this passes review quickly, I want to snapshot the regression tables right before rollout.

stbenjam:
> Regression tracking can only be enabled for one view for each release,
-main is assumed. This allows us to always show the mainline regression
tracking data regardless what custom report you might be viewing.

Do we have any plans to change that?

/lgtm

dgoodwin:
/override lint
/hold cancel

Snapshot was taken.

PR #1986, opened 2024-09-10
TRT-1807: CVE-2024-45296 update path-to-regexp

stbenjam:
There's a CVE in path-to-regexp. Ran `npm audit fix`.  In the end, the fixes were backported to work with react-router v5 but it wasn't clear that was going to happen: we need to move to v6.

Also the CVE itself is a nothing burger, we might want to move to something else to let us skip CVE's that aren't applicable (npm audit out of the box doesn't do that).

### DIFF
--- a/sippy-ng/package-lock.json
+++ b/sippy-ng/package-lock.json
@@ -4143,10 +4143,11 @@
       "dev": true
     },
     "node_modules/@jridgewell/trace-mapping": {
-      "version": "0.3.19",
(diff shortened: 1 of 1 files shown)


stbenjam:
This broke some pages... will come up with a different solution tomorrow.

/hold

stbenjam:
/hold cancel

It's all working now

xueqzhan:
/lgtm

Wow, how many versions of versions of path-to-regexp are we running here!

PR #1991, opened 2024-09-12
Add regressions for 4.17

neisw:
Add regression to clear the regression signal and preserve the previous releases baseline

- Add JSON file
- Embed the JSON in intentional_regressions.go
- Import the regressions
- If the file exists and regressions are already being imported new regressions can be added to the existing release specific regressions file

```
//go:embed regressions_4.17.json
var regressions4_17 []byte

func init() {
importIntentionalRegressions(release417, regressions4_17)
}
```

### DIFF
--- a/pkg/regressionallowances/regressions_4.17.json
+++ b/pkg/regressionallowances/regressions_4.17.json
@@ -29,7 +29,7 @@
     "TestID": "openshift-tests:d6b41cee7afca1c2a0b52f9e6975425f",
     "TestName": "[bz-kube-apiserver][invariant] alert/KubeAPIErrorBudgetBurn should not be at or above info",
     "JiraBug": "https://issues.redhat.com/browse/OCPBUGS-42083",
-    "ReasonToAllowInsteadOfFix": "TBD Reason",
+    "ReasonToAllowInsteadOfFix": "Degraded near the end of stabilization.  Investigation indicated not a release blocker.",
     "variant": {
       "variants": {
         "Network": "ovn",
--- a/pkg/regressionallowances/regressions_4.17.json
+++ b/pkg/regressionallowances/regressions_4.17.json
@@ -20,7 +20,32 @@
     "PreviousSuccesses": 66,
     "PreviousFailures": 0,
     "PreviousFlakes": 1,
-    "RegressedSuccesses": 27,
+    "RegressedSuccesses": 25,
+    "RegressedFailures": 3,
+    "RegressedFlakes": 0
+  },
+  {
+    "JiraComponent": "Management Console",
+    "TestID": "openshift-tests:f9552d2b4b817d19bdcd666a7b8d1ece",
+    "TestName": "[bz-Management Console] clusteroperator/console should not change condition/Available",
+    "JiraBug": "https://issues.redhat.com/browse/OCPBUGS-39580",
+    "ReasonToAllowInsteadOfFix": "Test flaked in 4.16, became more strict in 4.17",
+    "variant": {
+      "variants": {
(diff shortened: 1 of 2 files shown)


pkg/regressionallowances/regressions_4.17.json
```diff
@@ -0,0 +1,77 @@
+[
+  {
+    "JiraComponent": "Management Console",
+    "TestID": "openshift-tests:f9552d2b4b817d19bdcd666a7b8d1ece",
+    "TestName": "[bz-Management Console] clusteroperator/console should not change condition/Available",
+    "JiraBug": "https://issues.redhat.com/browse/OCPBUGS-39580",
+    "ReasonToAllowInsteadOfFix": "TBD Reason",
```
candita, line 7:
```suggestion
    "ReasonToAllowInsteadOfFix": "Test flaked in 4.16, became more strict in 4.17",
```

pkg/regressionallowances/regressions_4.17.json
```diff
@@ -0,0 +1,77 @@
...
+    "RegressedSuccesses": 18,
+    "RegressedFailures": 3,
+    "RegressedFlakes": 0
+  },
+  {
+    "JiraComponent": "Management Console",
+    "TestID": "openshift-tests:f9552d2b4b817d19bdcd666a7b8d1ece",
+    "TestName": "[bz-Management Console] clusteroperator/console should not change condition/Available",
+    "JiraBug": "https://issues.redhat.com/browse/OCPBUGS-39580",
+    "ReasonToAllowInsteadOfFix": "TBD Reason",
```
candita, line 32:
```suggestion
    "ReasonToAllowInsteadOfFix": "Test flaked in 4.16, became more strict in 4.17",
```

pkg/regressionallowances/regressions_4.17.json
```diff
@@ -0,0 +1,52 @@
...
+    "RegressedSuccesses": 27,
+    "RegressedFailures": 4,
+    "RegressedFlakes": 0
+  },
+  {
+    "JiraComponent": "kube-apiserver",
+    "TestID": "openshift-tests:d6b41cee7afca1c2a0b52f9e6975425f",
+    "TestName": "[bz-kube-apiserver][invariant] alert/KubeAPIErrorBudgetBurn should not be at or above info",
+    "JiraBug": "https://issues.redhat.com/browse/OCPBUGS-42083",
+    "ReasonToAllowInsteadOfFix": "TBD Reason",
```
brenton, line 32: Since I filed this bug I just want to chime with my opinion:

This bug was found during the final stabilization period of our release.  We believe this is a real problem and must be fixed, but investigation suggested it was not severe enough to change our release date.

neisw:
/hold

Preference is for responsible teams to open separate PRs allowing regressions.
TRT can help produced the JSON and work with teams validating their PRs.

candita:
@neisw please note that the regressions for https://issues.redhat.com/browse/OCPBUGS-39580 are no longer needed.  The issue was traced to an environmental problem with a DNS forwarder and I closed the bug.

neisw:
Thank you @candita .  [CR](https://sippy.dptools.openshift.org/sippy-ng/component_readiness/main?regressedModal=1&view=4.17-main) still shows Management Console as regressed so until the failures are outside of the sample window we will likely still need to add a regression to get the 'all clear'.  It looks like at least two of the failures might roll off tomorrow.  If not, then the reason can be added explaining it was environmental.

neisw:
/close
In favor of https://github.com/openshift/sippy/pull/2015

PR #1993, opened 2024-09-15
TRT-1811: fallback queries

neisw:
- Queries previous releases without variant, capability or testid filtering
- Caches results so navigating down through capabilities doesn't cause the query to run again
- Compares stats from previous releases based on testIdentification and selects the one with the highest pass rate
- Includes Advanced Option for `Historical release analysis` default is exclude currently
- Fallback test details shows both the queried base release status and fallback when applicable
  - Override Release Assessment
  - Fallback Release Assessment
  - Fallback Release basis with dates
  - Override Release basis with dates (new tab)

image

When included the regressed since column will be empty since regression tracking is currently not aware of these

image

Test Details Report will show tabs for the requested basis as well as the fallback / historical basis.  It will indicate an override as well as an explanation

image

### DIFF
--- a/pkg/api/componentreadiness/component_report.go
+++ b/pkg/api/componentreadiness/component_report.go
@@ -719,19 +719,15 @@ func (f *fallbackTestQueryReleasesGenerator) getTestFallbackReleases(ctx context
 	var selectedReleases []*crtype.Release
 	fallbackRelease := f.BaseRelease
 
-	// Get base plus up to 3 fallback releases
-	for i := 0; i < 4; i++ {
+	// Get up to 3 fallback releases
+	for i := 0; i < 3; i++ {
 		var crRelease *crtype.Release
 
-		// currently fetching the baseRelease via fallback to
-		// perform validation of the test data
-		// this gets uncommented when ready
-		// and the section at the bottom removed
-		// fallbackRelease, err = previousRelease(fallbackRelease)
-		// if err != nil {
-		// 	log.WithError(err).Errorf("Failure determining fallback release for %s", fallbackRelease)
-		// 	continue
-		// }
+		fallbackRelease, err := previousRelease(fallbackRelease)
+		if err != nil {
+			log.WithError(err).Errorf("Failure determining fallback release for %s", fallbackRelease)
+			break
+		}
 
 		for i := range releases {
 			if releases[i].Release == fallbackRelease {
@@ -742,16 +738,6 @@ func (f *fallbackTestQueryReleasesGenerator) getTestFallbackReleases(ctx context
 
 		if crRelease != nil {
 			selectedReleases = append(selectedReleases, crRelease)
-		}
-
-		// Attempt to get the previous release
-		// remove this section if / when we uncomment previousRelease call above
-		// to bypass getting the original base fallback results for validation
-		var err error
-		fallbackRelease, err = previousRelease(fallbackRelease)
-		if err != nil {
-			log.WithError(err).Errorf("failure determining fallback release for %s", fallbackRelease)
-			break // Stop attempting if previousRelease fails
 		}
 	}
 
@@ -1796,26 +1782,6 @@ func (c *componentReportGenerator) generateComponentTestReport(ctx context.Conte
 				if approvedRegression == nil {
 					resolvedIssueCompensation, triagedIncidents = c.triagedIncidentsFor(ctx, testID)
 				}
-			}
-
-			// this goes away and is just to validate
-			// that the fallback data for the base release
-			// isn't missing any tests
-			if c.cachedFallbackTestStatuses != nil {
-				// get fallback for the current base, make sure we don't have any misses'
-				if cachedTestStatuses, ok := c.cachedFallbackTestStatuses.Releases[c.BaseRelease.Release]; ok {
-					if cTestStats, ok := cachedTestStatuses.Tests[testIdentification]; ok {
-						// obviously we need more analysis here to determine if the
-						// previous release(s) stats should override
-						baseStats = cTestStats
-						baseReleaseMatches++
-					} else {
-						baseReleaseMisses++
-					}
-				}
-
-			} else if c.IncludeMultiReleaseAnalysis {
-				log.Error("Expected cached fallback statuses but was missing")
 			}
 
 			// this is where we look to see if a previous release has a higher pass rate
--- a/pkg/api/componentreadiness/component_report.go
+++ b/pkg/api/componentreadiness/component_report.go
@@ -11,6 +11,8 @@ import (
 	"sync"
 	"time"
 
+	"github.com/openshift/sippy/pkg/util"
+
 	"cloud.google.com/go/bigquery"
 	"github.com/apache/thrift/lib/go/thrift"
 	fischer "github.com/glycerine/golang-fisher-exact"
@@ -115,6 +117,7 @@ const (
 					ELSE jobs.prowjob_job_name
 		    	END AS variant_registry_job_name,
 `
+	fallbackQueryTimeRoundingOverride = 12 * time.Hour
 )
 
 type GeneratorType string
@@ -146,6 +149,13 @@ var (
 	DefaultIgnoreMissing    = false
 	DefaultIgnoreDisruption = true
 )
+
+func newFallbackReleases() crtype.FallbackReleases {
+	fb := crtype.FallbackReleases{
+		Releases: map[string]crtype.ReleaseTestMap{},
+	}
+	return fb
+}
 
 func getSingleColumnResultToSlice(ctx context.Context, query *bigquery.Query) ([]string, error) {
 	names := []string{}
@@ -179,6 +189,14 @@ func GetComponentTestVariantsFromBigQuery(ctx context.Context, client *bqcachedc
 
 	return api.GetDataFromCacheOrGenerate[crtype.TestVariants](ctx, client.Cache, cache.RequestOptions{},
 		api.GetPrefixedCacheKey("TestVariants~", generator), generator.GenerateVariants, crtype.TestVariants{})
+}
+
+func GetReleaseDatesFromBigQuery(ctx context.Context, client *bqcachedclient.Client) ([]crtype.Release, []error) {
+	generator := componentReportGenerator{
+		client: client,
+	}
+
+	return api.GetDataFromCacheOrGenerate[[]crtype.Release](ctx, client.Cache, cache.RequestOptions{}, api.GetPrefixedCacheKey("CRReleaseDates~", generator), generator.GenerateReleaseDates, []crtype.Release{})
 }
 
 func GetJobVariantsFromBigQuery(ctx context.Context, client *bqcachedclient.Client,
@@ -224,14 +242,16 @@ func GetComponentReportFromBigQuery(ctx context.Context, client *bqcachedclient.
 // is marshalled for the cache key and should be changed when the object being
 // cached changes in a way that will no longer be compatible with any prior cached version.
 type componentReportGenerator struct {
-	ReportModified *time.Time
-	client         *bqcachedclient.Client
-	prowURL        string
-	gcsBucket      string
-	cacheOption    cache.RequestOptions
-	BaseRelease    crtype.RequestReleaseOptions
-	SampleRelease  crtype.RequestReleaseOptions
-	triagedIssues  *resolvedissues.TriagedIncidentsForRelease
+	ReportModified             *time.Time
+	client                     *bqcachedclient.Client
+	prowURL                    string
+	gcsBucket                  string
+	cacheOption                cache.RequestOptions
+	BaseRelease                crtype.RequestReleaseOptions
+	BaseOverrideRelease        crtype.RequestReleaseOptions
+	SampleRelease              crtype.RequestReleaseOptions
+	triagedIssues              *resolvedissues.TriagedIncidentsForRelease
+	cachedFallbackTestStatuses *crtype.FallbackReleases
 	crtype.RequestTestIdentificationOptions
 	crtype.RequestVariantOptions
 	crtype.RequestAdvancedOptions
@@ -267,6 +287,24 @@ func (c *componentReportGenerator) GenerateVariants(ctx context.Context) (crtype
 		Upgrade:  columns["upgrade"],
 		Variant:  columns["variants"],
 	}, errs
+}
+
+func (c *componentReportGenerator) GenerateReleaseDates(ctx context.Context) ([]crtype.Release, []error) {
+	releases, err := api.GetReleasesFromBigQuery(ctx, c.client)
+	if err != nil {
+		return nil, []error{err}
+	}
+	crReleases := []crtype.Release{}
+	for _, release := range releases {
+		crRelease := crtype.Release{Release: release.Release}
+		if release.GADate != nil {
+			prior := util.AdjustReleaseTime(*release.GADate, true, "30", c.cacheOption.CRTimeRoundingFactor)
+			crRelease.Start = &prior
+			crRelease.End = release.GADate
+		}
+		crReleases = append(crReleases, crRelease)
+	}
+	return crReleases, nil
 }
 
 func (c *componentReportGenerator) GenerateJobVariants(ctx context.Context) (crtype.JobVariants, []error) {
@@ -361,7 +399,7 @@ func (c *componentReportGenerator) GenerateComponentReportTestStatus(ctx context
 }
 
 // getCommonTestStatusQuery returns the common query for the higher level summary component summary.
-func (c *componentReportGenerator) getCommonTestStatusQuery(allJobVariants crtype.JobVariants, isSample bool) (string, string, []bigquery.QueryParameter) {
+func (c *componentReportGenerator) getCommonTestStatusQuery(allJobVariants crtype.JobVariants, isSample, isFallback bool) (string, string, []bigquery.QueryParameter) {
 	// Parts of the query, including the columns returned, are dynamic, based on the list of variants we're told to work with.
 	// Variants will be returned as columns with names like: variant_[VariantName]
 	// See fetchTestStatus for where we dynamically handle these columns.
@@ -426,45 +464,49 @@ func (c *componentReportGenerator) getCommonTestStatusQuery(allJobVariants crtyp
 		queryString += ` AND NOT 'Disruption' in UNNEST(capabilities)`
 	}
 
-	variantGroups := c.IncludeVariants
-	// potentially cross-compare variants for the sample
-	if isSample && len(c.VariantCrossCompare) > 0 {
-		variantGroups = c.CompareVariants
-	}
-	if variantGroups == nil { // server-side view definitions may omit a variants map
-		variantGroups = map[string][]string{}
-	}
-
-	for group, variants := range variantGroups {
-		queryString += " AND ("
-		first := true
-		for _, variant := range variants {
-			if first {
-				queryString += fmt.Sprintf(`jv_%s.variant_value = '%s'`, group, variant)
-				first = false
-			} else {
-				queryString += fmt.Sprintf(` OR jv_%s.variant_value = '%s'`, group, variant)
-			}
+	// fallback queries get all variants with no filtering
+	// so all tests are fetched then cached
+	if !isFallback {
+		variantGroups := c.IncludeVariants
+		// potentially cross-compare variants for the sample
+		if isSample && len(c.VariantCrossCompare) > 0 {
+			variantGroups = c.CompareVariants
 		}
-		queryString += ")"
-	}
+		if variantGroups == nil { // server-side view definitions may omit a variants map
+			variantGroups = map[string][]string{}
+		}
 
-	for k, v := range c.RequestedVariants {
-		queryString += fmt.Sprintf(` AND jv_%s.variant_value = '%s'`, k, v)
-	}
-	if c.Capability != "" {
-		queryString += " AND @Capability in UNNEST(capabilities)"
-		commonParams = append(commonParams, bigquery.QueryParameter{
-			Name:  "Capability",
-			Value: c.Capability,
-		})
-	}
-	if c.TestID != "" {
-		queryString += ` AND cm.id = @TestId`
-		commonParams = append(commonParams, bigquery.QueryParameter{
-			Name:  "TestId",
-			Value: c.TestID,
-		})
+		for group, variants := range variantGroups {
+			queryString += " AND ("
+			first := true
+			for _, variant := range variants {
+				if first {
+					queryString += fmt.Sprintf(`jv_%s.variant_value = '%s'`, group, variant)
+					first = false
+				} else {
+					queryString += fmt.Sprintf(` OR jv_%s.variant_value = '%s'`, group, variant)
+				}
+			}
+			queryString += ")"
+		}
+
+		for k, v := range c.RequestedVariants {
+			queryString += fmt.Sprintf(` AND jv_%s.variant_value = '%s'`, k, v)
+		}
+		if c.Capability != "" {
+			queryString += " AND @Capability in UNNEST(capabilities)"
+			commonParams = append(commonParams, bigquery.QueryParameter{
+				Name:  "Capability",
+				Value: c.Capability,
+			})
+		}
+		if c.TestID != "" {
+			queryString += ` AND cm.id = @TestId`
+			commonParams = append(commonParams, bigquery.QueryParameter{
+				Name:  "TestId",
+				Value: c.TestID,
+			})
+		}
 	}
 	return queryString, groupString, commonParams
 }
@@ -480,9 +522,8 @@ type baseQueryGenerator struct {
 
 // getBaseQueryStatus builds the basis query, executes it, and returns the basis test status.
 func (c *componentReportGenerator) getBaseQueryStatus(ctx context.Context,
-	allJobVariants crtype.JobVariants) (map[string]crtype.
-	TestStatus, []error) {
-	baseQuery, baseGrouping, baseParams := c.getCommonTestStatusQuery(allJobVariants, false)
+	allJobVariants crtype.JobVariants) (map[string]crtype.TestStatus, []error) {
+	baseQuery, baseGrouping, baseParams := c.getCommonTestStatusQuery(allJobVariants, false, false)
 	generator := baseQueryGenerator{
 		client: c.client,
 		cacheOption: cache.RequestOptions{
@@ -504,6 +545,40 @@ func (c *componentReportGenerator) getBaseQueryStatus(ctx context.Context,
 	}
 
 	return componentReportTestStatus.BaseStatus, nil
+}
+
+func (c *componentReportGenerator) getFallbackBaseQueryStatus(ctx context.Context, allJobVariants crtype.JobVariants, release string, start, end time.Time) []error {
+	baseQuery, baseGrouping, baseParams := c.getCommonTestStatusQuery(allJobVariants, false, true)
+	generator := fallbackTestQueryReleasesGenerator{
+		client: c.client,
+		cacheOption: cache.RequestOptions{
+			ForceRefresh: c.cacheOption.ForceRefresh,
+			// increase the time that fallback queries are cached for
+			// could formalize as input flag
+			CRTimeRoundingFactor: fallbackQueryTimeRoundingOverride,
+		},
+		commonQuery:     baseQuery,
+		groupByQuery:    baseGrouping,
+		BaseRelease:     release,
+		BaseStart:       start,
+		BaseEnd:         end,
+		queryParameters: baseParams,
+	}
+
+	// default to main cache
+	// but if defined will use persistent cache
+	fcache := c.client.Cache
+	if c.client.PersistentCache != nil {
+		fcache = c.client.PersistentCache
+	}
+	cachedFallbackTestStatuses, errs := api.GetDataFromCacheOrGenerate[*crtype.FallbackReleases](ctx, fcache, generator.cacheOption, api.GetPrefixedCacheKey("FallbackReleases~", generator), generator.getTestFallbackReleases, &crtype.FallbackReleases{})
+
+	if len(errs) > 0 {
+		return errs
+	}
+
+	c.cachedFallbackTestStatuses = cachedFallbackTestStatuses
+	return nil
 }
 
 func (b *baseQueryGenerator) queryTestStatus(ctx context.Context) (crtype.ReportTestStatus, []error) {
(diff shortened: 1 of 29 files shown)


sippy-ng/src/component_readiness/AdvancedOptions.js
```diff
@@ -123,6 +129,13 @@ export default function AdvancedOptions(props) {
               name="ignoreDisruption"
               color="primary"
             />
+            <p>Fallback Basis: {ignoreFallbackBasis ? 'ignore' : 'keep'}</p>
+            <Switch
+              checked={ignoreFallbackBasis}
+              onChange={handleChangeIgnoreFallbackBasis}
+              name="ignoreFallbackBasis"
+              color="primary"
+            />
```
sosiouxme, line 138: could it please have a tooltip to explain what it's for

sippy-ng/src/component_readiness/CompReadyTestReport.js
```diff
@@ -504,87 +438,41 @@ View the test details report at ${document.location.href}
...
-      </TableContainer>
+      {isBaseOverride ? (
+        <Fragment>
+          <Tabs
+            value={activeTabIndex}
+            onChange={handleTabChange}
+            aria-label="Test Report Tabs"
+          >
+            <Tab label="Basis Tests" {...tabProps(0)} />
+            <Tab label="Override Tests" {...tabProps(1)} />
```
sosiouxme, line 449: These titles are too ambiguous for a newb. It's admittedly difficult to come up with better. Maybe:
```suggestion
            <Tab label="Severest Result" {...tabProps(0)} />
            <Tab label="Specified Result" {...tabProps(1)} />
```
Again some kind of explanation for newbs would be helpful... though I'm not sure tooltips work well on tabs

sosiouxme (commented):
rebase to include https://github.com/openshift/sippy/pull/2029

pkg/api/componentreadiness/component_report.go
```diff
@@ -608,6 +677,180 @@ func (s *sampleQueryGenerator) queryTestStatus() (crtype.ReportTestStatus, []err
...
+				if release.GADate != nil && release.GA30DayPrior != nil {
+					end = *release.GADate
+					start = *release.GA30DayPrior
+				}
+				break
+			}
+		}
+
+		wg.Add(1)
+		go func(queryRelease *crtype.Release, queryStart, queryEnd time.Time) {
```
stbenjam, line 736: This isn't the only case where we're not doing it, but we need to start propagating the context from the http Request down into component readiness and using it.  When a client disconnects, these still continue executing which isn't ideal.

At the minimum it makes sippy a target for a ddos attack (clients can quickly reconnect over and over again and spawn off many goroutines that stay running).

i.e.:

```golang
func (s *Server) jsonComponentReportFromBigQuery(w http.ResponseWriter, req *http.Request) {
    ctx := req.Context()

    // [....]
    componentreadiness.GetComponentReportFromBigQuery(ctx, ...)
    // [...]
}

func GetComponentReportFromBigQuery(ctx context.Context, ...) {
    // [...]
    go func() {
        select {
        case <-ctx.Done():
            return // terminate when context does
        default:
            // actual body of work to do
        }
    }()
    // [...]
}

```

stbenjam (commented):
- Can you share a localhost link to a page with a fallback? It'll make it easier if I'm running locally 

- Are you seeing any errors with redis locally?  About 25% of the time things fail to cache - maybe it's my local network, the redis is running elsewhere

```
INFO[2024-10-03T07:28:16.163-04:00] Fallback (4.15) QueryTestStatus completed in 23.094652875s with 180541 base results from db
WARN[2024-10-03T07:28:16.797-04:00] couldn't persist new item to cache            error="write tcp 192.168.1.157:62924->192.168.1.215:6379: i/o timeout"
```

- Is there any way to wire up ignoreFallback to the cache being unavailable? Running locally if you don't have a cache, it's very very slow


Some UI comments:
- I think we should focus on only showing the user one square / one assessment per test.  Having 2 muddies the waters: which is the right one? You can put it in a tooltip that it was overridden and what the original result was

- Would it be possible to show all the releases in tabs? e.g. the tab titles could be the basis release, and let me see all the ones we have (e.g. all 4).  It'd help for archeology, i.e. to see how the test did over time

pkg/api/componentreadiness/component_report.go
```diff
@@ -608,6 +677,180 @@ func (s *sampleQueryGenerator) queryTestStatus() (crtype.ReportTestStatus, []err
...
+		// 	continue
+		// }
+
+		start := f.BaseStart
+		end := f.BaseEnd
+		var crRelease *crtype.Release
+
+		for _, release := range releases {
+			if release.Release == fallbackRelease {
+				crRelease = &release
```
stbenjam, line 726:
```suggestion
		for i, release := range releases {
			if release.Release == fallbackRelease {
				crRelease = &releases[i]
```

Or [update go.mod to require 1.22](https://go.dev/blog/loopvar-preview)

sippy-ng/src/component_readiness/AdvancedOptions.js
sosiouxme: also i would find this less confusing if the boolean were reversed and so we had "Enable Fallback Basis" defaulting to false (for now)

pkg/api/componentreadiness/component_report.go
stbenjam: I filed a separate card for this: https://issues.redhat.com/browse/TRT-1845

pkg/api/componentreadiness/component_report.go
```diff
@@ -608,6 +677,180 @@ func (s *sampleQueryGenerator) queryTestStatus() (crtype.ReportTestStatus, []err
...
+
+		}(crRelease, start, end)
+
+		// this gets removed when we stop loading fallback data for the current base release
+		// and the section up to gets uncommented
+		fallbackRelease, err = previousRelease(fallbackRelease)
+		if err != nil {
+			log.WithError(err).Errorf("Failure determining fallback release for %s", fallbackRelease)
+			continue
+		}
```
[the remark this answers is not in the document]
neisw: Nice catch, thank you!

pkg/api/componentreadiness/component_report.go
```diff
@@ -1321,6 +1579,96 @@ func (c *componentReportGenerator) getRequiredConfidence(testID string, variants
...
+
+					if priorTestStats.ReportStatus < baseTestStats.ReportStatus {
+						baseStats = cTestStats
+						baseTestStats = priorTestStats
+						baseRelease = priorRelease
+					}
+				}
+			} else {
+				// if we miss on a prior release we stop
+				break
```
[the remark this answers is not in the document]
neisw: In the case of previousReleae err would be != nil and the for loop would end but I can make it more explicit.

pkg/api/componentreadiness/component_report.go
```diff
@@ -608,6 +677,180 @@ func (s *sampleQueryGenerator) queryTestStatus() (crtype.ReportTestStatus, []err
...
+		end := f.BaseEnd
+		var crRelease *crtype.Release
+
+		for _, release := range releases {
+			if release.Release == fallbackRelease {
+				crRelease = &release
+				if release.GADate != nil && release.GA30DayPrior != nil {
+					end = *release.GADate
+					start = *release.GA30DayPrior
+				}
```
dgoodwin, line 730: So unless if the fallback feature is enabled, regardless what base date range you choose, we may swap it out to the 30 days prior to the selected release's GA?

I'm wondering how we can smooth over the UX there. We've got the relative strings somewhat available via the views work (now-7d, ga-30d), but users can't specify that in the UI today. Seems like we *must* preserve ability to select specific date ranges, we can't just assume to let someone type/select now/ga-Xd because someone will eventually want a specific past date that is neither. Personally I would not be opposed to just typing in ga-30d in a textfield and adding support for 2024-10-14-30d (or some less dash ridden format) Frankly it's quicker than a date picker IMO.

stbenjam: image

I wish the basis release was a multi-select rather than having a fallback toggle of some kind.

For the dates, I'm not that convinced a text box is a great UI for someone who hasn't used CR before.  You could just disable the date picker when multi-release is selected and force GA. Date picker is more natural. Jira does dropdown filters by default, and has an advanced jql input field, which could also be an option.

BTW, we need to fix 30 days vs 4 weeks (28) and pick just one

neisw: We only override dates for fallback queries.  We don't change the date for the selected basis.

We might consider something like `default`, or allow users to select specific releases / with dates.

I think much of this is likely separate from the initial effort though.  Lots we can do to standardize date selection / ui / etc.

pkg/api/componentreadiness/component_report.go
```diff
@@ -415,45 +453,49 @@ func (c *componentReportGenerator) getCommonTestStatusQuery(allJobVariants crtyp
...
-	// potentially cross-compare variants for the sample
-	if isSample && len(c.VariantCrossCompare) > 0 {
-		variantGroups = c.CompareVariants
-	}
-	if variantGroups == nil { // server-side view definitions may omit a variants map
-		variantGroups = map[string][]string{}
-	}
+	// fallback queries get all variants with no filtering
+	// so all tests are fetched then cached
+	if !isFallback {
```
dgoodwin, line 458: Is there a in-code replacement for the removed IncludeVariants query filtering? Presumably if we skip it here we must do it in code later, but I couldn't find it in the PR. (searching for uses of IncludeVariants)

And followup question, can this function just change to always NOT filter on variants/capability here, all the time, regardless if it's fallback. Wondering if we can avoid a logic fork and have just one path through here, and always rely on our own code for the filtering.

pkg/api/componentreadiness/component_report.go
```diff
@@ -350,7 +388,7 @@ func (c *componentReportGenerator) GenerateComponentReportTestStatus() (crtype.R
 }
 
 // getCommonTestStatusQuery returns the common query for the higher level summary component summary.
-func (c *componentReportGenerator) getCommonTestStatusQuery(allJobVariants crtype.JobVariants, isSample bool) (string, string, []bigquery.QueryParameter) {
+func (c *componentReportGenerator) getCommonTestStatusQuery(allJobVariants crtype.JobVariants, isSample, isFallback bool) (string, string, []bigquery.QueryParameter) {
```
dgoodwin, line 391: Our production redis pod looks to run pretty consistently around 500mb.

Locally with a little testing I'm around 2.2gb. I believe this is with all main views cached so that number isn't horrible, but it looks like you will need to bump the redis memory request up from 1gb to maybe 5-10, otherwise we'll be cache missing too often as things may get evicted.

pkg/api/componentreadiness/component_report.go
```diff
@@ -1321,6 +1579,96 @@ func (c *componentReportGenerator) getRequiredConfidence(testID string, variants
 	return c.RequestAdvancedOptions.Confidence
 }
 
+func (c *componentReportGenerator) matchBaseRegression(testID crtype.ReportTestIdentification, baseRelease string, baseStats crtype.TestStatus) (crtype.TestStatus, string) {
```
dgoodwin, line 1582: Taking a pass at godoc would be helpful on some of these more important functions.

pkg/api/componentreadiness/component_report.go
```diff
@@ -1347,20 +1706,46 @@ func (c *componentReportGenerator) generateComponentTestReport(baseStatus map[st
...
+			} else if !c.IgnoreFallback {
+				log.Error("Expected cached fallback statuses but was missing")
+			}
+
+			// this is where we look to see if a previous release has a higher pass rate
+			baseRelease := c.BaseRelease.Release
+			baseStats, baseRelease, testStats = c.matchBestBaseStats(testID, testIdentification, baseRelease, baseStats, sampleStats, requiredConfidence, approvedRegression, resolvedIssueCompensation)
+
+			if baseRelease != c.BaseRelease.Release {
+				log.Infof("Overrode base stats using release %s for Test: %s - %s", baseRelease, baseStats.TestName, testIdentification)
```
dgoodwin, line 1746: There's now the explanations field on a regressed test that could be used to show this info in the tooltip and test details page if desired.

sippy-ng/src/component_readiness/AdvancedOptions.js
dgoodwin: Agreed on the boolean reversal. I don't know if we can fit it but to me this is best described as Fallback To Best Past Release. Some version of that may help a little in the variable/param naming as well. (fallbackToBestPastRelease)

neisw: I don't really disagree and likely wouldn't have coded it this way without prior art but what I did is consistent with what is [already there](https://github.com/openshift/sippy/blob/4fd58999afbe0caddd520af91f2b17d0dd029db2/sippy-ng/src/component_readiness/AdvancedOptions.js#L24) so I'm a little hesitant to refactor.

pkg/api/componentreadiness/component_report.go
```diff
@@ -1347,20 +1706,46 @@ func (c *componentReportGenerator) generateComponentTestReport(baseStatus map[st
...
+						baseReleaseMisses += 1
+					}
+				}
+
+			} else if !c.IgnoreFallback {
+				log.Error("Expected cached fallback statuses but was missing")
+			}
+
+			// this is where we look to see if a previous release has a higher pass rate
+			baseRelease := c.BaseRelease.Release
```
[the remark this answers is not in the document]
neisw: Hmm, but it may not actually be the requestedBaseRelease, the next line down looks to return a fallback if that is determined higher threshold.  Would 'matchedBaseRelease' work / be clearer?

sippy-ng/src/component_readiness/CompReadyTestReport.js
neisw: I haven't looked into the tooltip yet but how does this look:

image

neisw:
> * Can you share a localhost link to a page with a fallback? It'll make it easier if I'm running locally
> 
>     * Are you seeing any errors with redis locally?  About 25% of the time things fail to cache - maybe it's my local network, the redis is running elsewhere
> 
> 
> ```
> INFO[2024-10-03T07:28:16.163-04:00] Fallback (4.15) QueryTestStatus completed in 23.094652875s with 180541 base results from db
> WARN[2024-10-03T07:28:16.797-04:00] couldn't persist new item to cache            error="write tcp 192.168.1.157:62924->192.168.1.215:6379: i/o timeout"
> ```
> 
>     * Is there any way to wire up ignoreFallback to the cache being unavailable? Running locally if you don't have a cache, it's very very slow
> 
> 
> Some UI comments:
> 
>     * I think we should focus on only showing the user one square / one assessment per test.  Having 2 muddies the waters: which is the right one? You can put it in a tooltip that it was overridden and what the original result was
> 
>     * Would it be possible to show all the releases in tabs? e.g. the tab titles could be the basis release, and let me see all the ones we have (e.g. all 4).  It'd help for archeology, i.e. to see how the test did over time

Links for [Component Readiness](http://localhost:3000/sippy-ng/component_readiness/main?baseEndTime=2024-08-30%2023%3A59%3A59&baseRelease=4.17&baseStartTime=2024-08-02%2000%3A00%3A00&columnGroupBy=Platform%2CArchitecture%2CNetwork&confidence=95&ignoreDisruption=1&ignoreFallbackBasis=0&ignoreMissing=0&includeVariant=Architecture%3Aamd64&includeVariant=FeatureSet%3Adefault&includeVariant=Installer%3Aipi&includeVariant=Installer%3Aupi&includeVariant=Owner%3Aeng&includeVariant=Platform%3Aaws&includeVariant=Platform%3Aazure&includeVariant=Platform%3Agcp&includeVariant=Platform%3Ametal&includeVariant=Platform%3Avsphere&includeVariant=Topology%3Aha&includeVariant=CGroupMode%3Av2&includeVariant=ContainerRuntime%3Arunc&minFail=3&pity=5&regressedModal=0&sampleEndTime=2024-09-28%2023%3A59%3A59&samplePRNumber=&samplePROrg=&samplePRRepo=&sampleRelease=4.18&sampleStartTime=2024-09-21%2000%3A00%3A00&view=) and [Test Details](http://localhost:3000/sippy-ng/component_readiness/test_details?Aggregation=none&Architecture=amd64&Architecture=amd64&FeatureSet=default&FeatureSet=default&Installer=ipi&Installer=ipi&Network=ovn&Network=ovn&NetworkAccess=default&Platform=azure&Platform=azure&Scheduler=default&SecurityMode=default&Suite=unknown&Suite=unknown&Topology=ha&Topology=ha&Upgrade=none&Upgrade=none&baseEndTime=2024-08-30%2023%3A59%3A59&baseRelease=4.17&baseStartTime=2024-08-02%2000%3A00%3A00&capability=operator-conditions&columnGroupBy=Platform%2CArchitecture%2CNetwork&component=Etcd&confidence=95&dbGroupBy=Platform%2CArchitecture%2CNetwork%2CTopology%2CFeatureSet%2CUpgrade%2CSuite%2CInstaller&environment=amd64%20default%20ipi%20ovn%20azure%20unknown%20ha%20none&ignoreDisruption=true&ignoreFallbackBasis=false&ignoreMissing=false&includeVariant=Architecture%3Aamd64&includeVariant=CGroupMode%3Av2&includeVariant=ContainerRuntime%3Arunc&includeVariant=FeatureSet%3Adefault&includeVariant=Installer%3Aipi&includeVariant=Installer%3Aupi&includeVariant=Owner%3Aeng&includeVariant=Platform%3Aaws&includeVariant=Platform%3Aazure&includeVariant=Platform%3Agcp&includeVariant=Platform%3Ametal&includeVariant=Platform%3Avsphere&includeVariant=Topology%3Aha&minFail=3&pity=5&sampleEndTime=2024-09-28%2023%3A59%3A59&sampleRelease=4.18&sampleStartTime=2024-09-21%2000%3A00%3A00&testBasisRelease=4.15&testId=Operator%20results%3A45d55df296fbbfa7144600dce70c1182&testName=operator%20conditions%20etcd) should work for you.  It is the Etcd regression for operator condition that is hitting fallback.  I haven't noticed issues persisting to the cache but will keep an eye out for it.  I removed the extra status icon and changed the tabs to reflect the version of the Basis they are for.  I could see a uses case for showing them all (like in your screenshot for selecting multiple releases) we can discuss more but at the moment I'm thinking it would be new work and not part of the first effort.

pkg/api/componentreadiness/component_report.go
neisw: Capturing the discussion we had on this, swapping out the basis for the sample in the loop leaves us unable to detect missing samples.  Will need some work there to report on tests that existed in the base but not the sample when we look to refactor this.

neisw:
135a2c0711cf31b74918a9267f6e83f1a13c3992 introduces both a compression and persistent layer for caching.

There are currently some redundant config options for dataset and project that I anticipate will go away if/when we formalize this by adding the cache dataset to our main project.  

Additionally it takes a separate 'persistent-warm-cache' url to allow a separate cache from the main cache, if desired, to reduce evicting items from the main cache.  

There is also the option to be 'read-only'  for the persistent cache in the case of main sippy being read only.  We would need to coordinate another process like regression tracking to populate the persistent cache data.

```
--redis-url=redis://localhost:6379 --persistent-cache-dataset=fsbabcock_test --persistent-cache-project=openshift-ci-data-analysis --persistent-warm-cache-url=redis://localhost:6380 --persistent-cache-duration=12h --persistent-cache-write=true 
```

pkg/api/componentreadiness/component_report.go
neisw: Let's discuss 135a2c0711cf31b74918a9267f6e83f1a13c3992 and possibly a separate redis deployment for these large caches so we don't impact the main report cache.  Having redis in the persistent cache layer is optional but does speed things up.  So nice to have but if something is evicted we can still get it quicker from the persistent backend than completely redoing the query.

sippy-ng/src/component_readiness/AdvancedOptions.js
neisw: With the move to include both the fallback basis as well as the requested basis I'm leaning towards removing this as an option and letting the default config (view?) determine if we use fallback or not.

neisw: I see the [views](https://github.com/openshift/sippy/blob/master/config/views.yaml#L58)  are used to control the advanced options so I will investigate that vs. removal.

sippy-ng/src/component_readiness/CompReadyTestReport.js
xueqzhan: I like this one better. I was confused with the terms as well. But this makes it clear.

pkg/cache/compressed/cache.go
```diff
@@ -0,0 +1,106 @@
...
+
+const (
+	cachePrefix = "cc:"
+)
+
+type Cache struct {
+	Cache cache.Cache
+}
+
+func NewCompressedCache(c cache.Cache) (*Cache, error) {
```
dgoodwin, line 24: Doesn't seem to be used at this time? Assume this was an attempt at helping us stick with just redis.

pkg/cache/bigquerycache/bigquery.go
```diff
@@ -0,0 +1,223 @@
...
+// https://cloud.google.com/bigquery/quotas#streaming_inserts
+// Maximum row size 	10 MB 	Exceeding this value causes invalid errors.
+// HTTP request size limit 	10 MB Exceeding this value causes invalid errors.
+const (
+	cachedTable     = "cached_data"
+	partitionColumn = "modified_time"
+	chunkSize       = 7000000 // ~7MB to stay under the max row limit
+)
+
+type Cache struct {
```
dgoodwin, line 26: Godoc would be good here, what this is for, why we did it, how it interacts with the warm redis cache.

pkg/api/componentreadiness/component_report.go
```diff
@@ -504,6 +545,40 @@ func (c *componentReportGenerator) getBaseQueryStatus(ctx context.Context,
...
+		BaseEnd:         end,
+		queryParameters: baseParams,
+	}
+
+	// default to main cache
+	// but if defined will use persistent cache
+	fcache := c.client.Cache
+	if c.client.PersistentCache != nil {
+		fcache = c.client.PersistentCache
+	}
```
dgoodwin, line 571: This feels a little off in ways that will be confusing to explain to someone. The passing of a cache, but there's another cache embedded on the bigquery client that can be used for other purposes but also proxy off the original cache. Could we come up with a more cohesive mechanism for this module?

Passing in the bigquery cache separately would be one step clearer in my mind, its unexpected it would be hanging off the client as the two then depend on each other.

But ideally a single cache that could handle all the logic would be really nice. It could know to make use or the bigquery cache or not based on the size of the data and the duration requested to cache for. Though I don't know how we could tell it not to fallback to the bigquery cache for most regular pre-existing keys.

pkg/cache/bigquerycache/bigquery.go
```diff
@@ -0,0 +1,223 @@
...
+type CacheRecord struct {
+	Key        string    `bigquery:"key"`
+	UUID       string    `bigquery:"uuid"`
+	Modified   time.Time `bigquery:"modified_time"`
+	Expiration time.Time `bigquery:"expiration"`
+	Data       []byte    `bigquery:"data"`
+	ChunkIndex int       `bigquery:"chunk_index"`
+}
+
+func (c Cache) Get(ctx context.Context, key string) ([]byte, error) {
```
dgoodwin, line 53: It would be very helpful here to instrument this and Set. Prometheus counters for cache hits and missed on warm vs cold. Histograms for the size of the keys being set if you could predict ~10 bucket sizes. I'd settle for logging though. 

Being able to see how it's behaving would be quite beneficial.

dgoodwin (commented):
Looking great. My main concern is the increase in complexity so most of these comments are related to trying to keep things as clear as is reasonable.

pkg/flags/cache.go
```diff
@@ -23,6 +34,43 @@ func (f *CacheFlags) BindFlags(fs *pflag.FlagSet) {
...
+		os.Getenv("PERSISTENT_CACHE_PROJECT"),
+		"Project containing the dataset used for persistent caching")
+
+	// keep this separate from the main redis cache
+	// in case we want to run separate cache instances
+	// for main and persistent warm cache
+	fs.StringVar(&f.PersistentWarmCacheURL,
+		"persistent-warm-cache-url",
+		os.Getenv("PERSISTENT_WARM_CACHE_URL"),
+		"Redis URL for persistent warm caching")
```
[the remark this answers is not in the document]
neisw: Yep, or I can drop this config for now.  I wanted to allow for separate instances but if we don't want it currently we can simplify and always add it back later.

pkg/api/componentreadiness/component_report.go
neisw: If there is support for the change we could move everything to a single cache that would only attempt to persist items that have a retention time longer than X and/or byte length > Y, etc.

dgoodwin: That would be excellent and really nicely encapsulate the complexity.

pkg/api/componentreadiness/component_report.go
neisw: Done, thanks!

pkg/flags/cache.go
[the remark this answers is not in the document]
neisw: Done

pkg/cache/compressed/cache_test.go
```diff
@@ -0,0 +1,74 @@
...
+	return nil
+}
+
+func TestPseudoCache(t *testing.T) {
+	data := "It is useful mainly in compressed network protocols, to ensure that a remote reader has enough data to reconstruct a packet. Flush does not return until the data has been written. If the underlying writer returns an error, Flush returns that error. "
+
+	cache, err := NewCompressedCache(&PseudoCache{cache: make(map[string][]byte)})
+
+	if err != nil {
+		t.Fatalf("Failed to create compression cache")
```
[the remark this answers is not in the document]
neisw: Thanks!

sippy-ng/src/component_readiness/AdvancedOptions.js
neisw: Reversed the boolean and changed the label / parameter naming

pkg/cache/compressed/cache.go
```diff
@@ -0,0 +1,106 @@
...
+	// strip off the last 16 bytes which is the checksum
+	data := b[:dataLen-16]
+	var checksum [16]byte
+	copy(checksum[:], b[dataLen-16:])
+	return uncompress(data, checksum)
+}
+
+func (c Cache) Set(ctx context.Context, key string, content []byte, duration time.Duration) error {
+	// append the checksum
+	data, checksum, err := compress(content)
```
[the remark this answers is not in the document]
neisw: Should see the following entries now:
```
"Compressed cache start size: 468901.25k, after compression: 12475.21k.  Ratio: 99.97%"
```

pkg/cache/compressed/cache.go
neisw: Updated to actually use the function and not just init the struct.

pkg/cache/compressed/cache.go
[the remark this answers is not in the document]
dgoodwin: Very cool!

dgoodwin:
Great changes, thanks.

This looks good to try to me.

/lgtm
/hold

pkg/bigquery/client.go
```diff
@@ -11,9 +11,10 @@ import (
 )
 
 type Client struct {
-	BQ      *bigquery.Client
-	Cache   cache.Cache
-	Dataset string
+	BQ              *bigquery.Client
+	Cache           cache.Cache
+	PersistentCache cache.Cache
```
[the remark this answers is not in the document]
neisw: Simplified to single cache and doc'd the bigquery cache

pkg/api/componentreadiness/component_report.go
neisw: Had to modify the Get url to include duration but added logic in the bigquery cache to not persist data below a min duration and collapsed down to a single cache from the component_report perspective.

pkg/cache/bigquerycache/bigquery.go
neisw: Added some for hits, misses and sets.  Data sizes are in the logs now but we can pursue histograms if there is value.
Sample values from deleting redis pod and also forcing refresh
```
# HELP sippy_persistent_backend_cache_get Number of persistent cache get that query the backend cache.
# TYPE sippy_persistent_backend_cache_get counter
sippy_persistent_backend_cache_get 1
# HELP sippy_persistent_cache_backend_set Number of persistent cache backend set calls.
# TYPE sippy_persistent_cache_backend_set counter
sippy_persistent_cache_backend_set 5
# HELP sippy_persistent_cache_get Number of persistent cache get requests.
# TYPE sippy_persistent_cache_get counter
sippy_persistent_cache_get 47
# HELP sippy_persistent_warm_cache_miss Number of persistent cache gets that were not in the warm cache.
# TYPE sippy_persistent_warm_cache_miss counter
sippy_persistent_warm_cache_miss 15
```

pkg/cache/bigquerycache/bigquery.go
```diff
@@ -0,0 +1,302 @@
...
+// Providing regular cache updates to the read only process and allowing for outages as well.
+//
+// minExpiration is the minimum duration that will be written to the bigquery cache table
+// items that are cached with a Set duration shorter than minExpiration will skip the biqquery
+// caching and rely on the main caching layer.  The minExpiration is also used to determine if a Get
+// that misses the main caching layer should check the bigquery cache table.  If the duration is below the min
+// it will not check for an entry but instead return a miss.
+type Cache struct {
+	client        *sippybq.Client
+	readOnly      bool
```
neisw, line 84: Wondering if we need an inverse of this that is `primaryWrite` indicating skip the warm cache lookup and go to the backend cache.  In the case above where maxExpiration is shortened for the cache write process that won't impact the read from the warm cache which would lead to it not considering the data expired and therefore not refreshing the cache as expected.

pkg/cache/bigquerycache/bigquery.go
```diff
@@ -0,0 +1,302 @@
...
+	Data       []byte    `bigquery:"data"`
+	ChunkIndex int       `bigquery:"chunk_index"`
+}
+
+func (c Cache) Get(ctx context.Context, key string, duration time.Duration) ([]byte, error) {
+
+	persistentCacheGetMetric.WithLabelValues().Inc()
+
+	// if we have it in our warm cache use it
+	if c.client.Cache != nil {
```
neisw, line 113: && (duration < c.minExpiration || !c.primaryWrite)

dgoodwin:
/lgtm

Love the metrics, thanks, they may actually prove useful in questions like, was someone hitting comp readiness hard over the weekend.

PR #2006, opened 2024-09-24
Add 4.17 intentional regressions file

kevinrizza:
Also tracks a regression for alert/KubeAPIErrorBudgetBurn

### DIFF
--- a/pkg/regressionallowances/regressions_4.17.json
+++ b/pkg/regressionallowances/regressions_4.17.json
@@ -1,0 +1,52 @@
+[
+    {
+      "JiraComponent": "kube-apiserver",
+      "TestID": "openshift-tests:d6b41cee7afca1c2a0b52f9e6975425f",
+      "TestName": "[bz-kube-apiserver][invariant] alert/KubeAPIErrorBudgetBurn should not be at or above info",
+      "JiraBug": "https://issues.redhat.com/browse/OCPBUGS-42083",
+      "ReasonToAllowInsteadOfFix": "Degraded enough to trip test near the end of stabilization. Still no root cause determined, and the team is still actively investigating. We are marking this as not a blocker because we did not see signals indicating that this does not stabilize after install is completed, although that has not yet been confirmed.",
+      "variant": {
+        "variants": {
+          "Network": "ovn",
+          "Upgrade": "none",
+          "Architecture": "amd64",
+          "Platform": "azure",
+          "FeatureSet": "default",
+          "Suite": "serial",
+          "Topology": "ha",
+          "Installer": "ipi"
+        }
+      },
+      "PreviousSuccesses": 74,
+      "PreviousFailures": 0,
+      "PreviousFlakes": 0,
+      "RegressedSuccesses": 20,
+      "RegressedFailures": 3,
+      "RegressedFlakes": 0
+    },
+    {
+      "JiraComponent": "kube-apiserver",
+      "TestID": "openshift-tests-upgrade:d6b41cee7afca1c2a0b52f9e6975425f",
+      "TestName": "[bz-kube-apiserver][invariant] alert/KubeAPIErrorBudgetBurn should not be at or above info",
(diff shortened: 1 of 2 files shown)


kevinrizza:
@neisw @dgoodwin this supersedes parts of https://github.com/openshift/sippy/pull/1991

Can someone please review? The intention is to remove the test from the dashboard for now, and we can reenable it later once the regression is tracked down.

neisw:
/lgtm

I ran this locally to validate

pkg/regressionallowances/regressions_4.17.json
```diff
@@ -0,0 +1,52 @@
+[
+    {
+      "JiraComponent": "kube-apiserver",
+      "TestID": "openshift-tests:d6b41cee7afca1c2a0b52f9e6975425f",
+      "TestName": "[bz-kube-apiserver][invariant] alert/KubeAPIErrorBudgetBurn should not be at or above info",
+      "JiraBug": "https://issues.redhat.com/browse/OCPBUGS-42083",
+      "ReasonToAllowInsteadOfFix": "Degraded near the end of stabilization.  Investigation indicated not a release blocker.",
```
deads2k, line 7: Can we get some more reasoning here about why this isn't a release blocker versus just the assertion?  For instance, has the investigation reveal proximate causes?

kevinrizza: I added some more detail about what's happening here and why this choice was made

pkg/regressionallowances/regressions_4.17.json
```diff
@@ -0,0 +1,52 @@
+[
+    {
+      "JiraComponent": "kube-apiserver",
+      "TestID": "openshift-tests:d6b41cee7afca1c2a0b52f9e6975425f",
+      "TestName": "[bz-kube-apiserver][invariant] alert/KubeAPIErrorBudgetBurn should not be at or above info",
+      "JiraBug": "https://issues.redhat.com/browse/OCPBUGS-42083",
+      "ReasonToAllowInsteadOfFix": "Degraded near the end of stabilization.  Investigation indicated not a release blocker based on the fact that in our obseved error case this measurement stabilizes under the failure threshold after cluster installation completes.",
```
deads2k, line 7: Do we know what happens during installation to produce this problem?  For instance, we know that it isn't only caused by etcd leader loss since on some occasions some requests are still functioning.

kevinrizza: We do not know that for certain, no.

pkg/regressionallowances/regressions_4.17.json
```diff
@@ -0,0 +1,52 @@
+[
+    {
+      "JiraComponent": "kube-apiserver",
+      "TestID": "openshift-tests:d6b41cee7afca1c2a0b52f9e6975425f",
+      "TestName": "[bz-kube-apiserver][invariant] alert/KubeAPIErrorBudgetBurn should not be at or above info",
+      "JiraBug": "https://issues.redhat.com/browse/OCPBUGS-42083",
+      "ReasonToAllowInsteadOfFix": "Degraded enough to trip test near the end of stabilization. Still no root cause determined, and the team is still actively investigating. We are marking this as not a blocker because we did not see signals indicating that this does not stabilize after install is completed, although that has not yet been confirmed.",
```
vrutkovs, line 7:
```suggestion
      "ReasonToAllowInsteadOfFix": "Degraded enough to trip test near the end of stabilization. This alert fires on installations, when kube-apiserver doesn't get full list of etcds and may lose etcd connectivity for several minutes. We are marking this as not a blocker because it affects fresh installats only and the alert goes away shortly after installation is complete.",
```

neisw:
/close
[sippy#2015](https://github.com/openshift/sippy/pull/2015) included this commit and resolved merge conflicts with [sippy#2013](https://github.com/openshift/sippy/pull/2013)

PR #2007, opened 2024-09-24
TRT-1808: fix flakes link to include variants in job runs filter

stbenjam:
When clicking the Flakes icon, it doesn't filter the job run list by the variants for that row on the tests table. We were just using the wrong filter model to pass in.


sosiouxme (approved):
/lgtm

PR #2014, opened 2024-09-26
TRT-1771: Test Pass Rate Modes for Component Readiness

dgoodwin:
Adds two new options to component readiness reports, visible in advanced options, and definable in views.

Pass rate required for new tests: enables a hybrid mode where if we have no basis data for a test, implying it is a new test (or all too commonly a renamed test that wasn't updated in the mapping repo), we fall back to a required pass rate (expecting 99%). Any new test then appears in the report as regressed if below this threshold. minFail is not respected here, we just check the pass rate.

Pass rate required for all tests: This was a future goal that just looked doable while I was in here, so I lumped it in. This should be considered experimental, we're a long way from ever being able to flip this on. Regardless in this mode, ALL tests are only compared based on pass rate, fisher's no longer runs.

A view is included which enables both these options, 99% requirement for new tests, 95% for all pre-existing.

The report regressed tests struct now looks like:

```json
          "regressed_tests": [
            {
              "component": "Installer / openshift-installer",
              "capability": "Other",
              "test_name": "install should succeed: other",
              "test_suite": "cluster install",
              "test_id": "cluster install:2ec4097c2122a2e07e7270c2d7cb3cc6",
              "variants": {
                "Architecture": "amd64",
                "FeatureSet": "default",
                "Installer": "ipi",
                "Network": "ovn",
                "Platform": "metal",
                "Suite": "unknown",
                "Topology": "ha",
                "Upgrade": "micro"
              },
              "opened": "2024-10-15T14:12:01.75474Z",
              "status": -4,
              "comparison": "fisher_exact",
              "explanations": [
                "Significant regression detected.",
                "Fishers Exact probability of a regression: 99.99%.",
                "Test pass rate dropped from 99.54% to 94.03%."
              ],
              "sample_stats": {
                "release": "4.18",
                "success_rate": 0.9402985074626866,
                "success_count": 63,
                "failure_count": 4,
                "flake_count": 0
              },
              "fisher_exact": 0.011530704731398564,
              "base_stats": {
                "release": "4.17",
                "success_rate": 0.9954337899543378,
                "success_count": 218,
                "failure_count": 1,
                "flake_count": 0
              }
            },


or for pass rate:

          "regressed_tests": [
            {
              "component": "Machine Config Operator",
              "capability": "OCPFeatureGate:ManagedBootImages",
              "test_name": "[sig-mco][OCPFeatureGate:ManagedBootImages][Serial] Should update boot images on all MachineSets when configured [apigroup:machineconfigur
ation.openshift.io] [Suite:openshift/conformance/serial]",
              "test_suite": "openshift-tests",
              "test_id": "openshift-tests:8adfa1323bb3d0e6e45a603c517ce87e",
              "variants": {
                "Architecture": "amd64",
                "FeatureSet": "default",
                "Installer": "ipi",
                "Network": "ovn",
                "Platform": "gcp",
                "Suite": "serial",
                "Topology": "ha",
                "Upgrade": "none"
              },
              "opened": null,
              "status": -4,
              "comparison": "pass_rate",
              "explanations": [
                "Test has a 95.24% pass rate, but 98.00% is required."
              ],
              "sample_stats": {
                "release": "4.18",
                "success_rate": 0.9523809523809523,
                "success_count": 20,
                "failure_count": 1,
                "flake_count": 0
              }
            },

```

The comparison field is new, fisher_exact and base_stats now may not be in the response.

I also added the explanations list as I found we were forking UI behavior, as well as expanding where we try to represent this data in human readable form to more places, jira templates, and the automated bugs. The hope is that this lets us have the server describe why it did what it did, and views can remain dumb and not have to interpret what comparison was used to display things differently.


pkg/api/componentreadiness/component_report.go
```diff
@@ -1594,120 +1649,179 @@ func (c *componentReportGenerator) assessComponentStatus(requiredConfidence, sam
...
-			Release: c.SampleRelease.Release,
-			TestDetailsTestStats: crtype.TestDetailsTestStats{
-				SuccessRate:  getSuccessRate(sampleSuccess, sampleFailure, sampleFlake),
-				SuccessCount: sampleSuccess,
-				FailureCount: sampleFailure,
-				FlakeCount:   sampleFlake,
-			},
-		},
-		BaseStats: crtype.TestDetailsReleaseStats{
+	if baseTotal == 0 && c.RequestAdvancedOptions.PassRateRequiredNewTests > 0 {
```
dgoodwin, line 1652: Core logic here, I tried to push almost everything into assessComponentStatus, and broke out fishers vs pass rate functionality into separate functions.

config/views.yaml
```diff
@@ -3,8 +3,8 @@ component_readiness:
   - name: 4.18-main
     base_release:
       release: "4.16"
```
stbenjam, line 5: Should we move this to 4.17 yet?

pkg/api/componentreadiness/component_report_test.go
```diff
@@ -1481,7 +1495,7 @@ func Test_componentReportGenerator_assessComponentStatus(t *testing.T) {
 			baseFlake:              1,
 			numberOfIgnoredSamples: 10,
 			expectedStatus:         -3,
-			expectedFischers:       1,
+			expectedFischers:       thrift.Float64Ptr(1),
```
stbenjam, line 1498: Nice find, I usually end up writing my own helpers for this kind of thing. Thrift looks pretty lightweight too

stbenjam (commented):
First pass, just trying it out.  Some UI comments

Should we make it obvious the red square is because of pass percentage? I think I do want to know without clicking through.  Maybe even a new status, and new icon(s)?

Something like...

image

Also on the table, should we show the comparison type?

image

config/views.yaml
dgoodwin: I think the intentional regressions are still being worked out, so a comparison to 4.17 relative to now could well make some things disappear that shouldn't like the console operator and kubeapi error budget burn. Definitely at ga time, but we could also do it once the intentional regressions are all merged.

dgoodwin:
I like the icon and I'm also craving knowing it was a pass rate comparison in the table. Is the icon sufficient for indicating that? Or were you thinking an additional column?

I was also thinking about playing with the tooltip on the icon to give a LOT more info so you didn't always have to drill down.

stbenjam:
Icon and tooltip are enough, we probably don't need another column in that table.

Do you want the SVG for the icon I showed? I have a few options I sent

config/views.yaml
```diff
@@ -51,10 +51,56 @@ component_readiness:
...
+          - azure
+          - gcp
+          - metal
+          - vsphere
+        Topology:
+          - ha
+    advanced_options:
+      minimum_failure: 3
+      pass_rate_required_new_tests: 95
+      pass_rate_required_all_tests: 80
```
dgoodwin, line 103: I'll remove this view before merge.

config/views.yaml
```diff
@@ -51,10 +51,56 @@ component_readiness:
       pity_factor: 5
       ignore_missing: false
       ignore_disruption: true
+      pass_rate_required_new_tests: 95
```
dgoodwin, line 54: This I'd like to enable and I don't think is too controversial. It's two additional regressions today.

pkg/api/componentreadiness/component_report.go
```diff
@@ -1220,7 +1221,6 @@ func triagedIssuesFor(releaseIncidents *resolvedissues.TriagedIncidentsForReleas
 	impactedJobRuns := sets.NewString() // because multiple issues could impact the same job run, be sure to count each job run only once
 	numJobRunsToSuppress := 0
 	for _, triagedIncident := range triagedIncidents {
-		startNumRunsSuppressed := numJobRunsToSuppress
```
dgoodwin, line 1223: This was unused.

pkg/api/componentreadiness/component_report.go
```diff
@@ -1360,6 +1360,8 @@ func (c *componentReportGenerator) getRequiredConfidence(testID string, variants
 	return c.RequestAdvancedOptions.Confidence
 }
 
+// TODO: break this function down and remove this nolint
+// nolint:gocyclo
```
dgoodwin, line 1364: TODO as I didn't want to further complicate this PR.

config/views.yaml
dgoodwin: This morning there were two, now there are none below 95%, you can set to 99% to see/test the mixed response in advanced options if none are presently showing up.

dgoodwin:
I think this is ready for review again.

pkg/api/componentreadiness/component_report_test.go
```diff
@@ -1496,16 +1524,124 @@ func Test_componentReportGenerator_assessComponentStatus(t *testing.T) {
...
+			sampleFlake:                 0,
+			baseTotal:                   100,
+			baseSuccess:                 97,
+			baseFlake:                   0,
+			numberOfIgnoredSamples:      0,
+			requiredPassRateForAllTests: 95,
+			expectedStatus:              crtype.NotSignificant,
+		},
+		{
+			name:                        "pass rate mode significant regression ignores minimum failures",
```
dgoodwin, line 1617: Note I do not consider min failures here, it felt contradictory to pass rate testing. If you've got a new test and it's only run 7x in a week, and one failed, that should not meet the bar.

pkg/api/componentreadiness/component_report.go
neisw: Isn't is used below?  So at the start of the for loop iteration set startNumRunsSuppressed to the current numJobRunsToSuppress, then loop through the job runs for the current triaged incident and if matches are found increment numJobRunsToSuppress

Outside that loop check to see if numJobRunsToSupress is now greater than startNumRunsSupressed and if so append the triagedIncident.  Otherwise don't include it because it isn't relevant and we don't want it in the UI.

What you have changed it to will include all triaged incidents, even ones with job runs that didn't match, any time there is at least one numJobRunsToSuppress

dgoodwin: Good catch thanks, pushing a restoration momentarily.

pkg/api/componentreadiness/component_report.go
```diff
@@ -1430,17 +1435,56 @@ func (c *componentReportGenerator) generateComponentTestReport(ctx context.Conte
...
 			return crtype.ComponentReport{}, err
 		}
+
+		// Check for approved regressions, and triaged incidents, which may adjust our counts and pass rate:
+		var testStats crtype.ReportTestStats
+		var triagedIncidents []crtype.TriagedIncident
+		var resolvedIssueCompensation int // triaged job run failures to ignore
+		// look for corresponding regressions we can account for in the analysis
+		approvedRegression := regressionallowances.IntentionalRegressionFor(c.SampleRelease.Release, testID.ColumnIdentification, testID.TestID)
+		baseRegression := regressionallowances.IntentionalRegressionFor(c.BaseRelease.Release, testID.ColumnIdentification, testID.TestID)
```
neisw, line 1453: Do you need this?  We don't have basis in this context and you pass in 0's below.  Should we remove this and just pass nil?

dgoodwin: Makes sense, changing to nil.

PR #2034, opened 2024-10-03
Remove fixed GA date map in favor of cached lookup

dgoodwin:
Found this when I flipped the view to use 4.17 ga which wasn't added to the map as we didn't have a step before.

We believe this is safe to assume bigquery connectivity now. I think the fallback to postgres was probably broken.

Added caching for speed.


pkg/sippyserver/metrics/installupgradesuccess.go
```diff
@@ -3,14 +3,14 @@ package metrics
 import (
 	"math"
 
+	v12 "github.com/openshift/sippy/pkg/apis/sippy/v1"
```
dgoodwin, line 6: auto refactor, fixing

sosiouxme (approved):
/lgtm
but
/hold
because i'm not sure if you want to YOLO on a friday afternoon

dgoodwin:
lint failing on npm security fixes, we will address separately.

/override ci/prow/lint

PR #2037, opened 2024-10-04
TRT-1830: Move regression tracking to a command/cronjob

dgoodwin:
This was done in the sippy pods themselves as they already query the
component report, so it seemed most efficient.

I moved to fetchdata, then to the daemon, then decided it makes the most sense as a sub-command that can run once in a kube cronjob. This would give us scheduling control, retries, clean logs per run, no resources consumed while not running, ability to keep multiple pod logs around. Didn't make sense to sleep for hours.

Corrected the bug that completely broke tracking when views launched.

And refactored tracking to snapshot all regressions for all releases each time we run (if any changes were needed), instead of trying to update rows, which gets very painful with bigquery's streaming buffer. Now we query only the latest snapshot time.


neisw:
Stephen brought up a good point about some of Ken's work to look into utilizing [daemon_server](https://github.com/openshift/sippy/blob/a025777dcd9b83c57fa14dfc03196394d6f67547/pkg/sippyserver/daemon_server.go)  Noted your comments in the alerts channel regarding fetchdata getting slower and slower.  Should we think about not putting this in fetchdata?

dgoodwin:
yeah that seems reasonable to me, I'll get it moved over this week.

Makefile
```diff
@@ -30,10 +30,10 @@ test: builddir npm
 
 lint: builddir npm
 	./hack/go-lint.sh run ./...
+	cd sippy-ng; npx eslint .
 	# See https://github.com/facebook/create-react-app/issues/11174 about
 	# why we only audit production deps:
 	cd sippy-ng; npm audit --production
-	cd sippy-ng; npx eslint .
```
dgoodwin, line 36: Making sure the audit doesn't prevent regular lint from running.

cmd/sippy/serve.go
```diff
@@ -33,9 +33,8 @@ type ServerFlags struct {
 	ProwFlags               *flags.ProwFlags
 	ComponentReadinessFlags *flags.ComponentReadinessFlags
 
-	ListenAddr               string
-	MetricsAddr              string
-	MaintainRegressionTables bool
```
dgoodwin, line 38: This one could be removed now as Luke moved it to the auth deployment recently.

config/views.yaml
```diff
@@ -283,4 +283,4 @@ component_readiness:
     metrics:
       enabled: true
     regression_tracking:
-      enabled: false
+      enabled: true
```
dgoodwin, line 286: Turn on tracking for 4.16, seemed reasonable.

pkg/api/componentreadiness/regressiontracker.go
```diff
@@ -0,0 +1,363 @@
...
+	"github.com/openshift/sippy/pkg/apis/cache"
+	v1 "github.com/openshift/sippy/pkg/apis/sippy/v1"
+	sippybigquery "github.com/openshift/sippy/pkg/bigquery"
+	"github.com/pkg/errors"
+	log "github.com/sirupsen/logrus"
+	"google.golang.org/api/iterator"
+)
+
+const (
+	testRegressionsTable = "test_regressions_append"
```
dgoodwin, line 22: New table so I could clear out all the nullable fields due to evolving schema. Old table will be fully deleted, the data is bad by this point.

pkg/componentreadiness/tracker/tracker.go
```diff
@@ -1,283 +0,0 @@
...
-		}
-		regressions = append(regressions, regression)
-	}
-	return regressions, nil
-
-}
-func (bq *BigQueryRegressionStore) OpenRegression(ctx context.Context, view crtype.View, newRegressedTest crtype.ReportTestSummary) (*crtype.TestRegression, error) {
-	id := uuid.New()
-	newRegression := &crtype.TestRegression{
-		View:         bigquery.NullString{StringVal: view.Name},
```
dgoodwin, line 89: Not specifying Valid: true here is why the table completely broke.

pkg/api/componentreadiness/regressiontracker.go
```diff
@@ -0,0 +1,363 @@
...
+					}
+				} else {
+					rt.logger.WithField("snapshot", snapshotTime).Warnf(
+						"SKIPPED syncing %d total regressions for all releases to bigquery due to dry-run flag",
+						len(newRegressionSnapshot))
+				}
+
+				// TODO:
+				rt.logger.Info("sleeping...")
+				time.Sleep(2 * time.Minute)
```
dgoodwin, line 179: This makes me think it should be a kube cronjob running on a schedule...

pkg/sippyserver/metrics/metrics.go
```diff
@@ -18,7 +18,6 @@ import (
 	"github.com/openshift/sippy/pkg/apis/cache"
 	v1 "github.com/openshift/sippy/pkg/apis/sippy/v1"
 	bqclient "github.com/openshift/sippy/pkg/bigquery"
-	"github.com/openshift/sippy/pkg/componentreadiness/tracker"
```
dgoodwin, line 21: For the changes in this file, metrics and regression tracking are now decoupled.

pkg/api/componentreadiness/regressiontracker.go
stbenjam: Should this have said 2 hours?

PR #2040, opened 2024-10-07
TRT-1836: wire up contexts for http requests

stbenjam:
goroutines need to be terminated when an HTTP request ends, currently without a context they keep running.   We also need to pass these into `bigquery.Query()` so the DB query terminates too.

If you wanted to test this PR:
1. Start up component readiness and click on an uncached square
2. Immediately close the browser tab -- notice the queries are terminated (you'll see context cancelled in the logs)

If you try this without this PR, the queries will continue executing. 

New goroutines should follow this pattern:

```go

	go func() {
                wg.Add(1)
		defer wg.Done()
		select {
		case <-ctx.Done():
			logrus.Infof("Context canceled while fetching base job run test status")
			return
		default:
                       // work to do
		}
```

### DIFF
--- a/hack/go-lint.sh
+++ b/hack/go-lint.sh
@@ -19,8 +19,14 @@ else
     exit 1
   fi
 
+  # Check if running on Linux
+  VOLUME_OPTION=""
+  if [[ "$(uname -s)" == "Linux" ]]; then
+    VOLUME_OPTION=":z"
+  fi
+
   $DOCKER run --rm \
-    --volume "${PWD}:/go/src/github.com/openshift/sippy:z" \
+    --volume "${PWD}:/go/src/github.com/openshift/sippy${VOLUME_OPTION}" \
     --workdir /go/src/github.com/openshift/sippy \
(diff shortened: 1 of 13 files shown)


hack/go-lint.sh
```diff
@@ -19,8 +19,14 @@ else
...
 
+  # Check if running on Linux
+  VOLUME_OPTION=""
+  if [[ "$(uname -s)" == "Linux" ]]; then
+    VOLUME_OPTION=":z"
+  fi
+
   $DOCKER run --rm \
-    --volume "${PWD}:/go/src/github.com/openshift/sippy:z" \
+    --volume "${PWD}:/go/src/github.com/openshift/sippy${VOLUME_OPTION}" \
```
stbenjam, line 29: Fix for running lint on a mac

dgoodwin:
Context looks simple enough to deal with when we conflict.

/lgtm

Sorry @neisw as we'll both pay for this one :)

stbenjam:
/override ci/prow/lint

Known issue

PR #2041, opened 2024-10-07
TRT-1475: Implement Jira integration with component readiness regressions

xueqzhan:
I want to get some initial feedback on this.

Relevant doc: https://docs.google.com/document/d/1q_7IPhpR2f_sKWGFr6qnDMZ0W8l-TQ9g7WOiD_evViY/edit

Sample Jira card created: https://issues.redhat.com/browse/OCPBUGS-45537

A few questions I have:
1. The doc mentioned about capability group to jira component mapping. As of now, there is no capability group to jira component mapping. CR component is the same as jira-component. So I am using component for now. 
2. For regressions with jira card, do we need to update every day? There will be lots of daily comments.
3. Right now jira integration is only enabled it on 4.18-main view. Need to come up with a list of views to enable
4. Do we have an existing jira account to use? 
5. Are we ready to turn this on org wide? 

/hold


pkg/componentreadiness/jiraintegrator/jiraintegrator.go
```diff
@@ -0,0 +1,524 @@
...
+	jiraCustomFieldReleaseBlockerName          = "customfield_12319743"
+	jiraCustomFieldReleaseBlockerValueApproved = "Approved"
+
+	jiraStatusNew        = "New"
+	jiraStatusInProgress = "In Progress"
+	jiraStatusAssigned   = "Assigned"
+	jiraStatusModified   = "Modified"
+	jiraStatusQA         = "QA"
+	jiraStatusVerified   = "Verified"
+	jiraStatusClosed     = "Closed"
```
stbenjam, line 36: Suggest moving out jira stuff like this to our own Jira package, like in pkg/apis/jira/v1. We use these values in multiple places, like pkg/dataloader/jiraloader/jiraloader.go.  Suggest matching the case to what's actually used in Jira.

stbenjam (commented):
I'm not sure if it's just my Mac, but GitHub doesn't handle this PR well enough for me to review it, both Chrome and Safari lock up because of it's size. If you move all the vendoring changes temporarily to their own commit, the web viewer lets me view by commit so I can get a smaller diff.... something like:

```
git reset HEAD^
git add vendor/
git commit -m "Vendor updates"
git add .
git commit -m "Non-vendor changes"
```

cmd/sippy/component_readiness.go
```diff
@@ -231,6 +267,43 @@ func (f *ComponentReadinessFlags) runServerMode() error {
 				panic(err)
 			}
 		}()
+
+		if f.IntegrateJira {
```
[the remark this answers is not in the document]
dgoodwin: I'm moving the regression detection there as we speak, which means we'll be pulling the component report already and you can tie in with that.

pkg/componentreadiness/jiraintegrator/jiraintegrator.go
```diff
@@ -0,0 +1,524 @@
...
+// a. adding a new comment containing a CR link with the most recently analyzed time window where the regression is still manifesting.
+// b. if pre-release, label the ticket as a Release Blocker if someone removed it
+func (j JiraIntegrator) updateExistingJiraIssue(view crtype.View, existing *jira.Issue) error {
+	absUrl, _, err := j.getComponentReadinessURLsForView(view)
+	if err != nil {
+		return err
+	}
+	comment := fmt.Sprintf(`This bug is still seen in component readiness. Here is [the current link|%s] for your convenience`, absUrl)
+
+	_, err = j.jiraClient.AddComment(existing.ID, &jira.Comment{Body: comment})
```
dgoodwin, line 209: It would be cool if this listed all the tests regressed (and their variant details) for that component each time.

pkg/componentreadiness/jiraintegrator/jiraintegrator.go
```diff
@@ -0,0 +1,524 @@
...
+				}
+			}
+		}
+		// test code. Only process one component for testing purposes
+		break
+	}
+	return nil
+}
+
+func (j JiraIntegrator) IntegrateJira() error {
```
dgoodwin, line 514: Nit: across the board I would suggest renaming anything about "integrating" jira to perhaps "automated jiras" or "automated component readiness jiras", something like that. The latter is verbose but much clearer.

pkg/componentreadiness/jiraintegrator/jiraintegrator.go
```diff
@@ -0,0 +1,524 @@
...
+	log.Infof("Start integrating component readiness regressions with Jira")
+	for _, view := range j.views {
+		err := j.integrateJiraForView(view)
```
dgoodwin, line 517: You'll want to add configuration on the View type for automated jiras, there are examples there already for turning on metrics and regression tracking.

pkg/componentreadiness/jiraintegrator/jiraintegrator.go
```diff
@@ -0,0 +1,524 @@
...
+	return report, nil
+}
+
+func getProbabilityString(status crtype.Status, fisherExact float64) string {
+	if status == crtype.SignificantRegression || status == crtype.ExtremeRegression {
+		return fmt.Sprintf("Probability of significant regression: %.2f%%\n\n", (1-fisherExact)*100)
+	} else if status == crtype.SignificantImprovement {
+		return fmt.Sprintf("Probability of significant improvement: %.2f%%\n\n", (1-fisherExact)*100)
+	}
+	return "There is no significant evidence of regression\n\n"
```
dgoodwin, line 132: I realized in my pass rate PR where sometimes we won't be returning fisher results, perhaps the server should provide an "explanation" string in the regressed tests API response with this kind of text, so we don't have to fork the same logic in multiple places. I'm going to test adding that to the pass rates PR and if we do we can use that here so we just render a human readable explanation provided by the server. It's going to get complicated otherwise. I'm going to try that out and maybe we can leverage here.

cmd/sippy/automatejira.go
```diff
@@ -0,0 +1,157 @@
...
+func (f *AutomateJiraFlags) BindFlags(fs *pflag.FlagSet) {
+	f.BigQueryFlags.BindFlags(fs)
+	f.GoogleCloudFlags.BindFlags(fs)
+	f.CacheFlags.BindFlags(fs)
+	f.ComponentReadinessFlags.BindFlags(fs)
+	f.JiraOptions.AddFlags(flag.CommandLine)
+	fs.AddGoFlagSet(flag.CommandLine)
+	fs.StringVar(&f.SippyURL, "sippy-url", f.SippyURL, "The Sippy URL prefix to be used to generate sharable Sippy links")
+	fs.StringVar(&f.ComponentWhiteListStr, "component-white-list", f.ComponentWhiteListStr, "The whitelist of comma seperated jira components to file issues against")
+	fs.StringArrayVar(&f.VariantBasedComponentRegressionThresholdStrs, "variant-based-component-regression-threshold", f.VariantBasedComponentRegressionThresholdStrs, "A list of variants which we will check for a column of red cells in component readiness report. If the number of red cells of a relevant column is over the threshold, we will a create jira card for the component corresponding to the specified variant (e.g. Bare Metal Hardware Provisioning for metal platform. The format of string is [variant]:[value]:[threshold] (e.g. Platform:metal:3)")
```
xueqzhan, line 59: I don't really like this name VariantBasedComponentRegressionThresholdStrs. Any better suggestions?

pkg/componentreadiness/jiraintegrator/jiraintegrator.go
xueqzhan: I might need to take care of this tomorrow.

pkg/componentreadiness/jiraintegrator/jiraintegrator.go
xueqzhan: I don't remember what we said about this.

cmd/sippy/automatejira.go
dgoodwin: How about ColumnThresholds

pkg/componentreadiness/jiraautomator/jiraautomator.go
```diff
@@ -0,0 +1,597 @@
...
+			description += fmt.Sprintf("\n* This regression might be related to [%s|%s]. Feel free to link it if found related.\n", linkedIssue.Key, linkedIssue.Self)
+		}
+
+		absUrl, viewUrl, err := j.getComponentReadinessURLsForView(view)
+		if err != nil {
+			return nil, err
+		}
+		description += fmt.Sprintf("\n h4. Useful Component Readiness Links:\n")
+		description += fmt.Sprintf("\nWe are proving the following two links for your convenience:\n")
+		description += fmt.Sprintf("\n- Click [here|%s] to access the component readiness page generated at the time this issue was created.\n", absUrl)
```
dgoodwin, line 340: suggest: "to access a snapshot of the component readiness page at the time this issue was generated"

pkg/componentreadiness/jiraautomator/jiraautomator.go
```diff
@@ -0,0 +1,597 @@
...
+		description += fmt.Sprintf("\n- Click [here|%s] to access the component readiness page that will be generated at the time when it is clicked. This is useful for developers to verify their fixes.\n", viewUrl)
```
dgoodwin, line 341: to access the component readiness page with latest data

pkg/componentreadiness/jiraautomator/jiraautomator.go
```diff
@@ -0,0 +1,597 @@
...
+// b  adding the Regression label defined by LabelJiraAutomator.
+// c. setting a description with links to CR
+// d. for pre-release, setting "Release Blocker" label to Approved
+func (j JiraAutomator) createNewJiraIssueForRegressions(view crtype.View, component string, tests []crtype.ReportTestSummary, linkedIssue *jira.Issue) (*jira.Issue, error) {
+	if len(tests) > 0 {
+		description := `Component Readiness has found a potential regression in the following tests:`
+		for i, test := range tests {
+			// Only show stats for the worst regression
+			if i == 0 {
+				description += fmt.Sprintf("\n h4. Most Regressed Test:\n{code}%s{code}\n", test.TestName)
```
dgoodwin, line 316: I still worry that strongly separating out a most regressed test sends the wrong message, and could lead to someone focusing on one issue and ignoring the others. From our perspective we really don't care, they're all equal in my eyes. 

I can't remember what we decided when this was discussed, if we decided to go ahead with this then ignore me. :)

pkg/componentreadiness/jiraautomator/jiraautomator.go
```diff
@@ -0,0 +1,597 @@
...
+				},
+				Labels: []string{jiratype.LabelJiraAutomator},
+			},
+		}
+		created, err := j.jiraClient.CreateIssue(&issue)
+		if err != nil {
+			return created, err
+		}
+		// Set Release Blocker field. Jira does not allow setting those during creation. So do it in separate step.
+		return j.updateReleaseBlocker(created, view.SampleRelease.Release)
```
dgoodwin, line 372: I think we need to add an Instructions section. Things like how we expect teams to handle breaking out the jiras, do not close this jira or another one will be automatically created, if/when we'll automatically close this jira, any other guidelines we want to convey.

pkg/componentreadiness/jiraautomator/jiraautomator.go
```diff
@@ -0,0 +1,597 @@
...
+	for component, tests := range componentRegressedTests {
+		sort.Slice(tests, func(i, j int) bool {
+			return tests[i].ReportStatus < tests[j].ReportStatus
+		})
+		componentRegressedTests[component] = tests
+	}
+	return componentRegressedTests, nil
+}
+
+func (j JiraAutomator) integrateJiraForView(view crtype.View) error {
```
dgoodwin, line 453: Suggest automateJirasForView

pkg/componentreadiness/jiraautomator/jiraautomator.go
```diff
@@ -0,0 +1,597 @@
...
+	logger.Info("jira integration for view")
+
+	report, err := j.getComponentReportForView(view)
+	if err != nil {
+		logger.WithError(err).Error("error getting report for view")
+	}
+
+	componentRegressedTests, err := j.groupRegressedTestsByComponents(report)
+	if err != nil {
+		logger.WithError(err).Error("error getting regressed tests from report")
```
dgoodwin, line 465: This will continue processing with a bad componentRegressedTests? Suspect you want to return the err here.

pkg/componentreadiness/jiraautomator/jiraautomator.go
```diff
@@ -0,0 +1,597 @@
...
+	}
+	for component, tests := range componentRegressedTests {
+		if component != "Test Framework" {
+			continue
+		}
+		// fetch jira bugs
+		// resetting this to our component to test
+		component = "Test Framework"
```
dgoodwin, line 473: Looks like this is no longer needed.

pkg/componentreadiness/jiraautomator/jiraautomator.go
```diff
@@ -0,0 +1,597 @@
...
+	}
+
+	componentRegressedTests, err := j.groupRegressedTestsByComponents(report)
+	if err != nil {
+		logger.WithError(err).Error("error getting regressed tests from report")
+	}
+	for component, tests := range componentRegressedTests {
+		if component != "Test Framework" {
+			continue
+		}
```
dgoodwin, line 470: This might not be needed either with the whitelist now?

pkg/componentreadiness/jiraautomator/jiraautomator.go
```diff
@@ -0,0 +1,597 @@
...
+func (j JiraAutomator) createNewJiraIssueForRegressions(view crtype.View, component string, tests []crtype.ReportTestSummary, linkedIssue *jira.Issue) (*jira.Issue, error) {
+	if len(tests) > 0 {
+		description := `Component Readiness has found a potential regression in the following tests:`
+		for i, test := range tests {
+			// Only show stats for the worst regression
+			if i == 0 {
+				description += fmt.Sprintf("\n h4. Most Regressed Test:\n{code}%s{code}\n", test.TestName)
+				description += getProbabilityString(test.ReportStatus, *test.FisherExact)
+				description += getStatsString("Sample (being evaluated)", test.SampleStats, view.SampleRelease.RelativeStart, view.SampleRelease.RelativeEnd)
+				description += getStatsString("Base (historical)", *test.BaseStats, view.BaseRelease.RelativeStart, view.BaseRelease.RelativeEnd)
```
dgoodwin, line 319: With the merger of pass rate testing this text gets much more complicated. There may be no basis. There may be no probability string. There *is* however the new "explanations" slice of strings, which the server should populate with WHY it marked this a regression. I did this to hopefully prevent proliferation of if/else logic for how we present this info to users. I suggest scaling back the details here to just show the explanations, avoiding any assumption about fishers.

pkg/componentreadiness/jiraautomator/jiraautomator.go
```diff
@@ -0,0 +1,597 @@
...
+		}
+		return j.jiraClient.UpdateIssue(&issue)
+	}
+	return issue, nil
+}
+
+// groupRegressedTestsByComponents groups the regressed tests in the report by components.
+// It also takes into consideration a special column grouping. The idea is if a certain column variant (e.g. metal) has more
+// red cells, we will not want to create a jira card for all the components affected. Instead, we will just create
+// one jira card for metal platform.
```
dgoodwin, line 395: I expected this was a "we will ALSO create a card for the metal platform". I expected the regular behavior to continue. What do you think?

pkg/componentreadiness/jiraautomator/jiraautomator.go
```diff
@@ -0,0 +1,597 @@
...
+		}
+
+		// No existing issues, create new one
+		if len(issues) == 0 {
+			_, err := j.createNewJiraIssueForRegressions(view, component, tests, nil)
+			if err != nil {
+				log.WithError(err).Error("error creating jira issue")
+			}
+		} else {
+			selected := issues[0]
```
dgoodwin, line 489: The contents of this else look intensive, might be good to break out an updateJiraForRegressions to match the above.

cmd/sippy/automatejira.go
stbenjam: The data in this command line option sounds like it's starting to get into config file territory.

cmd/sippy/automatejira.go
```diff
@@ -0,0 +1,157 @@
...
+
+type AutomateJiraFlags struct {
+	BigQueryFlags                                *flags.BigQueryFlags
+	GoogleCloudFlags                             *flags.GoogleCloudFlags
+	CacheFlags                                   *flags.CacheFlags
+	ComponentReadinessFlags                      *flags.ComponentReadinessFlags
+	JiraOptions                                  prowflagutil.JiraOptions
+	SippyURL                                     string
+	ComponentWhiteListStr                        string
+	ComponentWhiteList                           sets.String
```
stbenjam, line 34: Industry is moving away from white/black list terminology. We should use allowlist / blocklist

e.g. https://developers.google.com/style/inclusive-documentation#replace

pkg/componentreadiness/jiraautomator/jiraautomator.go
```diff
@@ -0,0 +1,597 @@
...
+									log.WithError(err).Error("error creating jira issue")
+								}
+							}
+						}
+					}
+				}
+			}
+		}
+		// test code. Only process one component for testing purposes
+		break
```
dgoodwin, line 582: No longer needed with the whitelist?

pkg/apis/jira/v1/types.go
```diff
@@ -1,5 +1,30 @@
...
+	StatusClosed         = "Closed"
+	StatusModified       = "MODIFIED"
+)
+
+const (
+	ProjectKeyOCPBugs = "OCPBUGS"
+)
+
+const (
+	LabelJiraAutomator = "Regression"
```
dgoodwin, line 20: ComponentAutomatedRegression
Something a little more specific would be good here, doesn't have to be this, but clarity that this came from our code.

pkg/componentreadiness/jiraintegrator/jiraintegrator.go
xueqzhan: We are going to keep daily update with summary. Will collect feedback when this is deployed and tested by a team.

cmd/sippy/automatejira.go
xueqzhan: We should discuss about this too. Right now it is hardcoded in the code.

cmd/sippy/automatejira.go
xueqzhan: ComponentAllowList doesn't seem to convey the same as ComponentWhiteList. I think I am going to use includeComponents.

pkg/componentreadiness/jiraautomator/jiraautomator.go
xueqzhan: Will keep all tests on the same level.

pkg/componentreadiness/jiraautomator/jiraautomator.go
xueqzhan: How about the following:

Recommendations

This is an automatically generated Jira card against the component based on stats generated from component readiness dashboard. Here are some useful recommendations when dealing with this card:
1. It is best to use this card as a placeholder for all the regressed tests for your component. A linked issue can be created for each regressed test. Please only close this issue when all known regressed tests are believed fixed.
2. Please do not remove “Release Blocker” label. The bot will automatically add it back if regressed tests continues showing for the component.

Anything else?

pkg/componentreadiness/jiraautomator/jiraautomator.go
xueqzhan: Do we really want to create for both? I thought the goal was to avoid bombarding teams when there is a column of red cells due to other root cause.

pkg/componentreadiness/jiraautomator/jiraautomator.go
xueqzhan: WIll do.

dgoodwin:
/lgtm

Let's merge and iterate, several pases made already, most recent commits look good in isolation.

PR #2050, opened 2024-10-22
Add a new LayeredProduct variant

stbenjam:
This is to support having a variant that classifies jobs as enabled for virtualization or other layered products.


stbenjam:
/hold

Want to discuss more

PR #2064, opened 2024-10-28
Add Procedure and RarelyRun variants

dgoodwin:
These are tentative new additions for PoC'ing the support for including longer timeframes of rarely run jobs in component readiness.


pkg/variantregistry/ocp.go
```diff
@@ -407,6 +410,17 @@ func (v *OCPVariantLoader) IdentifyVariants(jLog logrus.FieldLogger, jobName str
...
 	}
 
+	if etcdScaling.MatchString(jobName) {
+		variants[VariantProcedure] = "etcd-scaling"
+		variants[VariantRarelyRun] = "true"
+	} else if cpuPartitioning.MatchString(jobName) {
+		variants[VariantProcedure] = "cpu-partitioning"
+		variants[VariantRarelyRun] = "true"
+	} else {
+		variants[VariantProcedure] = "none"
```
stbenjam, line 420: VariantNoValue

pkg/variantregistry/ocp.go
```diff
@@ -407,6 +410,17 @@ func (v *OCPVariantLoader) IdentifyVariants(jLog logrus.FieldLogger, jobName str
 		variants[VariantSuite] = "unknown" // parallel perhaps but lots of jobs aren't running out suites
 	}
 
+	if etcdScaling.MatchString(jobName) {
+		variants[VariantProcedure] = "etcd-scaling"
+		variants[VariantRarelyRun] = "true"
+	} else if cpuPartitioning.MatchString(jobName) {
+		variants[VariantProcedure] = "cpu-partitioning"
+		variants[VariantRarelyRun] = "true"
```
stbenjam, line 418: Could we find a way to avoid a boolean?

What about putting jobs into tier buckets? Experimental jobs might be automatically excluded from component readiness, and it would let us remove the SQL hacks that filter out specific jobs. Maybe call it JobTier?

| **Category**    | **Description**                                      | **Default Analysis**          |
|-----------------|------------------------------------------------------|------------------------|
| `rare`          | Runs less than daily, used for rarely run analysis   | Rarely run    |
| `standard`      | Runs at least daily, not on payloads                 | Fishers, pass rate             |
| `informing`     | Informing on payloads                                | Fishers, pass rate               |
| `blocking`      | Blocking on payloads                                 | Fishers, pass rate               |
| `experimental`  | Default ignored in component readiness (CR)          | None                |

stbenjam: The data we should already have, too. For informing/blocking, Ken added that data in the Sippy config a while back: https://github.com/openshift/sippy/blob/master/config/openshift.yaml#L8409

dgoodwin: I like this, I'll at least add the variant and rare for now and we can expand with the others in a followup. may get it all today, will see.

dgoodwin: Pushed!

dgoodwin:
/hold

Ideally I want to roll this out with the config flag ready in DPCR as updating 14k jobs will take forever.

pkg/variantregistry/ocp.go
dgoodwin: Fantastic idea.

dgoodwin:
New variants look to have rolled out smoothly.

PR #2074, opened 2024-10-31
TRT-1338: Potential replacement for the remaining 35 minute matview: Test analysis by job

dgoodwin:
In an attempt to ultimately allow the test details page to show more history than 2 weeks, I went after the last slow matview. The theory was we were wasting time recalculating past days that no longer change. I wanted to replace it with a daily summary, calculated by bigquery, stored in a permanent postgresql table. That new table is partitioned, new partitioned are created each day when we load for that day.

The loader will check the last imported day, and if none are found load the last two weeks. I've been populating prod's db while working through this, so we already have data going back to Oct 29. The loader will then load from whatever the last day we have is, up to yesterday. Partitions are not cleaned up at this time.

The analysis page now loads 30 days instead of 14, we'll see how that goes and adjust if needed. We could then add a  date picker to the filters to let a user willing to wait to see more data do so.

The by variants view is updated to use the new by_job partitioned table, the queries now must specify a date as we're no longer querying the entire matview that was limited to two weeks.

This will eliminate the last expensive matview running around 30 mins each hour, replacing it with a daily update that would typically be for just one day, run once a day, and taking 12 mins on my local system.


stbenjam:
Did you try any bigger batch sizes? These rows are small, I bet you could insert 50K at once

dgoodwin:
> Did you try any bigger batch sizes? These rows are small, I bet you could insert 50K at once

I tried 10k and quickly ran into that parameter size problem we always used to see on postgres. I'll test 5 or so.

stbenjam:
Oh ok, probably not a huge deal, we can just let it run over a weekend or so

dgoodwin:
Down to 17m per day by wrapping the whole creation in one transaction instead of many smaller ones. This would only be for one fetchdata per day once we cross the 8am UTC threshold.

dgoodwin:
How do we feel about the general approach?  I've got it loading up a weeks worth of data to prod now, if we like it I can  try to seed 2 months manually letting it run for approx 18 hours.

dgoodwin:
I've loaded up a week of data in prod db in 2 hours. I then setup testview to do the variant query against this new table and it still returns immediately. I suspect this approach will let us chart much longer.

dgoodwin:
Added a couple new commits that import one day at a time transactionally, and autocreate partitions for each day. Then will test how queries perform with this in place, hopefully we can do our standard 2 week query and allow the user to select longer ranges and wait if desired, as we discussed.

Creation of the partition table is manual, I'll have to figure that out before this can go in.

```sql
CREATE TABLE test_analysis_by_job_by_dates (                                                                                                                             
    date timestamp with time zone,                                                                                                                                       
    test_id bigint,                                                                                                                                                      
    release text,                                                                                                                                                        
    job_name text,                                                                                                                                                       
    test_name text,                                                                                                                                                      
    runs bigint,                                                                                                                                                         
    passes bigint,                                                                                                                                                       
    flakes bigint,                                                                                                                                                       
    failures bigint                                                                                                                                                      
) PARTITION BY RANGE (date);                                                                                                                                             
                                                                                                                                                                         
CREATE UNIQUE INDEX test_release_date                                                                                                                                    
ON test_analysis_by_job_by_dates (date, test_id, release, job_name); 
```

```
INFO[2024-11-13T13:54:28.094-04:00] loading variants from bigquery...                                                                                                    INFO[2024-11-13T13:54:35.234-04:00] variants loaded from bigquery in 7.140694503s  jobs=15069
INFO[2024-11-13T13:54:46.678-04:00] job cache created with 7714 entries from database 
INFO[2024-11-13T13:54:46.678-04:00] starting 1 loaders...                                                                                                                
INFO[2024-11-13T13:54:46.678-04:00] starting loader "prow" with metrics wrapper                                                                                          
INFO[2024-11-13T13:54:46.753-04:00] importing test analysis by job for dates: [2024-10-29 2024-10-30 2024-10-31 2024-11-01 2024-11-02 2024-11-03 2024-11-04 2024-11-05 20
24-11-06 2024-11-07 2024-11-08 2024-11-09 2024-11-10 2024-11-11 2024-11-12]       
INFO[2024-11-13T13:54:47.398-04:00] job cache created with 7714 entries from database            
INFO[2024-11-13T13:54:52.402-04:00] test cache created with 125532 entries from database         
INFO[2024-11-13T13:54:52.402-04:00] Loading test analysis by job daily summaries  date=2024-10-29
INFO[2024-11-13T13:54:52.402-04:00] CREATE TABLE IF NOT EXISTS test_analysis_by_job_by_dates_2024_10_29 PARTITION OF test_analysis_by_job_by_dates
                FOR VALUES FROM ('2024-10-29') TO ('2024-10-30');  date=2024-10-29
INFO[2024-11-13T13:54:52.433-04:00] partition created                             date=2024-10-29
INFO[2024-11-13T13:55:15.442-04:00] inserting 1406285 rows                        date=2024-10-29                                                                        INFO[2024-11-13T14:27:41.439-04:00] insert complete after 32m25.997654534s        date=2024-10-29
INFO[2024-11-13T14:27:41.439-04:00] Loading test analysis by job daily summaries  date=2024-10-30
INFO[2024-11-13T14:27:41.439-04:00] CREATE TABLE IF NOT EXISTS test_analysis_by_job_by_dates_2024_10_30 PARTITION OF test_analysis_by_job_by_dates
                FOR VALUES FROM ('2024-10-30') TO ('2024-10-31');  date=2024-10-30                                                                                       
INFO[2024-11-13T14:27:41.521-04:00] partition created                             date=2024-10-30
INFO[2024-11-13T14:28:06.475-04:00] inserting 1542754 rows                        date=2024-10-30
        INFO[2024-11-13T14:58:18.866-04:00] insert complete after 30m12.390888535s        date=2024-10-30
INFO[2024-11-13T14:58:18.866-04:00] Loading test analysis by job daily summaries  date=2024-10-31                                                                        INFO[2024-11-13T14:58:18.866-04:00] CREATE TABLE IF NOT EXISTS test_analysis_by_job_by_dates_2024_10_31 PARTITION OF test_analysis_by_job_by_dates
                FOR VALUES FROM ('2024-10-31') TO ('2024-11-01');  date=2024-10-31                                                                                       
INFO[2024-11-13T14:58:18.914-04:00] partition created                             date=2024-10-31
INFO[2024-11-13T14:58:39.345-04:00] inserting 1427864 rows                        date=2024-10-31
INFO[2024-11-13T15:23:49.438-04:00] insert complete after 25m10.093027879s        date=2024-10-31                     
```

dgoodwin:
First test, oct 29 - nov 14, 11s with no caching.
Daily imports take 20-30 minutes. I'm manually triggering them each day. Soon we'll have a good bit more than 2 weeks and can see how it would work if someone wanted to extend the date range.

dgoodwin:
I think this is ready for final review now.

dgoodwin:
e2e is now timing out because I stopped running the loader every day, so it's trying to populate from late Nov to now and there isn't enough time. Fortunately the job does a day at a time transactionally so there should be no impact on the data itself, but I think if I get it running today through the last few weeks, e2e should work.

dgoodwin:
I checked with a dev backup that is not including these tables, how the app behaves locally. The test details page just shows the overall line on the by job by variant graphs. For dev I think this seems reasonably ok but if you want to load those, it's going to take some time.

As for this e2e problem, I could break them out from the prow loader into a separate one, but this means it's now possible to load from prow without getting the by job and by variant test breakdowns. 

Another option would be to add a way to configure it to just populate one day, but that doesn't make for much of a graph. 

The bigquery query used here does not filter on release, and in e2e we only use 4.16 specifically for reasonable load times. I will try filtering to just the releases sippy knows about, this might mean we could load two weeks in a reasonable timeframe.

side note, this matview being replaced is at times over 60 minutes all by itself now.

dgoodwin:
e2e and lint should now be fixed.

dgoodwin:
THanks, first update could take a number of hours, which will block subsequent fetches. I might release hold for the weekend.

dgoodwin:
/hold cancel

I've synced everything up to 14th so the first fetchdata will need to do one update.

PR #2088, opened 2024-11-06
Respect minFail in pass rate comparisons

dgoodwin:
We think this will be more fair and realistic for component readiness
stop ship decisions.


stbenjam:
/lgtm

It's consistent from a UI perspective to obey the flag and avoid frustration with red squares from infra flakes, but we end up softening the 95% goal.  Hopefully we can figure out a way to boost sample sizes cheaply.

PR #2108, opened 2024-11-14
TRT-1107: migrate openshift-trt github bot to app

sosiouxme:
This is just for sippy-daemon which uses the creds to make Risk Analysis comments.

It should continue to work with the github token, can use the app when configured with a client key.


sosiouxme:
/hold
i put a wrong id in :roll_eyes:

PR #2125, opened 2024-11-20
TRT-1772: Hacky solution to test rarely run jobs results

dgoodwin:
Getting this to work fully in some kind of generic fashion across the UI
and query params will be extremely difficult. We're not even sure the
idea will work, so for now this PR adds a hacky temporary solution.

A scheduled query is doing a full table scan every day and copying
rarely run junit results to a much smaller table in bigquery,
junit_rarely_run_jobs. This allows us to scan it without the immense
cost.

For this temporary view, I added a hack where if we see you're querying
for JobTier rare, we will flip to using this other table.

Side effect, if someone uses the UI and adds in this JobTier, suddenly
their results will be fully busted.

### DIFF
--- a/pkg/api/componentreadiness/component_report.go
+++ b/pkg/api/componentreadiness/component_report.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"regexp"
+	"slices"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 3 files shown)


pkg/api/componentreadiness/component_report.go
```diff
@@ -382,6 +386,19 @@ func (c *componentReportGenerator) getCommonTestStatusQuery(allJobVariants crtyp
 		jobNameQueryPortion = pullRequestDynamicJobNameCol
 	}
 
+	// TODO: this is a temporary hack while we explore if rarely run jobs approach is actually going to work.
+	// A scheduled query is copying rarely run job results to a separate much smaller table every day, so we can
+	// query 3 months without spending a fortune. If this proves to work, we will work out a system of processing
+	// this as generically as we can, but it will be difficult.
+	junitTable := defaultJunitTable
```
neisw, line 393: https://github.com/openshift/sippy/pull/1993#discussion_r1834310375 makes me chuckle :-)

PR #2145, opened 2024-11-29
Display last failure time for regressed tests

dgoodwin:
New column for regressed tests with the last failure time, in the same x days ago format we use for regressed since.

I'd like to get bug data into this chart but its logistically more challenging, it could be done, but it ties the app to the jira caching structure Brad built, or we pull it from our postgres which may not be enabled if the deployment is just running component readiness.

### DIFF
--- a/pkg/api/componentreadiness/component_report.go
+++ b/pkg/api/componentreadiness/component_report.go
@@ -12,6 +12,7 @@ import (
 	"sync"
 	"time"
 
+	"cloud.google.com/go/civil"
 	"github.com/openshift/sippy/pkg/util"
 
 	"cloud.google.com/go/bigquery"
@@ -454,6 +455,7 @@ func (c *componentReportGenerator) getCommonTestStatusQuery(allJobVariants crtyp
 						COUNT(cm.id) AS total_count,
 						SUM(adjusted_success_val) AS success_count,
 						SUM(adjusted_flake_count) AS flake_count,
+						MAX(CASE WHEN adjusted_success_val = 0 THEN modified_time ELSE NULL END) AS last_failure,
 						ANY_VALUE(cm.component) AS component,
 						ANY_VALUE(cm.capabilities) AS capabilities,
 					FROM (%s)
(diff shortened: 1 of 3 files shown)


pkg/api/componentreadiness/component_report.go
```diff
@@ -1889,6 +1905,9 @@ func (c *componentReportGenerator) generateComponentTestReport(ctx context.Conte
 				testStats.ReportStatus = crtype.NotSignificant
 			}
 		}
+		if !sampleStats.LastFailureTime.IsZero() {
+			testStats.LastFailure = &sampleStats.LastFailureTime
```
neisw, line 1909: The linter has complaints.  Maybe
```
lastFailureTime := sampleStats.LastFailureTime
testStats.LastFailure = &lastFailuretime
```

pkg/apis/api/componentreport/types.go
```diff
@@ -230,6 +231,9 @@ type ReportTestStats struct {
 
 	// BaseStats may not be present in the response, i.e. new tests regressed because of their pass rate.
 	BaseStats *TestDetailsReleaseStats `json:"base_stats,omitempty"`
+
+	// LastFailure is the last time the regressed test failed.
+	LastFailure *time.Time `json:"last_failure"`
```
neisw, line 236: I wasn't going to mention it but since the linter had issues..  I was curious about consistency here with [TestStatus.LastFailureTime](https://github.com/openshift/sippy/pull/2145/files#diff-5d08c27e4f3c98d1a7feba2a80cb7b626430eea406630aa85857aaea13d72368R127) it's minor and not worth a lot of effort so either way really.

dgoodwin: Good catch thanks, made them consistent with the shorter form.

pkg/api/componentreadiness/component_report.go
dgoodwin: Pushed a fix.,

PR #2150, opened 2024-12-04
TRT-1781: sanitize sippy-server parameter inputs

sosiouxme:
this also helps a bit with [TRT-1890](https://issues.redhat.com//browse/TRT-1890) (normalization) since values all go into named parameters instead of directly in the query.

### DIFF
--- a/pkg/util/param/param.go
+++ b/pkg/util/param/param.go
@@ -1,0 +1,67 @@
+package param
+
+import (
+	"net/http"
+	"regexp"
+	"strings"
+
+	log "github.com/sirupsen/logrus"
+)
+
+// Cleanse removes unexpected characters from an input parameter value.
+// This is useful for sanitizing dynamic SQL queries built from user input.
+func Cleanse(name string) string {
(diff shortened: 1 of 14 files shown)


pkg/util/param/param.go
```diff
@@ -0,0 +1,68 @@
...
+	"samplePRNumber":   numRegexp,
+}
+
+// SafeRead returns the value of a query parameter only if it matches the given regexp.
+// this should be used to validate query parameters that are not otherwise validated.
+func SafeRead(req *http.Request, name string) string {
+	re, ok := paramRegexp[name]
+	if !ok {
+		logrus.Warnf("code BUG: request for unknown param %s", name)
+		return ""
```
dgoodwin, line 60: This looks like it might trip someone up in a way that could cost a lot of time trying to figure out why. I'd advocate for log.Fatal here, its a clear programmer error and they need to know asap, wdyt?

sosiouxme: yep, sounds good!

PR #2162, opened 2024-12-10
Refactor component readiness query generators to  generate entire queries

dgoodwin:
Previously the bulk of the query was calculated outside the query generator, and then passed into it. This PR breaks query generators out into a separate file, encapsulates all query generation within the generator, and introduces constructors to help ensure callers are specifying required parameters.

This *should* just be refactoring with no impact to behavior. In my testing, this seems to have no impact on cache keys as the query was a private field on the objects and the ComponentReportGenerator is what's actually getting serialized for the key. 

This is being done to prep for work for rarely run queries, you may notice a couple benign hints in the code as to what's coming there.


pkg/api/componentreadiness/component_report.go
```diff
@@ -929,6 +483,22 @@ func (c *componentReportGenerator) getTestStatusFromBigQuery(ctx context.Context
...
+				VariantKey:          "JobTier",
+				VariantValue:        "rare",
+				TableName:           "junit_rarely_run",
+				RelativeStartOffset: "90d",
+			},
+		}
+
+	*/
+	// TODO: Perform additional parallel queries for any table overrides based on the included variants,
+	// and merge the results.
```
dgoodwin, line 500: This is the portion I started on for rarely run jobs, I will resume here after merging if we can live with this disabled for the time being.

pkg/apis/api/componentreport/types.go
```diff
@@ -161,6 +161,20 @@ type ReportRow struct {
...
+	VariantKey   string `json:"variant_key"`
+	VariantValue string `json:"variant_value"`
+	// TableName is an alternative table to look for query results for this variant. If empty, we will
+	// use the default table.
+	TableName string `json:"table_name"`
+	// RelativeStartOffset will be used to alter the lookback for the sample query
+	// when this variant is included in the report and we redirect to another table.
+	// If empty, we will use the requests sample start time.
+	RelativeStartOffset string `json:"relative_start_offset"`
+}
```
dgoodwin, line 176: Another hint as to the plan for rarely run jobs.

dgoodwin:
Needs a rebase, working on it.

pkg/api/componentreadiness/querygenerators.go
```diff
@@ -311,7 +340,184 @@ func (f *fallbackTestQueryGenerator) getFallbackBaseQueryStatus(ctx context.Cont
...
+			commonParams = append(commonParams, bigquery.QueryParameter{
+				Name:  "TestId",
+				Value: c.TestID,
+			})
+		}
+	}
+	return queryString, groupString, commonParams
+}
+
+func fetchTestStatus(ctx context.Context, query *bigquery.Query) (map[string]crtype.TestStatus, []error) {
```
sosiouxme, line 586: `fetchTestStatus` is confusingly close to `queryTestStatus`. i might like something to indicate that it's just executing the query and returning results, like `fetchTestStatusResults`

`fetchJobRunTestStatus` is already distinct enough but for consistency with the above might be `fetchJobRunTestStatusResults`

sosiouxme (commented):
Glad to see this section getting some love, increasing clarity. I like the name updates. The following are suggestions for you consideration... I wouldn't see a problem merging as-is.

sosiouxme:
/lgtm
/hold
(remove as you see fit)

pkg/api/componentreadiness/querygenerators.go
```diff
@@ -0,0 +1,742 @@
...
+			{
+				Name:  "Repo",
+				Value: s.ComponentReportGenerator.SampleRelease.PullRequestOptions.Repo,
+			},
+			{
+				Name:  "PRNumber",
+				Value: s.ComponentReportGenerator.SampleRelease.PullRequestOptions.PRNumber,
+			},
+		}...)
+	}
```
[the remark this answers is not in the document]
dgoodwin: If it's alright leave this one for the future as I'd like to move to the next phase and it's a little larger, but it's a good idea. Others should now be addressed.

PR #2166, opened 2024-12-11
Improve test reporting for feature gates

stbenjam:
Sippy has historically showed the "working" percentage (success +
flake), but we're ready to deprecate this concept and show pass rate
only (strict success).  This is the first part: only in core sippy.  CR will come later.

We already do this for the o/api feature gate analyzer, which uses strict success. This
results in confusion when people try to do archeology in Sippy and see the rates
reported (working) don't match o/api's requirements (success).

While Sippy users can change the view from Working -> Passing, it's not
very discoverable.

This PR:

- Adds a "Feature Gates" page per release, that links to the tests-by-variant
- Deprecates "Working" reporting from sippy -- only show success pass rates
- Caches test result pages for speedier loading

### DIFF
--- a/pkg/apis/api/types.go
+++ b/pkg/apis/api/types.go
@@ -9,6 +9,7 @@ import (
 	"cloud.google.com/go/bigquery"
 	"cloud.google.com/go/civil"
 	"github.com/lib/pq"
+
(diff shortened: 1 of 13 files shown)


stbenjam:
YOLO

/hold cancel

PR #2167, opened 2024-12-11
component readiness: enforce techpreview in 4.19+

stbenjam:
Starting in 4.19 we'd like to enforce TP reliability by having it in the main dashboard.

### DIFF
--- a/config/views.yaml
+++ b/config/views.yaml
@@ -29,6 +29,7 @@ component_readiness:
           - amd64
         FeatureSet:
           - default
+          - techpreview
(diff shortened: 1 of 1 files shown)


config/views.yaml
```diff
@@ -29,6 +29,7 @@ component_readiness:
           - amd64
         FeatureSet:
           - default
+          - techpreview
```
neisw, line 32: Wondered if we should include it by default for sno, other views but it isn't the only option that only applies to main..

PR #2172, opened 2024-12-12
TRT-1912: Treat flake as failures in component readiness

xueqzhan:
/hold for now.

### DIFF
--- a/pkg/api/componentreadiness/component_report.go
+++ b/pkg/api/componentreadiness/component_report.go
@@ -1671,7 +1671,7 @@ func (c *componentReportGenerator) matchBaseRegression(testID crtype.ReportTestI
 		baseRegression = regressionallowances.IntentionalRegressionFor(baseRelease, testID.ColumnIdentification, testID.TestID)
 
 		// This could go away if we remove the option for ignoring fallback
-		if baseRegression != nil && baseRegression.PreviousPassPercentage() > getTestStatusSuccessRate(baseStats) {
+		if baseRegression != nil && baseRegression.PreviousSuccessPercentage() > getTestStatusSuccessRate(baseStats) {
 			// override with  the basis regression previous values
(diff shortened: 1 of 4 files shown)


sosiouxme:
I tried it locally, and AFAICS it works great. Flipping the flag changes the results, both in CR and when drilling down to a test result.

I haven't reviewed the code and don't know that we want major changes to sippy right before the break, but it would give David something to enjoy fiddling with :)

pkg/api/componentreadiness/component_report.go
```diff
@@ -1102,7 +1099,7 @@ func (c *componentReportGenerator) matchBaseRegression(testID crtype.ReportTestI
 		baseRegression = regressionallowances.IntentionalRegressionFor(baseRelease, testID.ColumnIdentification, testID.TestID)
 
 		// This could go away if we remove the option for ignoring fallback
-		if baseRegression != nil && baseRegression.PreviousPassPercentage() > getTestStatusSuccessRate(baseStats) {
+		if baseRegression != nil && baseRegression.PreviousPassPercentage(c.FlakeAsFailure) > c.getTestStatusPassRate(baseStats) {
```
neisw, line 1102: Shouldn't we update places like these as well so that flakes can be counted as failures when c.FlakeAsFailure is enaled?

https://github.com/openshift/sippy/blob/b218d0b853064bf3c54625f482ad1940473d5b09/pkg/api/componentreadiness/component_report.go#L1466

https://github.com/openshift/sippy/blob/b218d0b853064bf3c54625f482ad1940473d5b09/pkg/api/componentreadiness/component_report.go#L1458

neisw: Ahh, nevermind so you are separating out the success, failure and flake and then checking the flag in getPassRate

PR #2181, opened 2024-12-16
Use default go version from dnf install

xueqzhan:
The one installed by dnf install is go1.22.7. We don't need to manually install this in the dockerfile.


stbenjam:
Is there a benefit to using the one from `dnf`? It's just going to lag behind again when we're ready to go to a newer one.

PR #2242, opened 2025-01-09
component readiness: single node to blocking for 4.19-main

stbenjam:
Establish regression protection on single node, the current handful of SNO regressions look tackable before 4.19 GA.


stbenjam:
/hold

Pending discussion

PR #2252, opened 2025-01-13
TRT-1941: update cross-compare views

sosiouxme:
also fix `rosa` no longer being a valid installer.

### DIFF
--- a/config/views.yaml
+++ b/config/views.yaml
@@ -66,6 +66,64 @@ component_readiness:
       enabled: true
     regression_tracking:
       enabled: true
+  - name: 4.19-x86-vs-arm
+    base_release:
+      release: "4.19"
+      relative_start: now-7d
+      relative_end: now
+    sample_release:
+      release: "4.19"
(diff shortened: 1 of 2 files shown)


config/views.yaml
```diff
@@ -67,7 +66,7 @@ component_readiness:
       enabled: true
     regression_tracking:
       enabled: true
-  - name: 4.19-ha-vs-single
+  - name: 4.19-x86-vs-arm
```
sosiouxme, line 69: not very exciting yet...
image

config/views.yaml
```diff
@@ -113,10 +109,10 @@ component_readiness:
           - runc
           - crun
       compare_variants:
-        Topology:
-          - single
+        Architecture:
+          - arm64
```
sosiouxme, line 113: note that `multi-a-a` jobs count as arm64, but other heterogeneous jobs (e.g. `multi-x-ax`) do not.

PR #2255, opened 2025-01-14
[WIP] TRT-1922: Use JobTier for filtering jobs

stbenjam:
Start setting `JobTier` to excluded on jobs we don't know about in Sippy's config. This makes Sippy's config authoritative for both CR and Sippy -- it should allow the data to match between the two, and for jobs we don't explicitly opt-in to to stop showing up in Component Readiness.

### DIFF
--- a/config/views.yaml
+++ b/config/views.yaml
@@ -25,6 +25,10 @@ component_readiness:
         Topology: {}
         Upgrade: {}
       include_variants:
+        JobTier:
+          - informing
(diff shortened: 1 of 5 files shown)


dgoodwin:
Any thoughts on how we could get past the fact that the jobtier is set based on the sippy config generator code. Would we pull that code in here and let it run once a day when the variant registry gets updated?

PR #2259, opened 2025-01-15
TRT-1924: fix refactor and handle missing jobs

sosiouxme:
Turns out my [macro-driven refactor](https://github.com/openshift/sippy/pull/2112/files) removed some `return` statements. Oops.

Respond correctly for jobs that request RA but sippy does not import. They still won't get RA but at least it will be clear why not.

The 3 commits are distinct small changes in this file.

### DIFF
--- a/pkg/sippyserver/server.go
+++ b/pkg/sippyserver/server.go
@@ -975,6 +975,7 @@ func (s *Server) jsonJobRunRiskAnalysis(w http.ResponseWriter, req *http.Request
 		jobRunID, err := strconv.ParseInt(jobRunIDStr, 10, 64)
 		if err != nil {
 			failureResponse(w, http.StatusBadRequest, "unable to parse prow_job_run_id: "+err.Error())
+			return
 		}
 
 		logger = logger.WithField("jobRunID", jobRunID)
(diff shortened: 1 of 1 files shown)


pkg/sippyserver/server.go
```diff
@@ -1011,12 +1015,24 @@ func (s *Server) jsonJobRunRiskAnalysis(w http.ResponseWriter, req *http.Request
...
+				// sippy does not import all jobs to its prow_jobs table; for context see
+				// https://redhat-internal.slack.com/archives/C02K89U2EV8/p1736972504210699
+				// for example, PR jobs on GA releases are excluded, and various other jobs
+				errMsg := fmt.Sprintf("ProwJob '%s' is not included in imported jobs so risk analysis will not run.", jobRun.ProwJob.Name)
+				result := apitype.ProwJobRunRiskAnalysis{
+					OverallRisk: apitype.JobFailureRisk{Level: apitype.FailureRiskLevelUnknown, Reasons: []string{errMsg}},
+				}
+				api.RespondWithJSON(http.StatusOK, w, result)
+			} else {
+				failureResponse(w, http.StatusBadRequest, fmt.Sprintf("unable to find ProwJob '%s': %v", jobRun.ProwJob.Name, res.Error))
```
neisw, line 1033: Nit really but, is that the right message?  Seems like ErrRecordNotFound is the 'unable to find ProwJob'  but other errors are 'Failure locating ProwJob...

sosiouxme: Any error here is an error while running the search for a prow job. the first case is just one that's recognized (i.e., it isn't there, no point retrying). if something else goes wrong (connection to db lost or whatever) then that's passed through as a normal error.

neisw: Yep, so you are -1 on changing the failure message :-)

PR #2262, opened 2025-01-16
Require special approval for intentional regs in consecutive releases.

dgoodwin:
Surface the problem via a unit test which will fail PRs adding an
intentional regression for two releases in a row, for the same test.

To work past this, a new overrides.json file is added in a new directory
with very limited approvers. Test output directs to a README in this
directory signalling how to add to the overrides.json, and reach out to
get approval.


dgoodwin:
First run will intentionally fail so we can see what it looks like.

dgoodwin:
Example failure: https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_sippy/2262/pull-ci-openshift-sippy-master-unit/1879892721640935424

stbenjam:
Nit, could we get it to produce junits so it's more obvious in spyglass? `gotestsum` can do it

stbenjam:
> Nit, could we get it to produce junits so it's more obvious in spyglass? gotestsum can do it

It can also be a separate thing to do, it's not exactly related to this.


/lgtm

dgoodwin:
> Nit, could we get it to produce junits so it's more obvious in spyglass? `gotestsum` can do it
Pushed a commit to try, I fear it wont be in the src image we use to run make test

sosiouxme:
yup...
`make: gotestsum: No such file or directory`
it works locally though

dgoodwin:
Better failure now: https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_sippy/2262/pull-ci-openshift-sippy-master-unit/1880027527498436608

PR updated to remove the forced failure now and should be ready for review.

PR #2264, opened 2025-01-16
Generate junit artifacts via gotestsum

dgoodwin:


dgoodwin:
/override ci/prow/junit

Can't pass until this merges

PR #2276, opened 2025-01-22
Add -sno in singleNodeRegex

liangxia:
Some jobs are using `sno (single node OCP)` to indicate single node.


pkg/variantregistry/ocp.go
```diff
@@ -300,7 +300,7 @@ var (
 	s390xRegex              = regexp.MustCompile(`(?i)-s390x|-multi-z-z`)
 	sdnRegex                = regexp.MustCompile(`(?i)-sdn`)
 	serialRegex             = regexp.MustCompile(`(?i)-serial`)
-	singleNodeRegex         = regexp.MustCompile(`(?i)-single-node`)
+	singleNodeRegex         = regexp.MustCompile(`(?i)-single-node|-sno`)
```
neisw, line 303: Might want to make that -sno- so we are sure it isn't a substring of something else.  Also adding some [tests](https://github.com/openshift/sippy/blob/bdfb262d429af922d344d7baa50100beaaa03ac1/pkg/variantregistry/ocp_test.go)  validating some of the job names would be good.

pkg/variantregistry/ocp_test.go
```diff
@@ -223,6 +223,36 @@ func TestVariantSyncer(t *testing.T) {
 				VariantLayeredProduct:   VariantNoValue,
 			},
 		},
+		{
+			job: "periodic-ci-openshift-release-master-nightly-4.16-e2e-aws-ovn-sno-serial",
+			variantsFile: map[string]string{
+				"Topology": "single", // should be ignored
```
neisw, line 229: I don't think this applies as the expected Topology is in fact 'single'.  Unfortunately it looks like the test above is incorrect as well and the likely source.  Could you remove the variantesFile in both places just confirm it isn't picking that up.  Thanks!

liangxia: Updated.  The e2e passed.

PR #2283, opened 2025-01-23
Stop creating ignored tests in db.

dgoodwin:
We check if they're ignored before creating test results, but not when
we check if the test record itself needs to be created, leading to many
thousands of tests in the db we do not want.


dgoodwin:
Tested locally, with this change we stop creating Run multi-stage tests.

Now for cleanup...

PR #2292, opened 2025-01-27
trt-1959: sanitize event target

neisw:


neisw:
/hold
testing security

PR #2315, opened 2025-02-05
add test exceptions for 4.18 GA

dmwest2:
adding tests that are red on component readiness for 4.18 GA exceptions

### DIFF
--- a/pkg/regressionallowances/consecutiveoverrides/overrides.json
+++ b/pkg/regressionallowances/consecutiveoverrides/overrides.json
@@ -1,1 +1,51 @@
-{}
+{    
+   "JiraComponent": "kube-apiserver",
+   "TestID": "openshift-tests:d6b41cee7afca1c2a0b52f9e6975425f",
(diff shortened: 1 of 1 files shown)


pkg/regressionallowances/consecutiveoverrides/overrides.json
```diff
@@ -1 +1,51 @@
-{}
+{    
+   "JiraComponent": "kube-apiserver",
```
dgoodwin, line 2: This file looks like you want it a directory above in regressions_4.18.json. 

Only the budget burn needs to be in this file, and the syntax is simpler, see the README in this dir.

dgoodwin:
I'd like to get a primer on how we can gain insight into a watch count failure as some of the numbers are still quite large prior to our lgtm on this. https://issues.redhat.com/browse/OCPBUGS-42087

neisw:
/close
past the ga window at this point.

PR #2317, opened 2025-02-05
TRT-1703: Consider high risk evaluation result from other unrelated prs

xueqzhan:
With this there are still a few caveats:

1. Some jobs are missing risk analysis file. Those jobs failed to contact sippy during analysis time. I assume those are the ones without external network access. A few examples:
- pull-ci-openshift-installer-main-e2e-vsphere-host-groups-ovn-custom-no-upgrade
- pull-ci-openshift-console-master-okd-scos-e2e-aws-ovn
2. RA does not deal with junits created by sippy (e.g. install units) or the ones created by gather extra (e.g. operator related). But PR commenting still picks them up. Eventually we probably want to filter them out.
3. RA does not deal with some non-openshift-tests tests. Some examples:
- Hypershift tests: pull-ci-openshift-cluster-version-operator-main-e2e-hypershift
- pull-ci-openshift-cluster-ingress-operator-master-e2e-aws-operator


pkg/sippyserver/pr_commenting_processor.go
```diff
@@ -502,6 +512,16 @@ func (aw *AnalysisWorker) processRiskAnalysisComment(prPendingComment models.Pul
 	log.Debugf("Comment added to pendingComments: %s/%s/%s", pendingComment.org, pendingComment.repo, pendingComment.sha)
 }
 
+func setRiskAnalysisHighRiskMetrics(org, repo, number, jobName, jobID string, summary RiskAnalysisSummary) {
+	for _, testSummary := range summary.TestRiskAnalysis {
+		if summary.RiskLevel == api.FailureRiskLevelHigh {
+			riskAnalysisPRTestRiskMetric.WithLabelValues(org, repo, number, jobName, jobID, testSummary.Name).Set(float64(testSummary.Risk.Level.Level))
+		} else {
+			riskAnalysisPRTestRiskMetric.DeleteLabelValues(org, repo, number, jobName, jobID, testSummary.Name)
```
xueqzhan, line 520: I struggled to unset the metric gauge. At first I thought this would work. But in reality, failed test might pass in the next run and will not be evaluated for risk analysis. That means they will not be in the summary.

neisw: I wonder if we need a separate daemon processor for metrics.  Something that could look at the previous 24 hours and track the active count until it dropped off to 0 for 24 hours or so.

pkg/api/job_runs.go
```diff
@@ -280,6 +287,7 @@ func JobRunRiskAnalysis(dbc *db.DB, jobRun *models.ProwJobRun, jobRunTestCount i
 			logger.WithError(err).Errorf("Failed to find matching jobIds for: %s", jobRun.ProwJob.Name)
 		}
 	}
+	logger.Infof("Found %d unfilered matching jobs %d runs for: %s\njobs %+v", len(jobNames), totalJobRuns, jobRun.ProwJob.Name, jobNames)
```
neisw, line 290: `s/unfilered/unfiltered`

pkg/sippyserver/pr_commenting_processor.go
neisw: Could be better to leave the metrics out and focus that work on [TRT-1704](https://issues.redhat.com/browse/TRT-1704)

pkg/api/job_runs.go
```diff
@@ -161,7 +167,7 @@ func findReleaseMatchJobNames(dbc *db.DB, jobRun *models.ProwJobRun, compareRele
 			}
 
 			if len(jobs) > 0 {
-				logger.Infof("Found %d matches with: %s", len(jobs), name)
+				logger.Infof("Found %d matches with: %s\njobs %+v", len(jobs), name, jobs)
```
neisw, line 170: This is a lot of logging isn't it?  An array of jobs each time within a loop...

pkg/api/job_runs.go
```diff
@@ -553,6 +572,55 @@ func runTestRunAnalysis(failedTest models.ProwJobRunTest, jobRun *models.ProwJob
...
+		if err == iterator.Done {
+			break
+		}
+		if err != nil {
+			log.WithError(err).Error("error parsing number of high risk items from bigquery")
+			return false
+		}
+		rowCount = values[0].(int64)
+		if rowCount > 0 {
+			log.Infof("High risk items found in other PRs for job %s test '%s'", jobRun.ProwJob.Name, failedTest.Test.Name)
```
neisw, line 616: How about adding the count to the logging?  Maybe even returning the count so we could use it in a metric

pkg/sippyserver/pr_commenting_processor.go
neisw: At the moment, we only set this for FailureRiskLevelHigh.  I would be interested in a metric that published the count for how many failures we have seen recently for a High Risk Failure, the greater the count the more likely we are detecting a regression.  Again, this work would likely be better in the other card.

pkg/api/job_runs.go
```diff
@@ -540,7 +548,18 @@ func runTestRunAnalysis(failedTest models.ProwJobRunTest, jobRun *models.ProwJob
...
 	if (testResultsVariants != nil && testResultsVariants.CurrentRuns > 0) || (testResultsJobNames != nil && testResultsJobNames.CurrentRuns > 0) {
 		// select the 'best' test result
-		analysis.Risk = selectRiskAnalysisResult(testResultsJobNames, testResultsVariants, jobNames, compareRelease)
+		risk := selectRiskAnalysisResult(testResultsJobNames, testResultsVariants, jobNames, compareRelease)
+		if compareOtherPRs && risk.Level.Level >= apitype.FailureRiskLevelHigh.Level && len(jobRun.PullRequests) > 0 && isHighRiskInOtherPRs(bqc, failedTest, jobRun) {
+			// If the same test/job has high risk in other PRs, we override the risk level
+			analysis.Risk = apitype.TestFailureRisk{
+				Level: apitype.FailureRiskLevelNone,
+				Reasons: []string{
+					"High risk was identified in other PRs first",
```
neisw, line 557: I'm wondering if this shouldn't go straight to none but instead be set to Medium with the reason indicating that the test may have regressed external to the current PR.  `Potential external regression detected for High Risk Test analysis`

pkg/api/job_runs.go
```diff
@@ -106,6 +111,7 @@ func FetchJobRun(dbc *db.DB, jobRunID int64, logger *log.Entry) (*models.ProwJob
 	// Load the ProwJobRun, ProwJob, and failed tests:
 	// TODO: we may want to expand to analyzing flakes here in the future
 	res := dbc.DB.Joins("ProwJob").
+		Preload("PullRequests").
```
neisw, line 114: Was this something that is being used already and is performance boost or left over from previous work (I didn't see a new reference to pull requests but might be missing it.

pkg/api/job_runs.go
```diff
@@ -553,6 +572,55 @@ func runTestRunAnalysis(failedTest models.ProwJobRunTest, jobRun *models.ProwJob
 	return analysis, nil
 }
 
+func isHighRiskInOtherPRs(bqc *bigquery.Client, failedTest models.ProwJobRunTest, jobRun *models.ProwJobRun) bool {
```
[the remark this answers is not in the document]
xueqzhan: yeah, here is an example:

INFO[2025-02-27T22:09:43.361-05:00] 8 High risk item(s) found in other PRs for job pull-ci-openshift-hypershift-main-okd-scos-e2e-aws-ovn test '[sig-arch] Only known images used by tests'

pkg/api/job_runs.go
xueqzhan: I believe Stephen added it here: https://github.com/openshift/sippy/commit/83eba80ed5518368e416fdbeb459dbb8ddce1c2a. Don't think it was used though.

neisw: I just meant the Preload part, I missed why that was added but I think its `jobRun.PullRequests ` https://github.com/openshift/sippy/blob/383af8a28ec5ca37d526c8b73135f9577363c2e7/pkg/api/job_runs.go#L560

PR #2320, opened 2025-02-06
Attempt a fix for postgres CPU

dgoodwin:
In working on reducing matview CPU, I recently moved us to a partitioned
table test_analysis_by_job_by_dates, updated once a day, and we do not
prune old dates as we intended to allow configurable ranges to generate
the test pass rate graphs.

I analyzed running queries on postgres and found many risk analysis
queries, tracing back to this use of the new table without a filter on
dates. This means jobs are constantly running and requesting risk
analysis, and scanning this table that's growing larger and larger
without a date filter. Fixing this may well fix our problem with the
matviews as this could be what's taking all the cpu.

### DIFF
--- a/pkg/api/job_runs.go
+++ b/pkg/api/job_runs.go
@@ -345,7 +345,9 @@ func jobNamesTestResultFunc(dbc *db.DB) testResultsByJobNameFunc {
 			return nil, nil
 		}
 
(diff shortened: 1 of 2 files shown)


dgoodwin:
Tested and working locally, but unsure if it will fix the problems. It's a good possibility though.

sosiouxme (approved):
/lgtm

PR #2321, opened 2025-02-06
TRT-1999: Add API endpoint metrics

dgoodwin:
This allows us to track traffic, response times and status codes, both per endpoint and globally if you sum without labels.


stbenjam:
How does this work with two replicas?

PR #2328, opened 2025-02-07
TRT-1772: Variant Junit Table Overrides (aka Rarely Run Jobs)

dgoodwin:
I spent a lot of time trying to boil these very hacky one off requirements down into something with some semblance of being generic. The result is variant junit table overrides. 

* The server config file can now contain overrides for a variant to point to another junit table in bigquery if included in the request.
* Server side config is used because we obviously do not want the UI/api pointing to random tables.
* When present, we adjust the variants in our queries carefully to exclude this variant when querying the normal sample data, and then run another query to query this variant from the overridden table, and all the test results are then merged together. Because of the above variant changes, the test keys should not collide in the map. Normal sample query is skipped if the overridden variant value is the only value requested, in which case we only do the override query.
* Pass rate comparison is not forced yet. The existing view works fine because we flip on pass rate requirements. But if we want to merge this into the main view, more work will be needed. It's not easy to have a specific test result get a specific comparison today, it may be more so after some refactoring I want to do next. For now we're ok with the rarely-run view as it stands today.
* Old hack for rarely run table is removed.
* Feature does not work with variant cross compare. The API will error if you try and the UI goes blank as we do not today seem to handle API error display. I figure that should be a separate card. 

Unquestionably this is not pretty and further complicates an already complex portion of the code. It has led to a lot of ideas for refactoring I hope to try next but I do not know if those will be successful.

### DIFF
--- a/cmd/sippy/automatejira.go
+++ b/cmd/sippy/automatejira.go
@@ -16,6 +16,7 @@ import (
 	bqcachedclient "github.com/openshift/sippy/pkg/bigquery"
 	"github.com/openshift/sippy/pkg/componentreadiness/jiraautomator"
 	"github.com/openshift/sippy/pkg/flags"
+	"github.com/openshift/sippy/pkg/flags/configflags"
 	"github.com/openshift/sippy/pkg/util/sets"
 	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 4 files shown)


pkg/api/componentreadiness/component_report.go
```diff
@@ -395,18 +422,176 @@ func (c *componentReportGenerator) getTestStatusFromBigQuery(ctx context.Context
...
+			statusCh <- status
+			for _, err := range errs {
+				statusErrCh <- err
+			}
 		}
 
 	}()
 
-	wg.Wait()
+	// fork additional sample queries for the overrides
```
dgoodwin, line 441: I'm pretty sure this breaks if you tried to do more than one override due to the includeVariant logic being an AND.

neisw:
lgtm overall.  Appreciate your adding in the comments and roadmapping things you want to address next.  Looks like Stephen already commented where the linter had concerns.

dgoodwin:
Lint addressed and rebased. Ready for lgtm.

PR #2329, opened 2025-02-07
[wip] TRT-1984: Refactor variant registry code for job tiers 

stbenjam:


cmd/sippy/genocpvariants.go
stbenjam, line 1: Changes here for configflags is just to avoid a dependency cycle

PR #2332, opened 2025-02-08
Match -tp- for FeatureSet:TechPreview

yapei:
QE jobs are using `-tp-` as part of job names to indicate jobs enabled `TechPreviewNoUpgrade`


yapei:
@deads2k @DennisPeriquet May I have your review?

PR #2344, opened 2025-02-11
TRT-1994: Adjustments to feature gates page

stbenjam:
- Add "first seen in" columns to know when a feature gate was born
- This makes sort and filtering server-side so filtering on things like the enabled column actually work
- Default view is now only tech/dev preview for recent releases
- Adds bookmarks
- Also adds the download link

### DIFF
--- a/pkg/apis/api/types.go
+++ b/pkg/apis/api/types.go
@@ -862,3 +862,45 @@ type FeatureGate struct {
 	UniqueTestCount int64          `json:"unique_test_count"`
 	Enabled         pq.StringArray `json:"enabled" gorm:"type:text[]"`
 }
+
+func (fg FeatureGate) GetFieldType(param string) ColumnType {
+	switch param {
+	case "id":
+		return ColumnTypeNumerical
+	case "enabled":
+		return ColumnTypeArray
+	case "unique_test_count":
+		return ColumnTypeNumerical
+	default:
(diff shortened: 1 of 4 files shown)


stbenjam:
/hold

Some more changes probably coming

pkg/apis/api/types.go
```diff
@@ -856,9 +856,54 @@ type SippyViews struct {
...
+	}
+}
+
+func (fg FeatureGate) GetStringValue(param string) (string, error) {
+	switch param {
+	case "feature_gate":
+		return fg.FeatureGate, nil
+	case "release":
+		return fg.Release, nil
+	default:
```
neisw, line 888: Do you need support for FirstSeenIn here?

pkg/apis/api/types.go
```diff
@@ -856,9 +856,54 @@ type SippyViews struct {
...
+		return "", fmt.Errorf("unknown string field %s", param)
+	}
+}
+
+func (fg FeatureGate) GetNumericalValue(param string) (float64, error) {
+	switch param {
+	case "id":
+		return float64(fg.ID), nil
+	default:
```
neisw, line 897: FirstSeenInMajor, FirstSeenInMinor?

stbenjam: Fixed

PR #2359, opened 2025-02-17
Update regressions_4.18.json

dmwest2:
add exception for 50587

### DIFF
--- a/pkg/regressionallowances/regressions_4.18.json
+++ b/pkg/regressionallowances/regressions_4.18.json
@@ -1,1 +1,26 @@
-[]
+[{
+    "JiraComponent": "Etcd",
+    "TestID": "Operator results:45d55df296fbbfa7144600dce70c1182",
+    "TestName": "Operator results.operator conditions etcd",
+    "JiraBug": "https://issues.redhat.com/browse/OCPBUGS-50587",
+    "ReasonToAllowInsteadOfFix": "This is caused by a race between the installer controller and node status controller. The code/API level flaw is that we have a multi-writer field that we didn't catch until now.  The race requires multiple updates and cache updates to happen in a specific sequence that is common, but not guaranteed.  In our testing we can see this race unblock itself multiple times in three hours. This makes the revision rollout extremely slow (this is bad), but not permanently stuck.Since removing masters is a rare activity and since the system unsticks itself eventually, we will ship 4.18.0 with this problem and correct it in a close following z-stream update.",
+    "variant": {
+      "variants": {
+        "Network": "ovn",
+        "Upgrade": "none",
+        "Architecture": "amd64",
+        "Platform": "azure",
+        "FeatureSet": "default",
+        "Suite": "unknown",
+        "Topology": "ha",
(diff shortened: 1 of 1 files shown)


neisw:
@dmwest2 Looks like this regression has shifted to GCP now.  I think we should keep azure and add in gcp

```
[
  {
    "JiraComponent": "Etcd",
    "TestID": "Operator results:45d55df296fbbfa7144600dce70c1182",
    "TestName": "operator conditions etcd",
    "JiraBug": "https://issues.redhat.com/browse/OCPBUGS-50587",
    "ReasonToAllowInsteadOfFix": "This is caused by a race between the installer controller and node status controller. The code/API level flaw is that we have a multi-writer field that we didn't catch until now.  The race requires multiple updates and cache updates to happen in a specific sequence that is common, but not guaranteed.  In our testing we can see this race unblock itself multiple times in three hours. This makes the revision rollout extremely slow (this is bad), but not permanently stuck.Since removing masters is a rare activity and since the system unsticks itself eventually, we will ship 4.18.0 with this problem and correct it in a close following z-stream update.",
    "variant": {
      "variants": {
        "Network": "ovn",
        "Upgrade": "none",
        "Architecture": "amd64",
        "Platform": "azure",
        "FeatureSet": "default",
        "Suite": "unknown",
        "Topology": "ha",
        "Installer": "ipi"
      }
    },
    "PreviousSuccesses": 355,
    "PreviousFailures": 1,
    "PreviousFlakes": 0,
    "RegressedSuccesses": 81,
    "RegressedFailures": 5,
    "RegressedFlakes": 0
  },
  {
    "JiraComponent": "Etcd",
    "TestID": "Operator results:45d55df296fbbfa7144600dce70c1182",
    "TestName": "operator conditions etcd",
    "JiraBug": "https://issues.redhat.com/browse/OCPBUGS-50587",
    "ReasonToAllowInsteadOfFix": "This is caused by a race between the installer controller and node status controller. The code/API level flaw is that we have a multi-writer field that we didn't catch until now.  The race requires multiple updates and cache updates to happen in a specific sequence that is common, but not guaranteed.  In our testing we can see this race unblock itself multiple times in three hours. This makes the revision rollout extremely slow (this is bad), but not permanently stuck.Since removing masters is a rare activity and since the system unsticks itself eventually, we will ship 4.18.0 with this problem and correct it in a close following z-stream update.",
    "variant": {
      "variants": {
        "Network": "ovn",
        "Upgrade": "none",
        "Architecture": "amd64",
        "Platform": "gcp",
        "FeatureSet": "default",
        "Suite": "unknown",
        "Topology": "ha",
        "Installer": "ipi"
      }
    },
    "PreviousSuccesses": 445,
    "PreviousFailures": 1,
    "PreviousFlakes": 0,
    "RegressedSuccesses": 67,
    "RegressedFailures": 6,
    "RegressedFlakes": 0
  }
]
```

dmwest2 (commented):
updated with gcp from review comments.

neisw:
/close
past the ga window at this point

PR #2363, opened 2025-02-18
Exclude virt layered product until we're ready

dgoodwin:


dankenigsberg:
Should this PR be reverted once OpenShift Virt is compliant (as https://issues.redhat.com/browse/CNV-52649 resolves)?

dgoodwin:
Yes, could you add a card on the TRT project in Jira and we'll take a look at your data and see if we can bring it in. If there is already a card just let myself and Forrest Babcock know in the release oversight channel.

PR #2365, opened 2025-02-18
Remove legacy unused watchlist and metadata extractor

dgoodwin:
remove dead code from two unused features.  The watchlist was to give us a tests tab to monitor for anything that had a linked bug. Metadata extractor was used to pull fine grained data our of generic test failure output so we could see how often specific failure modes were  happening. Neither was ever used and both are probably quite broken now.


neisw:
/lgtm
/hold
release when ready

PR #2369, opened 2025-02-19
TRT-1983: Add test for snapshotting job variants

stbenjam:
I'm making some changes to the variant registry, and I want to understand the impacts of those changes. This adds a snapshot test that if an existing job changes variants, alerts you like this:

```
ocp_test.go:1125: Changes for job periodic-ci-openshift-openshift-tests-private-release-4.15-multi-nightly-azure-ipi-proxy-tp-amd-f28-destructive:
  Changed variant: FeatureSet (techpreview → default)
```

You will have to accept these changes by running `make update-variants`, which will record the changes in a snapshot YAML file and make the test pass.  This makes it harder to introduce unintended changes.

If you'd like to test this:
- Change a variant defintions in `pkg/variantregistry/ocp.go`, maybe remove "-tp-" from the techpreview regex
- Run `go test -count 1 ./pkg/variantregistry -run 'TestVariantsSnapshot'` and notice a failure
- Run `make update-variants` and then `go test -count 1 ./pkg/variantregistry -run 'TestVariantsSnapshot'` -- it will pass


dgoodwin:
/lgtm
/hold

All looks fine to me. Release when ready.

PR #2376, opened 2025-02-26
Introduce middleware for release fallback and intentional regressions (attempt 2)

dgoodwin:
- **Reapply "[TRT-1982](https://issues.redhat.com//browse/TRT-1982): Introduce middleware starting with ReleaseFallback impl"**
- **Fix cache key bug with base/sample test details queries**

Found the bug, the private reqOptions on the base/sample query generators meant cache keys were serializing as: time="2025-02-26T14:05:46.953-04:00" level=info msg="cache hit" key="BaseJobRunTestStatus~{\"BaseRelease\":\"4.18\",\"BaseStart\":\"2025-01-27T00:00:00Z\",\"BaseEnd\":\"2025-02-26T16:00:00Z\"}" type=componentreport.JobRunTestReportStatus

Causing all kinds of weird and horrifying wires to be crossed when you loaded the test details pages for subsequent tests. Fix is to make the field public and thus serialized as it was intended, but I missed.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,7 +11,6 @@ RUN mkdir -p /config
 COPY --from=builder /go/src/sippy/sippy /bin/sippy
 COPY --from=builder /go/src/sippy/sippy-daemon /bin/sippy-daemon
 COPY --from=builder /go/src/sippy/scripts/fetchdata.sh /bin/fetchdata.sh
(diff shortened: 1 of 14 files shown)


xueqzhan:
/lgtm

I should have noticed that. :(

dgoodwin:
Me too. :) Actually really easy to miss in review I think, but I should have noticed in testing.

PR #2377, opened 2025-02-26
TRT-1874: rename master to main

sosiouxme:


sosiouxme:
/hold
until the actual name change

PR #2379, opened 2025-02-28
TRT-2017: Port regression tracking from bigquery to postgresql

dgoodwin:
This change will allow us to begin introducing triage / jira assignments without the snapshot the universe hacks and transactional issues that arise due to bigquerys inability to update rows in the streaming buffer for ~90 minutes. My hope is our future direction will be bigquery for big data, postgres for small data that requires updates. 

Component readiness does not require postgres today, this remains the case with this pr, but regression tracking will be disabled in that case. Going forward, we are looking more and more tied.

Includes migration, when the regression tracking command runs it checks if no entries exist in postgres, and assumes this is an indicator to transfer over from bigquery. I'll remove that code in a future PR.

This PR also includes some additional testing on the postgresql portion via the e2e framework we seldom use.

### DIFF
--- a/hack/go-lint.sh
+++ b/hack/go-lint.sh
@@ -28,6 +28,6 @@ else
   $DOCKER run --rm \
     --volume "${PWD}:/go/src/github.com/openshift/sippy${VOLUME_OPTION}" \
     --workdir /go/src/github.com/openshift/sippy \
-    docker.io/golangci/golangci-lint:v1.56.2 \
+    docker.io/golangci/golangci-lint:v1.57.0 \
     golangci-lint "${@}"
 fi
--- a/pkg/api/componentreadiness/component_report.go
+++ b/pkg/api/componentreadiness/component_report.go
@@ -20,6 +20,7 @@ import (
 	"github.com/openshift/sippy/pkg/api/componentreadiness/utils"
 	configv1 "github.com/openshift/sippy/pkg/apis/config/v1"
 	v1 "github.com/openshift/sippy/pkg/apis/sippy/v1"
+	"github.com/openshift/sippy/pkg/db"
 	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
 	"google.golang.org/api/iterator"
@@ -107,6 +108,7 @@ func GetJobVariantsFromBigQuery(ctx context.Context, client *bqcachedclient.Clie
 func GetComponentReportFromBigQuery(
 	ctx context.Context,
 	client *bqcachedclient.Client,
+	dbc *db.DB,
 	prowURL, gcsBucket string,
 	reqOptions crtype.RequestOptions,
 	variantJunitTableOverrides []configv1.VariantJunitTableOverride,
@@ -123,6 +125,7 @@ func GetComponentReportFromBigQuery(
 		gcsBucket:                  gcsBucket,
 		ReqOptions:                 reqOptions,
 		triagedIssues:              nil,
+		dbc:                        dbc,
 		variantJunitTableOverrides: variantJunitTableOverrides,
 	}
 
@@ -144,6 +147,7 @@ func GetComponentReportFromBigQuery(
 type ComponentReportGenerator struct {
 	ReportModified             *time.Time
 	client                     *bqcachedclient.Client
+	dbc                        *db.DB
 	prowURL                    string
 	gcsBucket                  string
 	triagedIssues              *resolvedissues.TriagedIncidentsForRelease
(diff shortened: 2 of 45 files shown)


hack/go-lint.sh
```diff
@@ -28,6 +28,6 @@ else
   $DOCKER run --rm \
     --volume "${PWD}:/go/src/github.com/openshift/sippy${VOLUME_OPTION}" \
     --workdir /go/src/github.com/openshift/sippy \
-    docker.io/golangci/golangci-lint:v1.56.2 \
+    docker.io/golangci/golangci-lint:v1.57.0 \
```
dgoodwin, line 31: This lets make lint work locally as it now is compatible with our go mod version. However it picks up a couple new issues we're not addressing yet. I'll try to get this in a future PR.

pkg/api/componentreadiness/regressiontracker.go
```diff
@@ -116,83 +176,78 @@ type RegressionTracker struct {
 // Run iterates all views with regression tracking enabled and syncs the results of it's
 // component report to the regression tables in bigquery.
```
xueqzhan, line 177: Not bigquery any more

xueqzhan (commented):
Looks simpler than the BQ way!

pkg/api/componentreadiness/component_report.go
```diff
@@ -269,14 +273,17 @@ func (c *ComponentReportGenerator) GenerateReport(ctx context.Context) (crtype.C
 
 	// Load current regression data from bigquery, used to enhance the response with information such as how long
 	// this regression has been appearing in a tracked view.
-	bqs := NewBigQueryRegressionStore(c.client)
-	allRegressions, err := bqs.ListCurrentRegressions(ctx)
-	if err != nil {
-		log.WithError(err).Error("error listing current regressions")
-		errs = append(errs, err)
-		return crtype.ComponentReport{}, errs
+	// We only execute this if we were given a postgres database connection, it is still possible to run
```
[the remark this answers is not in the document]
dgoodwin: I was not able to maintain dual compatible implementations. There was an older bigquery implementation of the interface I brought back because it supported updating records, but we had to throw it away because of the streaming buffer and implement the snapshot the world approach. This resulted in a different interface that doesn't really match how we would work with postgresql. 

I guess perhaps we could do an interface at the RegressionStore level itself with just a very high level "Run", but we'd have an unused impl laying around with no tests which doesn't feel great either. We could always bring it back to life via git if a use case materializes though.

dgoodwin:
Rebased, addressed comment, and spotted a suspiciously unused param in FindOpenRegression, looks like a bug, I've addressed in separate commit.

dgoodwin:
It does indeed look so much simpler than bq was, quite funny going through it realizing the hoops we were jumping through.

hack/go-lint.sh
stbenjam: Which is what CI uses but only after merge

test/e2e/componentreadiness/regressiontracker/regressiontracker_test.go
```diff
@@ -0,0 +1,178 @@
...
+	if res.Error != nil {
+		log.Errorf("error deleting test regressions: %v", res.Error)
+	}
+}
+
+func Test_RegressionTracker(t *testing.T) {
+	require.NotEqual(t, "", os.Getenv("SIPPY_E2E_DSN"),
+		"SIPPY_E2E_DSN environment variable not set")
+
+	psqlFlags := flags.NewPostgresDatabaseFlags(os.Getenv("SIPPY_E2E_DSN"))
```
stbenjam, line 32: What's the reason behind using a different environment variable here? It'd be better if NewPostgresDatabaseFlags didn't take arguments.

scripts/e2e.sh
```diff
@@ -33,24 +38,25 @@ $DOCKER run --name $PSQL_CONTAINER -e POSTGRES_PASSWORD=password -p $PSQL_PORT:5
 echo "Wait 5s for postgresql to start..."
 sleep 5
 
+export SIPPY_E2E_DSN="postgresql://postgres:password@localhost:$PSQL_PORT/postgres"
```
stbenjam, line 41: Why can't this just be SIPPY_DATABSE_DSN?

dgoodwin: This e2e cleans out the regressions table after each test. I wanted to be exceedingly careful someone doesn't accidentally run it against prod, so I thought it made sense to make it a very explicit e2e var that isn't going to accidentally be set in any situation I could come up with. WDYT?

test/e2e/componentreadiness/regressiontracker/regressiontracker_test.go
dgoodwin: Maybe I can find a way to set the flag opt programatically without modifying the constructor.

dgoodwin: A simple flags.DSN did the trick, new change pushed

hack/go-lint.sh
dgoodwin: This implies fixing the lint warnings in this PR, which is probably ok, I'll do that now.

test/e2e/componentreadiness/regressiontracker/regressiontracker_test.go
dgoodwin: That's a good point, it's no code reuse at all really, I'll flip it to no use of flags.

pkg/api/componentreadiness/component_report.go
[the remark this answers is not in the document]
xueqzhan: I thought the same. Just checking. All frequent update tables should probably be postgres.

dgoodwin:
updated, the lint version now in sync with build root, addressed a few lint fixes that become required on new version, no longer using dbflags for postgres.

PR #2380, opened 2025-03-03
TRT-1966: add accessibility-mode to component readiness

smg247:
This mode, configurable next to 'night mode', changes the green icons to blue and red to orange. If there are further accessibility adjustments needed in the future, the toggle can control those as well.

### DIFF
--- a/sippy-ng/src/App.js
+++ b/sippy-ng/src/App.js
@@ -1,3 +1,5 @@
+import { Accessibility, DarkMode, LightMode } from '@mui/icons-material'
+import { AccessibilityModeProvider } from './AccessibilityModeProvider'
 import { CompReadyVarsProvider } from './component_readiness/CompReadyVars'
 import { createTheme, useTheme } from '@mui/material/styles'
 import {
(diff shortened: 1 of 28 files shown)


sippy-ng/src/App.js
```diff
@@ -307,391 +337,415 @@ export default function App(props) {
...
+                              onClick={colorMode.toggleColorMode}
+                              color="inherit"
+                            >
+                              {mode === 'dark' ? <LightMode /> : <DarkMode />}
+                            </IconButton>
+                          </Tooltip>
+                          <Tooltip
+                            title={
+                              accessibilityModeOn
+                                ? 'Toggle normal mode'
```
stbenjam, line 422: Suggest not calling this normal, but instead `Turn off accessibility mode`

smg247:
> LGTM but could we put a button in the component readiness sidebar too?
> 
> If you land on a component readiness link directly, it's not obvious there's a hidden sidebar with more options

This was actually more difficult to implement than I had anticipated, but I have it working now. I needed to add a custom Provider and a couple of components to reduce duplication.

smg247:
The linter failure is unrelated...
/retest

PR #2383, opened 2025-03-04
TRT-1825: limit comment size for new tests

sosiouxme:
just on principle, best not to allow comments to grow unbounded in size

### DIFF
--- a/pkg/sippyserver/pr_commenting_processor.go
+++ b/pkg/sippyserver/pr_commenting_processor.go
@@ -491,6 +491,8 @@ func (aw *AnalysisWorker) determinePrComment(prCommentProspect models.PullReques
 	log.Debugf("Comment added to preparedComments: %s/%s/%s", preparedComment.org, preparedComment.repo, preparedComment.sha)
 }
 
+const commentRowLimit = 20
+
 func buildCommentText(riskAnalyses []RiskAnalysisSummary, newTestRisks []*JobNewTestRisks, sha string) string {
 	sb := &strings.Builder{}
 	if len(riskAnalyses) == 0 && len(newTestRisks) == 0 {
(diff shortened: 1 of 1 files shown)


pkg/sippyserver/pr_commenting_processor.go
```diff
@@ -518,18 +518,30 @@ func buildNewTestRisksComment(sb *strings.Builder, jobRisks []*JobNewTestRisks,
 		SortByJobNameNT(notableJobRisks)
 		sb.WriteString(fmt.Sprintf("New Test Risks for sha: %s\n\n", sha))
 		sb.WriteString("| Job Name | New Test Risk |\n|:---|:---|\n")
+		rows := 0
 		for _, jr := range notableJobRisks {
 			for _, risk := range sortedTestRisks(jr.NewTestRisks) {
+				rows++
+				if rows > 20 {
```
smg247, line 525: You could define a const for `20`, and then subtract `1` from it for the check on line 541.

smg247 (commented):
I had to do something similar with `pj-rehearse`. I ended up writing out the rest of the details in a GCS bucket and providing a link to it. Do you think something like that could be useful here as well?

sosiouxme:
@smg247 that sounds like a strategy to keep in our back pocket, but honestly I don't think it will become relevant enough to make the effort. If your PR really has that much going on in new tests, you either know you're making big changes to test names, or you have no idea that your little change would cascade into this. Either way, the sheer volume and the link to our guidelines are the main message, and the details likely don't matter.

pkg/sippyserver/pr_commenting_processor.go
sosiouxme: introduced `commentRowLimit`

PR #2386, opened 2025-03-05
Prep test details for middleware use

dgoodwin:
Expand middleware to the test details pathway through ComponentReportGenerator, starting with release fallback. The query stage is implemented, and the analysis is primed to be moved to middleware but I ran into a roadblock in the form of triage, embedded deep within the internalGenerate method used for test details. I think this may have to go to middleware before I can complete release fallback for test details.

I did however find a way to clean up the fallback logic leaking into the API in a couple places, previously test details reports would return a report, with a nested subreport in base_override_report (which was actually the overridden report), and the UI would show that if present. The API now returns a list of analyses, the fallback report will be in the first slot if we triggered it, and the original in the second. The UI is not fully ported to handle any number of reports, but on the API side the problem is mostly cleaned out. 

Next step will be to port triage to middleware, then complete test details, then rarely run jobs.


pkg/api/componentreadiness/component_report.go
```diff
@@ -126,6 +127,10 @@ func GetComponentReportFromBigQuery(
 		variantJunitTableOverrides: variantJunitTableOverrides,
 	}
 
+	if os.Getenv("DEV_MODE") == "1" {
+		return generator.GenerateReport(ctx)
+	}
```
dgoodwin, line 132: This double caching, the report as well as the underlying bigquery queries, is a real pain for devel. I'd almost like to remove it and let CR recalc the report every time, but at least for development it's very helpful to benefit from bigquery caching, but not for report generation.

pkg/api/componentreadiness/test_details.go
```diff
@@ -50,60 +56,74 @@ func (c *ComponentReportGenerator) GenerateTestDetailsReport(ctx context.Context
...
+	for _, mw := range c.middlewares {
+		err = mw.TestDetailsAnalyze(&report)
+		if err != nil {
+			return crtype.ReportTestDetails{}, []error{err}
 		}
+	}
 
-		overrideReport := c.internalGenerateTestDetailsReport(ctx, componentJobRunTestReportStatus.BaseOverrideStatus, c.ReqOptions.BaseOverrideRelease.Release, &c.ReqOptions.BaseOverrideRelease.Start, &c.ReqOptions.BaseOverrideRelease.End, overrideSampleStatus)
+	// Generate the report for the fallback release if one was found:
+	// Move all this to the middleware.
```
dgoodwin, line 105: This is poised to move but the use of c.internalGenerateTestDetailsReport gets very complicated particularly with triage. I need to figure out how to extract that before I can proceed here, but picture this block of code in the middleware.

pkg/api/componentreadiness/component_report.go
neisw: IIRC we looked at removing the report caching but there is still a significant hit (in ui loading terms anyway) that made it worthwhile to keep it.

pkg/api/componentreadiness/test_details.go
neisw: Technically I don't think we need `var baseOverrideReport *crtype.ReportTestDetails` any longer, you could just append `overrideReport`.  But with your comment that there is further refactoring to be done seems fine to leave as is for now.

pkg/api/componentreadiness/test_details.go
```diff
@@ -30,6 +31,9 @@ func GetTestDetails(ctx context.Context, client *bigquery.Client, prowURL, gcsBu
 		gcsBucket:  gcsBucket,
 		ReqOptions: reqOptions,
 	}
+	if os.Getenv("DEV_MODE") == "1" {
```
neisw, line 34: I wish I could suggest something more elegant here.   Thought about tucking this into GetLastReportModifiedTime that is used in component_report or a way to override cacheDuration.  

I get you just want to disable the caching for the report rendering and not the underlying queries and this seems effective.  Maybe something for another day.

pkg/api/componentreadiness/middleware/regressionallowances/regressionallowances.go
```diff
@@ -98,3 +98,15 @@ func (r *RegressionAllowances) matchBaseRegression(testID crtype.ReportTestIdent
 
 	return baseStats, baseRelease
 }
+
+func (r *RegressionAllowances) QueryTestDetails(ctx context.Context, wg *sync.WaitGroup, errCh chan error, allJobVariants crtype.JobVariants) {
+	return
```
neisw, line 103: linter no likey

neisw:
I ran it locally and verified fallback works and navigated to a non regressed test as well.  lgtm overall

pkg/api/componentreadiness/component_report.go
dgoodwin: I'm showing about 1.5s request response times with this setup on a very fast laptop, so there's definitely some significant computation being done and could be slower still in prod.

pkg/api/componentreadiness/test_details.go
dgoodwin: I will note that for the next phase, ty!

dgoodwin:
Looks clear now @neisw !

/hold

I'll release tomorrow morning when I can be here to see.

PR #2387, opened 2025-03-05
TRT-2028: Remove prow and GCS bucket flags from Sippy

stbenjam:
This data is in the BQ tables, we don't need to specify it on the command line. This will facillitate jobs from different prows and buckets in one Sippy (i.e., engineering and QE data).

Sippy-daemon still requires GCS bucket, since it crawls the bucket for jobs.


stbenjam:
/hold

Component readiness works, have to see if I broke anything else.  A lot relies on this data.

stbenjam:
Looks ok in my testing, and e2e passed

/hold cancel

cmd/sippy-daemon/main.go
```diff
@@ -25,10 +25,11 @@ import (
...
 type SippyDaemonFlags struct {
-	BigQueryFlags    *flags.BigQueryFlags
-	CacheFlags       *flags.CacheFlags
-	DBFlags          *flags.PostgresFlags
-	GoogleCloudFlags *flags.GoogleCloudFlags
+	BigQueryFlags       *flags.BigQueryFlags
+	CacheFlags          *flags.CacheFlags
+	DBFlags             *flags.PostgresFlags
+	GoogleCloudFlags    *flags.GoogleCloudFlags
+	GoogleStorageBucket string
```
stbenjam, line 32: Sippy daemon keeps this flag, because it walks the buckets.  It's gone from everywhere else.

pkg/api/componentreadiness/query/querygenerators.go
```diff
@@ -405,12 +408,14 @@ func getTestDetailsQuery(
 						ANY_VALUE(cm.jira_component) AS jira_component,
 						ANY_VALUE(cm.jira_component_id) AS jira_component_id,
 						COUNT(*) AS total_count,
+						ANY_VALUE(jobs.prowjob_url) AS prowjob_url,
```
dgoodwin, line 411: ANY_VALUE good here? We're not doing a group by so it looks ok, but caught my eye, I can't recall why it's used here at all...

dgoodwin (commented):
Great simplification, just a question that is probably a non-issue. 

/lgtm
/hold

pkg/api/jobrunintervals/job_run_intervals.go
```diff
@@ -20,31 +21,16 @@ import (
...
-		path := parts[1][1:]
-		log.WithField("path", path).Debug("calculated gcs path")
-		gcsJobRun = gcs.NewGCSJobRun(bkt, path)
+	jobRun, err := api.FetchJobRun(dbc, jobRunID, false, logger)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to fetch job run %d", jobRunID)
 	}
+
+	bkt := gcsClient.Bucket(jobRun.GCSBucket)
+	gcsJobRun := gcs.NewGCSJobRun(bkt, gcsPath)
```
dgoodwin, line 33: Actually a small concern here. We won't have the bucket for pre-existing job runs, so intervals will not load on some jobs correct? This would become less and less important until we hit 3 months. But could we mass update all rows once this rolls out?

pkg/api/componentreadiness/query/querygenerators.go
stbenjam: I just followed the other examples here.  This query could maybe be fixed up to avoid it with group bys but as is bigquery complains, even though there's only one option for prowjob_url.  The links are showing up correctly for me.

```
Error: Return code = 500 (error querying component test details from big query: [googleapi: Error 400: SELECT list expression references jobs.prowjob_url which is neither grouped nor aggregated at [17:49], invalidQuery])
```

pkg/api/jobrunintervals/job_run_intervals.go
stbenjam: Good eye -- I plan to `update prow_job_runs set gcs_bucket = "test-platform-results";` after the schema gets updated in production

PR #2390, opened 2025-03-06
Increase timeout for golangci-lint

dgoodwin:
New version seems too slow for our CI system


dgoodwin:
/override ci/prow/e2e

Irrelevant for this pr

PR #2391, opened 2025-03-07
Fix group by problem in variant registry

stbenjam:
Always use the most recent prowjob_url and gcs_bucket for all rows, and use ANY_VALUE to get the gcs bucket. Fixes:

```
time="2025-03-07T08:05:16.994Z" level=fatal msg="could not execute root command" error="error querying primary list of all jobs: googleapi: Error 400: SELECT list expression references r.gcs_bucket which is neither grouped nor aggregated at [24:3], invalidQuery"
```


sosiouxme (approved):
TIL `ANY_VALUE`
/lgtm

sosiouxme:
/lgtm
(having to work around these edge cases is kinda ugly though)

PR #2395, opened 2025-03-11
OCPERT-33 add/refine qe view for releases 4.14+

rioliu-rh:
- use absolute datetime for base release
- time duration of base release is 30 days. queried earliest modified_time of auto release job from BigQuery
- only keep qe for variant owner, remove service-delivery
- base release version is same as sampel release

### DIFF
--- a/config/qe-views.yaml
+++ b/config/qe-views.yaml
@@ -1,10 +1,169 @@
 ---
 component_readiness:
-  - name: 4.17-qe-main
+  - name: 4.14-qe-main
+    base_release:
+      release: "4.14"
+      relative_start: "2024-10-16T00:00:00Z"
+      relative_end: "2024-11-16T23:59:59Z"
+    sample_release:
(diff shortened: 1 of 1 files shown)


config/qe-views.yaml
```diff
@@ -1,10 +1,169 @@
...
+      db_group_by:
+        Architecture: {}
+        FeatureSet: {}
+        Installer: {}
+        Network: {}
+        Platform: {}
+        Suite: {}
+        Topology: {}
+        Upgrade: {}
+      include_variants:
```
stbenjam, line 26: To show just auto-release you can add this to the views:

```
Procedure:
  - automated-release
```

rioliu-rh: @stbenjam Can you review again, thx

rioliu-rh:
seems like the failed job is not caused by changes in this PR

PR #2399, opened 2025-03-12
Fix missing regressed since values

dgoodwin:
Missed passing the db connection through to the metrics loop here. If the
auth sippy which runs the component-readiness entrypoint command
happened to be the one which found the cache expired, it would populate
it without regressed since values. This would happen roughly half the
time.


dgoodwin:
/override ci/prow/lint

Not this PR, we've picked up npm audit problems. I'll try to fix separately.

PR #2406, opened 2025-03-14
trt-2039: restore triagedIncidentsFor call

neisw:


deads2k:
Tracking it through to impact on the sample calculation and the report status https://github.com/openshift/sippy/blob/372919e49414b7c619489655b29da1f95a5844c0/pkg/api/componentreadiness/component_report.go#L1475-L1479, this lgtm

/lgtm

PR #2409, opened 2025-03-18
TRT-2047: CR UI Updates

neisw:
- Support relative sample times for TestDetails api triage
- Show triage icon for any actively triaged regression
- Include triaged incidents in the Regression Badge counts
- UI cleanup

### DIFF
--- a/pkg/api/componentreadiness/component_report.go
+++ b/pkg/api/componentreadiness/component_report.go
@@ -741,9 +741,12 @@ func updateCellStatus(rowIdentifications []crtype.RowIdentification,
 	}
 }
 
+// getTriagedIssuesFromBigquery will
+// fetch triaged issue from big query if not already present on the ComponentReportGenerator
+// and check the triaged issue for matches with the current testID
 func (c *ComponentReportGenerator) getTriagedIssuesFromBigQuery(ctx context.Context,
 	testID crtype.ReportTestIdentification) (
-	int, []crtype.TriagedIncident, []error) {
+	int, bool, []crtype.TriagedIncident, []error) {
 	generator := triagedIncidentsGenerator{
 		ReportModified: c.GetLastReportModifiedTime(ctx, c.client, c.ReqOptions.CacheOption),
 		client:         c.client,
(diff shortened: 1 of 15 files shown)


sippy-ng/src/component_readiness/CompCapRow.js
```diff
@@ -85,5 +86,5 @@ CompCapRow.propTypes = {
   results: PropTypes.array.isRequired,
   columnNames: PropTypes.array.isRequired,
   filterVals: PropTypes.string.isRequired,
-  accessibilityMode: PropTypes.bool.isRequired,
+  accessibilityMode: PropTypes.bool,
```
neisw, line 89: Debugger was throwing a warning that accessiblityMode was marked required but undefined.  Looks like [default](https://github.com/openshift/sippy/blob/e5957bd255c9570802fd21327f9e8e5f7edcfcbf/sippy-ng/src/component_readiness/CompReadyUtils.js#L123) is set so removed isRequired

pkg/api/componentreadiness/component_report.go
```diff
@@ -1092,16 +1112,17 @@ func (c *ComponentReportGenerator) generateComponentTestReport(ctx context.Conte
 		var testStats crtype.ReportTestStats // This is the actual stats we return over the API
 		var triagedIncidents []crtype.TriagedIncident
 		var resolvedIssueCompensation int
+		activeProductRegression := false
```
smg247, line 1115: nit:
```suggestion
		var activeProductRegression bool
```

sippy-ng/src/component_readiness/CompCapRow.js
neisw: I didn't see the same warning when I tried running the debugger from the prod ui.  I was thinking the initial default value was undefined at the time the debugger was loading and it was throwing the warning.

neisw: I backed these changes out since I don't see them in production.

When debugging locally I see:
```
Warning: Failed prop type: The prop `accessibilityMode` is marked as required in `CompSeverityIcon`, but its value is `undefined`.
```

But we can leave this for another day if it is a regular issue.

neisw: Missed one

PR #2412, opened 2025-03-19
TRT-1983: Set release and job tier for presubmits

stbenjam:
In order to remove the config, we need to correctly classify presubmits with the right job tier and release variants. This would let us replace the regexp part of the sippy configuration[1].

[1] https://github.com/openshift/sippy/blob/0563dd59fab4b2877ddcced48b63d09deab23929/config/openshift-customizations.yaml#L16-L19


pkg/variantregistry/ocp.go
```diff
@@ -611,14 +619,19 @@ func (v *OCPVariantLoader) setJobTier(_ logrus.FieldLogger, variants map[string]
 
 	// Determine job tier from release configuration
 	release := variants[VariantRelease]
-	if util.StrSliceContains(v.config.Releases[release].BlockingJobs, jobName) {
+	switch {
+	case release == "Presubmits":
```
neisw, line 623:
```suggestion
	case release == "Presubmits",  v.config.Releases[release].Jobs[jobName]:
```

neisw: If we wanted to shorted the statement up a little..

PR #2413, opened 2025-03-19
Parallelize job variant registry

stbenjam:
Processing all the jobs with their cluster data serially takes 2.5 hours in CI.  That doesn’t really matter much for something that runs once a day, but I’ve had to use it a few times locally, and I wished it was faster. This parallelizes generating variants, bringing it to around 20 minutes.

### DIFF
--- a/pkg/variantregistry/ocp.go
+++ b/pkg/variantregistry/ocp.go
@@ -6,6 +6,8 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
+	"sync"
+	"sync/atomic"
 	"time"
 
 	"cloud.google.com/go/bigquery"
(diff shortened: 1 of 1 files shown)


pkg/variantregistry/ocp.go
```diff
@@ -115,18 +117,15 @@ ORDER BY j.prowjob_job_name;
...
-	count := 0
 	for {
 		// TODO: last run but not necessarily successful, this could be a problem for cluster-data file parsing causing
 		// our churn. We can't flip the query to last success either as we wouldn't have variants for non-passing jobs at all.
 		// Two queries? Use the successful one for cluster-data?
-		jlr := prowJobLastRun{}
-		err := it.Next(&jlr)
-		if err == iterator.Done {
+		jlr := new(prowJobLastRun)
+		err := it.Next(jlr)
```
sosiouxme, line 127: ... how would this work without passing a reference to be filled in? (i may be missing something...)

sosiouxme (commented):
seems fine other than my question

pkg/variantregistry/ocp.go
stbenjam: Can you clarify what you mean? new creates a pointer to an instance prowJobLastRun{}.   This is mostly existing code, I just prefer to use `new` in for loops to show I'm explicitly creating a new instance in each iteration of the loop.

PR #2432, opened 2025-03-27
TRT-1984: Replace sippy mat view with bigquery tables

xueqzhan:
/hold

A summary of what was done:

1. Created two scheduled query ([junit-7day-comparison](https://console.cloud.google.com/bigquery/scheduled-queries/locations/us/configs/67d1c8e3-0000-26e8-acd6-d4f547f13e84/runs?invt=AbtExw&project=openshift-gce-devel) and [junit-2day-comparison](https://console.cloud.google.com/bigquery/scheduled-queries/locations/us/configs/67d33f44-0000-2b70-b5d9-c82add7f4ec0/runs?invt=AbtExw&project=openshift-gce-devel)) that produced two tables (junit_7day_comparison and junit_2day_comparison) and refresh daily.
2. Those two tables mimic what our existing mat views do. I had to play around shifting some work between the scheduled query and API query back and forth. 
3. The scheduled query runs 1m each and processes 60GB data
4. The API queries themselves take a few seconds (1s for collapsed and 4s for variants) and process 600-800MB of data
5. But the overall APIs take much longer time (2s for collapsed and 16s for variants). I believe most of the time it was due to data transfer. The existing matview based test page suffer the same slowness.
6. I created a PR to hold current state. The change is only done for api/tests endpoint. Because of this, I had to duplicate some structs to make it work for other endpoints as well. 
7. You can play around with it next week when you have some time. I will come back and finish the other endpoints. 
8. Any feedback at this early stage will be appreciated.


pkg/api/tests.go
```diff
@@ -249,3 +336,167 @@ func BuildTestsResults(dbc *db.DB, release, period string, collapse, includeOver
...
+	if fil != nil {
+		rawFilter, processedFilter = fil.Split([]string{"name", "variants"})
+	}
+
+	table := "junit_7day_comparison"
+	if period == "twoDay" {
+		table = "junit_2day_comparison"
+	}
+
+	dataSet := "openshift-gce-devel.ci_analysis_us"
```
neisw, line 356: Can we use [BigQueryFlags](https://github.com/openshift/sippy/blob/main/pkg/flags/bigquery.go#L23) instead

sippy-ng/src/App.js
```diff
@@ -192,6 +195,27 @@ export default function App(props) {
     [setCookie]
   )
 
+  const testTableDBSourcePreference = cookies['testTableDBSource']
+  const [testTableDBSource, setTestTableDBSource] = React.useState(
+    testTableDBSourcePreference
+  )
+  const testTableDBSourceToggle = React.useMemo(
```
neisw, line 202: Might be good to demo this to the team.  I wasn't expecting a live option to switch the new queries on.  I don't recall if we still have sippy_qe where I would expect this would fail currently and I'm curious if others thought it might be confusing or not to end users who discovered the settting and wonder what portions of the app it affects, etc.

pkg/api/tests.go
neisw: Thank you for the demo!  I think we just need to validate the dataset is infact `openshift-gce-devel.ci_analysis_us` since that is the only place we currently have the scheduled query in place.  I can create a separate card for the adhoc query support without the scheduled query that we discussed.

pkg/sippyserver/server.go
```diff
@@ -784,6 +784,18 @@ func (s *Server) jsonTestsReportFromDB(w http.ResponseWriter, req *http.Request)
 	}
 }
 
+func (s *Server) jsonTestsReportFromBigQuery(w http.ResponseWriter, req *http.Request) {
+	// Fall back to postgres if dataset is not ci_analysis_us
+	if s.bigQueryClient == nil || s.bigQueryClient.Dataset != "ci_analysis_us" {
```
xueqzhan, line 789: @neisw I like this redirect better. But I also have another version returning 500 error. Let me know which one you prefer.

PR #2433, opened 2025-03-27
TRT-1936: Triage Phase 1

dgoodwin:
Implements a restful api for component readiness triage records, intended to be somewhat easy to use over http/curl for as long as we need it while we sort out the UI. 

Structure is a little different than the bigquery impl today, Triage records link to regressions from the regression tracker, potentially many to one though this will be rarer. All are stored in postgres which we want for live user edits and to avoid the 90 minute update delays due to streaming buffer in bigquery, or the snapshot the world approach to work around that.

Did some work on e2e test framework so I could validate everything was working as I went. 

hack/triage is a small chatgpt script for easier working with the api, including letting us manage triages with yaml even though the api will be json. 

A new server flag is added to enable write apis. We'll only enable that in the sippy-auth deployment. 

This does not yet feature job run triage that we have in bigquery. As a result we cannot switch to this mechanism yet until that is implemented, as we need it right now for current release so that teams have a clear path to green immediately. With what Luke has been working on I'm not going to touch this, we'll sync up on getting his solution in with this. Right now this data isn't used for anything, I just want to get the first section reviewed.

After this we will hook in job run triage via some means, add middleware to replace the older temporary triage data and start returning it in the APIs as appropriate, and work on the UI to let users start doing some of this.


dgoodwin:
@smg247 @neisw this isn't urgent when you get a chance your review would be helpful.

pkg/api/componentreadiness/triage.go
```diff
@@ -0,0 +1,174 @@
...
+		log.WithError(res.Error).Error("error creating triage record")
+		return triage, res.Error
+	}
+	log.WithField("triageID", triage.ID).Info("triage record created")
+
+	injectHATEOASLinks(&triage)
+	return triage, nil
+}
+
+func UpdateTriage(dbc *db.DB, triage models.Triage) (models.Triage, error) {
```
neisw, line 108: What is the plan for audit create / update?  Maybe we want to get this in and build off of it but do we intend for the audit tracking to be present before enabling?

dgoodwin: Definitely by the time we roll out for others. I do not have a design in mind yet, something attached to the triage record that covers user and what they did though, including addition/removal of job runs, all stored in postgres.

pkg/dataloader/bugloader/bugloader.go
```diff
@@ -118,6 +137,15 @@ func (bl *BugLoader) Load() {
 		}
 		allBugs[b.ID] = b
 	}
+	for _, b := range triageBugs {
+		if _, ok := allBugs[b.ID]; ok {
+			allBugs[b.ID].Jobs = b.Jobs
```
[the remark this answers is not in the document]
dgoodwin: Well spotted!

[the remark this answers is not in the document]
dgoodwin: The whole thing looks like a bad copy, I'm removing it for now, I have more work to do on importing bugs in once used in triage.

pkg/flags/api.go
```diff
@@ -0,0 +1,26 @@
...
+import (
+	"github.com/spf13/pflag"
+)
+
+// APIFlags holds configuration information for Sippy API servers.
+type APIFlags struct {
+	EnableWriteEndpoints bool
+	ListenAddr           string
+	MetricsAddr          string
+	RedisURL             string
```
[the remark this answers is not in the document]
dgoodwin: Looks like another bad copy my bad, this is covered in cache.go

dgoodwin:
Updated with review feedback

pkg/api/componentreadiness/triage.go
neisw: Could I suggest that we work out a minimal design that we use to record who (for the script we just us the service account email id currently) and when?  At least that way we have history of updates and then we can build out specific update types.  Doesn't have to be this PR but feels like it should be a priority before switching over.

dgoodwin: Yes I'll add a jira and get thinking on this asap.

dgoodwin: https://issues.redhat.com/browse/TRT-2075

PR #2434, opened 2025-03-27
upkeep: update variants for tna/tnf

eggfoobar:
updated variants to split out two node arbiter and two node fencing 
manually updated generated file to use names that will be available when re-sync runs


eggfoobar:
/hold

Probably want to wait for the automated sync for the job names, updated manually for unit tests to pass.

eggfoobar:
/unhold

Name change PR has merged in

PR #2435, opened 2025-03-27
NO-JIRA: add a basic error boundary to sippy

smg247:
This keeps us from displaying a blank page when an unknown error occurs


neisw:
/lgtm
/hold
release when ready

PR #2439, opened 2025-03-28
API to enable querying job artifacts

sosiouxme:
Search a set of job runs for artifacts matching a glob and (optionally) content matching a string. API params are:

- `prowJobRuns`: a comma-separated list of prow job run IDs to query
- `pathGlob`: a glob pattern to match against the GCS path (ref. https://cloud.google.com/storage/docs/json_api/v1/objects/list#list-object-glob)
- `textContains` (optional): a string to search for in the contents of the artifacts, returning the containing line

Returns a JSON `jobartifacts.QueryResponse` object listing the job runs, the artifacts that matched the glob, and (optionally) lines within those artifacts that matched. Errors that occurred are also listed per job run. To prevent (accidental) DOS, the number of artifacts and matches returned are limited.

Example query: http://localhost:8080/api/jobs/artifacts?prowJobRuns=1905396175427604480,1905379393581092864,1905368974137233408,1905409701445636096,1905375855236878336,1905396953936564224&pathGlob=artifacts/*e2e*/gather-extra/build-log.txt&textContains=ClusterVersion:

### DIFF
--- a/pkg/api/jobartifacts/apitypes.go
+++ b/pkg/api/jobartifacts/apitypes.go
@@ -2,29 +2,29 @@ package jobartifacts
 
 type QueryResponse struct {
 	Errors  []JobRunError `json:"errors,omitempty"`
-	JobRuns []JobRun      `json:"jobRuns,omitempty"`
+	JobRuns []JobRun      `json:"job_runs,omitempty"`
 }
 
 type JobRun struct {
 	// ID is string because some parsers translate long ints into scientific notation
 	ID      string `json:"id"`
 	URL     string `json:"url"`
-	JobName string `json:"jobName"`
+	JobName string `json:"job_name"`
 	// NOTE: limited per maxJobFilesToScan, sets Truncated if more files match
 	Artifacts             []JobRunArtifact `json:"artifacts"`
-	ArtifactListTruncated bool             `json:"artifactListTruncated"`
+	ArtifactListTruncated bool             `json:"artifact_list_truncated"`
 }
 
 type JobRunError struct {
-	ID    string `json:"jobRunId,omitempty"`
+	ID    string `json:"job_run_id,omitempty"`
 	Error string `json:"error"`
 }
 
 type JobRunArtifact struct {
-	JobRunID    string `json:"jobRunId"`
-	ArtifactURL string `json:"artifactUrl"`
+	JobRunID    string `json:"job_run_id"`
+	ArtifactURL string `json:"artifact_url"`
 	// NOTE: limited per maxFileMatches, sets Truncated if file has more matches
-	MatchedContent   []string `json:"matchedContents,omitempty"`
-	MatchesTruncated bool     `json:"matchesTruncated"`
+	MatchedContent   []string `json:"matched_contents,omitempty"`
+	MatchesTruncated bool     `json:"matches_truncated"`
 	Error            string   `json:"error,omitempty"`
 }
--- a/pkg/api/jobartifacts/apitypes.go
+++ b/pkg/api/jobartifacts/apitypes.go
@@ -1,0 +1,30 @@
+package jobartifacts
+
+type QueryResponse struct {
+	Errors  []JobRunError `json:"errors,omitempty"`
+	JobRuns []JobRun      `json:"jobRuns,omitempty"`
+}
+
+type JobRun struct {
+	// ID is string because some parsers translate long ints into scientific notation
+	ID      string `json:"id"`
+	URL     string `json:"url"`
+	JobName string `json:"jobName"`
+	// NOTE: limited per maxJobFilesToScan, sets Truncated if more files match
+	Artifacts             []JobRunArtifact `json:"artifacts"`
+	ArtifactListTruncated bool             `json:"artifactListTruncated"`
+}
+
+type JobRunError struct {
+	ID    string `json:"jobRunId,omitempty"`
+	Error string `json:"error"`
+}
+
+type JobRunArtifact struct {
+	JobRunID    string `json:"jobRunId"`
+	ArtifactURL string `json:"artifactUrl"`
+	// NOTE: limited per maxFileMatches, sets Truncated if file has more matches
+	MatchedContent   []string `json:"matchedContents,omitempty"`
+	MatchesTruncated bool     `json:"matchesTruncated"`
+	Error            string   `json:"error,omitempty"`
+}
--- a/pkg/api/jobartifacts/manager.go
+++ b/pkg/api/jobartifacts/manager.go
@@ -77,7 +77,7 @@ func (m *Manager) Query(ctx context.Context, query *JobArtifactQuery) (result Qu
 	// the listener before dumping requests to the request channel, since that will block.
 	go func() {
 	responseLoop:
-		for {
+		for len(remaining) > 0 { // done if we received a response for every request
 			select {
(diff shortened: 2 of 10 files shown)


pkg/api/jobartifacts/manager.go
```diff
@@ -0,0 +1,250 @@
...
+	})
+	return
+}
+
+// a jobRunRequest channels a single job run to be processed and returned with matching file paths
+type jobRunRequest struct {
+	jobRun      int64
+	query       *JobArtifactQuery
+	jobRunsChan chan jobRunResponse // when processing is done, send response to this channel
+	ctx         context.Context     // enable cancellation/timeout of query
```
stbenjam, line 118: I usually try to design things where contexts aren't stored in structs. It might be unavoidable in this design, but more context (heh) about this here https://go.dev/blog/context-and-structs

pkg/api/jobartifacts/manager.go
```diff
@@ -0,0 +1,250 @@
...
+		case req := <-m.jobRunChan:
+			jobLog := logger.WithField("jobRunId", req.jobRun).WithContext(req.ctx)
+			jobLog.Debug("Received from jobRunChan")
+			select {
+			case <-req.ctx.Done(): // query is starting too late, request already cancelled
+				jobLog.WithError(ctx.Err()).Warn("Context aborted request for job run artifacts - getting backlogged?")
+				req.jobRunsChan <- jobRunResponse{
+					jobRun: req.jobRun,
+					error:  fmt.Errorf("job run request was not serviced soon enough: %v", req.ctx.Err()),
+				}
```
stbenjam, line 145: Is a continue missing here?

sosiouxme: I don't think so? Once the case is done, should just finish the select and drop out to the next `for` iteration...

Anyway I tried adding it explicitly and no difference.

pkg/api/jobartifacts/manager.go
sosiouxme: If you think of these structs (which are only passed over channels) as a set of parameters for the worker loop, it's certainly not far in spirit from the advice given to always pass as a param. I can't think of another way to get an ephemeral context into an existing goroutine listening on a channel for time-limited/cancellable work orders.

Maybe I'm thinking about it wrong. What I really want is to limit (server-wide) the number of these requests that are running at any given time. Maybe it should just spin up a goroutine for each, and just moderate the global limit on how many are running via the Manager. The way I'm doing it now with long-running listeners doesn't seem to be common.

stbenjam: I prefer fixed numbers of workers over one-off goroutines, I like the design overall. 

Do you have an API call that's reproducing it getting stuck? I tried with a ton of job ID's and a wide path glob and it all returned for me.

`go run -race ./cmd/sippy` didn't report any potential raciness either.  If you can reproduce it getting stuck you can also try pprof to see where the goroutine is stuck.

stbenjam: Actually, I did notice matching on non-existing paths causes a timeout:

```
curl http://localhost:8080/api/jobs/artifacts\?prowJobRuns\=1905561996145725440,1905575305355988992,1905561997831835648,1898704060324777984\&pathGlob\=artifacts\*/e2e\*/gather-extra/\*.zqrsx\&textContainers\=ClusterVersion: | json_reformat
```

sosiouxme: Yes, I found if there are no artifacts then the artifact worker hangs waiting for the first response :roll_eyes: easily solved. I don't think the same thing is happening for jobrun workers.

[This hangs consistently for me](http://localhost:8080/api/jobs/artifacts?prowJobRuns=1905396175427604480,1905379393581092864,1905368974137233408,1905409701445636096,1905375855236878336,1905396953936564224,1905431900567113728,1905384598624276480,1905437951282647040,1905397245008678912,1905425871439663104,1905437951257481216,1905429907559682048,1905396426725134336,1905437951358144512,1905375933192212480,1905427821925240832,1905397301208158208,1905409773596053504,1905434931228577792,1905375906361249792,1905430904625434624,1905433458352590848,1905425871586463744,1905368956546322432,1905396882369155072,1905433924696281088,1905425871506771968,1905397260963811328,1905409647049707520,1905396958948757504,1905390730587672576,1905396933774544896,1905427823036731392,1905369245143797760,1905375961717673984,1905414546382655488,1905397219834466304,1905437951395893248,1905399198543515648&pathGlob=**/build-log.txt&textContains=ClusterVersion:)

sosiouxme: I figured out what was going on here -- didn't start receiving until all were sent, and send was blocking.

pkg/api/jobartifacts/apitypes.go
```diff
@@ -0,0 +1,30 @@
...
+}
+
+type JobRun struct {
+	// ID is string because some parsers translate long ints into scientific notation
+	ID      string `json:"id"`
+	URL     string `json:"url"`
+	JobName string `json:"jobName"`
+	// NOTE: limited per maxJobFilesToScan, sets Truncated if more files match
+	Artifacts             []JobRunArtifact `json:"artifacts"`
+	ArtifactListTruncated bool             `json:"artifactListTruncated"`
```
stbenjam, line 15: I don't think we've written this down anywhere, but Sippy tends to use snake case for the API's, e.g. component readiness responses in `pkg/apis/api/componentreport`.

Again, very inconsistent but many API response types are in `pkg/apis`.

pkg/util/utils.go
```diff
@@ -177,3 +185,47 @@ func AdjustReleaseTime(relTime time.Time, isStart bool, daysAdjustment string, c
 	}
 	return relTime
 }
+
```
stbenjam, line 188: Testing helpers feel like they should be separated from the app's helpers, maybe `pkg/util/testing.go`?

pkg/sippyserver/server.go
```diff
@@ -1286,6 +1333,12 @@ func (s *Server) Serve() {
 			Capabilities: []string{LocalDBCapability},
 			HandlerFunc:  s.jsonJobBugsFromDB,
 		},
+		{
+			EndpointPath: "/api/jobs/artifacts",
+			Description:  "Queries job artifacts and their contents",
+			Capabilities: []string{LocalDBCapability},
+			HandlerFunc:  s.queryJobArtifacts,
+		},
```
stbenjam, line 1341: Should this be cached? If we share the links around, it might be nice if the response was immediate on subsequent queries. Just needs a CacheTime entry to turn it on

stbenjam (commented):
Looks good, just small things

pkg/sippyserver/server.go
sosiouxme: I don't like the idea of caching at this level given there's some possibility of returning  transient error responses among otherwise successful content.

My experience is that responses are pretty quick already (a few seconds even with quite a lot of results), although that's just in testing against my own system.

Also I suspect most usage of this API will be indirect via the UI, where the user may be tweaking the set of job runs included; it may make more sense to cache responses at the individual job run level (and filter out any that saw errors).

I think we can wait until performance becomes problematic and decide how to cache then based on actual usage.

pkg/util/utils.go
sosiouxme: Good idea, makes the leading comment more visible too.

PR #2473, opened 2025-04-09
enrich test details results

sosiouxme:
test_details report data now includes the original sample/base job names (if relevant) and a new attr for job run ID. So it now looks like:

```
{
...
      "job_stats": [
        {
          "job_name": "periodic-ci-openshift-cluster-control-plane-machine-set-operator-release-X.X-periodics-e2e-aws-techpreview",
          "sample_job_name": "periodic-ci-openshift-cluster-control-plane-machine-set-operator-release-4.19-periodics-e2e-aws-techpreview",
          "base_job_name": "",
          "sample_stats": {...},
          "base_stats": {...},
          "sample_job_run_stats": [
            {
              "job_url": "https://prow.ci.openshift.org/view/gs/test-platform-results/logs/periodic-ci-openshift-cluster-control-plane-machine-set-operator-release-4.19-periodics-e2e-aws-techpreview/1906587761641525248",
              "job_run_id": "1906587761641525248",
              "test_stats": {
                "success_rate": 1,
                "success_count": 1,
                "failure_count": 0,
                "flake_count": 0
              }
            },
```

### DIFF
--- a/pkg/apis/api/componentreport/types.go
+++ b/pkg/apis/api/componentreport/types.go
@@ -299,7 +299,8 @@ type TestDetailsJobStats struct {
 }
 
 type TestDetailsJobRunStats struct {
-	JobURL string `json:"job_url"`
+	JobURL   string `json:"job_url"`
+	JobRunID string `json:"job_run_id"`
 	// TestStats is the test stats from one particular job run.
 	// For the majority of the tests, there is only one junit. But
 	// there are cases multiple junits are generated for the same test.
@@ -327,6 +328,7 @@ type JobRunTestIdentification struct {
 
 type JobRunTestStatusRow struct {
 	ProwJob         string   `bigquery:"prowjob_name"`
+	ProwJobRunID    string   `bigquery:"prowjob_run_id"`
 	ProwJobURL      string   `bigquery:"prowjob_url"`
 	TestID          string   `bigquery:"test_id"`
 	TestName        string   `bigquery:"test_name"`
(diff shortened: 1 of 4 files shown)


pkg/apis/api/componentreport/types.go
```diff
@@ -290,7 +290,9 @@ type TestDetailsTestStats struct {
 }
 
 type TestDetailsJobStats struct {
-	JobName           string                   `json:"job_name"`
+	NormalizedJobName string                   `json:"job_name"`
```
neisw, line 293: This saves you from having to update the ui but shouldn't it be  `json:"normalized_job_name"`

neisw: I guess my question would be is it worth renaming at all, it looked like an outlier to me and I suspect someone *might* ask about it later.  That said, I see

```
TriagedIncidents []TriagedIncident     `json:"incidents,omitempty"
ProwJob         string   `bigquery:"prowjob_name"`
```

So there is prior art :smile:

neisw:
/lgtm
/hold
release when ready

pkg/apis/api/componentreport/types.go
sosiouxme: True, smells a little weird. We have a lot of things named "ProwJob" and "JobName" and such scattered around the code, I wanted it to be a little more obvious that this particular usage wasn't actually a job name, and is only used as a map key to correlate base and sample jobs. I'll likely rename the json property too in my follow-on UI change.

sosiouxme:
OOMkilled again... what's up w TP?
/test lint

sosiouxme:
linter updated
/test lint

pkg/apis/api/componentreport/types.go
sosiouxme: https://github.com/openshift/sippy/pull/2480 gets rid of it entirely

PR #2480, opened 2025-04-10
change layout of `test_details` results

sosiouxme:
Rather than a "normalized" job name taking up an oversized column on the left, the real job names are interleaved with their results. Also if a sample or basis job is not present, name and zeroed stats are omitted. I think this is clearer, and it compacts better than the previous layout (where the long job name would become many lines).

image

The normalized name no longer has a use in the UI so omit it.


neisw:
/lgtm
/hold
release when ready
never liked that normalizedjobname anyway :stuck_out_tongue_winking_eye:

PR #2489, opened 2025-04-13
Improve middleware and port regression tracker over

dgoodwin:
Improve the middleware design by moving Transform from acting on db results, which differ between main reports and test details reports requiring two implementations. Instead we move it further back, right before we perform analysis on ReportTestStats. Middleware is given the stats via pointer and has their chance to adjust base stats (fallback / regression analysis), adjust required confidence (regression tracker hysteresis), and inject explanations for what they did.

Regressed tests now include more open regression details from the tracker, though we may want to trim this down later as a little bit is duplicated.

Well poised for new triage impl, I hope.


pkg/api/componentreadiness/middleware/interface.go
```diff
@@ -20,15 +20,8 @@ type Middleware interface {
...
-
-	// TransformTestDetails gives middleware opportunity to adjust the queried base and sample job run data
-	// before we proceed to analysis.
-	TransformTestDetails(status *crtype.JobRunTestReportStatus) error
-
-	// TestDetailsAnalyze gives middleware opportunity to analyze data and adjust the final report before
-	// being returned over the API.
-	TestDetailsAnalyze(report *crtype.ReportTestDetails) error
+	// PreAnalysis gives middleware opportunity to adjust test analysis data prior to running analysis.
+	// Implementations should can alter base/sample data as needed, request confidence levels, and add explanations for
```
neisw, line 24: ~~should~~

pkg/api/componentreadiness/middleware/releasefallback/releasefallback.go
```diff
@@ -73,50 +79,41 @@ func (r *ReleaseFallback) Query(ctx context.Context, wg *sync.WaitGroup, allJobV
...
 		// if we fail to determine the previous release then stop
 		if err != nil {
-			return baseStats
+			break
 		}
 
 		// if we hit a missing release then stop
 		if cachedReleaseTestStatuses, ok = r.cachedFallbackTestStatuses.Releases[priorRelease]; !ok {
-			return baseStats
+			break
```
neisw, line 116: > cc:FallbackBaseTestStatus~{\"BaseRelease\":\"4.17\",\"BaseStart\":\"2024-09-01T00:00:00Z\",\"BaseEnd\":
cc:FallbackBaseTestStatus~{\"BaseRelease\":\"4.17\",\"BaseStart\":\"2024-09-01T00:00:00Z\",\"BaseEnd\":
cc:FallbackBaseTestStatus~{\"BaseRelease\":\"4.17\",\"BaseStart\":\"2024-09-01T00:00:00Z\",\"BaseEnd\":

I haven't fully debugged it yet but noticed that fallback seems to call the same release (4.17) 3 times instead of 4.17, 4.16, 4.15.  

Also noted that the PreAnalysis seems to get called before the fallback query (I think), which may be intended, and breaks here.  Still walking through it.

dgoodwin: I went back to main to check, and this seems reproducible there, to be honest I don't think I've ever seen a 4.16 result match, only ever 4.17. Is it possible this has been like this for awhile?

dgoodwin: going to see if i can repro with a test somehow

neisw: It is possible it came in with the first middleware refactor (or other).  I know it used to go back 3 releases.  Was going to step through it more.  Don't mind if we go with a separate fix, just noticed it when testing locally.

dgoodwin: Let me know what you've spotted for PreAnalysis vs fallback query, the fallback query should be run in middleware Query which definite should not be running after PreAnalysis which comes much later.

dgoodwin: that is an exceptional find then thank you. We have a test verifying double fallback, so perhaps it's in the query side of things from a past refactor, I'll get it fixed.

neisw: https://github.com/dgoodwin/sippy/blob/7199677326c87ccb3fe6044ba3dd12f4e722df47/pkg/api/componentreadiness/middleware/releasefallback/releasefallback.go#L287

Should be `fallbackRelease, err = utils.PreviousRelease(fallbackRelease)` I believe

neisw: So it updates https://github.com/dgoodwin/sippy/blob/7199677326c87ccb3fe6044ba3dd12f4e722df47/pkg/api/componentreadiness/middleware/releasefallback/releasefallback.go#L287

neisw: What I think I was seeing is that this code path correctly goes back to 4.16, etc.  However, since the query does not there are no results and we break out of the loop.  If you can patch up the query I suspect it will be ok.

dgoodwin: Found and fixed with a unit test for the future. A sneak := that was supposed to be an =.

neisw: If you run this locally it will pop up a new [etcd regression](http://localhost:3000/sippy-ng/component_readiness/test_details?Architecture=amd64&Architecture=amd64&FeatureSet=default&FeatureSet=default&Installer=ipi&Installer=ipi&Network=ovn&Network=ovn&NetworkAccess=default&Platform=azure&Platform=azure&Scheduler=default&SecurityMode=default&Suite=unknown&Suite=unknown&Topology=ha&Topology=ha&Upgrade=micro&Upgrade=micro&baseEndTime=2025-04-24%2023%3A59%3A59&baseRelease=4.18&baseStartTime=2025-03-25%2000%3A00%3A00&capability=Alerts&columnGroupBy=Architecture%2CNetwork%2CPlatform%2CTopology&component=Etcd&confidence=95&dbGroupBy=Platform%2CArchitecture%2CNetwork%2CTopology%2CFeatureSet%2CUpgrade%2CSuite%2CInstaller&environment=amd64%20default%20ipi%20ovn%20azure%20unknown%20ha%20micro&flakeAsFailure=false&ignoreDisruption=true&ignoreMissing=false&includeMultiReleaseAnalysis=true&includeVariant=Architecture%3Aamd64&includeVariant=CGroupMode%3Av2&includeVariant=ContainerRuntime%3Acrun&includeVariant=ContainerRuntime%3Arunc&includeVariant=FeatureSet%3Adefault&includeVariant=FeatureSet%3Atechpreview&includeVariant=Installer%3Aipi&includeVariant=Installer%3Aupi&includeVariant=JobTier%3Ablocking&includeVariant=JobTier%3Ainforming&includeVariant=JobTier%3Astandard&includeVariant=LayeredProduct%3Anone&includeVariant=Network%3Aovn&includeVariant=Owner%3Aeng&includeVariant=Owner%3Aservice-delivery&includeVariant=Platform%3Aaws&includeVariant=Platform%3Aazure&includeVariant=Platform%3Agcp&includeVariant=Platform%3Ametal&includeVariant=Platform%3Arosa&includeVariant=Platform%3Avsphere&includeVariant=Topology%3Aha&includeVariant=Topology%3Amicroshift&minFail=3&passRateAllTests=0&passRateNewTests=95&pity=5&sampleEndTime=2025-04-24%2023%3A59%3A59&sampleRelease=4.19&sampleStartTime=2025-04-17%2000%3A00%3A00&testBasisRelease=4.15&testId=openshift-tests-upgrade%3Ae2534e24395e9cf57959cf34805c549a&testName=%5Bbz-etcd%5D%5Binvariant%5D%20alert%2FetcdHighCommitDurations%20should%20not%20be%20at%20or%20above%20info) that looks like it is falling back to 4.15 which had no basis and fails due to 94% pass rate.  Can't say if this is because of these changes but since we fixed fallback wondering if falling back to min pass rate is something we considered before...

neisw: Everything else looks spot on comparing to prod CR

dgoodwin:
Found it, the test details flow is not fully ported to middleware. Looks tricky. PR is not complete without it though so continuing to hold.

PR #2494, opened 2025-04-15
TRT-2080: Include dev start date with Release

neisw:
Pass along DevelStartDate so callers can determine if development has started on the release yet.  Allowing us to add releases prior to actual branch cut / dev start dates

### DIFF
--- a/pkg/api/releases.go
+++ b/pkg/api/releases.go
@@ -508,7 +508,7 @@ func GetReleasesFromBigQuery(ctx context.Context, client *bqcachedclient.Client)
 	return releases, nil
 }
 
-// transformRelease converts the BQ release row to v1.Relesae type
+// transformRelease converts the BQ release row to v1.Release type
 func transformRelease(r apitype.ReleaseRow) v1.Release {
 	release := v1.Release{Release: r.Release, Status: r.ReleaseStatus.String()}
 	if r.GADate.Valid {
@@ -517,7 +517,7 @@ func transformRelease(r apitype.ReleaseRow) v1.Release {
 	}
 	if r.DevelStartDate.IsValid() {
 		develStartDate := r.DevelStartDate.In(time.UTC)
-		release.DevelStartDate = &develStartDate
+		release.DevelopmentStartDate = &develStartDate
 	}
 	return release
 }
--- a/pkg/api/releases.go
+++ b/pkg/api/releases.go
@@ -503,12 +503,21 @@ func GetReleasesFromBigQuery(ctx context.Context, client *bqcachedclient.Client)
 			log.WithError(err).Error("error parsing release row from bigquery")
 			return releases, err
 		}
-		release := v1.Release{Release: r.Release, Status: r.ReleaseStatus.String()}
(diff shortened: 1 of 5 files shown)


pkg/api/releases.go
```diff
@@ -503,12 +503,21 @@ func GetReleasesFromBigQuery(ctx context.Context, client *bqcachedclient.Client)
...
-			gaDate := r.GADate.Date.In(time.UTC)
-			release.GADate = &gaDate
-		}
-		releases = append(releases, release)
+		releases = append(releases, transformRelease(r))
 	}
 	return releases, nil
 }
+
+// transformRelease converts the BQ release row to v1.Relesae type
```
stbenjam, line 511:
```suggestion
// transformRelease converts the BQ release row to v1.Release type
```

pkg/apis/api/types.go
```diff
@@ -714,9 +714,10 @@ type TestOutput struct {
 }
 
 type Releases struct {
-	Releases    []string             `json:"releases"`
-	GADates     map[string]time.Time `json:"ga_dates"`
-	LastUpdated time.Time            `json:"last_updated"`
+	Releases        []string             `json:"releases"`
+	GADates         map[string]time.Time `json:"ga_dates"`
+	DevelStartDates map[string]time.Time `json:"devel_start_dates"`
+	LastUpdated     time.Time            `json:"last_updated"`
```
stbenjam, line 720: I feel like if we're tracking a second set of dates, we might someday add a third, or fourth.   What if we want to track support end dates, for example?

Also s/Devel/Development/.  The extra characters saved I don't think are offset by the reduction in clarity.

WDYT?

```golang
type ReleaseDates struct {
	GA               *time.Time `json:"ga,omitempty"`
	DevelopmentStart *time.Time `json:"development_start,omitempty"`
}

type Releases struct {
	Releases          []string                 `json:"releases"`
	Dates             map[string]ReleaseDates  `json:"dates"`
	DeprecatedGADates map[string]time.Time     `json:"ga_dates"` // for backward compatibility
	LastUpdated       time.Time                `json:"last_updated"`
}
```

pkg/sippyserver/server.go
```diff
@@ -806,8 +808,19 @@ func (s *Server) jsonReleasesReportFromDB(w http.ResponseWriter, req *http.Reque
...
+			releaseDate.GA = release.GADate
+			addDate = true
+		}
+		if release.DevelopmentStartDate != nil {
+			releaseDate.DevelopmentStart = release.DevelopmentStartDate
+			addDate = true
+		}
+		if addDate {
+			response.Dates[release.Release] = releaseDate
 		}
```
sosiouxme, line 824: total nit but it seems like unnecessary logic when you could just do
```suggestion
			response.Dates[release.Release] = releaseDate
		}
		if release.DevelopmentStartDate != nil {
			releaseDate.DevelopmentStart = release.DevelopmentStartDate
			response.Dates[release.Release] = releaseDate
		}
```

stbenjam:
/lgtm
/hold

Release when ready

PR #2516, opened 2025-04-23
TRT-2073: test_details job artifact search

sosiouxme:
Beginnings of a UI for searching up job artifact content from the test_details page.
image

The UX needs a lot of work but it's good enough to play with, and does some really useful stuff already.


dgoodwin:
/lgtm
Not really enough time to do a full in depth review, but it looks low risk and I'd really like to start using this. 
I say we merge and iterate as needed.
/hold

cancel if you're ok with that.

sosiouxme:
works for me
/hold cancel

PR #2517, opened 2025-04-23
TRT-2045: add a description field to `triage`

smg247:
This also formats the link to Jira better so that it takes up less space

### DIFF
--- a/sippy-ng/src/component_readiness/CompReadyUtils.js
+++ b/sippy-ng/src/component_readiness/CompReadyUtils.js
@@ -72,6 +72,8 @@ export const cancelledDataTable = {
     },
   ],
 }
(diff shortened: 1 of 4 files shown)


sippy-ng/src/component_readiness/CompReadyUtils.js
```diff
@@ -72,6 +72,8 @@ export const cancelledDataTable = {
     },
   ],
 }
+export const jiraUrlPrefix = 'https://issues.redhat.com/browse/'
```
stbenjam, line 75: Optional, things like this are good candidates for the environment https://github.com/openshift/sippy/blob/main/sippy-ng/.env.development

Might be useful if you ever want to play around with the staging jira environment

PR #2523, opened 2025-04-25
Revert "Improve middleware and port regression tracker over"

neisw:
Reverts openshift/sippy#2489

I don't think we intended for this to go in yet based on the comments


bradmwilliams (approved):

neisw:
/override ci/prow/e2e

prod is currently panicking, can't be worse

PR #2543, opened 2025-05-01
Time filtering for intervals charts

dgoodwin:
The existing time filtering worked poorly, it was impossible to get to precise times, often buggy, and would still include mass empty rows between intervals actually in the frame.

Add actual interval time filtering, prepopulated with the first/last intervals times. You can then adjust precisely if desired and see only intervals in that timeframe. A reset button is included, and these are working as query params so you can send someone a link to just what you see.

Also took the time to sort the Categories/Sources buttons, and wrap them so we no longer have to scroll to the right all the time.

### DIFF
--- a/sippy-ng/src/prow_job_runs/IntervalsChart.js
+++ b/sippy-ng/src/prow_job_runs/IntervalsChart.js
@@ -493,6 +493,12 @@ export default function IntervalsChart(props) {
     setEnd(event.target.value)
   }
 
+  const resetTimeFilters = (event) => {
+    console.log('reset time filters')
+    setStart(eventIntervals[0].from)
+    setEnd(eventIntervals[eventIntervals.length - 1].to)
+  }
+
   // handleSegmentClicked is called whenever an individual interval in the chart is clicked.
   // Used to display details on the interval and locator in a way that a user can copy if needed.
   function handleSegmentClicked(segment) {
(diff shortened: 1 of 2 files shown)


sippy-ng/src/prow_job_runs/IntervalsChart.js
```diff
@@ -305,19 +327,31 @@ export default function ProwJobRun(props) {
...
   useEffect(() => {
-    updateFiltering()
+    const isReady =
+      selectedSources != null &&
+      eventIntervals != null &&
+      eventIntervals.length > 0 &&
+      overrideDisplayFlag != null // adjust this check as needed
+
+    if (isReady) {
+      console.log('useEffect on intervals/sources')
```
smg247, line 337: do we want to leave this log here?

sippy-ng/src/prow_job_runs/IntervalsChart.js
```diff
@@ -468,23 +532,29 @@ export default function ProwJobRun(props) {
         Loaded {eventIntervals.length} intervals from GCS, filtered down to{' '}
         {filteredIntervals.length}.
       </p>
-      <div>
+      <div style={{ paddingBottom: '1rem' }}>
```
smg247, line 535: nit: `paddingBottom: '1rem'` is used in three separate places, you could make a styleClass for it

PR #2550, opened 2025-05-02
NO-JIRA: enable unused import linting

smg247:
This updates the necessary deps and installs the `eslint-plugin-unused-imports`. There are a couple of rules that need to be updated when installing that, and I had to turn off a new rule from the `react` linter: `react/no-unknown-property` as it was reporting that we weren't allowed to use the `align` property on divs, but there are a few cases where we have done so, and it does have an effect.

The second commit includes all the necessary changes to satisfy the new rule.


stbenjam:
Nice

/lgtm

PR #2564, opened 2025-05-07
NO-JIRA: fix the test checkbox by updating the location of the regression_id in the JSON

smg247:
The API return value changed in https://github.com/openshift/sippy/commit/64b4c66791c10dbc025de8264a08c5ca9653c3a0


dgoodwin:
/lgtm
/override ci/prow/e2e

I'm eager, my bad, thank you for the fix!

PR #2569, opened 2025-05-07
TRT-2100: import css directly to fix styling after unused import removal

smg247:


smg247:
/override ci/prow/e2e

e2e is irrelevant here

PR #2578, opened 2025-05-09
Triage Resolution

dgoodwin:
This is the quick implementation of a dramatically simpler approach to triage. User specify a resolution time on the triage record. (in future, perhaps we detect when the jira transitions state automatically) This immediately gets you a new icon indicating believed to be fixed, but critically we do not artificially make you disappear from the board, now we wait for statistical evidence to clear you from the board. This is both simpler and much safer in that we require actual evidence of a problem disappearing before we let it go green.

If we see a last failure time AFTER the resolution time, we again change to a new state, the dreaded pants on fire. This indicates something is very wrong, either a failed fix or a new problem hitting the same test. 

- **Make room for new status codes**
- **Add resolved time for triage records**
- **Adjust status codes for triaged regressions**
- **Ensure triage/regression mapping works on query**
- **Load regressions with triage data**
- **Fix missing regressions when queried**
- **Display fixed-waiting and fix-failed icons**
- **Fix pants on fire icon not appearing**
- **Allow a page refresh to pickup new triage updates without hitting bq again**


sippy-ng/src/component_readiness/CompReadyTestReport.js
```diff
@@ -293,9 +293,6 @@ export default function CompReadyTestReport(props) {
   //   If there are triage entries, and the status isn't already triaged
   //   (which it could be due to existing incidents during this interim period) increment by 2
   let status = data.analyses[0].status
-  if (triageEntries.length > 0 && status <= -4) {
```
smg247, line 296: We can remove the associated comment now

pkg/api/componentreadiness/middleware/regressiontracker/regressiontracker.go
```diff
@@ -82,6 +83,55 @@ func (r *RegressionTracker) PreAnalysis(testKey crtype.ReportTestIdentification,
...
+				}
+			}
+
+			switch {
+			case allTriagesResolved && testStats.LastFailure != nil && lastResolution.Before(*testStats.LastFailure):
+				// claimed fixed but does not appear to be
+				// aka liar liar pants on fire
+				testStats.ReportStatus = crtype.FailedFixedRegression
+				testStats.Explanations = append(testStats.Explanations, fmt.Sprintf(
+					"Regression is triaged and believed fixed as of %s but failures have been observed as late as %s.",
```
smg247, line 113:
```suggestion
					"Regression is triaged, and believed fixed as of %s, but failures have been observed as recently as %s.",
```

pkg/api/componentreadiness/middleware/regressiontracker/regressiontracker.go
```diff
@@ -82,6 +83,55 @@ func (r *RegressionTracker) PreAnalysis(testKey crtype.ReportTestIdentification,
...
+					"Regression is triaged and believed fixed as of %s.",
+					lastResolution.Format(time.RFC3339)))
+			case testStats.ReportStatus == crtype.SignificantRegression:
+				testStats.ReportStatus = crtype.SignificantTriagedRegression
+				testStats.Explanations = append(testStats.Explanations,
+					"Regression has been triaged to one or more bugs.")
+			case testStats.ReportStatus == crtype.ExtremeRegression:
+				testStats.ReportStatus = crtype.ExtremeTriagedRegression
+				testStats.Explanations = append(testStats.Explanations,
+					"Regression has been triaged to one or more bugs.")
```
smg247, line 128: Maybe:
```suggestion
					"Extreme regression has been triaged to one or more bugs.")
```

pkg/db/query/triage_queries.go
```diff
@@ -28,3 +28,17 @@ func TriagesForRegressionID(dbc *db.DB, regressionID string) ([]models.Triage, e
 	}
 	return triages, res.Error
 }
+
+func ListRegressions(dbc *db.DB, release string) ([]*models.TestRegression, error) {
```
smg247, line 32: Should this be called `ListOpenRegressions`?

PR #2579, opened 2025-05-09
MCO-1652: Add MCO suite tracking

yuqi-zhang:
Follow up to https://github.com/openshift/release/pull/64752 and https://github.com/openshift/origin/pull/29776

Just a draft for now, but since the original tests were added to serial directly and have reached the necessary pass rates before, I went ahead and added it to the views as well for 4.20.

### DIFF
--- a/pkg/variantregistry/ocp.go
+++ b/pkg/variantregistry/ocp.go
@@ -704,6 +704,7 @@ func setProcedure(_ logrus.FieldLogger, variants map[string]string, jobName stri
 		{"-cert-rotation-shutdown-", "cert-rotation-shutdown"},
 		{"-console-operator-", "console-operator"},
 		{"-ipsec", "ipsec"},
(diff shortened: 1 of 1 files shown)


pkg/variantregistry/ocp.go
```diff
@@ -443,6 +443,7 @@ func setSuite(_ logrus.FieldLogger, variants map[string]string, jobName string)
 		{"-etcd-scaling", "etcd-scaling"},
 		{"conformance", "parallel"}, // Jobs with "conformance" but no explicit serial are probably parallel
 		{"usernamespace", "usernamespace"},
+		{"machine-config-operator", "machine-config-operator"},
```
dgoodwin, line 446: I would use Procedure variant here, we're looking to remove Suite. It can't be reliably detected.

yuqi-zhang:
Updated to variant per suggestion and removed view since we don't need a decided view yet

PR #2586, opened 2025-05-12
TRT-2101: improvements to jobartifacts backend

sosiouxme:
* error / timeout API reporting improved
* add caching for job artifact searches

UI improvements planned to take advantage of these...

### DIFF
--- a/pkg/api/jobartifacts/query.go
+++ b/pkg/api/jobartifacts/query.go
@@ -4,6 +4,7 @@ import (
 	"bufio"
 	"compress/gzip"
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"strconv"
@@ -11,9 +12,11 @@ import (
 	"time"
 
 	"cloud.google.com/go/storage"
+	"github.com/openshift/sippy/pkg/apis/cache"
 	"github.com/openshift/sippy/pkg/db"
 	"github.com/openshift/sippy/pkg/db/models"
 	"github.com/openshift/sippy/pkg/util"
+	"github.com/openshift/sippy/pkg/util/sets"
 	log "github.com/sirupsen/logrus"
 	"google.golang.org/api/iterator"
 )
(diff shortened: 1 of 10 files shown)


pkg/api/jobartifacts/query.go
```diff
@@ -159,5 +211,88 @@ func (q *JobArtifactQuery) getFileContentMatches(jobRunID int64, file *storage.O
...
+	if err != nil {
+		log.Errorf("CacheKeyForJobRun should never fail to serialize the key: %v", err)
+		panic(err) // an engineer causing this somehow should find out ASAP
+	}
+	return string(jsonBytes)
+}
+
+const cacheExpiration = 4 * time.Hour
+
+func (q *JobArtifactQuery) SetJobRunCache(ctx context.Context, jobRunID int64, response JobRun) error {
```
dgoodwin, line 260: Caching is just the match data, not full files that matched, do I have that right?

pkg/api/jobartifacts/query_test.go
```diff
@@ -10,72 +10,110 @@ import (
...
-		artifact.MatchedContent.(ContentLineMatches).Matches[0].Match,
+		artifact.MatchedContent.ContentLineMatches.Matches[0].Match,
 		"localhost kernel: GPT: Use GNU Parted to correct GPT errors.",
 		"expected to scan uncompressed text")
 }
 
+func TestFunctional_RequeryCachedTimedOutJobArtifacts(t *testing.T) {
+	ctx := context.Background()
+	jobRunID := int64(1898704060324777984)
+	mgr := NewManager(ctx)
```
dgoodwin, line 80: Are these tests hitting gcs? If so they should move to the test/e2e/ directory and thus the e2e job.

pkg/api/jobartifacts/query.go
sosiouxme: correct.

pkg/api/jobartifacts/query_test.go
sosiouxme: i'm not opposed to that, but be aware they expect production creds, connection, and data to be available during the test. so i've just been running them manually as unit tests (which are skipped during automated unit tests). although production GCS contents this far in the past shouldn't change, it feels a little careless to leave tests around that will break years in the future when it gets cleaned up...

dgoodwin: Tricky. If these are not running as part of the normal unit tests then that's ok for now.

sosiouxme: they compile and run as part of automated unit tests (so we at least get compile failures if code changes signatures), but they immediately self-skip when they find they do not have credentials. and now that i think of it, some of the lookups here are expecting jobRun data from the prod DB (to match the GCS content) and would probably fail against the local DB (unless that's using a backup from prod...)

sosiouxme:
/retest e2e
sippy load timed out...

PR #2589, opened 2025-05-12
TRT-2089: utilize triages returned in the main test_details report

smg247:
There is a caveat here, we don't have the regression data included in the response for tests other than the one we are displaying. This means that we cannot display other tests that are associated to this regression in the table. I believe this is fine, however.


neisw:
/lgtm
/hold
release when ready

PR #2594, opened 2025-05-13
TRT-2105: intervals for payload-job runs

sosiouxme:
the intervals page insisted on finding a job run in the DB even though it has what it needs in most cases from the parameters. rather than break for jobs that don't have entries in the DB (like when invoked with `/payload-job`), the DB lookup is now optional (but authoritative if present).

i think this covers all of the cases we will actually see. we could futher fall back to looking up the job in BQ `jobs` table, but it looks unnecessary.

also there's now a link to the job run.


smg247 (commented):
/lgtm

PR #2603, opened 2025-05-15
minor beautification

sosiouxme:
Two improvements:
* improve some slightly ugly UI code that bothered me
* proper workaround for dynamic links that snyk won't complain about

### DIFF
--- a/sippy-ng/src/component_readiness/CompReadyUtils.js
+++ b/sippy-ng/src/component_readiness/CompReadyUtils.js
@@ -79,34 +79,39 @@ export const cancelledDataTable = {
 export const jiraUrlPrefix = 'https://issues.redhat.com/browse/'
 
 // Make one place to create the Component Readiness api call
-export function getAPIUrl() {
-  return process.env.REACT_APP_API_URL + '/api/component_readiness'
+export function getAPIUrl(endpoint) {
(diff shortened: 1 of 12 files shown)


sippy-ng/src/component_readiness/CompReadyUtils.js
```diff
@@ -79,34 +79,39 @@ export const cancelledDataTable = {
...
-  return process.env.REACT_APP_API_URL + '/api/component_readiness/test_details'
+  return getAPIUrl('component_readiness/test_details')
 }
 
-export function getTriagesAPIUrl() {
-  return process.env.REACT_APP_API_URL + '/api/component_readiness/triages'
+export function getTriagesAPIUrl(id) {
+  return getAPIUrl(
+    id ? `component_readiness/triages/${id}` : 'component_readiness/triages'
+  )
```
smg247, line 106: I'm not sure I like this change. It isn't clear, without understanding the internals of the function, that the `id` arg can be omitted. Maybe we could have a separate function that doesn't take the arg?

sosiouxme: I couldn't include the type indication since i don't really understand how my IDE/linter treat typescript vs javascript but at least including the `null` default is pretty clear.

PR #2617, opened 2025-05-19
Hypershift should be blocking in CR for 4.20+

stbenjam:
I noticed in 4.19, we're not monitoring for regressions in hypershift in CR.  This adds it to the main view for 4.20+


deads2k:
I agree.

/hold cancel

PR #2632, opened 2025-05-22
TRT-2050: allow the ability to delete triage records and remove regressions from them

smg247:
This also adds the `Last Change` field to the modal and triage details page. There is some refactoring of functionalities into separate components included here, as the `UpsertTriageModal` began to get too complex.

### DIFF
--- a/pkg/api/componentreadiness/triage.go
+++ b/pkg/api/componentreadiness/triage.go
@@ -12,13 +12,16 @@ import (
 	"gorm.io/gorm"
 )
 
-func GetTriage(dbc *db.DB, id int) (models.Triage, error) {
-	existingTriage := models.Triage{}
-	res := dbc.DB.Preload("Bug").Preload("Regressions").First(&existingTriage, id)
+func GetTriage(dbc *db.DB, id int) (*models.Triage, error) {
+	existingTriage := &models.Triage{}
+	res := dbc.DB.Preload("Bug").Preload("Regressions").First(existingTriage, id)
 	if res.Error != nil {
+		if errors.Is(res.Error, gorm.ErrRecordNotFound) {
+			return nil, nil
+		}
 		log.WithError(res.Error).Errorf("error looking up existing triage record: %d", id)
 	}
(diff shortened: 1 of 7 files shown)


pkg/sippyserver/server.go
```diff
@@ -1278,7 +1278,7 @@ func (s *Server) jsonTriages(w http.ResponseWriter, req *http.Request) {
 		if triageID > 0 {
 			existingTriage, err := componentreadiness.GetTriage(s.db, triageID)
 			if err != nil {
-				failureResponse(w, http.StatusInternalServerError, err.Error())
+				failureResponse(w, http.StatusNotFound, err.Error())
```
neisw, line 1281: Wondering if you have seen / it is worth explicitly checking explicitly to see if the error is record not found:
`if !errors.Is(err, gorm.ErrRecordNotFound)` 

Could potentially be applied to https://github.com/openshift/sippy/blob/main/pkg/api/componentreadiness/triage.go#L15 and skip the error logging if it is reasonable for it not to be present.  If we don't expect that case then leaving the error logging is fine.

pkg/api/componentreadiness/triage.go
```diff
@@ -164,6 +164,15 @@ func UpdateTriage(dbc *db.DB, triage models.Triage) (models.Triage, error) {
 		triage.BugID = &bug.ID
 	}
 
+	// When we have removed regressions, we must replace the associations directly, as Save will not handle it.
+	// This results in rows being deleted and re-created even when there are no changes, but this should happen
+	// enough that it makes no difference.
```
neisw, line 169: Total nit -> `but this shouldn't happen
	// enough that it makes a difference.`

pkg/sippyserver/server.go
smg247: This is a good idea. Normal navigation shouldn't result in `404's`, but if someone deletes the triage record, and then refreshes the details page, it will result in one. They could also get to that state by navigating to a triage details page manually (with a non-existant id) or by using an old link to one that has since been deleted. None of those are error states.

PR #2633, opened 2025-05-22
don't delete bugs anymore

sosiouxme:
triage records reference bugs and we don't want to remove those. there's no real reason to remove stale bugs, just filter them out of our queries.

### DIFF
--- a/pkg/dataloader/bugloader/bugloader.go
+++ b/pkg/dataloader/bugloader/bugloader.go
@@ -90,45 +90,54 @@ func (bl *BugLoader) Errors() []error {
 }
 
 func (bl *BugLoader) Load() {
+	logger := log.WithField("func", "bugloader.Load")
+	addError := func(err error, msg string) {
+		logger.WithError(err).Error(msg)
+		bl.errors = append(bl.errors, errors.Wrap(err, msg))
+	}
+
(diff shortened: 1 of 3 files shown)


pkg/dataloader/bugloader/bugloader.go
```diff
@@ -90,45 +90,54 @@ func (bl *BugLoader) Errors() []error {
...
 	testCache, err := query.LoadTestCache(bl.dbc, []string{"TestOwnerships"})
 	if err != nil {
-		bl.errors = append(bl.errors, err)
+		addError(err, "error loading test cache")
 		return
 	}
 	testBugs, err := bl.getTestBugMappings(context.TODO(), testCache)
 	if err != nil {
-		panic(err)
+		addError(err, "error loading test bug mappings")
```
neisw, line 109: Are you sure you want to remove the panics?  I believe they were intended to note unrecoverable failures and no further processing should continue.

I don't think we were hitting these cases with what you are fixing so wondering if we really want to change the behavior now?

neisw: If you reviewed with Stephen B and all is good I'm good to.  Just saw the change and wanted to double check.

PR #2643, opened 2025-05-26
Instant Test Details

dgoodwin:
This PR will provide for immediate test details page loads provided you're working off a view with cache priming enabled. This was done in a cost effective manner via the realization that we scan the same date ranges for regressed tests, thus if we could query once, for all regressed tests, we could pay the cost of a single test details page load, but have all test details pages in the cache. 

Essentially an old test details query would do something like this:

```sql
variant_registry_job_name LIKE 'periodic-%'
     OR variant_registry_job_name LIKE 'release-%'
     OR variant_registry_job_name LIKE 'aggregator-%')
    AND cm.id = "openshift-tests:f782f0945697da7b4f7f68ef6f4849b9"
    AND jv_CGroupMode.variant_value IN UNNEST(["v2"])
    AND jv_ContainerRuntime.variant_value IN UNNEST(["crun", "runc"])
    AND jv_JobTier.variant_value IN UNNEST(["blocking", "informing", "standard"])
    AND jv_LayeredProduct.variant_value IN UNNEST(["none"])
    AND jv_Owner.variant_value IN UNNEST(["eng", "service-delivery"])
    AND jv_Architecture.variant_value = "amd64"
    AND jv_FeatureSet.variant_value = "techpreview"
    AND jv_Installer.variant_value = "ipi"
    AND jv_Network.variant_value = "ovn"
    AND jv_Platform.variant_value = "vsphere"
    AND jv_Suite.variant_value = "unknown"
    AND jv_Topology.variant_value = "ha"
    AND jv_Upgrade.variant_value = "none"
    AND branch = "4.19"
```

And a new one will nest multiple of these in brackets, and OR them up. Our query then comes back with the job run rows for each test/variant combo, we slice them up in code, and call into the internal report generation methods with that slice.

To do this I modified the report generator to no longer assume one set of testid/capability/component filters, rather multiple. From the API, this is not exposed, you can only set one. Internally however, currently from the prime-cache command, you can set multiple, one for every regressed test found in a parent report. 

The component-readiness-cache loader command does one top level report, then reconstructs a multi-test ID report generator for test details, runs it, gets the results back for all those tests, and then slices them up by test/variant combos, runs the actual test analysis logic for each, and caches the result with a key that matches what you'd get in actual usage.

The loader will be run as a kube cronjob. We cache our reports with a rounded 4h time into the future UTC, using that time for both the sample end date, and the actual cache duration. i.e. if you cache miss at say 10:30am UTC, we will use 12:00 UTC as the sample end date, and a cache duration of 90 minutes. This means that we can schedule our kube cronjob to run every 4h on the hour, 4am, 8am, 12am etc, all UTC. Thankfully Kube cronjobs are scheduled in UTC. This means that for a view with cache priming enabled, all but the most unluck soul requesting a report between say 4:00 and 4:01, should always cache hit. 

This got brutally complicated with release fallback as for those, we are not scanning the same partitions. I applied the same approach there just within the releasefallback middleware. I slice up all the tests that had the same fallback release, lets say 4.15 4.16 and 4.17, then do ONE query for each of those releases with all tests regressed which fell back to that release. Then we split them up and analyze as before.

This is working and providing identical results to prod sippy. 

If we assume we load at least one test details page per cache period of 4 hours, we would be cost equal. In reality we load many test details pages through the day, so this is likely to be a significant cost savings.

I found a number of cache key bugs, chief among them the force refresh option was using that force in it's cache key, so you'd get a force refreshed report, but no one else would, and if you reloaded the page you'd lose the newer data too. I've implemented a safer means of getting consistent cache keys.

This change does NOT pre-cache the drilldown into a cell/component on the main report, that actually affects the queries quite substantially in a way I haven't figured out how to crack yet. However test details is our main pain point so that comes first.

The prime cache command takes about 45s per view on my laptop.

The resulting redis cache contains about 150mb with 30 regressed tests, over 500mb with 600 regressed tests. We only cache the resulting report which is not a huge chunk of data.


smg247 (commented):
Reviewed the first few files only. What I've seen looks good. I will continue to look if this isn't merged by Monday.

pkg/api/componentreadiness/query/querygenerators.go
```diff
@@ -412,67 +439,90 @@ func getTestDetailsQuery(
...
+	} else {
+		queryString += filterByCrossCompareVariants(c.VariantOption.VariantCrossCompare, includeVariants, &commonParams)
 	}
+	return queryString, groupString, commonParams
+}
+
+// addTestFilters injects query params to limit to one test and variants combo.
+func addTestFilters(
+	testIDOption crtype.RequestTestIdentificationOptions,
+	i int,
```
smg247, line 477: nit: since this has been factored into a function of its own:
```suggestion
	index int,
```

pkg/api/componentreadiness/query/querygenerators.go
```diff
@@ -764,37 +817,133 @@ func (s *sampleTestDetailsQueryGenerator) QueryTestStatus(ctx context.Context) (
...
+			logger.WithError(err).Error("error parsing component from bigquery")
 			errs = append(errs, errors.Wrap(err, "error parsing prowjob from bigquery"))
 			continue
 		}
-		prowName := utils.NormalizeProwJobName(testStatus.ProwJob, reqOptions)
-		status[prowName] = append(status[prowName], testStatus)
+
+		jobRunTestStatusRow, err := deserializeRowToJobRunTestReportStatus(row, it.Schema)
+		if err != nil {
+			err2 := errors.Wrap(err, "error deserializing row from bigquery")
```
smg247, line 885: nit:
```suggestion
			deserializationErr := errors.Wrap(err, "error deserializing row from bigquery")
```

pkg/sippyserver/metrics/metrics.go
```diff
@@ -226,9 +226,12 @@ func updateComponentReadinessMetricsForView(ctx context.Context, client *bqclien
 
 	// Get component readiness report
 	reportOpts := crtype.RequestOptions{
-		BaseRelease:    baseRelease,
-		SampleRelease:  sampleRelease,
-		TestIDOption:   crtype.RequestTestIdentificationOptions{},
+		BaseRelease:   baseRelease,
+		SampleRelease: sampleRelease,
+		// TODO: Needed to match API cache key
```
smg247, line 231: Was this just left over, or is there some action to take in the future here? If it is the latter, can you elaborate on it?

pkg/api/componentreadiness/test_details.go
```diff
@@ -25,55 +25,195 @@ import (
...
-	for _, v := range c.ReqOptions.VariantOption.DBGroupBy.List() {
-		if _, ok := c.ReqOptions.VariantOption.RequestedVariants[v]; !ok {
-			return crtype.ReportTestDetails{}, []error{fmt.Errorf("all dbGroupBy variants have to be defined for test details: %s is missing in %v", v, c.ReqOptions.VariantOption.RequestedVariants)}
+	// Give middleware their chance to adjust the result
+	for _, mw := range c.middlewares {
+		testKey := crtype.ReportTestIdentification{
+			RowIdentification:    report.RowIdentification,
+			ColumnIdentification: report.ColumnIdentification,
+		}
+		for ai := range report.Analyses {
```
smg247, line 63: why is this var named `ai`? analysis index?

pkg/api/componentreadiness/test_details.go
```diff
@@ -25,55 +25,195 @@ import (
...
+
+	reports := []crtype.ReportTestDetails{}
+	for _, tOpt := range c.ReqOptions.TestIDOptions {
+		testKey := crtype.TestWithVariantsKey{
+			TestID:   tOpt.TestID,
+			Variants: tOpt.RequestedVariants,
+		}
+		testKeyStr := testKey.KeyOrDie()
+		if statuses, ok := testKeyTestJobRunStatuses[testKeyStr]; ok {
+			report, errs2 := c.GenerateDetailsReportForTest(ctx, tOpt, statuses)
```
smg247, line 166: nit:
```suggestion
			report, generateReportErrs := c.GenerateDetailsReportForTest(ctx, tOpt, statuses)
```

pkg/api/componentreadiness/test_details.go
```diff
@@ -25,55 +25,195 @@ import (
...
+		return crtype.ReportTestDetails{}, errs
+	}
+
 	now := time.Now()
 	componentJobRunTestReportStatus.GeneratedAt = &now
 
+	// TODO: this is the spot, here we would have base and sample status, for all MultiTestIDOptions
+	// sort them by test and variant
+	// invoke report for each
+	// refactor so we can get multiple reports from one query, break out everything below.
```
smg247, line 207: was this meant to be left here?

pkg/dataloader/crcacheloader/crcacheloader.go
```diff
@@ -0,0 +1,245 @@
...
+			// a regression.
+			for _, triaged := range col.TriagedIncidents {
+				allRegressedTests = append(allRegressedTests, triaged.ReportTestSummary)
+			}
+		}
+	}
+	rLog.Infof("found %d regressed tests in report", len(allRegressedTests))
+	testIDOptions := []crtype.RequestTestIdentificationOptions{}
+	for _, regressedTest := range allRegressedTests {
+		// regressedTest.BaseStats.Release // start // end
```
smg247, line 127: did you mean to leave this here?

pkg/api/componentreadiness/component_report.go
```diff
@@ -241,9 +309,7 @@ func (c *ComponentReportGenerator) initializeMiddleware() {
 	// TODO: Should middleware constructors do the interpretation of the request
 	// and decide if they want to take part? Return nil if not?
 	c.middlewares = append(c.middlewares, regressionallowances2.NewRegressionAllowancesMiddleware(c.ReqOptions))
-	if c.ReqOptions.AdvancedOption.IncludeMultiReleaseAnalysis || c.ReqOptions.BaseOverrideRelease.Release != c.ReqOptions.BaseRelease.Release {
```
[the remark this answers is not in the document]
dgoodwin: Good catch, it's guarded in PreAnalysis, but nowhere else, which I think means we'd do the work of querying even when the feature is disabled. That's a clear indicator how dangerous it is to pull this check out of here, I'll restore it and fix whatever issue it was causing. We don't want to have to check in every interface impl method.

pkg/api/componentreadiness/component_report.go
```diff
@@ -241,9 +309,7 @@ func (c *ComponentReportGenerator) initializeMiddleware() {
 	// TODO: Should middleware constructors do the interpretation of the request
 	// and decide if they want to take part? Return nil if not?
```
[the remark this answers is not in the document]
dgoodwin: I'm going to remove the TODO for now, probably not a big deal for time being.

pkg/api/componentreadiness/test_details.go
dgoodwin: Yeah, maybe too overloaded :) I'll change it

pkg/api/componentreadiness/test_details.go
dgoodwin: My bad, that was where I originally thought it had to happen.

PR #2649, opened 2025-05-28
TRT-2094: add a link to the test_details report for each triage record

smg247:
This adapts functionality from the `RegressedTestsPanel` to utilize in the `TriagedRegressionTestList` on the modal as well. The adapted functionality now resides in `CompReadyUtils`.

This is not shown on the triage details page as we don't have the necessary info available there to determine the URL. This could be enhanced in the future if desired, but I don't think it is necessary.

### DIFF
--- a/sippy-ng/src/component_readiness/CompReadyUtils.js
+++ b/sippy-ng/src/component_readiness/CompReadyUtils.js
@@ -1,4 +1,5 @@
 import { alpha, InputBase, Typography } from '@mui/material'
+import { CompReadyVarsContext } from './CompReadyVars'
 import { formatInTimeZone } from 'date-fns-tz'
 import { safeEncodeURIComponent } from '../helpers'
 import { styled } from '@mui/styles'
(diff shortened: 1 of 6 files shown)


sippy-ng/src/component_readiness/CompReadyUtils.js
```diff
@@ -741,20 +741,40 @@ export function generateTestReport(
...
+  return sortQueryParams(retUrl)
+}
+
 export function generateRegressionCount(regressed_tests, triaged_incidents) {
   let regressedCount = regressed_tests ? regressed_tests.length : 0
   let triagedCount = triaged_incidents ? triaged_incidents.length : 0
   return regressedCount + triagedCount
 }
-
-export function safeExternalHref(url) {
```
smg247, line 750: This was a half-baked idea that I forgot to remove until now

sippy-ng/src/component_readiness/TriagedRegressions.js
```diff
@@ -17,7 +17,7 @@ export default function TriagedRegressions(props) {
     })
 
     if (
-      selectedTriagedEntry.regressions !== 'undefined' &&
```
smg247, line 20: this was actually throwing an error in the console that I hadn't noticed, this fixes it

PR #2662, opened 2025-06-02
TRT-2109: hide triages where regressions have rolled off

smg247:
We can achieve this by hiding any triage records from the listing that have **no** associated regressions among the `allRegressedTests` list.

This also seems to effectively filter them by the release included in the active view.


dgoodwin:
/hold

going to test local

PR #2675, opened 2025-06-05
TRT-1787: use query params for selections within the regressed test modal 

smg247:
Params are cleared when switching tabs within the modal (as they may no longer be relevant), and also when closing the modal.

This also uses the query params for other usages of the panels, but that is nice functionality. The only downside there is the way they are named, but I think keeping `modal` in the name is worth it for clarity within the source code itself.


dgoodwin:
Page and row!? Impressive.

/lgtm

PR #2683, opened 2025-06-06
TRT-2147: Improvements to triage pages

stbenjam:
- Triage details page title to show description
- Table hides created and updated columns by default; only show jira updated
  column
- Show timestamps as relative 
- Show resolution as a glanceable status icon
- Format all timestamps to second precision for consistency

image

image


smg247:
/lgtm
These are nice updates, thanks!

PR #2684, opened 2025-06-06
TRT-1787: test_details opens in new tab by default to fix router weirdness

smg247:
The first commit here is an un-revert. The second is the fix.


neisw:
/lgtm
/hold
up to you late on a friday...

PR #2685, opened 2025-06-06
TRT-2146: better ui for adding a regression to triage record

smg247:
Uses an `Autocomplete` component to select from the list of existing triage records to add to. This allows the user to type to filter the records.

Screenshot 2025-06-06 at 2 41 36 PM

I plan to follow this up with a way to add multiple regressions to an existing triage from the regressed tests modal, but that will require some careful consideration, and this part is distinct from that change.


dgoodwin:
/lgtm

Lovely

PR #2694, opened 2025-06-09
Change environment to include both variant names and values

xueqzhan:


xueqzhan:
Screenshot 2025-06-09 at 2 09 07 PM
Screenshot 2025-06-09 at 2 09 29 PM
Screenshot 2025-06-09 at 2 09 53 PM

dgoodwin:
Dupes continue but vsphere is no longer duped with none, so we get results on that regression.

/lgtm

stbenjam:
/retitle TRT-1736: Change environment to include both variant names and values

PR #2700, opened 2025-06-10
TRT-2007: Annotate job runs with labels

xueqzhan:
/Hold


cmd/sippy/annotatejobruns.go
```diff
@@ -0,0 +1,203 @@
...
+}
+
+func (f *AnnotateJobRunsFlags) BindFlags(fs *pflag.FlagSet) {
+	f.BigQueryFlags.BindFlags(fs)
+	f.PostgresFlags.BindFlags(fs)
+	f.GoogleCloudFlags.BindFlags(fs)
+	f.CacheFlags.BindFlags(fs)
+	f.ComponentReadinessFlags.BindFlags(fs)
+	f.ConfigFlags.BindFlags(fs)
+	fs.StringArrayVar(&f.VariantStr, "variant", f.VariantStr, "A variant to include to search for job runs. The format of the variant string is [variant]:[value] (e.g. Platform:metal).")
```
neisw, line 67: In the long term, nice to have category... Taking a URL and converting it to a variant set would be awesome.  But not for today.

PR #2704, opened 2025-06-10
TRT-1474: log the user making a mutating triage request in preparation for auditing

smg247:
I want to verify that we have access to this info without making changes to our oauth proxy pod


smg247:
It works:
```
time="2025-06-11T10:28:51.434Z" level=info msg="triage POST being made by user: sgoeddel, email: sgoeddel@cluster.local"
```

PR #2712, opened 2025-06-11
TRT-2158: remove (now) defunct triage incident logic from the frontend

smg247:
It has been removed from the server in https://github.com/openshift/sippy/pull/2708


sippy-ng/src/component_readiness/RegressedTestsModal.js
```diff
@@ -96,10 +94,11 @@ export default function RegressedTestsModal({
           >
             <Tab label="Unresolved" {...tabProps(0)} />
             <Tab label="Untriaged" {...tabProps(1)} />
-            {triageEntriesExist && <Tab label="Triaged" {...tabProps(2)} />}
-            {!triageEntriesExist && (
-              <Tab label="Triaged Incidents" {...tabProps(2)} />
-            )}
+            <Tab
+              label="Triaged"
+              disabled={!triageEntriesExist}
```
smg247, line 99: This fixes a bug where the "All" tab shows nothing if there are no active triage entries present in the view.

PR #2713, opened 2025-06-11
Fix double url parameters

xueqzhan:


sippy-ng/t
```diff
@@ -0,0 +1,397 @@
+WITH current AS (
```
dgoodwin, line 1: Looks like a mistaken file included.

xueqzhan:
/hold

Also needs a little more test.

xueqzhan:
/hold cancel

I was concerned about environment not capturing all variants needed in test details page. But that turns out to be not an issue. Good to go now.

PR #2719, opened 2025-06-13
regressionallowances: break regressions into dir structure

sosiouxme:
This way, each PR can add regressions independently of others.


dgoodwin:
Excellent, would you mind getting the add-intentional-regression.py script as well? It's AI generated so you could hand it to an AI and ask for the changes you want most likely :)

dgoodwin:
/lgtm
/label approved

No need for David/Nick approval here, no new regressions being added.

PR #2732, opened 2025-06-18
Make sippy-chat work for production by proxying websocket

stbenjam:
We have to proxy sippy-chat through the single api server to make the frontend work in production

  - We don't want to expose sippy-chat itself to the internet, we want it protected by the main api server's oauth
  - CORS means we shouldn't be having sippy hit a different URL
  - We can't configure the URL at runtime, its baked in with `make build`

Therefore this adds a websocket proxy to sippy core to handle proxying `/api/chat/stream` to sippy-chat.


dgoodwin:
/test e2e

Wonder if we're hitting the aws issues.

PR #2734, opened 2025-06-19
TRT-1474: Triage Auditing

smg247:
Using some helpful hooks from Gorm, we can add an `audit_log` record whenever a triage record is created or modified. 

The OpenShift oauth that runs in front of `auth-sippy` will tell us the user making the request in the `X-Forwarded-User` header, but we must set that directly during e2e testing. Setting the `DEV_MODE=1` env var will allow `developer` to be added to the `user` column for local development, but this is not necessary unless you actually want to test this out.

There is a little bit of complication due to the associated regression updates. I have handled this by doing the entire update in a single transaction, and only using the hooks on the final `Save` call.

The `old_data` and `new_data` columns are both `jsonb` for better indexing, and have been `GIN` indexed for quicker searching. They do not store all of the data, just the relevant parts that are actually modifiable by a user. This is done via a custom serializer.

There is currently no API (or plans for one) to access this data, it will simply be queried directly from the db when needed.

A very small portion of this code was created with some (heavily guided) AI assistance

### DIFF
--- a/pkg/api/componentreadiness/triage.go
+++ b/pkg/api/componentreadiness/triage.go
@@ -1,6 +1,7 @@
 package componentreadiness
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"time"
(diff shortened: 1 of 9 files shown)


dgoodwin:
This is a fantastic implementation, I didn't know most of this was possible in gorm. Its also testing fine locally. Lets roll out and keep an eye on it but it looks great to me.

/lgtm

PR #2745, opened 2025-06-23
component report type refactor

sosiouxme:
Untangle the types that are used in component reports and test details; separate them out into packages that each represent one coherent named concept:
- `reqopts`
- `crview`
- `crtest` (`view` and `test` are too generic and may conflict with other names)
- `bq`
- `testdetails`

With package names now providing context, shorten a lot of verbose names and reconsider confusingly-similar names (like `xxxStats` vs `xxxStatus`).

Finally, move things away that didn't really belong here, and remove cruft.

Each commit addresses a single coherent step of the refactor. There should be no change whatsoever in function.

### DIFF
--- a/pkg/api/componentreadiness/component_report.go
+++ b/pkg/api/componentreadiness/component_report.go
@@ -740,7 +740,7 @@ func updateCellStatus(
 	columnIdentifications []crtest.ColumnID,
 	testID crtest.Identification,
 	testStats testdetails.TestComparison,
-// use the inputs above to update the maps below (golang passes maps by reference)
+	// use the inputs above to update the maps below (golang passes maps by reference)
 	status map[crtest.RowIdentification]map[crtest.ColumnID]cellStatus,
 	allRows map[crtest.RowIdentification]struct{},
 	allColumns map[crtest.ColumnID]struct{},
--- a/pkg/apis/api/componentreport/crtest/types.go
+++ b/pkg/apis/api/componentreport/crtest/types.go
@@ -5,6 +5,9 @@ import (
 	"time"
 )
 
+// These are foundational types for tests, built only from basic golang types.
+// They are test data building blocks shared by test_details and component reports.
+
 type ColumnID string
 
 type Identification struct {
--- a/pkg/apis/api/componentreport/bq/types.go
+++ b/pkg/apis/api/componentreport/bq/types.go
@@ -8,17 +8,8 @@ import (
 	"github.com/openshift/sippy/pkg/apis/api/componentreport/crtest"
 )
 
-// TestStatus is an internal type used to pass data bigquery onwards to the actual
-// report generation. It is not serialized over the API.
-type TestStatus struct {
-	TestName     string   `json:"test_name"`
-	TestSuite    string   `json:"test_suite"`
-	Component    string   `json:"component"`
-	Capabilities []string `json:"capabilities"`
-	Variants     []string `json:"variants"`
-	crtest.Count
-	LastFailure time.Time `json:"last_failure"`
-}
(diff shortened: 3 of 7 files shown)


pkg/api/componentreadiness/component_report.go
```diff
@@ -183,18 +187,18 @@ func NewComponentReportGenerator(client *bqcachedclient.Client, reqOptions crtyp
...
-	BaseRelease    crtype.RequestReleaseOptions
-	SampleRelease  crtype.RequestReleaseOptions
-	VariantOption  crtype.RequestVariantOptions
-	AdvancedOption crtype.RequestAdvancedOptions
-	TestIDOptions  []crtype.RequestTestIdentificationOptions
+	BaseRelease    reqopts.Release
+	SampleRelease  reqopts.Release
+	VariantOption  reqopts.Variants
+	AdvancedOption reqopts.Advanced
+	TestIDOptions  []reqopts.TestIdentification
```
dgoodwin, line 201: Love this refactor/rename.

pkg/api/componentreadiness/component_report.go
```diff
@@ -144,15 +148,15 @@ func (c *ComponentReportGenerator) PostAnalysis(report *crtype.ComponentReport)
...
+				testKey := crtest.Identification{
 					RowIdentification:    col.RegressedTests[rti].RowIdentification,
 					ColumnIdentification: col.RegressedTests[rti].ColumnIdentification,
 				}
-				if err := c.middlewares.PostAnalysis(testKey, &report.Rows[ri].Columns[ci].RegressedTests[rti].ReportTestStats); err != nil {
+				if err := c.middlewares.PostAnalysis(testKey, &report.Rows[ri].Columns[ci].RegressedTests[rti].TestComparison); err != nil {
 					return err
 				}
-				if newStatus := report.Rows[ri].Columns[ci].RegressedTests[rti].ReportTestStats.ReportStatus; newStatus < initialStatus {
+				if newStatus := report.Rows[ri].Columns[ci].RegressedTests[rti].TestComparison.ReportStatus; newStatus < initialStatus {
```
dgoodwin, line 159: Love it.

pkg/api/componentreadiness/component_report.go
```diff
@@ -1118,3 +1122,11 @@ func (c *ComponentReportGenerator) getUniqueJUnitColumnValuesLast60Days(ctx cont
 func init() {
 	componentAndCapabilityGetter = testToComponentAndCapability
 }
+
+type TestVariants struct {
```
dgoodwin, line 1126: Why did this one land here rather than in apis/api/crtest ?

pkg/apis/api/componentreport/crtest/types.go
```diff
@@ -0,0 +1,110 @@
+package crtest
```
dgoodwin, line 1: Could you add a package doc to each of the types.go outline your vision for what should go in each, it will likely help me next time I need to figure out where to drop a type.

pkg/api/componentreadiness/component_report.go
sosiouxme: Its sole usage is in this file, `GenerateVariants` feeds it in as part of the cache key; it doesn't go out over the API at all, so I didn't think it made sense to mix it in. If it were to go anywhere there, I'd probably put it in `reqopts`? But now I'm thinking it would make sense to put it right next to `GenerateVariants` as it does sort of stick out here at the bottom.

pkg/apis/api/componentreport/crtest/types.go
sosiouxme: Yes! Good idea.

PR #2760, opened 2025-07-01
TRT-2160: auto-resolve triages when there are no more open regressions associated with them

smg247:
During regression tracking, when the final regression associated with a triage is closed, the triage will be immediately resolved with a resolution time of the closed time of the final regression. if any of the regressions are re-opened this will result in them going to pants on fire.

There will be a gap here as the new logic only occurs when regressions are closed. This means that there will be triages in the system that are unresolved, where all the regressions are already closed. This will be a one time situation, and can be resolved by manually updating the records.

Assisted by Augment (mostly for the e2e test case)


dgoodwin:
Regressions have that closed timestamp, could you get the one time cleanup to use max(r.closed) for all associated regressions if they are all non null? (as opposed to NOW())

PR #2766, opened 2025-07-03
TRT-2166: add the triages to the analysis regression in post-analysis

smg247:
This caused issues in the component report under certain conditions, and was reverted. Bringing it back with a nil check. It is as basic of logic as you can get, so there doesn't seem to be anything else that could have broken. It is unclear, to me, which circumstances would cause the `testStats.Regression` to be nil when we have gotten this far, but that is something that seems to occur (sometimes) in the component report.


pkg/api/componentreadiness/middleware/regressiontracker/regressiontracker.go
```diff
@@ -118,6 +118,10 @@ func (r *RegressionTracker) PostAnalysis(testKey crtest.Identification, testStat
 		}
 
 		if len(or.Triages) > 0 {
+			// triages need to be included, in case they are not in the cache, in order to show the list on the report
+			if testStats.Regression != nil {
+				testStats.Regression.Triages = or.Triages
+			}
```
dgoodwin, line 124: I think because we need to do that card where we sync the cache primer with the regression tracker, it still happens from time to time that a report has a regression which has not yet been processed by the regression tracker. it would also happen possibly all the time on views which do not use regression tracking.

dgoodwin: That may still boil back to the need to link up reg tracker and cache primer.

PR #2774, opened 2025-07-07
TRT-2130: sippy classic allows arbitrary release names

sosiouxme:
The release names come from sippy /api/releases which ultimately just draws from a BQ table. They could be anything and Sippy Classic would not really care.

Update param.go to just allow anything that looks like a word.


sosiouxme:
included somewhere else

PR #2775, opened 2025-07-07
TRT-2131: Set procedure to serial/parallel for jobs

xueqzhan:


pkg/variantregistry/ocp.go
```diff
@@ -707,6 +707,9 @@ func setProcedure(_ logrus.FieldLogger, variants map[string]string, jobName stri
 		{"-console-operator-", "console-operator"},
 		{"-ipsec", "ipsec"},
 		{"-machine-config-operator", "machine-config-operator"},
+		{"-serial", "serial"},
```
neisw, line 710: We should probably discuss how we handle multiple procedure values.  In this case all of the previous procedure values will also combine with either serial or parallel.  

We could consider parallel the default and omit it but we might still need to support `serial:ipsec`, etc.

PR #2777, opened 2025-07-08
TRT-2179: Add an explicit cache key for release fallback to limit unnecessary cache misses

smg247:
Very few of the exported fields within the `fallbackTestQueryGenerator` and `fallbackTestQueryReleasesGenerator` are actually used in their respective queries. These 2 new cache keys only utilize those fields.

I didn't use AI at all to write the code, but I did use Cursor (Claude-4-sonnet) to check my work:

> Looking at the `fallbackTestQueryGenerator` struct and its cache key, I need to examine which exported fields are used in the query logic but not included in the cache key.
> 
> Let me search for how `query.BuildComponentReportQuery` uses the `ReqOptions` parameter to understand what fields are actually needed:
> Ran tool
> 
> Let me examine the `BuildComponentReportQuery` function more closely to see what fields are actually used in fallback queries:
> Read file: pkg/api/componentreadiness/query/querygenerators.go
> 
> Read file: pkg/api/componentreadiness/query/querygenerators.go
> 
> Based on my analysis of the code, **no** - there are no exported fields within `fallbackTestQueryGenerator` that are used during the query logic but not included in `fallbackTestQueryGeneratorCacheKey`.
> 
> Here's why:
> 
> Looking at the `fallbackTestQueryGenerator` struct, the exported fields are:
> - `BaseRelease` - **included in cache key**
> - `BaseStart` - **included in cache key**
> - `BaseEnd` - **included in cache key**
> - `ReqOptions` - **partially included in cache key**
> 
> The key insight is in the `BuildComponentReportQuery` function. When `isFallback = true` (which is the case for fallback queries), it skips the entire variant filtering section that would otherwise use several fields from `ReqOptions`:
> 
> ```363:363:pkg/api/componentreadiness/query/querygenerators.go
> 	// fallback queries get all variants with no filtering
> 	// so all tests are fetched then cached
> 	if !isFallback {
> ```
> 
> This means the following fields from `ReqOptions` are **NOT** used in fallback queries:
> - `VariantOption.VariantCrossCompare`
> - `VariantOption.CompareVariants`
> - `TestIDOptions`
> - `SampleRelease.PullRequestOptions` (also not used because `isSample = false` for fallback)
> 
> The fields from `ReqOptions` that **ARE** used in fallback queries are:
> 1. `VariantOption.DBGroupBy` - used to construct query structure → **included in cache key as `VariantDBGroupBy`**
> 2. `AdvancedOption.IgnoreDisruption` - used for disruption filtering → **included in cache key as part of `AdvancedOptions`**
> 
> Therefore, the `fallbackTestQueryGeneratorCacheKey` correctly captures all the fields that would actually affect the fallback query results. The cache key is properly designed to avoid unnecessary cache misses while ensuring uniqueness for different query variations.

### DIFF
--- a/pkg/api/componentreadiness/middleware/releasefallback/releasefallback.go
+++ b/pkg/api/componentreadiness/middleware/releasefallback/releasefallback.go
@@ -498,10 +498,11 @@ func newFallbackBaseQueryGenerator(client *bqcachedclient.Client, reqOptions req
 }
 
 type fallbackTestQueryGeneratorCacheKey struct {
-	BaseRelease     string
-	BaseStart       time.Time
-	BaseEnd         time.Time
-	AdvancedOptions reqopts.Advanced
+	BaseRelease string
+	BaseStart   time.Time
+	BaseEnd     time.Time
+	// IgnoreDisruption is the only field within AdvancedOption that is used here
+	IgnoreDisruption bool
 	// VariantDBGroupBy is the only field within VariantOption that is used here
 	VariantDBGroupBy sets.String
 }
@@ -514,7 +515,7 @@ func (f *fallbackTestQueryGenerator) getCacheKey() fallbackTestQueryGeneratorCac
 		BaseRelease:      f.BaseRelease,
 		BaseStart:        f.BaseStart,
 		BaseEnd:          f.BaseEnd,
-		AdvancedOptions:  f.ReqOptions.AdvancedOption,
+		IgnoreDisruption: f.ReqOptions.AdvancedOption.IgnoreDisruption,
 		VariantDBGroupBy: f.ReqOptions.VariantOption.DBGroupBy,
 	}
 }
--- a/pkg/api/componentreadiness/middleware/releasefallback/releasefallback.go
+++ b/pkg/api/componentreadiness/middleware/releasefallback/releasefallback.go
@@ -15,6 +15,7 @@ import (
 	"github.com/openshift/sippy/pkg/apis/api/componentreport/crtest"
 	"github.com/openshift/sippy/pkg/apis/api/componentreport/reqopts"
 	"github.com/openshift/sippy/pkg/apis/api/componentreport/testdetails"
+	"github.com/openshift/sippy/pkg/util/sets"
 	log "github.com/sirupsen/logrus"
 
 	"github.com/openshift/sippy/pkg/api"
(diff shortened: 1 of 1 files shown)


pkg/api/componentreadiness/middleware/releasefallback/releasefallback.go
```diff
@@ -341,7 +342,31 @@ func newFallbackTestQueryReleasesGenerator(
 		ReqOptions:  reqOptions,
 	}
 	return generator
+}
 
+type fallbackTestQueryReleasesGeneratorCacheKey struct {
+	BaseRelease     string
+	BaseStart       time.Time
+	BaseEnd         time.Time
+	AdvancedOptions reqopts.Advanced
```
neisw, line 351: I'm curious about the use of AdvancedOptions in the cache key.  I think it is really just [IncludeMultiReleaseAnalysis](https://github.com/openshift/sippy/blob/c7ea161c348e62a7129bfe0fba43ba642eefc828/pkg/api/componentreadiness/middleware/releasefallback/releasefallback.go#L104) that impacts the cached result.

As an aside (not due to this pr) but querying for fallback and then ignoring the result when [IncludeMultiReleaseAnalysis](https://github.com/openshift/sippy/blob/c7ea161c348e62a7129bfe0fba43ba642eefc828/pkg/api/componentreadiness/middleware/releasefallback/releasefallback.go#L104)  is false leaves us doing extra work.  Think we could add a check and return nil in [https://github.com/openshift/sippy/blob/c7ea161c348e62a7129bfe0fba43ba642eefc828/pkg/api/componentreadiness/middleware/releasefallback/releasefallback.go#L372](https://github.com/openshift/sippy/blob/c7ea161c348e62a7129bfe0fba43ba642eefc828/pkg/api/componentreadiness/middleware/releasefallback/releasefallback.go#L372)?

pkg/api/componentreadiness/middleware/releasefallback/releasefallback.go
```diff
@@ -341,7 +342,31 @@ func newFallbackTestQueryReleasesGenerator(
...
+
+// getCacheKey creates a cache key using the generator properties that we want included for uniqueness in what
+// we cache. This provides a safer option than using the generator previously which carries some public fields
+// which would be serialized and thus cause unnecessary cache misses.
+func (f *fallbackTestQueryReleasesGenerator) getCacheKey() fallbackTestQueryReleasesGeneratorCacheKey {
+	return fallbackTestQueryReleasesGeneratorCacheKey{
+		BaseRelease:          f.BaseRelease,
+		BaseStart:            f.BaseStart,
+		BaseEnd:              f.BaseEnd,
+		AdvancedOptions:      f.ReqOptions.AdvancedOption,
```
neisw: That is interesting.  Agree that we need to honor `AdvancedOption.IgnoreDisruption` in the cache for the fallback queries even though it isn't likely to change (doubt anyone does) but things like pity would cause the TestDetails(?) to take along time to load on a cache miss for fallback and not change the results.

Doubt that happens a lot but if you were confident that was the only thing used it could be nice to reuse the cache in a case like that.

smg247: Yeah, I think it would make sense to only store `IgnoreDisruption` here

PR #2778, opened 2025-07-08
TRT-2130: hcm spike

sosiouxme:
Draft PR for my initial try at adding `hcm-integration` as a release. It would depend on https://github.com/openshift/ci-tools/pull/4624 as well as changes in `continuous-release-jobs` to provide the config file and start loading data for the new release. This sort of works, but we have some further issues to work out first.

### DIFF
--- a/pkg/api/utils.go
+++ b/pkg/api/utils.go
@@ -9,6 +9,7 @@ import (
 	"time"
 
 	"github.com/openshift/sippy/pkg/apis/api/componentreport/crtest"
+	configv1 "github.com/openshift/sippy/pkg/apis/config/v1"
 	"github.com/pkg/errors"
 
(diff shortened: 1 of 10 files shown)


pkg/api/utils.go
```diff
@@ -173,6 +172,13 @@ func GetReleases(ctx context.Context, bqc *bqclient.Client) ([]v1.Release, error
 		[]v1.Release{})
 	if len(errs) > 0 {
 		err = errs[0]
+	} else if config != nil {
+		// Add designated releases from config that aren't represented in the DB
+		for name, release := range config.Releases {
+			if release.AddRelease {
```
stbenjam, line 178: Why does it need to be gated by a boolean? If it's in the config shouldn't we assume it should be shown?

PR #2788, opened 2025-07-11
Revert "TRT-2191: Revert "Allow multiple payload to be used for sample""

xueqzhan:
Reverts openshift/sippy#2787


sosiouxme:
/test e2e
maybe CI won't be so borked today

PR #2797, opened 2025-07-15
NO-JIRA: remove `load-openshift-ci-bigquery` flag

smg247:
The `load-openshift-ci-bigquery` flag has long been the default option for prow loading. Now, the loader errors when that flag is not passed. Remove it.


dgoodwin:
Some other mentions and e2e usage in this repo that we should probably get as well.

PR #2809, opened 2025-07-18
trt-2131: test serial procedure

neisw:
Derivative work of https://github.com/openshift/sippy/pull/2775

I believe we only see one compound instance currently.  Though we will need to be careful with this type of pattern going forward if we support it.

```
periodic-ci-openshift-release-master-nightly-4.20-e2e-aws-ovn-serial-ipsec:
    Aggregation: none
    Architecture: amd64
    CGroupMode: v2
    ContainerRuntime: crun
    FeatureSet: default
    Installer: ipi
    JobTier: informing
    LayeredProduct: none
    Network: ovn
    NetworkAccess: default
    NetworkStack: ipv4
    Owner: eng
    Platform: aws
    Procedure: serial-ipsec
    Release: "4.20"
    ReleaseMajor: "4"
    ReleaseMinor: "20"
    Scheduler: default
    SecurityMode: default
    Suite: serial
    Topology: ha
    Upgrade: none
```


dgoodwin:
Could you throw in a quick unit test for this compound job name? That will help expose impacts if we need to change this in the future.

PR #2815, opened 2025-07-22
Adding OCL Variant

dkhater-redhat:
Defining OCL as its own variant to isolate how OCL specific jobs are performing


pkg/variantregistry/ocp.go
```diff
@@ -707,6 +707,7 @@ func setProcedure(_ logrus.FieldLogger, variants map[string]string, jobName stri
 		{"-console-operator-", "console-operator"},
 		{"-ipsec", "ipsec"},
 		{"-machine-config-operator", "machine-config-operator"},
+		{"-ocl", "on-cluster-layering"},
```
neisw, line 710: You probably want to build out test for what you expect here.  For a job name like `periodic-ci-openshift-machine-config-operator-release-4.20-periodics-e2e-aws-ovn-ocl` would you expect the procedure 'machine-config-operator' or 'on-cluster-layering'  I believe you will get the mco value which I'm guessing isn't the intention?  Changing the order in the list should flip that.

dkhater-redhat:
@neisw PTAL

PR #2821, opened 2025-07-24
Mark interop jobs as owner mpiit in variant registry

dgoodwin:
Also adds an experimental view for mpiit jobs, we'll see if they show up once this merges and has a chance to roll out.


pkg/variantregistry/ocp.go
```diff
@@ -102,6 +102,7 @@ WHERE j.prowjob_start > DATETIME_SUB(CURRENT_DATETIME(), INTERVAL 180 DAY) AND
         OR j.prowjob_job_name LIKE 'periodic-ci-shiftstack-%%'
         OR j.prowjob_job_name LIKE 'periodic-ci-redhat-chaos-prow-scripts-main-cr-%%'
         OR j.prowjob_job_name LIKE 'release-%%'
+        OR j.prowjob_job_name LIKE 'periodic-ci-%%-lp-interop-%%'
```
oharan2, line 105: This pattern is also required on line **86**

PR #2822, opened 2025-07-24
Add interop job in variant registry

oharan2:
- This PR resolves  #2821

### DIFF
--- a/config/views.yaml
+++ b/config/views.yaml
@@ -285,7 +285,7 @@ component_readiness:
       enabled: false
     regression_tracking:
       enabled: false
-  - name: 4.20-mpiit
+  - name: 4.20-LP-Interop
     base_release:
       release: "4.19"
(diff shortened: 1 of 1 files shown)


oharan2:
Hi @deepsm007 can you help me with this check? it fails with:
`Must provide path to GCS credential in GCS_SA_JSON_PATH env var`

It doesn't seem like an issue on my end since I passed `make update-variants` with:
`INFO[0000] variants successfully snapshotted`
Thank you

oharan2:
> Hi @deepsm007 can you help me with this check? it fails with: `Must provide path to GCS credential in GCS_SA_JSON_PATH env var`
> 
> It doesn't seem like an issue on my end since I passed `make update-variants` with: `INFO[0000] variants successfully snapshotted` Thank you

@stbenjam ^^ Thank you.

PR #2852, opened 2025-08-05
TRT-2198: prepare for sippy to accept hcm

sosiouxme:
This is the first phase to enable arbitrary releases that don't look like "x.y". The groundwork is laid in the backend; this can work with the existing schema / cache / UI, and allows updating the schema and Release table contents independently, running the variant refresh, and updating `BackendDisruptionPercentilesDeltaCurrentVsPrevGAV2`, without clearing old cache content.

The bits that actually need/use these updates will be in the next phase.

### DIFF
--- a/config/openshift-customizations.yaml
+++ b/config/openshift-customizations.yaml
@@ -18,3 +18,6 @@ releases:
       - "^pull-ci-openshift-.*-(master|main).*-e2e-.*"
       - "^pull-ci-operator-framework.*-(master|main).*-e2e-.*"
       - "^pull-ci-openshift-online.*-(master|main).*-(verify|unit).*"
+  aro-integration:
+      jobs:
+        periodic-ci-Azure-ARO-HCP-main-periodic-create-aro-hcp-in-integration: true
--- a/pkg/apis/sippy/v1/types.go
+++ b/pkg/apis/sippy/v1/types.go
@@ -3,6 +3,7 @@ package v1
 import (
 	"time"
 
+	"cloud.google.com/go/bigquery"
 	"cloud.google.com/go/civil"
 	bugsv1 "github.com/openshift/sippy/pkg/apis/bugs/v1"
 )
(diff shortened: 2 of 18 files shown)


config/openshift-customizations.yaml
```diff
@@ -18,3 +18,6 @@ releases:
       - "^pull-ci-openshift-.*-(master|main).*-e2e-.*"
       - "^pull-ci-operator-framework.*-(master|main).*-e2e-.*"
       - "^pull-ci-openshift-online.*-(master|main).*-(verify|unit).*"
+  aro-integration:
+      jobs:
+        periodic-ci-Azure-ARO-HCP-main-periodic-create-aro-hcp-in-integration: true
```
sosiouxme, line 23: aside from presubmits, jobs of interest are pulled from the release-controller. aro-hcp doesn't have a release-controller; so these need to be configured manually like presubmits.

pkg/dataloader/featuregateloader/featuregateloader.go
```diff
@@ -87,7 +87,7 @@ func (l *FeatureGateLoader) getTargetReleases() ([]string, error) {
 
 		v, err := version.NewVersion(release.Release)
 		if err != nil {
-			return nil, errors.Wrapf(err, "error parsing release version %s", release)
+			return nil, errors.Wrapf(err, "error parsing release version %s", release.Release)
```
sosiouxme, line 90: this hasn't been touched otherwise in this PR but for some reason it was failing in my local tests so i fixed it.

pkg/db/query/test_queries.go
```diff
@@ -126,7 +126,7 @@ WITH results AS (
            unnest(variants)        AS variant
     FROM prow_test_report_7d_matview
 	WHERE release = @release AND name ~* @testsubstrings
-          AND NOT(@excluded && variants)
+          AND NOT(variants is not null and @excluded && variants)
```
sosiouxme, line 129: when the variants are null, the `&&` operator weirdly seems to come back true. add a check so that null doesn't match exclusions. there are quite a lot of rows with null variants so this might add some results we've not been seeing in the health/overview/install areas. could maybe leave this change out but it helped with devel (not having variants defined for aro-integration) and it seems like a good idea.

pkg/variantregistry/ocp.go
```diff
@@ -838,6 +842,7 @@ func setPlatform(jLog logrus.FieldLogger, variants map[string]string, jobName st
 		{"-rosa", "rosa"}, // Keep above AWS as many ROSA jobs also mention AWS
 		{"-aws", "aws"},
 		{"-alibaba", "alibaba"},
+		{"-azure-aro-hcp", "aro"},
```
sosiouxme, line 845: I couldn't see that any variants other than owner, installer, platform made sense for `aro-integration`. maybe LayeredProduct? well, we can always adjust later.

sosiouxme:
/hold
while I check more consequences

sosiouxme:
/hold cancel
can merge at will

pkg/apis/sippy/v1/types.go
```diff
@@ -69,11 +70,30 @@ type FailureGroup struct {
...
+	ExcludeFetchData bool `json:"exclude_fetch_data"`
+	// ExcludeMetrics skips metrics collection and analysis for the release.
+	ExcludeMetrics bool `json:"exclude_metrics"`
+	// IncludePullRequests enables the Sippy Classic pull request UI for this release
+	IncludePullRequests bool `json:"include_pull_requests"`
+	// ExcludeFeatureGates omits sippy classic link for seeing release feature gates
+	ExcludeFeatureGates bool `json:"exclude_feature_gates"`
+	// ExcludePayloadTags omits sippy classic link for seeing release tags for payloads
+	ExcludePayloadTags bool `json:"exclude_payload_tags"`
 }
```
stbenjam, line 97: I am not a huge fan of booleans like this. The mixed Exclude and Include logic is also hard to follow. If you're going to use booleans, then I think they should all go one direction (i.e. Include or Exclude), but I'd rather not have any.  

Bigquery can do arrays (I think it calls them repeated fields), could we just have a single column that's a list of a release's capabilities?

sosiouxme:
/close
in favor of https://github.com/openshift/sippy/pull/2857

PR #2855, opened 2025-08-07
Integrate okd

kenjpais:


pkg/dataloader/releaseloader/types.go
```diff
@@ -85,3 +86,27 @@ type UpdatedImageCommits struct {
...
+// OCPRelease implements PlatformRelease for OpenShift Container Platform
+type OCPRelease struct{}
+
+// OKDRelease implements PlatformRelease for Origin Kubernetes Distribution
+type OKDRelease struct{}
+
+// PlatformRelease represents a platform that releases are built for (OCP, OKD, etc)
+type PlatformRelease interface {
+	GetName() string
+	GetAlias() string
```
sosiouxme, line 99: IMHO this method shouldn't exist. If the two objects need different behavior, they should each supply an interface method to implement the difference; nothing outside it should be looking at it and deciding how to act.

pkg/dataloader/releaseloader/releasesync.go
```diff
@@ -181,6 +215,11 @@ func releaseDetailsToDB(architecture string, tag ReleaseTag, details ReleaseDeta
 		release.Release = strings.Join(parts[:2], ".")
 	}
 
+	if platformAlias == "origin" {
+		// For origin, we need to add the -okd suffix to the release tag before saving it to the database ie. 4.15 -> 4.15-okd
+		release.Release = fmt.Sprintf("%v%s", release.Release, "-okd")
+	}
```
sosiouxme, line 221: e.g. I'd like this to be a method call, not a branch of logic

pkg/dataloader/releaseloader/releasesync.go
```diff
@@ -31,21 +31,21 @@ type ReleaseLoader struct {
...
 func New(dbc *db.DB, releases, architectures []string) *ReleaseLoader {
-	releaseStreams := make([]string, 0)
-	for _, release := range releases {
-		for _, stream := range []string{"nightly", "ci"} {
-			releaseStreams = append(releaseStreams, fmt.Sprintf("%s.0-0.%s", release, stream))
-		}
+	platformReleases, err := GetPlatformReleases("all")
+	if err != nil {
+		panic(err)
 	}
```
sosiouxme, line 43: if there isn't any way to configure which ones go in the loader, why even have a method call? just hardcode the result.

pkg/dataloader/releaseloader/platforms.go
```diff
@@ -0,0 +1,96 @@
...
+	"strings"
+)
+
+func (ocp *OCPRelease) GetName() string {
+	return "OpenShift Container Platform (OCP)"
+}
+
+func (ocp *OCPRelease) GetAlias() string {
+	return "ocp"
+}
```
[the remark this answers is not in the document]
kenjpais: Sure, makes sense. Made the changes.

pkg/dataloader/releaseloader/types.go
kenjpais: Just want to confirm, you are referring to both GetName() and GetAlias() right? Your previous comment seemed to mention only GetAlias(), so just wanted to double-check.

pkg/dataloader/releaseloader/releasesync.go
kenjpais: Sure, makes sense. Made the changes.

PR #2859, opened 2025-08-08
Variant updates

neisw:


neisw:
/hold
Let's see what thoughts are about https://github.com/openshift/sippy/pull/2860

PR #2872, opened 2025-08-12
TRT-2198: followup to releases change

sosiouxme:
It's likely easiest to review a commit at a time.

first commit is in response to [this comment](https://github.com/openshift/sippy/pull/2857/files?diff=unified&w=1#r2263027287). (I chose against the [other nit](https://github.com/openshift/sippy/pull/2857/files?diff=unified&w=1#r2262985126).)

the second commit uses the Releases table by default for loaders that need releases specified. this way we can remove all the `fetchdata` release params (although if supplied they are used) and not need to update the cronjob every time we add a release.

following commits expand on that a little bit to make the Releases table authoritative everywhere in the backend.

### DIFF
--- a/pkg/dataloader/featuregateloader/featuregateloader.go
+++ b/pkg/dataloader/featuregateloader/featuregateloader.go
@@ -1,7 +1,6 @@
 package featuregateloader
 
 import (
-	"context"
 	"fmt"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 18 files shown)


pkg/dataloader/featuregateloader/featuregateloader.go
```diff
@@ -75,21 +66,16 @@ func (l *FeatureGateLoader) Errors() []error {
...
+func (l *FeatureGateLoader) getTargetReleases() []string {
 	var targetReleases []string
-	releases, err := api.GetReleasesFromBigQuery(context.Background(), l.bqc)
-	if err != nil {
-		return nil, errors.Wrapf(err, "error querying releases from bq")
-	}
-
-	for _, release := range releases {
+	for _, release := range l.releaseConfigs {
 		if release.Capabilities[v1.FeatureGatesCap] {
```
sosiouxme, line 72: should this also filter out any with a future DevelStartDate? alternatively, we could just not add the `featureGates` capability until the branch occurs. either way we need a DB update

xueqzhan (commented):
looking good with small comment.

PR #2879, opened 2025-08-14
TRT-2161: display the computed component name in the FileBug dialog

smg247:
Also adds a note to the modal that the fields are editable post-submission
Screenshot 2025-08-14 at 11 20 53 AM


dgoodwin:
/lgtm 

Thank you!

PR #2890, opened 2025-08-20
move gcp-custom-dns to candidate jobTier

jianlinliu:


neisw:
/hold

Let's review, my understanding is the job is coming back with skips for the failing tests so we don't need to reduce the job tier.

PR #2892, opened 2025-08-20
TRT-2231: Add special handling for install related feature gates

xueqzhan:


sosiouxme (commented):
:sigh: not sure i can bring myself to say that having to hardcode a test name "looks good" to me.

sosiouxme:
guess there's no reason to delay this
/lgtm

PR #2899, opened 2025-08-21
TRT-2233: Add the ability to list potential matching regressions for a triage

smg247:
Provides a deep link to the potential matches when creating a new triage (assuming the user selects to see it). The potential matches are currently determined via two factors.

1. Similar test names, which is defined as having an edit distance of <= 5.
2. Same last failure time.

A confidence level is computed based on:
- which factors are matching
- how many currently associated regressions they match
- the actual edit distance (more confident if it is the same name vs. off by 5 chars)

Checkboxes are displayed alongside the list, and there is a button that will add the regressions selected to the triage. There is also filtering based on either of the matching criteria (if you want to hide same last failure time, for example).

Screenshot 2025-08-20 at 11 56 35 AM

Also, the regressed tests now show status and test_details link on triage details page.

There are some pretty substantial e2e changes buried in here. We needed to use the component_report to be able to display the status and link to test_details on each of the regressions in the potential matches list (and the triage details page in general). In order to verify that this works properly we must start using Redis in the e2e tests, and manipulate the cache to add the regressed_tests that we create within the e2e test to it. I created a util to do this, as I think the concept will be useful for future e2e testing as well.

Assisted by Cursor

### DIFF
--- a/scripts/e2e.sh
+++ b/scripts/e2e.sh
@@ -8,6 +8,8 @@
 DOCKER="podman"
 PSQL_CONTAINER="sippy-e2e-test-postgresql"
 PSQL_PORT="23433"
+REDIS_CONTAINER="sippy-e2e-test-redis"
+REDIS_PORT="23479"
 
 if [[ -z "$GCS_SA_JSON_PATH" ]]; then
     echo "Must provide path to GCS credential in GCS_SA_JSON_PATH env var" 1>&2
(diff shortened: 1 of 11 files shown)


scripts/e2e.sh
```diff
@@ -59,14 +72,15 @@ export SIPPY_ENDPOINT="127.0.0.1"
...
+  --redis-url="$REDIS_URL" > e2e.log 2>&1
 )&
 # store the child process for cleanup
 CHILD_PID=$!
 
-# Give it time to start up
+# Give it time to start up, and fill the redis cache
 echo "Waiting for sippy API to start on port $SIPPY_API_PORT, see e2e.log for output..."
-TIMEOUT=120
+TIMEOUT=600
```
smg247, line 83: This timeout increase doesn't actually make the test any longer. The time is saved when actually running the tests. It is a longer startup time, but overall is very similar in length.

scripts/e2e.sh
```diff
@@ -22,31 +24,42 @@ clean_up () {
...
+$DOCKER rm -i $REDIS_CONTAINER
 
 # start postgresql in a container:
 echo "Starting new sippy postgresql container: $PSQL_CONTAINER"
 $DOCKER run --name $PSQL_CONTAINER -e POSTGRES_PASSWORD=password -p $PSQL_PORT:5432 -d quay.io/enterprisedb/postgresql
 
-echo "Wait 5s for postgresql to start..."
+# start redis in a container:
+echo "Starting new sippy redis container: $REDIS_CONTAINER"
+$DOCKER run --name $REDIS_CONTAINER -p $REDIS_PORT:6379 -d quay.io/openshiftci/redis:latest
```
smg247, line 48: I have mirrored this image, I or DPTP can update it whenever necessary.

PR #2903, opened 2025-08-25
Add -lp-interop-virt LP filtering

oharan2:
This naming convention helps filter by the one specifically matching job that should be collected under the correlated component (`CNV-lp-interop`).
All other related virt data (collected from multiple jobs) can be remain under the `virt` filter.

### DIFF
--- a/pkg/variantregistry/ocp.go
+++ b/pkg/variantregistry/ocp.go
@@ -1079,6 +1079,7 @@ func setLayeredProduct(_ logrus.FieldLogger, variants map[string]string, jobName
 		{"-cnv", "virt"},
 		{"-kubevirt", "virt"},
 		{"-oadp-", "oadp"},
+		{"-lp-interop-cr-cnv", "lp-interop-virt"},
 	}
 
 	variants[VariantLayeredProduct] = VariantNoValue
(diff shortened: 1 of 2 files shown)


pkg/variantregistry/ocp.go
```diff
@@ -1075,6 +1075,7 @@ func setLayeredProduct(_ logrus.FieldLogger, variants map[string]string, jobName
 		substring string
 		product   string
 	}{
+		{"-lp-interop-cr-cnv", "lp-interop-virt"},
```
neisw, line 1078: This looks good but it will have the effect of dropping these out from your view.  Have a look at your [view definition](https://github.com/openshift/sippy/blob/51de5891c2659a9c40d48962fbf40d3ed89e25d4/config/views.yaml#L313) and update it to include `lp-interop-virt`, its up to you if you want to keep `none` and `virt` or not.

oharan2: Will it filter it by default (or only sets the option?)? 
If so, that's great, exactly what I needed.

neisw: As you have it only jobs with the `lp-interop-virt` layered product variant value will show in your view.  Note that after this merges there will be a delay between when your view is updated (right after it is deployed) and when your jobs are updated with the new variant value.  So give it overnight to see the variant value reflected

PR #2912, opened 2025-08-29
Move aks jobs to their own platform to differentiate from standard azure

dgoodwin:


neisw:
/hold

hypershift is platform:aws, topology:external which is how aks is setup, platform:azure, topology:external

pkg/variantregistry/ocp.go
```diff
@@ -870,7 +870,7 @@ func setPlatform(jLog logrus.FieldLogger, variants map[string]string, jobName st
 		{"-alibaba", "alibaba"},
 		{"-azure-aro-hcp", "aro"},
```
neisw, line 871: Look here for ARO, are you sure all AKS is ARO?

PR #2926, opened 2025-09-05
TRT-2319: Bootstrap an MCP Server for Sippy

stbenjam:
Bootstraps an MCP server for Sippy at `/mcp/v1`.  Eventually, we'll try to migrate all of sippy chat's tools to use MCP but this just adds the release one for now. I'm using the same MCP library as kube and [OCP](https://github.com/openshift/openshift-mcp-server)
 

Start sippy, ensuring you're pointing at a postgresdb

```
$ make sippy
$ ./sippy serve --mode ocp --google-service-account-credential-file <google service account json> --listen-metrics=""  --log-level=debug --database-dsn <postgres dsn>
```

Experiment with mcp using mcp-inspector:

```
$ npm install -g @modelcontextprotocol/inspector
$ mcp-inspector
```

Example:
image


sosiouxme (approved):
seems like a solid base.
/lgtm
just left small nits which need not hold this up if you don't want to.
/hold
in case ya do :)

PR #2927, opened 2025-09-05
TRT-2189: display triage audit logs on the Triage details page when requested

smg247:
Adds a new button to the triage details page to display audit logs in a user-friendly way in a new modal dialog. This information is only requested when a user clicks on the button. It displays a colorful diff of each change to the respective triage along with the time and user associated with the change.

Screenshot 2025-09-05 at 2 34 23 PM

The implementation here couples the logic to the current state of the `triage` data model. If we end up making breaking changes to that data model in the future, we will have to address them in the diffing logic. I think that has a pretty low likelihood of being an issue though.

Assisted by Cursor

### DIFF
--- a/pkg/sippyserver/server.go
+++ b/pkg/sippyserver/server.go
@@ -1468,6 +1468,25 @@ func (s *Server) jsonTriagePotentialMatchingRegressions(w http.ResponseWriter, r
 	api.RespondWithJSON(http.StatusOK, w, potentialMatches)
 }
 
+func (s *Server) jsonGetTriageAuditDetails(w http.ResponseWriter, req *http.Request) {
+	vars := mux.Vars(req)
+	idStr := vars["id"]
+
+	triageID, err := strconv.Atoi(idStr)
+	if err != nil {
+		failureResponse(w, http.StatusBadRequest, "invalid ID format: "+idStr)
+		return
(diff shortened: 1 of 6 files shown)


pkg/sippyserver/server.go
```diff
@@ -2066,6 +2085,13 @@ func (s *Server) Serve() {
 			Capabilities: []string{LocalDBCapability, ComponentReadinessCapability},
 			HandlerFunc:  s.jsonTriagePotentialMatchingRegressions,
 		},
+		{
+			EndpointPath: "/api/component_readiness/triages/{id}/audit",
```
stbenjam, line 2089: Should this be included in the `links` section of triage? I think if we're to settle on using HATEOAS, including `audit` in links would make sense

```
"links": {
  "self": "/api/component_readiness/triages/193",
  "audit": "/api/component_readiness/triages/193/audit"
}
```

WDYT?

stbenjam (commented):
Optional comment

/lgtm
/hold

pkg/sippyserver/server.go
smg247: Ah yes, that is a very good point. There are some others that also should be included; so I will do them all together in a follow-up.

PR #2928, opened 2025-09-05
Clean up styling on triaged regression test list

stbenjam:
Is this better? The rows are taller, but it fits more of the test name and variants, and matches the variant styling from jobs table. Also fixes a bug in pre-pending the array index on the variant (I don't think `formColumnName` fn is not needed here)

After: 

image

Before: 

image


smg247:
/lgtm
This looks much more legible!

PR #2950, opened 2025-09-12
TRT-2258: Reinstate "OKD-268 OKD backend integration"

sosiouxme:
Reverts openshift/sippy#2949 so that OKD releases will work and adds fixes/refactor to address what broke the last attempt.

### DIFF
--- a/pkg/dataloader/releaseloader/releasesync.go
+++ b/pkg/dataloader/releaseloader/releasesync.go
@@ -14,6 +14,7 @@ import (
 	v1 "github.com/openshift/sippy/pkg/apis/sippy/v1"
 	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
+	"golang.org/x/exp/maps"
 	"gorm.io/gorm/clause"
 
 	"github.com/openshift/sippy/pkg/apis/api"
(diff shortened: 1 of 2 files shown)


pkg/dataloader/releaseloader/releasesync.go
```diff
@@ -104,98 +109,114 @@ func (r *ReleaseLoader) Load() {
...
-		releaseName += "-" + architecture
+func (r *ReleaseLoader) resolveReleasePullRequests(pullRequests []models.ReleasePullRequest) []models.ReleasePullRequest {
+	if len(pullRequests) == 0 {
+		return pullRequests
+	}
+
+	type prKey struct{ url, name string }
+	prIndexMap := make(map[prKey]int, len(pullRequests))
+	orConditions := make([]string, 0, len(pullRequests))
+	args := make([]any, 0, len(pullRequests)*2)
```
smg247, line 139: Did you consider defining a new struct that contains all the info we need rather than using multiple maps and keeping them in sync with an index map?

sosiouxme: TBH I didn't write this part and only gave it a cursory look... now that you mention it, this has a code smell for me, there's rarely a good reason to look at indexes. There's only one map, but it just ought to map to the PRs.

PR #2959, opened 2025-09-17
UI: choose default release that is successor of latest GA

sosiouxme:
right now `4.19-okd` is the default PR, should be an OCP release.


sosiouxme:
/test unit
i know why this is failing and it's gonna be a pain but it's unrelated to this

PR #2963, opened 2025-09-19
Override Triage resolution icon with pants on fire

dgoodwin:


sippy-ng/src/component_readiness/Triage.js
```diff
@@ -82,6 +83,16 @@ export default function Triage({ id }) {
     }
   }
 
+  // Helper function to check if triage has any regressions with status -1000
+  const hasStatus1000Regression = () => {
```
smg247, line 87:
```suggestion
  const hasStatusMinus1000Regression = () => {
```

sippy-ng/src/component_readiness/TriagedRegressions.js
```diff
@@ -72,6 +74,29 @@ export default function TriagedRegressions({
     }
   }
 
+  // Helper function to check if triage has any regressions with status -1000
+  const hasStatus1000Regression = (triage) => {
```
smg247, line 78: It would be nice if we could have one common function that both places utilized. Can we combine these and put it in CompReadyUtils?

smg247 (commented):
/lgtm

PR #2967, opened 2025-09-22
TRT-2235: include LayeredProduct in DBGroupBy (4.21 views)

sosiouxme:
I can't see any reason this wouldn't work correctly now. Everything seems to be getting dbGroupBy from a view (the default would only be used in the absence of views) and passing it around as a parameter. I ran the cache primer with this locally and test_details linked from the CR modal came from cache.

### DIFF
--- a/config/views.yaml
+++ b/config/views.yaml
@@ -24,6 +24,7 @@ component_readiness:
         Suite: { }
         Topology: { }
         Upgrade: { }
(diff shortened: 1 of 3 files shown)


sosiouxme:
/test e2e
/hold
probably best to merge this and do a cache primer run once sippy builds

PR #2972, opened 2025-09-24
TRT-2284: informational comment on OCPBUGS cards used for triage

smg247:
Adds a simple comment to any OCPBUGS cards that are utilized for a triage record. This happens on initial creation, and when the triage is updated to point to a new Jira URL. We cannot comment on cards outside the OCPBUGS project, so there is simple validation there.

/hold merge after https://github.com/openshift/sippy/pull/2969


sosiouxme (approved):
/lgtm

PR #2978, opened 2025-09-26
TRT-2263: Indicate when feature gate test results are coming from included/excluded jobtiers

dgoodwin:
This does not enforce that feature gate test results are coming from component readiness included job tiers, but it does help display to the user whether they are or not in the default view presented when clicking through to the fg tests table.

JobTier is now first in the list, previously it was typically hidden as it wasn't in the first 8. It will be green if an included jobtier, red otherwise.
Hide Network (always ovn now)
Hide Owner (still visible in tooltip)

Tooltip will also get a warning if the result is from an excluded jobtier to help explain.

In future, a dropdown view for included/excluded CR would be nice both on this table, and the jobs table.

Assisted-by: Cursor AI

### DIFF
--- a/sippy-ng/src/tests/TestTable.js
+++ b/sippy-ng/src/tests/TestTable.js
@@ -43,6 +43,66 @@ const bookmarks = [
   },
 ]
 
+/**
+ * Chooses which variants to display in the table based on priority and limits.
+ * @param {Array<string>} variants - Array of variant strings in "key:value" format
+ * @returns {Array<string>} - Up to 8 selected variants
+ */
+function chooseVariantsToDisplay(variants) {
+  if (!variants || variants.length === 0) {
+    return []
+  }
+
+  // Filter out default variants first
+  const filteredVariants = variants.filter((item) => !item.endsWith(':default'))
+
+  // Priority order for variant keys
+  const priorityKeys = [
+    'JobTier',
+    'Platform',
+    'Architecture',
+    'NetworkStack',
+    'Topology',
+    'FeatureSet',
+    'Upgrade',
+  ]
+
+  // Parse variants into key-value pairs
+  const variantMap = new Map()
+  const remainingVariants = []
+
+  filteredVariants.forEach((variant) => {
+    const colonIndex = variant.indexOf(':')
+    if (colonIndex > 0) {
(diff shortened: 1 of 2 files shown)


sippy-ng/src/tests/TestTable.js
```diff
@@ -429,21 +508,37 @@ function TestTable(props) {
...
+          params.value.some(
+            (variant) =>
+              variant.startsWith('JobTier:') &&
+              !isComponentReadinessIncludedJobTier(variant)
+          )
+
+        const tooltipTitle = params.value
+          ? params.value.join('\n') +
+            (hasExcludedJobTier
+              ? '\n\nWARNING: Test results from jobs with this JobTier are not included in the main release blocking views for component readiness, and will not be monitored for regressions.'
```
stbenjam, line 526: The newlines don't work in this version without the former pre styling, so its all scrunched together in chrome:

image

Cursor suggested this solution to me:

```javascript
        const tooltipContent = params.value ? (
          <div>
            {params.value.map((variant, index) => (
              <div key={index}>{variant}</div>
            ))}
            {hasExcludedJobTier && (
              <>
                <br />
                <div>
                  <b>WARNING:</b> Test results from jobs with this JobTier are
                  not included in the main release blocking views for component
                  readiness, and will not be monitored for regressions.
                </div>
              </>
            )}
          </div>
        ) : (
          ''
        )
```

which results in:

image

sippy-ng/src/tests/TestTable.js
```diff
@@ -387,6 +456,16 @@ function TestTable(props) {
     </div>
   )
 
+  const getVariantStyle = (variant) => {
+    // Special treatment for JobTier to help users better understand if their test is feeding component readiness or not
+    if (isComponentReadinessIncludedJobTier(variant)) {
+      return { color: 'green' }
+    } else if (variant.startsWith('JobTier:')) {
+      return { color: 'darkred' }
+    }
```
stbenjam, line 465: Nit, you can use theme colors instead: `theme.palette.error.dark` and `theme.palette.success.dark` instead.  But this version is also fine, it's just more consistent the other way

stbenjam:
Optional nits

/lgtm
/hold

PR #2997, opened 2025-10-07
TRT-2150: fallback queries should filter by variants

sosiouxme:
include the IncludeVariants in the query and the cache key for fallbacks so those queries match what test details is looking at.


sosiouxme:
/hold
need to understand why we're getting cache misses in TD. this might muddy the water.

PR #2998, opened 2025-10-07
NO-JIRA: do not surface error when failing to comment on jira issue used for triage

smg247:
I noticed this today when triaging something to an issue that was restricted. We cannot comment on issues in this state. I don't think we should increase the permissions of the bot as that may open us up to leaks. Instead, we will just not throw errors as they are not actionable by the user and will just lead to confusion. Instead we can add warnings to the logs, and discover the problem if we notice failures to post the comment in other cases.


sosiouxme (approved):
/lgtm

PR #3290, opened 2026-02-23
perf: push filters into test report subqueries (~830x improvement)

stbenjam:
When collapse=false, TestsByNURPAndStandardDeviation builds a query that self-joins prow_test_report_7d_matview 3 times:

  1. Outer query - gets the raw rows
  2. pass_rates subquery - computes per-variant percentages
  3. stats subquery - computes AVG/STDDEV across variants

The name/variant filters were only applied to the outermost query. Subqueries 2 and 3 scanned all rows for the release to compute aggregates for every test, even when only a single test was requested.

For release 4.22 with a name filter, this meant:

  |                    | Before (outer only)     | After (pushed down) |
  |--------------------|-------------------------|---------------------|
  | Stats subquery     | Seq Scan, 1.28M rows    | Index Scan, 142     |
  | Estimated cost     | 802,603 - 1,137,530     | 7.53 - 1,371        |
  | Speedup            | -                       | ~830x               |

TestsByNURPAndStandardDeviation now accepts optional filter functions (variadic, backward-compatible) that are applied to both the stats and pass_rates subqueries. The filter is still also applied to the outer query, so results are identical.

## Summary by CodeRabbit

* **Chores**
  * Improved backend test-query filtering to preserve cross-variant statistics by default and allow more precise variant-specific filters.
  * Optimized subquery filtering and application paths for more accurate and efficient aggregation of test stats and pass rates.

* **Bug Fixes**
  * Fixed incorrect filter propagation so optional filters are applied to the intended subqueries, yielding more consistent results.

### DIFF
--- a/pkg/api/tests.go
+++ b/pkg/api/tests.go
@@ -14,6 +14,7 @@ import (
 	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
 	"google.golang.org/api/iterator"
+	"gorm.io/gorm"
 
 	apitype "github.com/openshift/sippy/pkg/apis/api"
 	bq "github.com/openshift/sippy/pkg/bigquery"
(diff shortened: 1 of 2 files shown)


stbenjam:
Closing — this PR has gone stale and is no longer being actively pursued. (Automated cleanup by agent)

PR #3409, opened 2026-04-07
TRT-2564: Cache Risk Analysis responses

sosiouxme:
Risk Analysis queries are fairly heavy and can impact  Sippy DB performance. During regression periods we are likely to see an increase in repetitive calls for the same job / test failures. So this adds a caching mechanism.

Both RA and Sippy classic test pages use the same underlying query mechanism, so the same cache is introduced for both (replacing the existing request-level cache for the test API). This caches the test queries for an hour by default, but cache entries are invalidated if the underlying matview is refreshed, so new data is surfaced faster.

1. Sippy's matview refresh code is updated to write a timestamp to redis whenever a matview refresh completes.
2. Query results are written to the cache with a timestamp.
3. On a cache hit, if the underlying matview timestamp is newer than the result timestamp, it is treated as a miss and regenerated. (In the case where the matview timestamp is missing, cached data is used until it expires normally.)

Individual commits separate the concepts that built this change, so reviewing them serially may be easiest.

https://github.com/openshift/origin/pull/30965 proposes one simple way to benchmark risk analysis requests.

## Summary by CodeRabbit

* **New Features**
  * Global caching support added across APIs, CLI commands, background processing and matview refresh tracking.

* **Changed**
  * Refresh/seed flows, server endpoints and background workers now accept and use optional cache clients; matview refreshes record cache timestamps; tests and risk/result generation were refactored to be cache-aware and unified.

* **Removals**
  * /api/canary endpoint removed.

* **Tests**
  * Extensive tests added for cache and matview invalidation/refresh behaviors.


sosiouxme:
@coderabbitai help

sosiouxme:
@coderabbitai review

pkg/api/cache.go
```diff
@@ -158,3 +158,79 @@ func isStructWithNoPublicFields(v interface{}) bool {
...
+			}
+		}
+
+		if lastRefresh.Before(cacheVal.Timestamp) {
+			// not invalidated by a newer refresh, so use it (if we don't know the last refresh, still use it)
+			return cacheVal.Val, nil
+		}
+		logrus.Debugf("matview %q refreshed at %v, will not use earlier cache entry from %v", matview, lastRefresh, cacheVal.Timestamp)
+	} else if strings.Contains(err.Error(), "connection refused") {
+		logrus.WithError(err).Fatalf("redis URL specified but got connection refused; exiting due to cost issues in this configuration")
```
petr-muller, line 222: `Fatal` in a library is brutal 😁 

could be a networking blip but I guess we can robustify if we run into this too often

cmd/sippy/load.go
petr-muller:
```suggestion
						return errors.Wrap(cacheErr,  "couldn't get cache client")
```

an LLM pointed me here, it's not that I read all code

pkg/api/tests.go
```diff
@@ -671,7 +687,7 @@ func BuildTestsResultsFromBigQuery(ctx context.Context, bqc *bq.Client, release,
 	log.WithFields(log.Fields{
 		"elapsed": elapsed,
 		"reports": len(testReports),
-	}).Debug("BuildTestsResults completed")
+	}).Debug("buildTestsResultsFromPostgres completed")
```
petr-muller:
```suggestion
	}).Debug("buildTestsResultsFromBigQuery completed")
```

pkg/api/tests.go
```diff
@@ -444,37 +429,64 @@ func GetJobRunTestsCountByLookback(dbc *db.DB, lookbackDays int) (int64, int64,
...
+func (spec *TestResultsSpec) buildTestsResultsFromPostgres(dbc *db.DB, cacheClient cache.Cache) (testResults, error) { //lint:ignore
+	matview := testReport7dMatView
+	if spec.Period == "twoDay" {
+		matview = testReport2dMatView
+	}
+
+	generator := func(ctx context.Context) (testResults, []error) {
+		return spec.buildTestsResults(dbc, matview)
+	}
+	result, errs := GetDataFromCacheOrMatview(context.TODO(), cacheClient,
```
petr-muller: there should be a request somewhere upstream, which means there should be a `ctx` somewhere too, would it be hard to plumb it here to avoid the `context.TODO()`?

pkg/api/tests.go
```diff
@@ -516,12 +530,14 @@ func BuildTestsResults(dbc *db.DB, release, period string, collapse, includeOver
 	log.WithFields(log.Fields{
 		"elapsed": elapsed,
 		"reports": len(testReports),
-	}).Debug("BuildTestsResults completed")
+	}).Debug("buildTestsResultsFromPostgres completed")
```
petr-muller:
```suggestion
	}).Debug("buildTestsResults completed")
```

petr-muller (approved):
/hold
/lgtm
/approve
/meow

Some comments inline, nothing serious. Feel free to unhold or address, your call.

pkg/api/cache.go
```diff
@@ -158,3 +158,79 @@ func isStructWithNoPublicFields(v interface{}) bool {
...
+				logrus.WithError(err).Warnf("failed to parse matview refresh timestamp %q for %q; cache will not be invalidated", lastRefreshBytes, matview)
+			} else {
+				lastRefresh = parsed
+			}
+		}
+
+		if lastRefresh.Before(cacheVal.Timestamp) {
+			// not invalidated by a newer refresh, so use it (if we don't know the last refresh, still use it)
+			return cacheVal.Val, nil
+		}
```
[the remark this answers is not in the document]
sosiouxme: it will make no difference in practice and i believe the PR code is more legible.

pkg/api/cache.go
sosiouxme: i do not know the history of why it was done this way and it did not seem likely to be a common problem that a blip would kill the server (but maybe it's been happening all the time and we just didn't notice because they're deployed redundantly)

pkg/api/tests.go
sosiouxme: agreed, i guess i was being lazy

pkg/api/tests.go
sosiouxme: yeah... actually it needs a name change anyway.

pkg/api/tests.go
sosiouxme: oops

scripts/updatesuites.go
```diff
@@ -150,6 +150,7 @@ func testNameWithoutSuite(dbc *gorm.DB) error {
 	log.Infof("update complete, total rows updated %d", rowsUpdated)
 
 	// Refresh materialized views
+	// NOTE: does not update timestamps to invalidate cached matview data; not clear if the use case for this script requires that.
 	sippyserver.RefreshData(&db.DB{
 		DB: dbc,
 	}, nil, false)
```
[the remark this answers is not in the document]
sosiouxme: that is fine with me. i doubt anyone uses this script against production. if we do and anyone cares, we'll have to add the ability to specify the cache.

sosiouxme (commented):
edits incoming

petr-muller (approved):

petr-muller:
Unhold at will

petr-muller:
@coderabbit approve

petr-muller (approved):
/lgtm

Looks good after the fixes. The `logrus.Fatalf` on Redis connection refused (pkg/api/cache.go) is still a concern for production resilience, but not a merge blocker — it can be revisited if it causes issues in practice.

petr-muller (approved):
/lgtm

PR #3422, opened 2026-04-09
Add 4.22-ha-vs-two-node-arbiter view.

jogeo:
Adds a 4.22 compare view between HA and Two Node with Arbiter clusters.

## Summary by CodeRabbit

* **Chores**
  * Added a new component readiness configuration for release 4.22 to compare HA vs two-node-arbiter topologies, including variant filters (architecture, feature set, installer, job tier, layered product, owner, platform, cgroup mode, container runtime) and lifecycle test filtering.
  * Configured advanced thresholds and regression tracking; metrics collection disabled.
  * Restricted an existing 4.22 fencing entry to platform "metal" only.

### DIFF
--- a/config/views.yaml
+++ b/config/views.yaml
@@ -879,6 +879,80 @@ component_readiness:
     enabled: false
   automate_jira:
     enabled: false
+- name: 4.22-ha-vs-two-node-arbiter
+  base_release:
+    release: "4.22"
+    relative_start: now-7d
+    relative_end: now
+  sample_release:
+    release: "4.22"
+    relative_start: now-7d
+    relative_end: now
+  variant_options:
+    column_group_by:
+      Architecture: { }
+      Network: { }
+      Platform: { }
+    db_group_by:
+      Architecture: { }
+      FeatureSet: { }
+      Installer: { }
+      Network: { }
+      Platform: { }
+      Suite: { }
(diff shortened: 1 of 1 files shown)


jaypoulz:
As part of this change, can you limit the view to just "metal" platform? (For both TNF and TNA)

eggfoobar:
This might be a silly question, but should we combine these as `ha-vs-two-node`? Since we can specify multiple topologies in a view we can just have a 3 column compare between HA metal and TNF/TNA in one view?

jaypoulz:
> This might be a silly question, but should we combine these as `ha-vs-two-node`? Since we can specify multiple topologies in a view we can just have a 3 column compare between HA metal and TNF/TNA in one view?

@eggfoobar would the regression list show regressions for both? I lean towards seperate  because I know what that would look like. I'm fine with HA vs Two-Node if it still possible to see regressions between HA and TNF plus HA and TNA, but I don't know if that would be the result of that.

jogeo:
@eggfoobar  The goal of these views is to help us show stability of TNA (and TNF) jobs. Once stable, they will be added to the main CR view. Therefore, I suggest we do not tie them together, as moving one to the main view will not depend on the other.

jaypoulz (approved):
/lgtm

PR #3433, opened 2026-04-14
TRT-2547: Track failed job runs per regression

dgoodwin:
This PR implements tracking of every job run we saw contribute to a regression, throughout it's lifespan. (just the failures) 
To do this, I needed access to the test details reports in the regression tracker. 

This made merging the cache primer and the regression tracker the most logical approach, I didn't want to deal with hitting the cache and hoping it wasn't missing, nor passing huge blobs for all views in memory. More efficient to do both at once and they are intricately related. 

I left a hack so the old cronjob command still works as before, the old loader names all map to the new loader. (--loader component-readiness-cache  --loader regression-tracker)

Previously we had a super minimal component readiness e2e just checking we could load the report for main view. This is expanded now to cover the cache primer, and regression tracker. 

NOTE: Today regressions can span views, this will add the job runs from all views, but will not duplicate. I think that's ok,  and I believe we'll be restoring regression to view mapping soon.

## Summary by CodeRabbit

* **New Features**
  * Regression tracking now persists detailed job-run records, merges/deduplicates runs, and can cache test-details for faster component-readiness responses; UI shows overlapping job-run percentages for potential matches.

* **Refactor**
  * Unified cache priming and regression tracking into a single regression-cache workflow; matching now prioritizes overlapping job-run evidence and updated confidence scoring.

* **Tests**
  * Added unit and E2E tests for job-run extraction, merge/persistence, cascade deletes, ordering, and cache-backed behavior.

* **Documentation**
  * Testing guidance added to avoid repeated expensive e2e runs and recommend local vet/test checks.


pkg/db/query/triage_queries.go
```diff
@@ -8,7 +8,7 @@ import (
 
 func ListTriages(dbc *db.DB) ([]models.Triage, error) {
 	var triages []models.Triage
-	res := dbc.DB.Preload("Bug").Preload("Regressions").Find(&triages)
+	res := dbc.DB.Preload("Bug").Preload("Regressions.JobRuns").Preload("Regressions").Find(&triages)
```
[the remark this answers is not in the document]
dgoodwin: I want this returned from those apis for AI analysis. It seldom exceeds a few dozen entries.

pkg/api/componentreadiness/triage.go
```diff
@@ -391,26 +397,63 @@ func determinePotentialMatch(regression models.TestRegression, triage *models.Tr
...
+	}
+}
+
 // GetRegressionPotentialMatchingTriages returns a list of PotentialMatchingTriage including all possible matching triages for a given
 // regression. It calculates this based on similarly named tests being regressed, and associated regressions that
 // have the same last failure time. It includes a confidence level for each match that states how likely the match is to be relevant.
 func GetRegressionPotentialMatchingTriages(regression models.TestRegression, triages []models.Triage, req *http.Request) ([]PotentialMatchingTriage, error) {
 	var potentialMatches []PotentialMatchingTriage
 	baseURL := sippyapi.GetBaseURL(req)
+	resolvedCutoff := time.Now().Add(-6 * 7 * 24 * time.Hour) // 6 weeks ago
```
dgoodwin, line 449: I threw this in, been noticing things like GCP mass failure triages from ages ago show up recommended often.

dgoodwin:
I think this is ready for review, reviewed the code myself, tested manually, expanded with some e2e coverage and unit tests as well.

sippy-ng/src/component_readiness/TriagePotentialMatches.js
```diff
@@ -400,7 +400,7 @@ export default function TriagePotentialMatches({
       headerName: (
         <Tooltip
           title={
-            'Confidence Level (0-10) - Higher values indicate higher likelihood of matching based on: Similar test names (edit distance scoring), Same last failure times (fails in the same job runs)'
+            'Confidence Level (0-10) - Higher values indicate higher likelihood of matching based on: Overlapping failed job runs (shared prow job runs between regressions), Similar test names (edit distance scoring)'
           }
```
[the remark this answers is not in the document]
dgoodwin: This was trimmed elsewhere, but it's now adjusted to be 1-10.

cmd/sippy/load.go
```diff
@@ -200,9 +203,26 @@ func NewLoadCommand() *cobra.Command {
 					if len(views.ComponentReadiness) == 0 {
 						return fmt.Errorf("no component readiness views provided")
 					}
-					loaders = append(loaders, crcacheloader.New(dbc, cacheClient, bqc, config, views, releaseConfigs,
-						f.ComponentReadinessFlags.CRTimeRoundingFactor))
 
+					// Build the regression store if any view uses regression tracking
+					var regressionStore componentreadiness.RegressionStore
+					for _, view := range views.ComponentReadiness {
+						if view.RegressionTracking.Enabled {
```
smg247: The logic here doesn't really seem necessary, when are we going to have no views with regression-tracking enabled? I think we should just always build the regressionStore.

smg247: ...unless this is like this for testing purposes, in which case, I think we should add a comment stating that.

pkg/api/componentreadiness/triage.go
```diff
@@ -376,12 +375,19 @@ func GetTriagePotentialMatches(triage *models.Triage, allRegressions []models.Te
...
 }
 
-// determinePotentialMatch decides if the given regression has the potential to be associated with the given triage
+// determinePotentialMatch decides if the given regression has the potential to be associated with the given triage.
+// It compares by job run overlap (strongest signal) and similar test names (supplementary signal).
 func determinePotentialMatch(regression models.TestRegression, triage *models.Triage) *PotentialMatch {
 	match := &PotentialMatch{}
+
+	// Build a set of the candidate regression's job run IDs for fast lookup
+	candidateRunIDs := make(map[string]bool, len(regression.JobRuns))
```
smg247: Can we use an `apimachinery.set` for this instead?

pkg/dataloader/regressioncacheloader/regressioncacheloader.go
```diff
@@ -0,0 +1,477 @@
...
+// RegressionCacheLoader is a unified loader that handles both component readiness cache priming
+// and regression tracking in a single pass. This eliminates duplicate BigQuery queries that
+// occurred when these were separate loaders.
+//
+// Behavior is controlled by per-view flags:
+//   - view.PrimeCache.Enabled: cache component report and test details in Redis
+//   - view.RegressionTracking.Enabled: sync regressions to Postgres, track job runs
+//
+// For backward compatibility, the legacy --loader flags "component-readiness-cache" and
+// "regression-tracker" both map to this single loader.
```
smg247: We probably don't need this comment since that will go away shortly after merge. It would be better to just have to change one place and not forget this

pkg/dataloader/regressioncacheloader/regressioncacheloader.go
```diff
@@ -0,0 +1,477 @@
...
+
+	// Cache priming deps
+	bqClient             *bigquery.Client
+	config               *configv1.SippyConfig
+	releases             []apiv1.Release
+	crTimeRoundingFactor time.Duration
+
+	// Regression tracking deps (nil if not needed)
+	regressionStore            componentreadiness.RegressionStore
+	variantJunitTableOverrides []configv1.VariantJunitTableOverride
```
smg247: related to earlier comment, will these ever really be nil?

pkg/dataloader/regressioncacheloader/regressioncacheloader.go
```diff
@@ -0,0 +1,477 @@
...
+
+	// Step 1: Generate the component report (one BQ query, cached)
+	report, err := l.generateAndCacheReport(ctx, generator)
+	if err != nil {
+		return nil, err
+	}
+
+	// Step 2: Sync regressions if enabled
+	var activeRegressions []*models.TestRegression
+	if view.RegressionTracking.Enabled && l.regressionStore != nil {
```
smg247: `l.regressionStore` will always be non-nil if regression-tracking is enabled

pkg/dataloader/regressioncacheloader/regressioncacheloader.go
```diff
@@ -0,0 +1,477 @@
...
+			return nil, fmt.Errorf("error syncing regressions for view %s: %w", view.Name, err)
+		}
+	}
+
+	// Step 3: Collect all regressed tests that need test details
+	regressedTests := l.collectRegressedTests(report)
+	vLog.Infof("found %d unresolved regressed tests in report", len(regressedTests))
+
+	// The MaxRegressionsToCache limit only applies to caching, not to the test details query
+	// itself. We always run the query so regression job run tracking works regardless.
```
smg247, line 201: This comment contradicts the comment on line 29:
```
// MaxRegressionsToCache caps the number of test details reports we cache to protect Redis memory.
// When regressions exceed this limit, test details caching and job run tracking are both skipped.
```

pkg/dataloader/regressioncacheloader/regressioncacheloader.go
```diff
@@ -0,0 +1,477 @@
...
+	cacheResults := view.PrimeCache.Enabled && len(regressedTests) <= MaxRegressionsToCache
+	if view.PrimeCache.Enabled && len(regressedTests) > MaxRegressionsToCache {
+		vLog.Warnf("skipping test_details caching: %d regressions exceeds max (%d), job run tracking will still proceed",
+			len(regressedTests), MaxRegressionsToCache)
+	}
```
smg247:
```suggestion
    primeCache := view.PrimeCache.Enabled
    maxRegressionsExceeded := len(regressedTests) > MaxRegressionsToCache
	cacheResults := primeCache && !maxRegressionsExceeded
	if primeCache && maxRegressionsExceeded {
		vLog.Warnf("skipping test_details caching: %d regressions exceeds max (%d), job run tracking will still proceed",
			len(regressedTests), MaxRegressionsToCache)
	}
```

pkg/dataloader/regressioncacheloader/regressioncacheloader.go
```diff
@@ -0,0 +1,477 @@
...
+			TestID:            report.TestID,
+			RequestedVariants: report.Variants,
+			Component:         report.Component,
+			Capability:        report.Capability,
+		}
+		if report.Analyses[0].BaseStats != nil && report.Analyses[0].BaseStats.Release != view.BaseRelease.Name {
+			newTIDOpts.BaseOverrideRelease = report.Analyses[0].BaseStats.Release
+		}
+		genCacheKey.TestIDOptions = []reqopts.TestIdentification{newTIDOpts}
+		tempKey := api.NewCacheSpec(genCacheKey, "TestDetailsReport~", nil)
```
smg247: "TestDetailsReport~" should be a const somewhere (if it isn't already)

pkg/db/models/triage.go
```diff
@@ -225,6 +225,24 @@ type TestRegression struct {
...
 	// Links contains HATEOAS-style links for this regression record (not stored in database)
 	Links map[string]string `json:"links,omitempty" gorm:"-"`
 }
+
+// RegressionJobRun represents a single job run observed during the lifetime of a regression.
+// It stores data from BigQuery so we don't depend on the job existing in PostgreSQL's prow_job_runs table.
+type RegressionJobRun struct {
+	ID           uint           `json:"id" gorm:"primaryKey"`
+	RegressionID uint           `json:"regression_id" gorm:"column:regression_id;not null;uniqueIndex:idx_regression_job_run"`
+	ProwJobRunID string         `json:"prowjob_run_id" gorm:"column:prow_job_run_id;not null;uniqueIndex:idx_regression_job_run"`
```
smg247, line 241: This is an ID from BQ, correct? I think that the way it is named implies it is an ID for a postgres row. I would like to see this more explicitly named, but I would settle for a detailed comment explaining it.

dgoodwin: I believe it's technically all of the above, prow, bigquery, and we also use as the postgres row ID IIRC.

pkg/dataloader/regressioncacheloader/regressioncacheloader.go
dgoodwin: Yup looks out of date, originally job run tracking was skipped, but I modified mid-way to always run the job run tracking. Updating on 29.

pkg/dataloader/regressioncacheloader/regressioncacheloader.go
dgoodwin: Moved to a nil check in New()

smg247 (commented):
/lgtm

PR #3449, opened 2026-04-17
TRT-2600: Determine OS variant from cluster data

petr-muller:
Reworks OS variant detection to derive the value from structured cluster data (the `os` object in cluster-data.json added in https://github.com/openshift/origin/pull/30827) instead of job name patterns. The cluster data contains OS image stream names for the default, control plane, and worker machine config pools. These stream names (e.g. `rhel-9.6`, `rhel-10.0`, `rhel-9-nvidia`) are mapped to variant values (`rhcos9`, `rhcos10`) using a regex that extracts the major version.

When all node pools agree on the OS, that value becomes the variant; when they disagree, the variant is `mixed`. Empty pool values are backfilled from the default, and if the default is also empty, it is inferred from the release major version. The only job-name pattern preserved is `rhcos9-10`, which takes precedence over cluster data.

For the variant snapshot (which has no access to real cluster data), jobs with `rhcos10` in their name get synthetic cluster data so the snapshot accurately reflects their expected OS variant.

## Summary by CodeRabbit

* **Improvements**
  * OS variant detection now reads cluster-data to determine ControlPlane/Workers/Optional OS, applies smarter defaults, maps rhel-* streams to rhcos variants, and correctly produces rhcos9, rhcos10, rhcos9-10, or mixed outcomes.

* **Tests**
  * Expanded unit and snapshot tests covering defaulting, backfilling, mixing behaviors, stream-name mappings, and multiple job scenarios.

### DIFF
--- a/pkg/variantregistry/ocp.go
+++ b/pkg/variantregistry/ocp.go
@@ -2,6 +2,7 @@ package variantregistry
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"regexp"
 	"sort"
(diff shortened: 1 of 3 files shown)


petr-muller:
@coderabbitai resume

PR #3456, opened 2026-04-21
Add support for using regex patterns in findSuite

oharan2:
Building on [ci-test-mapping #616](https://github.com/openshift-eng/ci-test-mapping/pull/616); this PR provides complmentry support for loading desired test suites based on a matching regex pattern, 
instead of holding list of static suite names, [ref](https://github.com/openshift/sippy/blob/main/pkg/db/suites.go#L19), and reducing the need to add every suite name explicitly.

Main changes:
- In addition to the `testSuite` list, add `testSuitePatterns` and compile them on init
- Add helper function `TestIsTestSuiteImportable` for find matches, and use it in `findSuite`
- Add unit tests

## Summary by CodeRabbit

* **New Features**
  * Importable test suites that match configured literal names or patterns are now auto-created instead of ignored.

* **Bug Fixes**
  * Missing suite lookups now consult importable-pattern rules and create the suite when permitted, improving reliability.

* **Tests**
  * Added coverage validating suite importability across multiple naming patterns and edge cases.

### DIFF
--- a/pkg/dataloader/prowloader/prow.go
+++ b/pkg/dataloader/prowloader/prow.go
@@ -1167,12 +1167,23 @@ func (pl *ProwLoader) findSuite(name string) *uint {
 	suite := &models.Suite{}
 	pl.dbc.DB.Where("name = ?", name).Find(&suite)
 	if suite.ID == 0 {
-		pl.suiteCache[name] = nil
-	} else {
-		id := suite.ID
-		pl.suiteCache[name] = &id
+		// No row - the exact suite name is not in the database (for example, by populateTestSuitesInDB)
+		if !db.IsTestSuiteImportable(name) {
+			pl.suiteCache[name] = nil
+			return nil
+		}
+		// If IsTestSuiteImportable() returns true (for example, pattern match), add the Suite row.
+		suite.Name = name
+		tx := pl.dbc.DB.Save(suite)
+		if tx.Error != nil {
+			log.WithError(tx.Error).Warningf("failed to create suite %q", name)
+			return nil
+		}
+		log.WithField("suite", name).Info("created new test suite")
 	}
-	return pl.suiteCache[name]
+	id := suite.ID
+	pl.suiteCache[name] = &id
+	return &id
 }
 
 func (pl *ProwLoader) prowJobRunTestsFromGCS(ctx context.Context, pj *prow.ProwJob, id uint, path string, junitPaths []string) ([]*models.ProwJobRunTest, int, sippyprocessingv1.JobOverallResult, error) {
--- a/pkg/db/suites.go
+++ b/pkg/db/suites.go
@@ -81,7 +81,7 @@ var testSuites = []string{
 var testSuitePatterns = []string{
 	// LP interop naming: `lp-interop-<product>--<suffix>`.
 	`^lp-interop-`,
-
+}
 
 var compiledTestSuitePatterns []*regexp.Regexp
 
--- a/pkg/db/suites.go
+++ b/pkg/db/suites.go
@@ -1,6 +1,8 @@
 package db
 
 import (
+	"regexp"
+
 	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
 	"gorm.io/gorm"
@@ -74,6 +76,44 @@ var testSuites = []string{
 	"tracing-uiplugin",
 }
 
+// testSuitePatterns are regular expressions (MatchString) for suite names that should be imported
+// without listing every literal name. Patterns are compiled in init().
+var testSuitePatterns = []string{
+	// LP interop naming: `lp-interop-<product>--<suffix>`.
+	`^lp-interop-`,
+
+
+var compiledTestSuitePatterns []*regexp.Regexp
+
+// Invalid regexes panic at process start.
+func init() {
+	compiledTestSuitePatterns = make([]*regexp.Regexp, len(testSuitePatterns))
+	for i, p := range testSuitePatterns {
+		compiledTestSuitePatterns[i] = regexp.MustCompile(p)
+	}
+}
+
+// Runs inside the prow loader on every unknown suite name
+// Reports whether junit from this testsuite name should be imported: either an
+// exact entry in testSuites or a match against testSuitePatterns.
+func IsTestSuiteImportable(name string) bool {
+	if name == "" {
+		return false
+	}
+	for _, s := range testSuites {
+		if s == name {
+			return true
+		}
+	}
+	for _, re := range compiledTestSuitePatterns {
+		if re.MatchString(name) {
+			return true
+		}
+	}
+	return false
+}
+
+// Runs when the DB is set up / migrated.
 func populateTestSuitesInDB(db *gorm.DB) error {
 	for _, suiteName := range testSuites {
 		s := models.Suite{}
--- a/pkg/db/suites_test.go
+++ b/pkg/db/suites_test.go
@@ -1,0 +1,28 @@
+package db
+
+import "testing"
+
+func TestIsTestSuiteImportable(t *testing.T) {
+	tests := []struct {
+		name string
+		want bool
+	}{
+		{"", false},
+		{"openshift-tests", true},
+		{"CNV-lp-interop", true},
+		{"ACS-lp-interop", true},
+		{"lp-interop-ACS--my-tests", true},
+		{"lp-interop-Foo", true},
+		{"CNV-lp-interop-extra-suffix", false},
+		{"random-suite", false},
+		{"-lp-interop", false},
+		{"lp-interop", false},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if got := IsTestSuiteImportable(tt.name); got != tt.want {
+				t.Errorf("IsTestSuiteImportable(%q) = %v, want %v", tt.name, got, tt.want)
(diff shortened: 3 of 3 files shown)


pkg/db/suites.go
```diff
@@ -74,6 +76,44 @@ var testSuites = []string{
 	"tracing-uiplugin",
 }
 
+// testSuitePatterns are regular expressions (MatchString) for suite names that should be imported
+// without listing every literal name. Patterns are compiled in init().
+var testSuitePatterns = []string{
+	// LP interop naming: `lp-interop-<product>--<suffix>`.
+	`^lp-interop-`,
```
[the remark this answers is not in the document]
oharan2: The pattern `lp-interop-<product>--<suffix>` in the comment is an example

oharan2:
@stbenjam Question regarding the best approach: would it be better to address the `IsTestSuiteImportable` check in the level of `prowloader`, and not in the `db/suites.go` level? Thanks

pkg/db/suites.go
etirta, line 21: We convert this array to a RegEx pattern array, remove the `sippy` entry and always create it in `populateTestSuitesInDB()`:
```go
var TestSuitePatterns = []*regexp.Regexp{
    regexp.MustCompile(`^openshift-tests$`),
     ...
    regexp.MustCompile(`^lp-interop-`),
}

func populateTestSuitesInDB(db *gorm.DB) error {
    s := models.Suite{Name: "sippy"}    // Sippy synthetic tests.
    if err := db.Clauses(clause.OnConflict{UpdateAll: true}).Create(&s).Error; err != nil {
        return errors.Wrapf(err, "error loading suite into db: sippy")
    }
    return nil
}
```

Advantage:
 1. Only 1 array to maintain.
 2. The footprint of `populateTestSuitesInDB()` is significantly reduced.

pkg/dataloader/prowloader/prow.go
```diff
@@ -1167,12 +1167,23 @@ func (pl *ProwLoader) findSuite(name string) *uint {
...
+		if tx.Error != nil {
+			log.WithError(tx.Error).Warningf("failed to create suite %q", name)
+			return nil
+		}
+		log.WithField("suite", name).Info("created new test suite")
 	}
-	return pl.suiteCache[name]
+	id := suite.ID
+	pl.suiteCache[name] = &id
+	return &id
```
etirta, line 1186: With a single array implementation, the solution is more straightforward (no call to an external function):
```suggestion
    if suite.ID == 0 {
        pl.suiteCache[name] = nil
        for _, re := range db.TestSuitePatterns {
            if re.MatchString(name) {
                suite.Name = name
                result := pl.dbc.DB.Where("name = ?", name).FirstOrCreate(suite)
                if result.Error == nil {
                    if result.RowsAffected > 0 {
                        log.WithField("suite", name).Info("created new test suite")
                    }
                    pl.suiteCache[name] = &suite.ID
                } else {
                    log.WithError(result.Error).Warningf("failed to create suite %q", name)
                }
                break
            }
        }
    } else {
        pl.suiteCache[name] = &suite.ID
    }
    return pl.suiteCache[name]
```

etirta (changes requested):
Hi Ori,

Thanks for working on this. However, the proposal adds complexity by requiring two arrays to be maintained going forward.

I think it's better to refactor the existing code to use regex patterns exclusively. Please see below suggestions. This actually reduces `pkg/db/suites.go`'s footprint instead of increasing it.

The only trade-off is a one-time migration: each existing exact name in `testSuites` becomes `^name$` — it is mechanical, low-risk, and never needs to be done again.

What do yo think @neisw?

pkg/db/suites.go
neisw: I'm going to opt for tread lightly.  Part of our e2e tests initialize a new db and load some synthetic data.  It is possible it is relying on some of the expected suites to be populated.


// Runs when the DB is set up / migrated.
func populateTestSuitesInDB(db *gorm.DB) error {

pkg/dataloader/prowloader/prow.go
```diff
@@ -1167,12 +1167,23 @@ func (pl *ProwLoader) findSuite(name string) *uint {
 	suite := &models.Suite{}
 	pl.dbc.DB.Where("name = ?", name).Find(&suite)
 	if suite.ID == 0 {
-		pl.suiteCache[name] = nil
-	} else {
-		id := suite.ID
-		pl.suiteCache[name] = &id
+		// No row - the exact suite name is not in the database (for example, by populateTestSuitesInDB)
+		if !db.IsTestSuiteImportable(name) {
```
neisw, line 1171: Could IsTestSuiteImportable do the save of the suite?  Claude noted the duplication of logic.  Wondering if the save could be refactored within db/suites so that both populateTestSuitesInDB and IsTestSuiteImportable used the same method.

  Duplicate pattern found in:                                                                                                                              
  - pkg/dataloader/prowloader/prow.go:1177-1181                                                                                                            
  - pkg/db/suites.go:85-92 (in populateTestSuitesInDB)                                                                                                     
                                                      
  Both locations contain identical suite creation logic:                                                                                                   
  suite.Name = name                                                                                                                                        
  tx := DB.Save(suite)                                                                                                                                     
  if tx.Error != nil {                                                                                                                                     
      log.WithError(...).Warningf(...)                                                                                                                     
      return ...                                                                                                                                           
  }                                                                                                                                                        
  log.WithField("suite", name).Info("created new test suite")                                                                                              
                                                                                                                                                           
  Recommendation: Extract into shared function:                                                                                                            
  // In pkg/db/suites.go                                                                                                                                   
  func createSuite(db *gorm.DB, name string) (*models.Suite, error)

pkg/dataloader/prowloader/prow.go
```diff
@@ -1167,12 +1167,23 @@ func (pl *ProwLoader) findSuite(name string) *uint {
...
+		suite.Name = name
+		tx := pl.dbc.DB.Save(suite)
+		if tx.Error != nil {
+			log.WithError(tx.Error).Warningf("failed to create suite %q", name)
+			return nil
+		}
+		log.WithField("suite", name).Info("created new test suite")
 	}
-	return pl.suiteCache[name]
+	id := suite.ID
```
neisw, line 1184: I was curious if ID would be set properly in the struct here since it is created during the save.  The indication is that it should be.  I wonder if it is still worth a check to verify it isn't 0 after the save though.  Worst case would be that the first time it is save that job would be ignored if the ID was still 0.  Next query it should exists.

● Analysis of suite.ID Population After Save()
                                                            
  You've identified a critical question! Let me trace through the code path:
                 
  suite := &models.Suite{}                           // ID = 0, Name = ""                                                                                  
  pl.dbc.DB.Where("name = ?", name).Find(&suite)     // Find fails, suite.ID still 0                                                                       
  if suite.ID == 0 {                                                                                                                                       
      suite.Name = name                               // ID = 0, Name = "lp-interop-foo"                                                                   
      tx := pl.dbc.DB.Save(suite)                    // INSERT operation                                                                                   
      // ... error handling ...                                                                                                                            
  }                                                                                                                                                        
  id := suite.ID                                      // ⚠️  Will this be populated?                                                                        
  pl.suiteCache[name] = &id                                                                                                                                
  return &id                                                                                                                                               
                                                                                                                                                           
  Expected GORM Behavior: YES, it should be populated  

 With GORM + PostgreSQL:                                                                                                                                  
  1. Save() on a struct with ID == 0 triggers an INSERT
  2. PostgreSQL generates the ID via nextval()                                                                                                             
  3. GORM uses the RETURNING id clause to retrieve it
  4. GORM populates suite.ID in the struct after the INSERT

pkg/db/suites.go
etirta: @neisw, do you have examples of the E2E tests you are concerned about? I have already used an AI to analyze this repository to determine how they are used, but I assume those E2E tests are hosted in other repos. If you can provide the locations of those E2E tests, I can run the exposure risk analysis.

My current proposal handles `sippy` as a synthetic test, but I suspect your concern pertains to other E2E test suites; is that correct?

etirta: Alternatively, we could take a more conservative approach. Instead of replacing `testSuites`, we could introduce:

```yaml
var testSuitePatterns = []*regexp.Regexp{
    regexp.MustCompile(`^lp-interop--`),
    ...
}
```

We can keep `populateTestSuitesInDB()` as is to pre-seed the database from `testSuites`. Then, we can implement the proposal outlined in [this PR comment](https://github.com/openshift/sippy/pull/3456/changes#r3120537147).

I believe this should work without introducing unnecessary complexity, while also addressing your concern that some tests may require the database to be pre-seeded. What do you think?

neisw: I did not run AI to evaluate the impact.  Just suggested minimizing impact.  [2e-scripts](https://github.com/oharan2/sippy/tree/1255211c29307e18d7dc6ed4d97dd92a64fa4546/e2e-scripts) looks like it is running a loader so it may work.

Seems like this would still have the duplication of code for saving the suite

neisw: Looking it over a bit more and thinking about the single array I would actually prefer we don't scan a large(ish) list of potential matches each time we miss a suite.  Seems keeping the list small to only the expected dynamic names is better.

Consider https://github.com/openshift/sippy/pull/3478 as an alternate approach

oharan2:
Will be completed in #3478

PR #3461, opened 2026-04-22
TRT-2588: Add synthetic release support to pin jobs to non-OCP releases

smg247:
Jobs in synthetic releases (rosa-stage, aro-*, etc.) containing OCP version numbers in their names were incorrectly assigned to OCP releases by name-based version extraction. Add a Synthetic flag to ReleaseConfig that gives explicit Jobs map entries priority over name-based matching in both the variant extraction and prow loader pipelines.

I am adding the planning document that I generated via Claude for this feature. I am not sure what structure we will want for these in the future, but I wanted it checked in for now and it can be moved later.

It appears that some `JobTier`s have changed. This is actually the intended behavior. These jobs were candidate before only because the tier lookup couldn't find them in any release config (since Release was unset). Now that they have a proper release assignment, the tier resolution works correctly. The jobs are explicitly listed in their release's Jobs map, which is the criterion for standard tier.

## Summary by CodeRabbit

* **New Features**
  * Added synthetic release overrides to pin jobs to synthetic releases across tooling

* **Bug Fixes**
  * Fixed inconsistent job-to-release routing and removed duplicate variant processing
  * Ensured release-derived variant fields remain consistent when synthetic overrides apply

* **Documentation**
  * Added [TRT-2588](https://redhat.atlassian.net/browse/TRT-2588) design/plan and updated DEVELOPMENT.md and Makefile guidance for snapshot updates

* **Tests**
  * Added tests for synthetic override construction and variant version selection


smg247:
@coderabbitai please verify that I have resolved all the issues you have opened and mark them as such.

config/openshift.yaml
stbenjam, line 1: This file is entirely generated, I think you would need to put this in https://github.com/openshift/sippy/blob/main/config/openshift-customizations.yaml which would be merged into the final config by sippy-config-generator

However, we are splitting some properties of releases now, we have capabilities in the release table in BQ, would it be better go there?

smg247: Good call on adding it to the releases table. I will look into that.

petr-muller:
@coderabbit if all your issues were resolved, why haven't you posted an approving review? I am NOT asking you to post it, I need to know the reason why it did not happen

petr-muller:
@coderabbit about

> @petr-muller: Resolving review comments and posting an approving review are two completely independent actions in CodeRabbit's workflow. When @smg247 asked me to verify and mark issues as resolved, I used the [resolve] action — which only closes the open review threads. It does not automatically trigger a PR approval.
> An approval from CodeRabbit requires a separate, explicit @coderabbitai approve command posted as a top-level PR comment. Since no such command was issued, no approving review was posted.

yeah but we do not want to tell you to approve the PR, that's your job. you should approve the PR once there is no open matter to address, without us telling you.

petr-muller:
> The current behavior — requiring an explicit @coderabbitai approve command — is a deliberate safety constraint in my design to prevent unintended approvals.

@coderabbit that sounds like a lie, this repository is configured with `request_changes_workflow: true`

petr-muller:
@coderabbit WTAF

https://github.com/openshift/sippy/blob/main/.coderabbit.yaml#L9

petr-muller:
@coderabbit are you reading the configuration file from the PR branch history? Because in the target branch (main) it does have `request_changes_workflow: true`

stbenjam:
@coderabbitai This branch does not change .coderabbit.yaml.  You seem to be fundamentally broken if you are not evaluating PR's against their merge base....

PR #3468, opened 2026-04-24
TRT-2574: Formalize regression-view associations and cross-compare isolation

smg247:
Adds regression_views junction table tracking which views each regression appears in, with per-view HATEOAS links and a CrossCompare flag to prevent identity collisions between standard and cross-compare views sharing the same sample release. Frontend updated to handle composite link keys and use view-based potential matches API.

## Summary by CodeRabbit

* **New Features**
  * Per-view regression tracking with per-view test-details links (keys like test_details:<view>), a unified "View" selector in the UI, and Potential Matches API now accepts a single view parameter.

* **Bug Fixes**
  * Rolled-off view associations are deactivated correctly; standard and cross-compare regressions are isolated to prevent cross-contamination.

* **Documentation**
  * Added an implementation plan describing regression-view associations and cross-compare behavior.

* **Tests**
  * Expanded unit and end-to-end coverage for view associations, per-view links, deactivation, and cross-compare isolation.

### DIFF
--- a/docs/plans/regression-view-association-plan.md
+++ b/docs/plans/regression-view-association-plan.md
@@ -1,0 +1,182 @@
+# Implementation Plan: Regression-View Junction Table
+
+This plan covers the implementation of a `regression_views` junction table that formally associates regressions with the views that detected them.
+
+**Updated 2026-04-24** to reflect completed implementation and remaining work for cross-compare view regression tracking.
+
+## Status Summary
+
+| Section | Status |
+|---------|--------|
+| 1. New Model and Schema | Done |
+| 2. RegressionStore Interface | Done |
+| 3. Unified Loader Changes | Done |
+| 4. HATEOAS Link Generation | Done |
+| 5. Preload Views on Queries | Done |
+| 6. Triage Expansion Updates | Done |
+| 7. Cleanup of Deprecated Functions | Done |
+| 8. Migration and Backfill | Done |
+| 9. Frontend Updates | Done |
+| 10. Potential Matches API | Done |
+| 11. Cross-Compare View Regression Tracking | Done |
+
+## 1–8: Backend Implementation (Complete)
+
+All backend changes from the original plan have been implemented:
+
+- **`regression_views` table** with composite PK `(test_regression_id, view_name)` and `active` boolean, with `ON DELETE CASCADE` from `test_regressions`
+- **`RegressionStore` interface** extended with `UpsertRegressionView` and `DeactivateRolledOffViews` (originally named `DeactivateStaleViews`, renamed for clarity)
+- **`RegressionCacheLoader`** tracks `activeViewMap` per release, calls `UpsertRegressionView` after sync, calls `DeactivateRolledOffViews` after closing
+- **HATEOAS links** on regression objects use composite keys `test_details:<viewName>` generated per active view from `InjectRegressionHATEOASLinks`
+- **Preloads** added for `Views` / `Regressions.Views` on all relevant queries
+- **`GetViewsForTriage`** reads from preloaded `regression.Views` instead of runtime release matching
+- **`getRegressedTestsForRegressions`** collects view names from preloaded `reg.Views`
+- **`FindViewByName`** exported for use by server.go
+- **Schema migration** handled by GORM `AutoMigrate`; backfill happens naturally on first loader run
+
+### Files changed (backend)
+
+| File | Changes |
+|------|---------|
+| `pkg/db/models/triage.go` | Added `RegressionView` model, added `Views` field with `ON DELETE CASCADE` to `TestRegression` |
+| `pkg/db/db.go` | Added `RegressionView` to auto-migrate list |
+| `pkg/api/componentreadiness/regressiontracker.go` | Extended `RegressionStore` interface, implemented `UpsertRegressionView` and `DeactivateRolledOffViews` |
+| `pkg/dataloader/regressioncacheloader/regressioncacheloader.go` | Track `activeViewMap`, call `UpsertRegressionView` after sync, call `DeactivateRolledOffViews` after closing |
+| `pkg/api/componentreadiness/triage.go` | Rewrote `InjectRegressionHATEOASLinks` with per-view composite keys, updated `GetViewsForTriage`, exported `FindViewByName` |
+| `pkg/db/query/triage_queries.go` | Added `.Preload("Views")` / `.Preload("Regressions.Views")` |
+| `pkg/sippyserver/server.go` | Updated expanded triage handler and `getRegressedTestsForRegressions` to use preloaded views |
+| `pkg/api/componentreadiness/middleware/linkinjector/linkinjector.go` | Kept plain `test_details` key (regressed tests in component reports are already scoped to a view) |
+
+## 9. Frontend Updates (Complete)
+
+### 9.1 `getTestDetailsLink` utility
+
+**File:** `sippy-ng/src/component_readiness/CompReadyUtils.js`
+
+Added a utility function that handles both link formats:
+
+```js
+export function getTestDetailsLink(links, viewName) {
+  if (!links) return null
+  if (links['test_details']) return links['test_details']
+  if (viewName && links[`test_details:${viewName}`]) {
+    return links[`test_details:${viewName}`]
+  }
+  const key = Object.keys(links).find((k) => k.startsWith('test_details:'))
+  return key ? links[key] : null
+}
+```
+
+Lookup order:
+1. Plain `test_details` — preferred (used by regressed tests in component reports)
+2. `test_details:<viewName>` — view-specific composite key (used by regression objects)
+3. First `test_details:*` — fallback to any composite key
+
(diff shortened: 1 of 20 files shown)


docs/plans/regression-view-association-plan.md
```diff
@@ -0,0 +1,182 @@
+# Implementation Plan: Regression-View Junction Table
```
dgoodwin, line 1: This doc is an awesome step, did it work well?

docs/plans/regression-view-association-plan.md
```diff
@@ -0,0 +1,182 @@
...
+| File | Changes |
+|------|---------|
+| `pkg/db/models/triage.go` | Added `CrossCompare bool` field with `gorm:"not null;default:false"` to `TestRegression` |
+| `pkg/api/componentreadiness/middleware/regressiontracker/regressiontracker.go` | Added `crossCompare bool` param to `FindOpenRegression` with `tr.CrossCompare != crossCompare` check. `PreAnalysis`/`PostAnalysis` pass `len(r.reqOptions.VariantOption.VariantCrossCompare) > 0`. |
+| `pkg/api/componentreadiness/regressiontracker.go` | `SyncRegressionsForReport` passes `crossCompare` from view config. `OpenRegression` sets `CrossCompare` on new regressions. |
+| `pkg/api/componentreadiness/middleware/regressiontracker/regressiontracker_test.go` | Updated all `FindOpenRegression` calls with `crossCompare` param. Added `TestFindOpenRegression_CrossCompareIsolation` (2 subtests) and "no match when cross-compare flag differs" case. |
+
+### 11.5 Enable regression tracking on cross-compare views
+
+All 6 cross-compare views already have `regression_tracking.enabled: true` in `config/views.yaml`. No config changes needed.
```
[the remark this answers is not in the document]
dgoodwin: Yeah I wouldn't think all cross compare views have this on, only one or two of the recent ones I would expect.

docs/plans/regression-view-association-plan.md
```diff
@@ -0,0 +1,182 @@
...
+| 8. Migration and Backfill | Done |
+| 9. Frontend Updates | Done |
+| 10. Potential Matches API | Done |
+| 11. Cross-Compare View Regression Tracking | Done |
+
+## 1–8: Backend Implementation (Complete)
+
+All backend changes from the original plan have been implemented:
+
+- **`regression_views` table** with composite PK `(test_regression_id, view_name)` and `active` boolean, with `ON DELETE CASCADE` from `test_regressions`
```
dgoodwin, line 27: The main thing I worry about is multi-view regressions, vs duplicating regressions per view.

If cross compare view regressions can't collide with -main, our primary use cases where regressions might appear in multiple views would be across ocp releases (5.0-main and 4.22-main, even back further), and rarer cases like the mass failure view that exists only for a time. 

If we look at generic tests, and remember that regression records recycle if the regression re-appears within something like 5 or 7 days, we reopen and re-use that regression record. With regressions for things like "install should succeed", it's already a little tough to get these regressions to close. If we span multiple releases, it's going to get even harder to get a regression to close. 

We might be ok when viewing a view live, as that presumably would only show active regressions within that view. (if a regression disappears from a view, does it remain in the db linked to that view? or is that link removed) 

But in terms of analysis of regression data itself, things get very skewed, our numbers for time to triage, how long regressions were open, all skew if an old less monitored release is not getting triaged or it's just a generic test that it's keeping the regression open due to bad luck. 

I'm wondering if regression per view, let them duplicate if necessary, use triage to sort it all out, would be a more stable and reliable plan claude could flip us over to? Let me know what you think.

smg247: The regressions are not going to be shared between views with different sample releases. The `active` field will be set to false for any view where the regression has rolled off, even if it is still active in some other view. I think we maybe need to take a look at how the interaction with the audit tables looks now for better reporting abilities.

I think that duplicating the regressions would be a lot more work for triage, and would require someone going into each view and manually triaging the regressions.

smg247: We can maybe look at also having a datetime field for when the regression started and ended for each view on the new linking table...

docs/plans/regression-view-association-plan.md
smg247: Yes, it works well. I had Claude update the doc at the end with some things that ended up being different once we got going with the implementation. No matter how much planning you do, you still end up with some things that need to be changed in practice.

docs/plans/regression-view-association-plan.md
dgoodwin: Within one release, this sounds like it will work. Aside from cross compare views, most others with regression tracking should be using variants we don't include in the main, thus the regressions can't match. It's really just the mass failure one that I know of right now. I think this makes sense and we should be good to go. Thanks!

smg247: That is also what I have noticed when running locally. I do think adding the date fields to help you better track when things were opened and closed on individual views is worthwhile. So I will look into doing that.

docs/plans/regression-view-association-plan.md
```diff
@@ -0,0 +1,182 @@
...
+- **`RegressionStore` interface** extended with `UpsertRegressionView` and `DeactivateRolledOffViews` (originally named `DeactivateStaleViews`, renamed for clarity)
+- **`RegressionCacheLoader`** tracks `activeViewMap` per release, calls `UpsertRegressionView` after sync, calls `DeactivateRolledOffViews` after closing
+- **HATEOAS links** on regression objects use composite keys `test_details:<viewName>` generated per active view from `InjectRegressionHATEOASLinks`
+- **Preloads** added for `Views` / `Regressions.Views` on all relevant queries
+- **`GetViewsForTriage`** reads from preloaded `regression.Views` instead of runtime release matching
+- **`getRegressedTestsForRegressions`** collects view names from preloaded `reg.Views`
+- **`FindViewByName`** exported for use by server.go
+- **Schema migration** handled by GORM `AutoMigrate`; backfill happens naturally on first loader run
```
dgoodwin, line 35: Will / could all existing closed regressions get aligned to release-main view during the backfill process? This is our best guess and will help keep my analysis tooling as healthy as we can. Until recently it was only main that could have regression tracking, and while the data for this release will be a little wonky it's the less bad option I think.

dgoodwin (commented):
I think my concerns are all met, thanks!

docs/plans/regression-view-association-plan.md
smg247: Writing a backfill script in the migration that handles this.

pkg/sippyserver/server.go
```diff
@@ -1834,22 +1824,22 @@ func (s *Server) jsonTriagePotentialMatchingRegressions(w http.ResponseWriter, r
...
+		failureResponse(w, http.StatusBadRequest, fmt.Sprintf("view not found: %s", viewName))
 		return
 	}
 	allReleases, err := s.getReleases(req.Context())
 	if err != nil {
 		failureResponse(w, http.StatusInternalServerError, err.Error())
 		return
 	}
-	regressions, err := componentreadiness.ListRegressions(s.db, sampleRelease, s.views.ComponentReadiness, allReleases, s.crTimeRoundingFactor, req)
+	regressions, err := componentreadiness.ListRegressions(s.db, view.SampleRelease.Name, s.views.ComponentReadiness, allReleases, s.crTimeRoundingFactor, req)
```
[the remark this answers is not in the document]
smg247: The views list is only used for HATEOAS link injection, not filtering. Regressions need links for all views they appear in.

pkg/dataloader/regressioncacheloader/regressioncacheloader.go
```diff
@@ -418,12 +453,12 @@ func (l *RegressionCacheLoader) closeStaleRegressions(release string, activeIDs
 		reg.Closed.Valid = true
 		reg.Closed.Time = now
 		if err := l.regressionStore.UpdateRegression(reg); err != nil {
-			return fmt.Errorf("error closing regression %d: %w", reg.ID, err)
+			return nil, fmt.Errorf("error closing regression %d: %w", reg.ID, err)
```
neisw: Seems like this has the potential to orphan some previously closedIDs and that list is used in part with DeactivateRolledOffViews.

Should this function return a list of errors and keep processing?

pkg/db/db.go
```diff
@@ -112,6 +113,10 @@ func (d *DB) UpdateSchema(reportEnd *time.Time) error {
 		}
 	}
 
+	if err := backfillClosedRegressionViews(d.DB); err != nil {
```
neisw, line 111: Sounds like this is temporary?  Mostly wondering if we want a comment indicating this is mostly just for a migration period and then can be removed.

smg247: This is a good idea. I also found another temporary migration I had commented on and hadn't removed yet so took care of that too.

neisw:
/lgtm
/hold
release when ready

smg247:
/unhold
Let's get this in today. If something goes bad, we can back out safely without needing to migrate data backwards as it is a new table and new column rather than a deletion or anything.

PR #3478, opened 2026-04-28
NO-JIRA: check for dynamic suites

neisw:
Derivative work of https://github.com/openshift/sippy/pull/3456

## Summary by CodeRabbit

* **New Features**
  * Dynamic suite name resolution using configurable pattern matching to broaden accepted suite names.

* **Bug Fixes**
  * Centralized suite ID lookup with improved caching and stricter handling of unknown or invalid suites for more consistent behavior.

* **Tests**
  * Added unit tests validating dynamic suite pattern matching across multiple positive and negative cases.


pkg/dataloader/prowloader/prow.go
```diff
@@ -1184,12 +1184,15 @@ func (pl *ProwLoader) findSuite(name string) *uint {
...
 	if suite.ID == 0 {
-		pl.suiteCache[name] = nil
-	} else {
-		id := suite.ID
-		pl.suiteCache[name] = &id
+		// No row - the exact suite name is not in the database (for example, by populateTestSuitesInDB)
+		// Check if this matches a dynamic suite pattern and create it if so
+		id := db.CheckForDynamicSuite(pl.dbc.DB, name)
+		pl.suiteCache[name] = id
+		return id
```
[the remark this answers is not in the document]
neisw: This is ok,  The prow loader runs on a cron schedule.  We don't want to retry over and over or get overly complicated.  New suites should be rare and if there is a db failure then the next import should pick it up if it isn't a persistent failure.

pkg/db/suites.go
```diff
@@ -74,22 +75,64 @@ var testSuites = []string{
 	"tracing-uiplugin",
 }
 
+// testSuitePatterns are regular expressions (MatchString) for suite names that should be imported
+// without listing every literal name. Patterns are compiled in init().
+var testSuitePatterns = []string{
+	// LP interop naming: `lp-interop-<product>--<suffix>`.
+	`^lp-interop-`,
+}
```
etirta: If we define this as `[]*regexp.Regexp` and use `regexp.MustCompile()`, the regex pattern compilation will occur at package initialization, eliminating the need for a separate `init()` function.

pkg/db/suites.go
```diff
@@ -74,22 +75,64 @@ var testSuites = []string{
...
+	}
+
+	for _, re := range compiledTestSuitePatterns {
+		if re.MatchString(name) {
+			return getOrCreateSuite(db, name)
+		}
+	}
+
+	return nil
+}
```
etirta: What do you think about completely confining the suite search algorithm within this module?
```suggestion
func GetSuiteID(db *gorm.DB, name string) *uint {
      if name == "" {
              return nil
      }

      suite := &models.Suite{}
      db.Where("name = ?", name).Find(&suite)
      if suite.ID != 0 {
              id := suite.ID
              return &id
      }

      for _, re := range compiledTestSuitePatterns {
              if re.MatchString(name) {
                      return getOrCreateSuite(db, name)
              }
      }

      return nil
}
```

This would allow the caller in `pkg/dataloader/prowloader/prow.go` to be simplified as follows:
```go
func (pl *ProwLoader) findSuite(name string) *uint {
    pl.suiteCacheMu.RLock()
    id, ok := pl.suiteCache[name]
    pl.suiteCacheMu.RUnlock()
    if ok {
        return id
    }

    id = db.GetSuiteID(pl.dbc.DB, name)

    pl.suiteCacheMu.Lock()
    pl.suiteCache[name] = id
    pl.suiteCacheMu.Unlock()
    return id
}
```

etirta (changes requested):
Hi @neisw, I support moving forward with this PR, as I agree with exposing the API in `pkg/db/suites.go` to handle the search algorithm.

However, I have a few suggestions to reduce the footprint and ensure the entire search algorithm is fully encapsulated within the `pkg/db/suites.go` module.

pkg/db/suites.go
```diff
@@ -75,38 +75,48 @@ var testSuites = []string{
...
 
-// testSuitePatterns are regular expressions (MatchString) for suite names that should be imported
-// without listing every literal name. Patterns are compiled in init().
-var testSuitePatterns = []string{
+// testSuitePatterns are regular expressions for suite names that should be imported
+// without listing every literal name. Invalid patterns panic at process start.
+var testSuitePatterns = []*regexp.Regexp{
 	// LP interop naming: `lp-interop-<product>--<suffix>`.
-	`^lp-interop-`,
+	regexp.MustCompile(`^lp-interop-`),
```
etirta: We use `--` as a prefix separator to distinguish it from word separators within the prefix itself.
```suggestion
regexp.MustCompile(`^lp-chaos--`),
regexp.MustCompile(`^lp-interop--`),
regexp.MustCompile(`^lp-ocp-compat--`),
```
We are also in the process of handing our current single LP test for cross-OCP versions back to the LP team. Consequently, we want to rename those (faux-)interop tests to `lp-ocp-compat` to reflect their actual nature. These tests only verify LP-OCP version compatibility rather than actual interoperability, which, by definition, should involve two or more LPs.

etirta (changes requested):
Rather than having another PR, let's define what we plan for all of our teams. Thx.

pkg/db/suites.go
```diff
@@ -74,22 +75,80 @@ var testSuites = []string{
...
+	}
+
+	// Check if this suite should be imported
+	if !isSuiteImportable(name) {
+		return nil
+	}
+
+	// Get existing or create new suite
+	return getOrCreateSuite(db, name)
+}
```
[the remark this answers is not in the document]
neisw: The code will try again on the next cron interval.  New suites being added should not be a frequent occurrence.  This is an edge case that is ok leaving as is.

pkg/db/suites.go
```diff
@@ -74,22 +75,89 @@ var testSuites = []string{
...
-			}
-			s = models.Suite{
-				Name: suiteName,
-			}
-			err := db.Clauses(clause.OnConflict{UpdateAll: true}).Create(&s).Error
-			if err != nil {
-				return errors.Wrapf(err, "error loading suite into db: %s", suiteName)
-			}
-			log.WithField("suite", suiteName).Info("created new test suite")
+		if getOrCreateSuite(db, suiteName) == nil {
```
oharan2, line 158: Should be `GetSuiteID` (wrapper func) as it has more checks.

```suggestion
		if GetSuiteID(db, suiteName) == nil {
```

oharan2 (changes requested):

pkg/db/suites.go
```diff
@@ -74,22 +75,89 @@ var testSuites = []string{
 	"tracing-uiplugin",
 }
 
+// testSuitePatterns are regular expressions for suite names that should be imported
+// without listing every literal name. Invalid patterns panic at process start.
+var testSuitePatterns = []*regexp.Regexp{
+	// LP interop naming: `lp-interop-<product>--<suffix>`.
```
etirta: Just one nit, depending on whether it matters for the comment to accurately reflect reality:
```suggestion
	// LP interop naming: `lp-interop--<product>--<suffix>`.
```
Either way, I’m signing off on this. Thanks for accommodating our needs!

pkg/db/suites.go
etirta: No, `populateTestSuitesInDB()` is only intended for pre-seeding. There is no need to check `testSuitePatterns` because pre-seeding requires a concrete suite name.

petr-muller (approved):
LGTM

PR #3486, opened 2026-04-30
Fix variant registry OS fallback for jobs without cluster data

petr-muller:
## Summary
- When the variant registry can't load cluster data for a job (no successful run, GCS issues), the OS variant always defaulted to `rhcos9` regardless of the job name.
- Now infers OS from the job name (`rhcos9`, `rhcos10`, `rhcos9-10`) as a fallback before the hard-coded default. Cluster data still takes precedence when available.
- Matches `rhcos9-10` before `rhcos9` to avoid false partial matches.

## Test plan
- [x] Unit tests for `osFromJobName` covering all patterns, ordering edge case, case insensitivity
- [x] Unit tests for `setOS` fallback behavior (job name used when no cluster data, cluster data wins when present)
- [x] All existing variant registry tests pass

## Release Notes

* **Bug Fixes**
  * Improved system resilience by enabling OS variant resolution when cluster data is unavailable
  * Enhanced support for inferring operating system configuration from job parameters
  * Added support for RHEL 9 and RHEL 10 operating system variant combinations

### DIFF
--- a/pkg/variantregistry/ocp.go
+++ b/pkg/variantregistry/ocp.go
@@ -1306,12 +1306,31 @@ func setLayeredProduct(_ logrus.FieldLogger, variants map[string]string, jobName
 	}
 }
 
-func (os clusterDataOS) setOS(_ logrus.FieldLogger, variants map[string]string, _ string) {
+func (os clusterDataOS) setOS(_ logrus.FieldLogger, variants map[string]string, jobName string) {
 	resolved := os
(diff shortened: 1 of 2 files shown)


petr-muller:
/hold

we'll need to revert https://github.com/openshift/sippy/pull/3449 and adapt job name-based detection instead

petr-muller:
/close 

Replaced by https://github.com/openshift/sippy/pull/3491

PR #3487, opened 2026-04-30
NO-JIRA: Classify os by job name if no cluster data was found

xueqzhan:
## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced OS variant detection for OpenShift releases: derives OS from cluster data, from job names when present (including RHCOS numeric suffixes), and falls back to sensible defaults when data is missing.

* **Tests**
  * Added test coverage validating job-name-driven variant parsing, RHCOS variants (e.g., rhcos9-10), and related configuration defaults (network, security, topology).


petr-muller:
/hold

we'll need to revert https://github.com/openshift/sippy/pull/3449 and adapt job name-based detection instead

petr-muller:
likely superseded by https://github.com/openshift/sippy/pull/3491

PR #3491, opened 2026-05-01
TRT-2648: Determine OS variant from job name with OCP version fallback

petr-muller:
## Summary
- Reverts [TRT-2600](https://redhat.atlassian.net/browse/TRT-2600) (cluster-data-based OS detection) and replaces it with job-name-based detection
- Detects `rhcos9-10`, `rhcos10`, and `rhcos9` fragments from job names, checked in order to avoid false matches (e.g. `rhcos9` matching `rhcos9-10`)
- When no rhcos fragment is present, falls back based on OCP major version: both OCP 4 and 5 default to `rhcos9`, with cases split explicitly for easy future bumping
- Jobs with `-main-` or `-master-` segments and no recognizable OCP version are treated as OCP 5
- Unknown major versions get `"unknown"` to surface misclassification instead of silently defaulting

## Summary by CodeRabbit

* **Refactor**
  * OS variant detection now relies solely on job-name patterns (ordered rhcos9-10, rhcos10, rhcos9) and release heuristics instead of per-job cluster config data.
* **Behavioral Changes**
  * Jobs without recognized RHCOS patterns may report "unknown" OS.
  * RHCOS for some release/job combos now defaults differently (e.g., rhcos9 for certain majors/main/master).
* **Tests & Snapshot**
  * Tests and snapshot data updated to match the new OS reporting.


pkg/variantregistry/ocp.go
```diff
@@ -1306,69 +1286,38 @@ func setLayeredProduct(_ logrus.FieldLogger, variants map[string]string, jobName
...
+		variants[VariantOS] = "rhcos9"
+	case "5":
+		// OCP 5 currently defaults to rhcos9. Update this when the default changes.
+		variants[VariantOS] = "rhcos9"
+	default:
+		if strings.Contains(jobNameLower, "-main-") || strings.Contains(jobNameLower, "-master-") {
+			variants[VariantOS] = "rhcos9"
+		} else {
+			variants[VariantOS] = "unknown"
+		}
```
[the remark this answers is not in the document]
petr-muller: @coderabbitai this is intentional, our master/main follow OCP5 now

pkg/variantregistry/ocp.go
```diff
@@ -1306,69 +1286,35 @@ func setLayeredProduct(_ logrus.FieldLogger, variants map[string]string, jobName
...
 		}
 	}
 
-	if len(seen) == 1 {
-		return cp
-	}
-	if len(seen) == 2 && seen["rhcos9"] && seen["rhcos10"] {
-		return "rhcos9-10"
+	// No explicit rhcos fragment in the job name: fall back based on OCP major version.
+	isMainBranch := strings.Contains(jobNameLower, "-main-") || strings.Contains(jobNameLower, "-master-")
```
neisw, line 1310: This has me wondering as I don't think we typically ingest jobs from main since we can't associate with a release.  But with the effort to support synthetic releases might we see more main type jobs and would there be cases where it wasn't aligned with ocp main.

Maybe unknown is better in that case?

neisw:
/lgtm
/hold
unhold when ready just pausing for comment review.  It isn't meant to block merging.

pkg/variantregistry/ocp.go
petr-muller: yeah this is very tricky - the advantage of reading the cluster data was that it reduced the need for guesswork like this, because there was evidence

I'll file a TRT card for this problem but let's merge this to fix Devan's data

PR #3498, opened 2026-05-05
TRT-2647: rhcos10 default for 5.0

neisw:
Staging for when the default changes over to rhcos10 for 5.0

Open question, should we remove `OS:` from the 5.0 main view so they are all picked up by default or explicitly specify rhcos9 and rhcos10?

## Summary by CodeRabbit

* **New Features**
  * RHCOS 10 is now included as a supported OS variant and becomes the default fallback for v5/main jobs.

* **Chores**
  * Reworked a v5 OS-comparison scenario: swapped comparison direction, changed feature-set, adjusted thresholds (minimum failures, pity factor, pass-rate) and toggles, and removed an obsolete comparison.

* **Tests**
  * Updated test expectations to reflect RHCOS 10 as the v5/main fallback.

### DIFF
--- a/config/views.yaml
+++ b/config/views.yaml
@@ -45,8 +45,6 @@ component_readiness:
       - virt
       Network:
       - ovn
-      OS:
-      - rhcos9
       Owner:
       - eng
       - service-delivery
@@ -135,8 +133,6 @@ component_readiness:
       - virt
       Network:
       - ovn
-      OS:
-      - rhcos9
       Owner:
       - eng
       - service-delivery
@@ -220,8 +216,6 @@ component_readiness:
       - virt
       Network:
       - ovn
-      OS:
-      - rhcos9
       Owner:
       - eng
       - service-delivery
@@ -405,7 +399,7 @@ component_readiness:
     flake_as_failure: false
   metrics:
     enabled: false
-- name: 5.0-techpreview-rhcos9-vs-rhcos10
+- name: 5.0-rhcos10-vs-rhcos9
   base_release:
     release: "5.0"
     relative_start: now-7d
@@ -435,6 +429,7 @@ component_readiness:
       - arm64
       - multi
       FeatureSet:
+      - default
       - techpreview
       Installer:
       - ipi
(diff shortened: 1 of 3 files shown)


neisw:
/hold
until we are ready to switch over

config/views.yaml
```diff
@@ -469,20 +464,22 @@ component_readiness:
...
     ignore_missing: false
     ignore_disruption: true
     flake_as_failure: false
+    pass_rate_required_new_tests: 90
+    include_multi_release_analysis: false
   metrics:
     enabled: false
   regression_tracking:
-    enabled: true
+    enabled: false
```
dgoodwin: This is interesting, looks like it flipped several settings to the "calm things down for a GA branch", is that intentional for the 9 vs 10 5.0 view? I would probably keep regression tracking on as well.

dgoodwin (commented):
Everything else looks good. 

Holding until we hear it's flipped sounds good.

config/views.yaml
neisw: @dgoodwin your jira card [TRT-2647](https://redhat.atlassian.net/browse/TRT-2647) indicates
```
Add an rhcos9 vs 10 cross compare view instead to minimally monitor that 9 doesn’t get worse than 10. Probably could detune it the way we do for ga -main views now. We don’t want to be overly sensitive.
```
I can enable regression_tracking here if you like.  Just let me know what you want the final config to look like.

dgoodwin: yes tracking would be useful here.

pkg/variantregistry/ocp.go
```diff
@@ -1313,8 +1313,7 @@ func setOS(_ logrus.FieldLogger, variants map[string]string, jobName string) {
 	case variants[VariantReleaseMajor] == "4":
 		variants[VariantOS] = "rhcos9"
 	case variants[VariantReleaseMajor] == "5" || isMainBranch:
```
petr-muller: we'll need the special treatment for update jobs as discussed on slack, right?

neisw: details details...

dgoodwin:
Should be live now but needs a rebase and make update-variants to resolve conflicts

dgoodwin:
/close

Replaced by https://github.com/openshift/sippy/pull/3619 which just has make update-variants run after a rebase

PR #3505, opened 2026-05-06
perf: test report cache priming, pagination, and query optimization

stbenjam:
## Cache priming for test results

The test report cache loader now primes both collapsed and non-collapsed results in Redis, eliminating cold-cache latency for the most common test report views. Previously only collapsed results were primed, leaving non-collapsed queries (the detailed per-variant NURP+ view) to hit the database on first access.

To avoid unnecessary work, the cache primer now only targets OCP development releases (identified by having the `payloadTags` capability and no GA date set). GA releases and non-OCP products (OKD, etc.) are skipped.

## Server-side pagination

The /api/tests endpoint previously returned all matching rows (up to 50k+ for uncollapsed views), causing the tests page to barely load. This adds server-side pagination following the existing pattern used by the job runs endpoint.

When perPage/page query parameters are present, the backend:
- Applies ORDER BY, COUNT, LIMIT, and OFFSET at the SQL level
- Returns a PaginationResult envelope with rows, total_rows, page_size, page
- Bypasses the cache (paginated queries are fast with LIMIT/OFFSET)

When pagination params are absent, existing behavior is preserved for backward compatibility.

Frontend changes switch the DataGrid to `paginationMode="server"` and send `perPage`/`page` params, following the `JobRunsTable` pattern.

## Filter pushdown (~830x improvement for filtered queries)

When collapse=false, TestsByNURPAndStandardDeviation builds a query that self-joins prow_test_report_7d_matview 3 times. Name/variant filters were only applied to the outermost query, causing subqueries to scan all rows for the release. Filters are now pushed into the stats and pass_rates subqueries, allowing index use on cache misses and paginated queries.

Replaces #3290 (rebased on main; original e2e failure was an unrelated sippy-load-job timeout).

## Release Notes

* **New Features**
  * Added server-side pagination to the Tests API endpoint with `perPage` and `page` parameters for efficient data retrieval
  * Test results table now supports server-side pagination for improved performance with large datasets
  * Pagination responses include total row count information for navigation

* **Documentation**
  * Updated Tests API documentation with pagination parameters and new response structure


stbenjam:
## Review Panel Verdict

**Disposition**: APPROVE — all BLOCKING findings from initial review resolved

---

### Specialist Findings

**Architecture Reviewer**: Cross-file impact is clean. `SubqueryFilter` is a justified abstraction for pushing filters into subqueries. Pagination path correctly bypasses cache (each page/sort combo has low cache reuse). `downloadDataFunc` now exports only the current page — acceptable tradeoff. No god functions, no circular dependencies.

**Security & Supply Chain Reviewer**: Sort direction uses an allowlist (`SortAscending`/`SortDescending` only — arbitrary strings cannot reach SQL). Sort field uses `pq.QuoteIdentifier()` which double-quote-escapes identifiers, preventing SQL injection. `perPage` bounded to 1–1000, `page` must be non-negative. No new dependencies added (`lib/pq` already vendored). No lockfile, build pipeline, or supply chain changes.

**UX & API Reviewer**: Backward compatible — when `perPage` is omitted, the legacy array response is preserved. When present, response is wrapped in `PaginationResult` envelope (`rows`, `total_rows`, `page_size`, `page`). Error messages for invalid pagination are clear (e.g., "perPage must be between 1 and 1000"). BigQuery path also wraps response in `PaginationResult` when pagination params are present. API docs updated in `pkg/api/README.md`.

**Codebase Consistency Reviewer**: `getPaginationParams` follows the same pattern as other parameter parsers in `parameters.go`. Test file `parameters_test.go` uses table-driven tests with `testify` assertions, consistent with the codebase. `SubqueryFilter` type is consistent with existing GORM query patterns. `PaginationResult` reuses the existing type from `pkg/apis/api/types.go`. `isVariantFilter` was inlined per suggestion.

**QA Engineer**: `parameters_test.go` covers 10 edge cases for `getPaginationParams`: no params, valid params, perPage without page, zero/negative perPage, negative page, exceeds max, non-numeric values, and boundary (max=1000). Sort direction validation is implicitly tested via the allowlist pattern (only `SortAscending` passes the guard). Backend pagination integration with the database requires integration tests (out of scope for this PR).

**Devil's Advocate**: Sort direction injection — resolved, allowlist prevents arbitrary tokens. BigQuery response shape — resolved, wraps in `PaginationResult`. COUNT + OFFSET TOCTOU — acceptable for materialized views (refreshed periodically, not subject to phantom rows). Double-fetch race in frontend (no AbortController) — pre-existing pattern across all tables in the codebase, not introduced by this PR. `ROW_NUMBER() OVER()` without ORDER BY gives non-deterministic IDs — harmless for React keys. Could not construct a failure scenario with the current validation.

**Technical Writer**: API docs in `pkg/api/README.md` updated: `perPage` (1–1000) and `page` (0+) parameters added to the Tests endpoint table. Pagination envelope format documented with example JSON. `limit` marked as legacy. Example response section renamed to clarify it shows the legacy format. No stale docs remain.

**DBA Expert (300 years PostgreSQL)**: Filter pushdown into stats/passRates subqueries is the key performance win (~830x for filtered queries). `pq.QuoteIdentifier` is the correct defense for identifier injection. Sort direction allowlist prevents syntax injection. `COUNT(*)` error is now checked and propagated. `perPage` capped at 1000 prevents full-table-scan via pagination API. LIMIT/OFFSET on materialized views is safe from phantom-row issues. Deep pagination degrades linearly with OFFSET, but acceptable for UI use (max page 1000 * 1000 = 1M offset, well within matview size). PostgreSQL cannot push LIMIT through the nested subquery joins, but the filter pushdown ensures the subqueries themselves are fast.

---

### Panel Synthesis

All eight specialists converged on the same five BLOCKING findings in the initial review, and all five have been resolved:

1. **Sort direction SQL injection** → Fixed with allowlist validation (only `asc`/`desc` reach SQL)
2. **BigQuery response incompatibility** → Fixed, both paths return `PaginationResult` when pagination params present
3. **Unbounded perPage/page** → Fixed, `perPage` validated to 1–1000, `page` >= 0
4. **COUNT error silently ignored** → Fixed, error checked and propagated
5. **API docs not updated** → Fixed, `README.md` updated with new parameters and response format

No remaining BLOCKING findings. The SUGGESTION items (inline `isVariantFilter`, separate `SubqueryFilter` doc comment) were also addressed. The DBA Expert confirmed the pagination approach is sound for materialized views and the filter pushdown is the correct optimization.

---

### Required Actions Before Merge

None.

---

### Optional Follow-ups

- Consider adding an `AbortController` to frontend fetch calls to cancel in-flight requests on rapid page changes (pre-existing pattern across all tables, not specific to this PR)
- Deep pagination could use keyset/cursor pagination if matview sizes grow significantly (current OFFSET approach is fine for UI use)
- `COUNT(*)` could be cached or use `pg_class.reltuples` approximation for very large result sets if count queries become a bottleneck

stbenjam:
## Benchmark: Local API with cache priming (release 5.0)

| Query | Cached | Uncached |
|-------|--------|----------|
| collapsed (default) | 0.130s | 3.426s |
| uncollapsed | 10.146s | 2m2s |
| collapsed + twoDay | 0.114s | - |
| uncollapsed + twoDay | 9.696s | - |
| collapsed + filter name contains sig-node | 0.225s | - |
| uncollapsed + filter name contains sig-node | 17.426s | - |
| collapsed + filter runs > 14 | 0.219s | - |
| uncollapsed + paginated page 0 | 8.358s | - |

Collapsed results see a ~26x improvement from caching. Uncollapsed results are ~12x faster cached vs uncached (2+ minutes down to ~10s). The uncollapsed cached time is dominated by deserializing the ~1GB JSON blob; the uncollapsed uncached query is a full self-joining matview scan that takes over 2 minutes.

dgoodwin:
This is huge and that benchmark comment answers the main question I had, having the option to see the uncompressed list again would be handy, I had to give up on that some time ago.

dgoodwin:
/lgtm
/hold

Just making sure I don't step on TRT's toes if they want a crack at this. Thanks for improving this, will help daily.

stbenjam:
I am not sure this is really the right approach, the caching is clunky and bad, @smg247 is going to look at it, if we can't figure out something better we could try to take this but it is really putting a tuxedo on a toad.

PR #3518, opened 2026-05-11
TRT-1989: schema migration for non gorm based tables

neisw:
Adding schema management for `test_analysis_by_job_by_dates` which previously was unmanaged.

## Summary by CodeRabbit

* **New Features**
  * Added CLI subcommands to view migration version/status, force-set a migration version, and roll back migrations
  * Introduced embedded, runtime-handled DB migrations and migration orchestration

* **Tests**
  * Added end-to-end migration tests covering up/down/force/version flows and connection resilience

* **Chores**
  * Bumped Go toolchain to 1.25 and upgraded dependencies
  * Removed automatic partitioned-table sync from the schema update path


neisw:
/hold
for further review / testing

neisw:
@coderabbitai review the current state of the pr and resolve all of your comments that have been addressed

go.mod
```diff
@@ -1,24 +1,25 @@
 module github.com/openshift/sippy
 
-go 1.25
+go 1.25.0
```
smg247: I think we noticed this before, but we should figure out why this has happened, and revert it.

test/e2e/db/migrate/migrate_test.go
```diff
@@ -0,0 +1,182 @@
...
+	"github.com/openshift/sippy/pkg/db/migrate"
+	"github.com/openshift/sippy/pkg/db/models"
+	"github.com/openshift/sippy/test/e2e/db/migrate/testmigrations"
+	"github.com/openshift/sippy/test/e2e/util"
+)
+
+// TestMigrations verifies that database migrations run successfully
+// and that the database connection remains usable after migration.
+// This test specifically validates the fix for the issue where m.Close()
+// was closing the shared gormDB connection.
```
smg247: nit: this part of the comment is overly specific and not particularly useful to leave around

go.mod
neisw: I added `When updating go module versioning only specify major and minor version. Never require a specific patch version` to [general.instructions.md](https://github.com/neisw/sippy/blob/b5d1a3b055a549671b019f71682ea5d82a5eaf5c/.apm/instructions/general.instructions.md)

But notice the .0 gets set when I run `go mod tidy` so I don't know that my `general.instructions` change is valid

neisw: I removed my instruction update.  I'm finding it is` go mod tidy` that is setting this and it is expected.

_Go 1.21 and later: The go directive in go.mod is now expected to be a full [semantic version](https://www.google.com/url?sa=i&source=web&rct=j&url=https://go.dev/ref/mod&ved=2ahUKEwiBj5Cml7mUAxV3K1kFHa1QAgIQy_kOegoIAAgACAEIDBAC&opi=89978449&cd&psig=AOvVaw2Rilxqd4Htl8gtY-g9PBwp&ust=1778861949544000) (e.g., 1.22.0). When you run go mod tidy, it will typically ensure the go version is listed as a full three-part version (like 1.23.0) if it wasn't already._

smg247: hmm, okay. I guess that is fine then.

smg247 (commented):
/lgtm

PR #3528, opened 2026-05-13
ROSAENG-1083: remove unrelated jobs from 4.22 section

bmeng:
Remove jobs from the 4.22 section which are not belonging 

All the existing rosa-e2e jobs should be handled in rosa-stage section now

## Summary by CodeRabbit

* **Chores**
  * Updated CI test configurations: removed outdated e2e test entries and added new load generator test configurations for 50, 100, and 1000 worker counts.


stbenjam:
This particular file is generated, I think the `ci-tools` sippy-config-generator will just add them back

PR #3539, opened 2026-05-18
TRT-2463: labels for payload job runs

sosiouxme:
* Update the sippy release loader to look up labels applied to a payload job and record them with the DB record
* Update the API and UI to display these

## Summary by CodeRabbit

* **New Features**
  * Job run labels are now fetched and displayed in the release payload view with expandable details showing label explanations and definitions.

* **Chores**
  * Enhanced label retrieval mechanisms with bulk operations for improved performance.


pkg/dataloader/releaseloader/labels.go
```diff
@@ -0,0 +1,73 @@
...
+	"google.golang.org/api/iterator"
+
+	bqcachedclient "github.com/openshift/sippy/pkg/bigquery"
+	"github.com/openshift/sippy/pkg/bigquery/bqlabel"
+)
+
+// GatherBulkLabelsFromBQ queries BigQuery for labels for multiple job runs in a single query.
+// The releaseTime is used to constrain the scan to a single date partition.
+// Returns a map of buildID → labels.
+func GatherBulkLabelsFromBQ(ctx context.Context, bqClient *bqcachedclient.Client, buildIDs []string, releaseTime time.Time) (map[string]pq.StringArray, error) {
```
neisw: pr review is pointing out the duplication with [prow.go](https://github.com/sosiouxme/sippy/blob/53356af3a59a46386871eae59131e344226ef36e/pkg/dataloader/prowloader/prow.go#L1102) any refactoring that could pull these closer together?

It is also complaining about missing unit tests but I'm not sure we have mocks for bqclient.

pkg/dataloader/releaseloader/releasesync.go
```diff
@@ -57,7 +61,9 @@ func New(dbc *db.DB, releases, architectures []string, releaseConfigs []v1.Relea
 	}
 
 	return &ReleaseLoader{
+		ctx:           ctx,
```
neisw, line 68: Issue: Storing context.Context in a struct violates Go best practices. Contexts should be passed as function parameters, not stored in structs.

pkg/dataloader/releaseloader/releasesync.go
```diff
@@ -463,6 +477,22 @@ func idFromURL(prowURL string) (uint, error) {
 	return uint(prowID), nil
 }
 
+// extractBuildIDFromURL extracts the build ID from a prow job URL
+// e.g., https://prow.ci.openshift.org/view/gs/test-platform-results/logs/periodic-ci-openshift-release-master-ci-4.16-e2e-gcp-ovn-upgrade/1234567890
+// returns "1234567890"
+func extractBuildIDFromURL(prowURL string) string {
```
neisw, line 491: This seems pretty testable

pkg/dataloader/releaseloader/labels.go
sosiouxme: mocking db calls is almost certainly more trouble than it's worth. i have a functional test locally to exercise the code, but for that you need creds. i suppose i could commit the test so it could be used manually (and skipped normally), but not sure how useful that would be? i guess somewhat useful.

i let claude write almost-duplicate code because there is something of a fundamental difference between the prow one-run approach versus the releaseLoader whole-payload approach. it's way too much refactoring to get the prow loader to operate on runs in bulk, but i guess it's only slightly awkward to have that single-build use case just call the multi-build code with a single build.

pkg/dataloader/releaseloader/releasesync.go
sosiouxme: not wrong, I had this objection myself. but claude pointed out that prowloader was already doing the same :)

I think I understand the spirit of this best practice. We don't want a context getting passed around / stored and then used where it doesn't really apply, and if you're not careful with object scope, that can happen.

At the same time, I think the point is that a context should be defined and relevant for a particular event that has a lifecycle and then goes away. I think if a struct is really an object which has a lifecycle, as a loader does - you define it, have it run some stuff that involves the context, and then it's finished - then passing around the context on every single method call (often just to pass it to the next method on the stack) is just clutter, the kind of clutter that an object is intended to hide in the first place.

All of which to say, I'm ok with violating a best practice here.

pkg/dataloader/releaseloader/releasesync_functional_test.go
```diff
@@ -0,0 +1,106 @@
...
+
+	t.Logf("Release %s has %d blocking jobs, %d informing jobs",
+		details.Name,
+		len(details.Results["blockingJobs"]),
+		len(details.Results["informingJobs"]),
+	)
+
+	// Run the function under test
+	loader := &ReleaseLoader{ctx: ctx, bqClient: bqClient}
+	jobRuns := loader.releaseJobRunsToDB(details, time.Now().Add(-48*time.Hour))
```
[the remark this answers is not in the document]
sosiouxme: no. just use a tag from the last 48 hours in your testing.

pkg/dataloader/releaseloader/releasesync.go
```diff
@@ -438,6 +426,36 @@ func releaseJobRunsToDB(details ReleaseDetails) []models.ReleaseJobRun {
...
+			}
+		}
+
+		buildIDs := make([]string, 0, len(buildIDToJobRun))
+		for buildID := range buildIDToJobRun {
+			buildIDs = append(buildIDs, buildID)
+		}
+
+		labelsByBuildID, err := prowloader.GatherLabelsFromBQ(r.ctx, r.bqClient, buildIDs, releaseTime)
+		if err != nil {
```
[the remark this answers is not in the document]
sosiouxme: that's not gonna happen, but it doesn't hurt to guard against some future weird release tag format that violates expectations.

PR #3568, opened 2026-06-02
Reduce Component Readiness updates from 6x/day to 2x/day

dgoodwin:
This is tuned for two updates around midnight and noon EST, technically
04:00 and 16:00 UTC. This will cut costs considerably, and matches the
expectation that QSE would look once or twice a day tops. This is not
our rapid regression detection mechanism, and most of the 4h updates
were probably going to waste.

Should merge with https://github.com/openshift/continuous-release-jobs/pull/1804

## Summary by CodeRabbit

* **New Features**
  * Added a new CLI flag to configure a component-readiness time-rounding offset.

* **Improvements**
  * Time-rounding now uses both a rounding factor and an offset, improving alignment of release time calculations, cache expiration boundaries, component-readiness reports, and related metrics and URLs.


smg247 (commented):
/lgtm

PR #3574, opened 2026-06-03
TRT-2660: Fix null explanations crash in regressed tests modal

openshift-trt:
## Summary
- Initialize `Explanations` as `[]string{}` instead of `nil` at both `TestComparison` creation sites (`component_report.go:616`, `test_details.go:444`) so Go serializes it as JSON `[]` instead of `null`.
- Add a defensive `null` check in `CompSeverityIcon.js` (`!= null` catches both `null` and `undefined`).

## Test plan
- [x] `make lint` passes
- [x] `make test` passes (14954 Go tests + JS tests)
- [ ] Open the regressed tests modal on the component readiness page and confirm it no longer crashes

## Bug Fixes
- Improved consistency in how explanation fields are initialized across component readiness reports and UI components, ensuring proper handling of empty values.

### DIFF
--- a/pkg/api/componentreadiness/component_report.go
+++ b/pkg/api/componentreadiness/component_report.go
@@ -613,7 +613,7 @@ func (c *ComponentReportGenerator) generateComponentTestReport(basisStatusMap, s
 	keySet := sets.NewString(slices.Collect(maps.Keys(basisStatusMap))...)
 	keySet.Insert(slices.Collect(maps.Keys(sampleStatusMap))...)
 	for testKeyStr := range keySet {
-		var cellReport testdetails.TestComparison // The actual stats we return over the API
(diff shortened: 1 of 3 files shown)


pkg/api/componentreadiness/component_report.go
```diff
@@ -613,7 +613,7 @@ func (c *ComponentReportGenerator) generateComponentTestReport(basisStatusMap, s
 	keySet := sets.NewString(slices.Collect(maps.Keys(basisStatusMap))...)
 	keySet.Insert(slices.Collect(maps.Keys(sampleStatusMap))...)
 	for testKeyStr := range keySet {
-		var cellReport testdetails.TestComparison // The actual stats we return over the API
+		cellReport := testdetails.TestComparison{Explanations: []string{}} // The actual stats we return over the API
```
smg247: can we remove the comment here? it is silly

PR #3588, opened 2026-06-08
TRT-2693: releasefallback panic

sosiouxme:
When requesting release dates for old releases, report releases without a GA date as an error instead of empty dates, and then handle those errors properly.

When requesting CR for a PR sample, don't compare against fallback releases.

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened validation of release date info to prevent processing when dates are missing.
  * Improved error collection and asynchronous handling during multi-release analysis to reduce lost errors and improve reliability.
  * Adjusted multi-release analysis gating so fallback processing only runs when appropriate options are present.

* **Documentation**
  * Fixed a typo in release-fallback documentation.

### DIFF
--- a/pkg/api/componentreadiness/middleware/releasefallback/releasefallback.go
+++ b/pkg/api/componentreadiness/middleware/releasefallback/releasefallback.go
@@ -46,7 +46,7 @@ func NewReleaseFallbackMiddleware(
 
 // ReleaseFallback middleware allows us to use the best pass rate data from the past
 // several releases for our basis instead of just the requested basis. This helps prevent
-// minor gradual degredation of quality, and also simplifies the process of accepting
+// minor gradual degradation of quality, and also simplifies the process of accepting
 // intentional regressions shortly before release, as we'll then automatically use the data
 // from prior releases.
 //
@@ -81,7 +81,6 @@ func (r *ReleaseFallback) Query(ctx context.Context, wg *sync.WaitGroup, allJobV
 			r.log.Infof("Context canceled while fetching fallback query status")
 			return
 		default:
-			// TODO: should we pass the same wg through rather than using another?
 			errs := r.getFallbackBaseQueryStatus(ctx, allJobVariants, r.reqOptions.BaseRelease.Name, r.reqOptions.BaseRelease.Start, r.reqOptions.BaseRelease.End)
 			if len(errs) > 0 {
 				for _, err := range errs {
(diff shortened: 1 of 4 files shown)


pkg/api/componentreadiness/middleware/releasefallback/releasefallback.go
```diff
@@ -226,7 +232,7 @@ func (r *ReleaseFallback) QueryTestDetails(ctx context.Context, wg *sync.WaitGro
 
 		start, end, err := utils.FindStartEndTimesForRelease(timeRanges, release)
 		if err != nil {
-			errCh <- err
+			errs = append(errs, err)
```
neisw: Are you expecting [releasefallback.go#L201](https://github.com/sosiouxme/sippy/blob/2097c9807864f8127f7dac5b16e81f0650f40483/pkg/api/componentreadiness/middleware/releasefallback/releasefallback.go#L201) to process this?  If it is nil at the time the closure is created will it process this list?

pkg/api/componentreadiness/middleware/releasefallback/releasefallback.go
neisw: does this still reference errs which is now bsErrs?

sosiouxme: oh yeah, good catch.

pkg/api/componentreadiness/middleware/releasefallback/releasefallback.go
sosiouxme: yes, and [yes](https://go.dev/play/p/o1hyyI4izxE). the closure captures the variable's scope in the environment, not whatever it points at when the closure is defined.

the problem was that we had two synchronous error handlers trying to push to a channel that nothing is listening to until this method returns. as soon as it had actual errors to report i saw it block. but i don't think it's safe to put the whole method call in a goroutine (it looks to me like it needs to complete before the rest of the goroutines in the caller scope do their work), and it leaves no other way to return an error... so it seemed safest to me to have one place that processes errors the right way.

it's admittedly a bit "subtle" as coderabbit says but... hard to think of a clearer way.

sosiouxme: actually.... i thought of a clearer way.

pkg/api/componentreadiness/utils/utils.go
```diff
@@ -333,3 +338,14 @@ func GenerateTestDetailsURL(
...
+// EnqueueAsync is a helper function to asynchronously enqueue something (like errors) on a channel from a synchronous context.
+func EnqueueAsync[T any](wg *sync.WaitGroup, channel chan T, things ...T) {
+	wg.Add(1)
+	go func() {
+		for _, thing := range things {
+			channel <- thing
+		}
+		wg.Done()
+	}()
+}
```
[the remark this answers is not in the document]
sosiouxme: no.

unit tests for something this simple are low bang for buck

and if you're calling it without wg or channel i don't know why we let you get your hands on the code

petr-muller (approved):
/hold

LGTM, holding for Forest who reviewed before, unhold as needed

sosiouxme:
I think Forrest is busy and one lgtm is enough :) 

/hold cancel

PR #3604, opened 2026-06-10
fix: base release times must not use TruncateAligned rounding

dgoodwin:
PR #3568 introduced 12h/4h TruncateAligned rounding (boundaries at
04:00/16:00 UTC) for sample release end times so that the cache
preloader and server agree on when the current window expires. However,
the change inadvertently applied the same rounding to base release
times, which should always use start-of-day (00:00:00 UTC) and
end-of-day (23:59:59 UTC).

The result was a cache key mismatch: the preloader stored base end as
e.g. 2026-06-10T04:00:00Z, but URLs in the frontend and HATEOAS links
on regressions expected 2026-06-09T23:59:59Z, causing every component
readiness view to be a cache miss.

Fix: pass rounding factor/offset of 0,0 for all "basis" release option
resolution in every caller of GetViewReleaseOptions and parseDateRange:
- pkg/api/componentreadiness/utils/queryparamparser.go
- pkg/api/componentreadiness/triage.go (HATEOAS link generation)
- pkg/dataloader/crcacheloader/crcacheloader.go
- pkg/dataloader/regressioncacheloader/regressioncacheloader.go (the
  loader actually invoked by --loader component-readiness-cache)
- pkg/sippyserver/server.go
- pkg/componentreadiness/jiraautomator/jiraautomator.go

Tests: update existing view-based tests to pass non-zero rounding and
verify base times remain start/end of day. Add TestHATEOASLinkCacheConsistency
which simulates the full preloader → HATEOAS link → browser click flow
and asserts the cache keys match, along with explicit assertions that
base start is 00:00 UTC and base end is 23:59:59 UTC.

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

## Summary by CodeRabbit

* **Bug Fixes**
  * Standardized how base-release timestamps are derived so queries and views use a consistent, predictable rounding behavior.
  * Ensured generated report URLs and view-derived requests produce matching cache keys.

* **Tests**
  * Added tests verifying cache-key consistency between HATEOAS links (URLs) and internal view-based cache preloading.

### DIFF
--- a/pkg/api/componentreadiness/queryparamparser_test.go
+++ b/pkg/api/componentreadiness/queryparamparser_test.go
@@ -1,6 +1,8 @@
 package componentreadiness
 
 import (
+	"encoding/json"
+	"fmt"
 	"net/http"
 	"net/url"
 	"strings"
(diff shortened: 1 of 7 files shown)


smg247 (commented):
/lgtm

PR #3606, opened 2026-06-10
TRT-1989: verify synthetic releases

neisw:
ProwLoader passes in [specified release](https://github.com/neisw/sippy/blob/main/cmd/sippy/load.go#L454) when present or falls back to known bq releases.  In the former case partitions are only created for the specified releases.

Synthetic releases are not checking the specified releases which leads to errors when attempting to create jobs for releases that are not configured.

Adding a verification check.

## Release Notes

* **Bug Fixes**
  * Improved validation of release configurations during job processing. The system now validates that release overrides reference existing, configured releases before processing jobs. Invalid release overrides trigger warnings and are skipped to prevent errors and maintain data integrity.


smg247 (commented):
/lgtm

PR #3614, opened 2026-06-11
Optimize test report matview refresh with pre-aggregation CTE

mstaeble:
## Summary

- Adds a `pre_agg` CTE to `testReportMatView` that groups `prow_job_run_tests` by `(prow_job_id, test_id, suite_id, prow_job_run_release)` before joining dimension tables, reducing the row count ~6.5x before the expensive sort and join stages
- Uses `SUM()::bigint` in the outer query instead of `COUNT()` to preserve the `bigint` column type contract
- Adds an upper-bound timestamp filter (`<= |||END|||`) in the pre_agg CTE to enable future partition pruning
- Removes the `ORDER BY` from the matview definition (matview physical order guarantees nothing to readers, and it forces a needless sort on every refresh)

Both the 7d and 2d matviews use the same template, so this fixes both.

**Note:** [TRT-1989](https://redhat.atlassian.net/browse/TRT-1989) (6d2aa7777) already landed the `JOIN prow_job_runs` removal and `created_at` filter removal on main. After the merge, the net diff of this PR is only the `pre_agg` CTE, the `::bigint` casts, the upper-bound filter, and the `ORDER BY` removal.

## Benchmarks

The benchmark numbers below measure the pre-aggregation CTE against the **original production query** (pre-TRT-1989), not against current main. The incremental gain of the pre-agg CTE over the [TRT-1989](https://redhat.atlassian.net/browse/TRT-1989) changes was not isolated separately, but the pre-agg is what shrinks the sort input ~6.5x, which is the dominant cost.

### Production database (1 billion rows in `prow_job_run_tests`)

| Query | Time |
|---|---|
| Original matview refresh (from production logs) | 25-35 min |
| Pre-agg CTE query (EXPLAIN ANALYZE on production) | **6 min 11 sec** |

### Staging replica

| Query | Time |
|---|---|
| Original query (all releases) | >60 min (killed after 1 hour) |
| Pre-agg CTE (all releases) | ~6 min |
| Pre-agg CTE (single release, 4.20) | ~33 sec |

## How it works

The original query joins all ~60M rows (14-day window) from `prow_job_run_tests` to dimension tables (`tests`, `suites`, `test_ownerships`, `jira_components`, `prow_jobs`), then aggregates. This forces PostgreSQL to sort 60M wide tuples, spilling gigabytes to disk.

The optimized query adds a `pre_agg` CTE that first groups by `(prow_job_id, test_id, suite_id, prow_job_run_release)` — all columns on `prow_job_run_tests` itself, no joins needed. This collapses 60M rows to ~10M. The outer query then joins 10M pre-aggregated rows to dimension tables using `SUM` instead of `COUNT`. This works because `COUNT` across partitions followed by `SUM` equals `COUNT` across the whole set.

## Correctness

Verified via `EXCEPT` comparison (both directions) producing zero differences across 661,266 matview rows on:
- 9.5M-row prod-like backup database
- Staging replica of production

Also independently verified by @stbenjam with a synthetic dataset stressing edge cases (ownership fan-out, NULL suite_id, boundary timestamps, multiple releases/partitions, non-standard statuses, shared variant arrays).

## Deploy note

The schema hash change will cause both matviews to be dropped and recreated `WITH NO DATA` on rollout. Queries against them will return no data until the first refresh completes (~6 minutes).

## Test plan

- [x] `gofmt` clean
- [x] `make test` passes (Go + Jest)
- [x] Correctness verified on prod-like database (zero differences)
- [x] Correctness verified on staging replica (zero differences)
- [x] Benchmarked on staging replica
- [x] Benchmarked on production database

Verification and benchmark SQL scripts: https://gist.github.com/mstaeble/e04d52210b1857d58950d0df63479b2b

Relates: [TRT-2676](https://issues.redhat.com/browse/TRT-2676)

## Summary by CodeRabbit

* **Refactor**
  * Optimized test report data aggregation logic to improve performance and maintainability of internal metrics computation.


not-stbenjam:
I checked out this PR and independently verified the query transformation — overall it's correct and a nice win, with a few notes below.

## Correctness — independently verified ✅

The pre-aggregation is sound for a structural reason: `pre_agg`'s group key (`prow_job_id, test_id, suite_id, release`) is a strict refinement of the outer GROUP BY's columns, so every outer group is a union of whole `pre_agg` groups, and any dimension-join fan-out (e.g. two `test_ownerships` rows for the same test+suite) multiplies pre-aggregated rows exactly the way it multiplied raw rows before.

I also verified empirically on a freshly migrated schema with a synthetic dataset stressing the edge cases: ownership fan-out to two jira components, NULL `suite_id`, a row timestamped exactly at BOUNDARY (counted in both windows by the inclusive BETWEENs — pre-existing quirk, faithfully preserved), two releases across two partitions, rows before START and after END, non-1/12/13 statuses, and two jobs sharing the same variants array. Old vs. new query compared with `EXCEPT ALL` in both directions: zero differences.

## PR description is stale on two of three claims

This commit was written against pre-TRT-1989 code. [TRT-1989](https://redhat.atlassian.net/browse/TRT-1989) (6d2aa7777) landed on main first and already removed the `JOIN prow_job_runs` and the `created_at` filter. After the merge of main, the net diff of this PR is only the `pre_agg` CTE. That also means the benchmark numbers (25–35 min → ~6 min) measure the combined effect of denormalization + pre-aggregation against the old production query — the incremental gain over current main wasn't isolated. Likely still substantial (the pre-agg is what shrinks the sort input ~6.5x), but worth re-benchmarking against current main's query if precision matters.

## One behavioral change the EXCEPT verification can't catch

`COUNT(*)` returns `bigint`, but `SUM(bigint)` returns `numeric` — so the 8 count columns in both matviews change type from `bigint` to `numeric`. I traced all the Go consumers and nothing breaks (they all re-aggregate or ratio these columns, and the codebase already scans `SUM(...)` numerics into `int` fields today). Still, I'd suggest `SUM(...)::bigint AS ...` casts: it preserves the schema contract for ad-hoc/external consumers, and the per-request API queries that `SUM()` these columns over the matview stay on the faster int8 aggregation path.

## CodeRabbit's SQL injection finding is a false positive

The flagged `reportEndFmt` concatenation isn't part of this PR's diff, and it isn't injectable: the value is a `time.Time` parsed via RFC3339 at flag-parse time and formatted with a fixed layout, so the output can only contain digits, `-`, `:`, and a space. The suggested parameterization also can't work — it's `CREATE MATERIALIZED VIEW` DDL, which PostgreSQL can't bind parameters into, and the literal must be baked into the view definition for pinned-date refreshes to work.

## Optional improvements to consider bundling

1. **Drop the trailing `ORDER BY` from the matview definition.** It forces a sort of the full result (~660k rows) on every refresh, and a matview's physical order guarantees nothing to readers — every consumer query does its own ordering. Since this PR already changes the schema hash (triggering a drop/recreate anyway), bundling it is free.
2. **Add `AND prow_job_run_timestamp <= |||END|||` to the `pre_agg` WHERE.** Practically a no-op since END is `NOW()` at refresh time, but it lets the planner prune pre-created future timestamp sub-partitions.
3. **Ops experiment (not a code change):** `pre_agg` groups by `prow_job_run_release`, the LIST partition key — `SET enable_partitionwise_aggregate = on` in the refresh session could let Postgres aggregate per-partition and shrink the sort further.
4. **Deploy note:** the hash change means both matviews get dropped and recreated `WITH NO DATA` on rollout, so queries against them fail until the first refresh completes. Existing behavior for any definition change, but with a ~6-minute refresh it's worth timing the deploy accordingly.

stbenjam:
/lgtm
/hold

In case @neisw wants to look

mstaeble:
> I checked out this PR and independently verified the query transformation — overall it's correct and a nice win, with a few notes below.
> <snip\>

Thanks for the thorough review and independent verification!

Addressed in 332faf57c:

  1. **`::bigint` casts** — added `SUM(...)::bigint` on all 8 count columns to preserve the schema contract.
  2. **Upper-bound filter** — added `AND prow_job_run_timestamp <= |||END|||` to the `pre_agg` WHERE clause.
  3. **Removed `ORDER BY`** — dropped from the matview definition.

Also updated the PR description to clarify that the benchmarks measure pre-agg CTE against the original pre-TRT-1989 query (not current main), and added the deploy note about the matview being empty until first refresh.

Re: `enable_partitionwise_aggregate` — worth experimenting with on staging. I'll follow up on that separately.

neisw:
I'm good if @mstaeble wants it to merge tonight he is free to `unhold`

mstaeble:
> I'm good if @mstaeble wants it to merge tonight he is free to `unhold`

I already pushed a new version to address some of Not-Stephen's comments. We can wait to merge.

PR #3616, opened 2026-06-12
TRT-2703: update jobRunsReportMatView post partitioning

neisw:
Updates matview to be aware of release and prow_job_run_timestamps
Adds fallback to prow_job_runs.test_failures if we are missing failed_test_results.test_count
Adds option to enable partitionwise planning optimizations

## Summary by CodeRabbit

* **New Features**
  * Added a configurable option/flag (`--enable-partitionwise`) to enable PostgreSQL partitionwise aggregate/join optimizations.
* **Improvements**
  * Updated the job-runs report to consolidate failed and flaked results, restrict processing to the last 90 days, and improve grouping/joins for more accurate counts.
* **Tests**
  * Added benchmarks covering materialized-view creation/indexing/refresh and overall data-refresh performance.

### DIFF
--- a/pkg/db/db.go
+++ b/pkg/db/db.go
@@ -52,7 +52,23 @@ func (w log2LogrusWriter) Printf(msg string, args ...any) {
 	w.entry.Debugf(msg, args...)
 }
 
-func New(dsn string, logLevel gormlogger.LogLevel) (*DB, error) {
+type Option func(*options)
+
+type options struct {
+	enablePartitionwise bool
+}
+
+func WithPartitionwise(enable bool) Option {
+	return func(o *options) {
+		o.enablePartitionwise = enable
+	}
+}
+
+func New(dsn string, logLevel gormlogger.LogLevel, opts ...Option) (*DB, error) {
+	var cfg options
+	for _, o := range opts {
+		o(&cfg)
+	}
 	gormLogger := gormlogger.New(
 		log2LogrusWriter{entry: log.WithField("source", "gorm")},
 		gormlogger.Config{
@@ -73,6 +89,10 @@ func New(dsn string, logLevel gormlogger.LogLevel) (*DB, error) {
 	// partitions. Custom plans use actual parameter values for partition pruning.
 	pgxConfig.RuntimeParams["plan_cache_mode"] = "force_custom_plan"
 	pgxConfig.RuntimeParams["work_mem"] = "128MB"
+	if cfg.enablePartitionwise {
+		pgxConfig.RuntimeParams["enable_partitionwise_aggregate"] = "on"
+		pgxConfig.RuntimeParams["enable_partitionwise_join"] = "on"
+	}
 
 	connPool := stdlib.OpenDB(*pgxConfig)
 
--- a/pkg/db/views.go
+++ b/pkg/db/views.go
@@ -187,6 +187,7 @@ pull_requests AS (
                 prow_job_run_prow_pull_requests ON prow_job_run_prow_pull_requests.prow_pull_request_id = prow_pull_requests.id
         INNER JOIN
                 prow_job_runs ON prow_job_run_prow_pull_requests.prow_job_run_id = prow_job_runs.id
+        WHERE prow_job_runs."timestamp" >= CURRENT_TIMESTAMP - interval '90 days'
         GROUP BY prow_job_runs.id, prow_pull_requests.link, prow_pull_requests.sha, prow_pull_requests.org, prow_pull_requests.repo, prow_pull_requests.author
 )
 SELECT prow_job_runs.id,
@@ -208,7 +209,7 @@ SELECT prow_job_runs.id,
    test_results.flaked_test_names AS flaked_test_names,
    test_results.flaked_test_count AS test_flakes,
    test_results.failed_test_names AS failed_test_names,
-   COALESCE(test_results.failed_test_count, prow_job_runs.test_failures) AS test_failures,
+   test_results.failed_test_count AS test_failures,
    pull_requests.link as pull_request_link,
    pull_requests.sha as pull_request_sha,
    pull_requests.org as pull_request_org,
--- a/pkg/db/views.go
+++ b/pkg/db/views.go
@@ -156,22 +156,26 @@ func syncPostgresViews(db *gorm.DB, reportEnd *time.Time) error {
 const jobRunsReportMatView = `
 WITH failed_test_results AS (
 	SELECT prow_job_run_tests.prow_job_run_id,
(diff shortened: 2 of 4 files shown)


neisw:
@coderabbitai - In general reviews of postgres_benchmarking_test.go are not necessary as these tests are not run by default and limited to explicit functional benchmark comparisons.  But the db error logging was a valid concern and has been addressed.

pkg/db/db.go
```diff
@@ -52,7 +52,23 @@ func (w log2LogrusWriter) Printf(msg string, args ...any) {
...
+func WithPartitionwise(enable bool) Option {
+	return func(o *options) {
+		o.enablePartitionwise = enable
+	}
+}
+
+func New(dsn string, logLevel gormlogger.LogLevel, opts ...Option) (*DB, error) {
+	var cfg options
+	for _, o := range opts {
+		o(&cfg)
```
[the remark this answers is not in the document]
neisw: The code does not have a path that allows for nil to be passed in.

pkg/flags/postgres_benchmarking_test.go
```diff
@@ -1088,6 +1095,145 @@ func Test_BenchmarkSingleReleaseMatview(t *testing.T) {
...
+			return nil
+		},
+	}, iterations))
+
+	indexName := fmt.Sprintf("idx_%s", matviewName)
+	indexCols := strings.Join(source.IndexColumns, ", ")
+	results = append(results, runBenchmarkCase(t, dbc, benchmarkCase{
+		name: "CreateIndex",
+		fn: func(dbc *db.DB) error {
+			dbc.DB.Exec(fmt.Sprintf("DROP INDEX IF EXISTS %s", indexName))
```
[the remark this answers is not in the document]
neisw: The following line creates the index again and captures the error if any.  This should be acceptable for the benchmarking cases.

neisw:
@coderabbitai can you do a followup review and update your responses based on the current state.  Have the ` DB connection errors` logging issues been resolved?

pkg/db/views.go
```diff
@@ -206,20 +205,21 @@ SELECT prow_job_runs.id,
...
    prow_job_runs.cluster AS cluster,
    prow_job_runs.labels as labels,
-   flaked_test_results.test_names AS flaked_test_names,
-   flaked_test_results.test_count AS test_flakes,
-   failed_test_results.test_names AS failed_test_names,
-   failed_test_results.test_count AS test_failures,
+   test_results.flaked_test_names AS flaked_test_names,
+   test_results.flaked_test_count AS test_flakes,
+   test_results.failed_test_names AS failed_test_names,
+   COALESCE(test_results.failed_test_count, prow_job_runs.test_failures) AS test_failures,
```
mstaeble: Can you help me to understand the purpose of this COALESCE? Are we trying to avoid a NULL in the scenario where there are no failed tests in the past 90 days? Or are we trying to use `prow_job_runs.test_failures` as a fallback in the case of data quality issues?

pkg/db/views.go
mstaeble, line 187: Have you looked into whether there would be improvements from adding the 90-day time constraint here, to allow the seq scan to eliminate rows that could not possibly be contenders?

pkg/flags/postgres_benchmarking_test.go
```diff
@@ -499,23 +500,29 @@ func getBenchmarkDBClient(t *testing.T) (*db.DB, string) {
...
-		DSN:      dsn,
+		LogLevel:            4,
+		DSN:                 dsn,
+		EnablePartitionwise: os.Getenv("enable_partitionwise") != "",
 	}
 
 	dbc, err := dbFlags.GetDBClient()
 	if err != nil {
-		t.Fatalf("couldn't get DB client: %v", err)
+		t.Fatal("couldn't get DB client")
```
mstaeble, line 510: What is the rationale behind removing the error from the log event?

neisw: This came from the concern that dsn might be leaked via the error.  I am open to err on the side of caution though not overly concerned by the previous fatal with error.

pkg/db/views.go
neisw: I see it as fallback.  If we don't have a value in our our query result then fallback to prow_job_runs value.  The main issue in the bug was a lack of explanation as to why the data wasn't present in the matview when we did have the data in the tables.

pkg/db/views.go
neisw: No but we want to prevent any full scans going forward as we look to bring in more historical data so I will look into that.

pkg/db/views.go
mstaeble: How does the value in `prow_job_runs.test_failures` compare with the value in `test_results.failed_test_count` in the happy scenario where the data is present? If the data is the same, can we always use `prow_job_runs.test_failures` instead? If the data is not the same, what are the implications for the consumer of the data if they get one versus the other?

Moreover, I am concerned that we are putting a bandaid in for an issue that we are fixing via other means. If an issue of this nature happens again in the future for another reason, will we lose visibility into their even being an issue by serving data from the other table? In other words, if this COALESCE had been present already, would we have even known that there was a problem with the matview that needed to be fixed?

neisw: Completely valid concerns.  At the moment I have removed the COALESCE.  I would expect both values to be the same and we likely need to review the matview definition entirely.

It looks to me as though the production values are updating properly again.  So these changes are now focused on adding optional support for `partitionwise` settings and a refactor to consolidate `prow_job_run_tests` scans and add timestamp range limits.

pkg/db/views.go
neisw: Done

pkg/db/views.go
```diff
@@ -206,20 +206,21 @@ SELECT prow_job_runs.id,
...
    prow_job_runs.cluster AS cluster,
    prow_job_runs.labels as labels,
-   flaked_test_results.test_names AS flaked_test_names,
-   flaked_test_results.test_count AS test_flakes,
-   failed_test_results.test_names AS failed_test_names,
-   failed_test_results.test_count AS test_failures,
+   test_results.flaked_test_names AS flaked_test_names,
+   test_results.flaked_test_count AS test_flakes,
+   test_results.failed_test_names AS failed_test_names,
+   test_results.failed_test_count AS test_failures,
```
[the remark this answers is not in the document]
neisw: With the 90-day filter now on both the CTE and the outer prow_job_runs query, the data windows are aligned — there's no scenario where a run exists in the outer query but its test results were excluded by a different time bound. When a run genuinely has no failed tests, NULL → 0 via GORM would handle the case anyway.

PR #3626, opened 2026-06-15
Add collapsed matview for default test results API path

mstaeble:
## Summary
- Adds a pre-collapsed materialized view (`prow_test_report_7d_collapsed_matview` and 2d variant) that pre-aggregates test results by `(suite_name, name, jira_component, jira_component_id, release)`, collapsing all variant combinations
- The `/api/tests` endpoint with `collapse=true` (the default "All Tests" page) reads from this matview directly instead of doing SUM + GROUP BY on the full per-variant matview
- Adds `RefreshPhase` field to `PostgresView` to control matview refresh ordering — all matviews in phase 0 refresh concurrently, then phase 1, etc.
- Moves `prow_test_report_2d_matview` to phase 1 (previously handled by sort-based ordering) to avoid CPU overload from refreshing concurrently with the 7d matview
- Collapsed matviews refresh in phase 2 after the base matviews they read from

## Benchmarks (staging, warm)

| Release | Rows | Before | After | Speedup |
|---------|------|--------|-------|---------|
| 4.18 | 290K → 7K | 2,115ms | 6ms | **360x** |
| 5.0 | 911K → 27K | 5,162ms | 24ms | **211x** |

Collapsed matview refresh cost: **~36s** (reads from already-refreshed base matview)

## Details

The base matview has ~5M rows (one per test × variant-combo × release). The collapsed matview reduces this to ~148K rows by summing counts across all variant combinations per test. This is the same aggregation the API currently does at query time on every request.

The `RefreshPhase` field replaces the previous sort-based ordering that ensured the 2d matview didn't refresh concurrently with the 7d matview. Matviews are sorted by phase and processed in a single pass — each phase's matviews refresh concurrently, then the next phase begins:
- **Phase 0**: 7d matview + other independent matviews
- **Phase 1**: 2d matview (avoids CPU overload with 7d)
- **Phase 2**: Collapsed matviews (read from base matviews completed in phases 0 and 1)

## Test plan
- [x] `go vet ./...` passes
- [x] `make test` passes (Go + Jest)
- [x] Row counts verified on staging (7,227 rows match for 4.18)
- [x] Spot-checked values match between current GROUP BY and collapsed matview
- [ ] Deploy to staging and verify All Tests page loads correctly

### DIFF
--- a/pkg/db/views.go
+++ b/pkg/db/views.go
@@ -2,6 +2,7 @@ package db
 
 import (
 	"fmt"
+	"slices"
 	"strings"
 	"time"
 
@@ -27,6 +28,7 @@ var PostgresMatViews = []PostgresView{
 		Name:         "prow_test_report_2d_matview",
 		Definition:   testReportMatView,
 		IndexColumns: []string{"release", "name", "id", "variants", "suite_name"},
+		RefreshPhase: 1, // avoid CPU overload from refreshing concurrently with the 7d matview
 		ReplaceStrings: map[string]string{
 			"|||START|||":    "|||TIMENOW||| - INTERVAL '9 DAY'",
 			"|||BOUNDARY|||": "|||TIMENOW||| - INTERVAL '2 DAY'",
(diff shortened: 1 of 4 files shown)


pkg/db/views.go
```diff
@@ -286,6 +305,21 @@ GROUP BY
...
+SELECT suite_name, name, jira_component, jira_component_id, release,
+    SUM(current_runs)::bigint AS current_runs,
+    SUM(current_successes)::bigint AS current_successes,
+    SUM(current_failures)::bigint AS current_failures,
+    SUM(current_flakes)::bigint AS current_flakes,
+    SUM(previous_runs)::bigint AS previous_runs,
+    SUM(previous_successes)::bigint AS previous_successes,
+    SUM(previous_failures)::bigint AS previous_failures,
+    SUM(previous_flakes)::bigint AS previous_flakes,
+    (array_agg(open_bugs))[1] AS open_bugs
```
mstaeble, line 344: `open_bugs` is a per-test count (not per-variant), so every row for the same test has the same value. We can't include it in the GROUP BY (it's not a grouping dimension) and SUM would over-count. `array_agg(...)[1]` picks the first value, which is the same as any other. This matches the pattern in [QueryTestSummer](https://github.com/openshift/sippy/blob/0963812d94cce075a9bb8fd30c6a1a9c2099d927/pkg/db/query/test_queries.go#L33).

mstaeble:
@coderabbitai review

PR #3632, opened 2026-06-16
Apply name filters via tests table lookup instead of matview ILIKE

mstaeble:
## Summary
- Move `name` filters (contains, startsWith, etc.) from running ILIKE directly on the matview (~700K+ rows per release) to running against the small `tests` table (~36K rows with a btree index on `name`)
- The matching test IDs are then used to filter the matview via an `IN` subquery, which uses the indexed `id` column
- This benefits both `collapse=true` (All Tests) and `collapse=false` (Tests By Variant) paths, as well as the risk analysis code path

## Benchmarks (staging, warm, release 4.21 with default filters)

| Path | Before | After | Speedup |
|------|--------|-------|---------|
| collapse=true (All Tests) | 27,753ms | 5,347ms | **5.2x** |

The improvement comes from running ILIKE against 36K rows instead of 700K+. The tests table has a unique btree index on `name` which makes pattern matching efficient.

## Why this works

The matview's `id` column matches `tests.id`. Instead of:
```sql
WHERE name NOT ILIKE '%.Overall%' -- scans 700K matview rows
```

We generate:
```sql
WHERE id IN (SELECT id FROM tests WHERE name NOT ILIKE '%.Overall%') -- scans 36K test rows, then indexed ID lookup
```

Variant filters (`has entry`) continue to be applied directly on the matview since they use the array `variants` column which only exists there.

## Test plan
- [x] `go vet ./...` passes
- [x] `make test` passes
- [x] Row counts verified identical (10,496 rows for both approaches on release 4.21)
- [x] Benchmarked on staging with full default filter set
- [x] Tested locally with sippy frontend against staging database


mstaeble:
Verified by running sippy locally connected to the staging database.

Response time of calls to /api/tests endpoint on the /tests page runs ~6-7s when loading 5.0 and 4.22 data.

PR #3643, opened 2026-06-17
TRT-2695: retroactively re-evaluate symptoms (API)

sosiouxme:
Currently just the backend API (frontend will come later).

## Release Notes

* **New Features**
  * Added `POST /api/jobs/runs/reevaluate` to retroactively re-evaluate completed job run symptoms, with `dry_run` support, request validation (max 50 job IDs), and updates across BigQuery, GCS, and PostgreSQL while preserving manually applied labels.

* **Documentation**
  * Documented the new endpoint in the API and symptoms feature docs, and published the backend implementation plan.

* **Tests**
  * Added unit tests plus gated end-to-end functional tests (including idempotency checks).

* **Chores**
  * Refreshed local artifact hash references and updated internal guidance.

### DIFF
--- a/pkg/api/jobrunscan/reevaluate.go
+++ b/pkg/api/jobrunscan/reevaluate.go
@@ -363,7 +363,7 @@ func (r *ReEvaluator) buildOutputs(matches []symptomMatch, buildID string, jobRu
 
 // loadLabelDefinitions fetches label content for the labels referenced by matched symptoms.
 func (r *ReEvaluator) loadLabelDefinitions(matches []symptomMatch) (map[string]jobrunscan.LabelContent, error) {
-	ids := sets.NewString()
+	ids := sets.New[string]()
 	for _, m := range matches {
 		ids.Insert(m.symptom.LabelIDs...)
 	}
@@ -372,8 +372,8 @@ func (r *ReEvaluator) loadLabelDefinitions(matches []symptomMatch) (map[string]j
 	}
 
 	var labels []jobrunscan.Label
-	if err := r.db.DB.Where("id IN ?", ids.List()).Find(&labels).Error; err != nil {
-		return nil, fmt.Errorf("querying label definitions for %v: %w", ids.List(), err)
+	if err := r.db.DB.Where("id IN ?", ids.UnsortedList()).Find(&labels).Error; err != nil {
+		return nil, fmt.Errorf("querying label definitions for %v: %w", ids.UnsortedList(), err)
 	}
 
 	result := make(map[string]jobrunscan.LabelContent, len(labels))
@@ -551,14 +551,14 @@ func (r *ReEvaluator) queryNonSymptomLabels(ctx context.Context, buildID string,
 
 // mergeLabels combines manual labels with newly applied symptom labels, deduplicating.
 func mergeLabels(manualLabels []string, bqLabels []models.JobRunLabel) []string {
-	merged := sets.NewString()
+	merged := sets.New[string]()
 	for _, l := range manualLabels {
 		merged.Insert(l)
 	}
 	for _, bl := range bqLabels {
 		merged.Insert(bl.Label)
 	}
-	return merged.List()
+	return merged.UnsortedList()
 }
 
 // jobRunPathFromURL extracts the GCS path from a ProwJobRun URL.
@@ -577,20 +577,20 @@ func jobRunPathFromURL(url string) string {
 
 // uniqueSymptomsMatched returns the sorted distinct symptom IDs from the matches.
 func uniqueSymptomsMatched(matches []symptomMatch) []string {
-	s := sets.NewString()
+	s := sets.New[string]()
 	for _, m := range matches {
 		s.Insert(m.symptom.ID)
 	}
-	return s.List()
+	return s.UnsortedList()
 }
 
 // uniqueLabels returns the unique label IDs from a set of BQ labels.
 func uniqueLabels(labels []models.JobRunLabel) []string {
-	s := sets.NewString()
+	s := sets.New[string]()
 	for _, l := range labels {
 		s.Insert(l.Label)
 	}
-	return s.List()
+	return s.UnsortedList()
 }
 
 // ValidateReEvalRequest validates the re-evaluation request parameters.
--- a/pkg/api/jobrunscan/reevaluate.go
+++ b/pkg/api/jobrunscan/reevaluate.go
@@ -123,7 +123,7 @@ func (r *ReEvaluator) ReEvaluateJobRuns(ctx context.Context, prowJobBuildIDs []s
 	if err != nil {
 		return nil, fmt.Errorf("loading symptoms: %w", err)
 	}
-	log.Debugf("symptom reEval: loaded %d active symptoms", len(symptoms))
+	log.WithField("activeSymptoms", len(symptoms)).Debug("symptom reEval: loaded active symptoms")
 
 	results := make([]ReEvaluationResult, 0, len(prowJobBuildIDs))
 	for _, buildID := range prowJobBuildIDs {
@@ -154,9 +154,9 @@ func filterRelevantSymptoms(symptoms []jobrunscan.Symptom) []jobrunscan.Symptom 
 		case jobrunscan.MatcherTypeString, jobrunscan.MatcherTypeRegex, jobrunscan.MatcherTypeFile:
 			filtered = append(filtered, s)
 		case jobrunscan.MatcherTypeCEL:
-			log.Warnf("symptom reEval: skipping symptom %q with unimplemented matcher_type %q (TRT-2466)", s.ID, s.MatcherType)
+			log.WithFields(log.Fields{"symptom": s.ID, "matcherType": s.MatcherType}).Warn("symptom reEval: skipping symptom with unimplemented matcher_type (TRT-2466)")
 		default:
-			log.Warnf("symptom reEval: skipping symptom %q with unknown matcher_type %q", s.ID, s.MatcherType)
+			log.WithFields(log.Fields{"symptom": s.ID, "matcherType": s.MatcherType}).Warn("symptom reEval: skipping symptom with unknown matcher_type")
 		}
 	}
 	return filtered
@@ -213,8 +213,12 @@ func (r *ReEvaluator) reEvaluateOne(ctx context.Context, buildID string, symptom
 	result.LabelsApplied = uniqueLabels(bqLabels)
 
 	if r.dryRun {
-		log.Infof("symptom re-evaluation dry run for %s: %d symptoms matched, %d BQ labels, %d GCS artifacts",
-			buildID, len(result.SymptomsMatched), len(bqLabels), len(bucketLabels))
+		log.WithFields(log.Fields{
+			"buildID":      buildID,
+			"matched":      len(result.SymptomsMatched),
+			"bqLabels":     len(bqLabels),
+			"gcsArtifacts": len(bucketLabels),
+		}).Info("symptom re-evaluation dry run")
 		result.Status = ReEvalSuccess
 		return result
 	}
@@ -249,7 +253,7 @@ func (r *ReEvaluator) evaluateSymptoms(ctx context.Context, jobRunID int64, symp
 	for _, symptom := range symptoms {
 		contentMatcher, err := ContentMatcherForSymptom(symptom.SymptomContent)
 		if err != nil {
-			log.WithError(err).Warnf("symptom reEval: skipping symptom %q due to matcher error", symptom.ID)
+			log.WithError(err).WithField("symptom", symptom.ID).Warn("symptom reEval: skipping symptom due to matcher error")
 			continue
 		}
 
@@ -443,12 +447,12 @@ func (r *ReEvaluator) clearBQLabels(ctx context.Context, buildID string, startTi
 	_, err := q.Read(ctx)
 	if err != nil {
 		if strings.Contains(err.Error(), "streaming buffer") {
-			log.Warnf("symptom reEval: BQ delete for %s hit streaming buffer, skipping: %v", buildID, err)
+			log.WithError(err).WithField("buildID", buildID).Warn("symptom reEval: BQ delete hit streaming buffer, skipping")
 			return nil
 		}
 		return fmt.Errorf("BQ delete for %s: %w", buildID, err)
 	}
-	log.Debugf("symptom reEval: cleared BQ symptom labels for build %s", buildID)
+	log.WithField("buildID", buildID).Debug("symptom reEval: cleared BQ symptom labels")
 	return nil
 }
 
@@ -473,7 +477,7 @@ func (r *ReEvaluator) clearGCSLabels(ctx context.Context, jobRunPath string) err
 			return fmt.Errorf("deleting GCS object %s: %w", attrs.Name, err)
 		}
 	}
-	log.Debugf("symptom reEval: cleared GCS label artifacts for %s", jobRunPath)
+	log.WithField("jobRunPath", jobRunPath).Debug("symptom reEval: cleared GCS label artifacts")
 	return nil
 }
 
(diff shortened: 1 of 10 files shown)


pkg/api/jobrunscan/reevaluate.go
```diff
@@ -0,0 +1,603 @@
...
+// clearBQLabels deletes existing symptom-originated labels from BQ for the given build ID.
+func (r *ReEvaluator) clearBQLabels(ctx context.Context, buildID string, startTime time.Time) error {
+	table := fmt.Sprintf("`%s.job_labels`", r.bqClient.Dataset)
+	queryStr := fmt.Sprintf(
+		`DELETE FROM %s
+		WHERE prowjob_build_id = @buildID
+		AND symptom_id IS NOT NULL AND symptom_id != ''
+		AND DATE(prowjob_start) = @startDate
+		AND (created_at IS NULL OR created_at < DATETIME_SUB(CURRENT_DATETIME(), INTERVAL 90 MINUTE))`,
+		table)
```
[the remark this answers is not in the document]
sosiouxme: true, this makes it impossible to have an airtight test for idempotency within 90 minutes, so i'll either need to drop that functional test, or i can relax the expectations around deleting. if the delete fails due to stream buffering, i think i'm ok with ignoring that and proceeding to add labels (there will just be duplicates)

.apm/instructions/backend.instructions.md
```diff
@@ -5,5 +5,6 @@ applyTo: "**/*.go"
 
 * When adding or updating APIs, **use HATEOAS** in responses to support discoverability and consistent client interaction.
 * Follow idiomatic Go practices.
+* Use `k8s.io/apimachinery/pkg/util/sets` (e.g. `sets.NewString()`) to deduplicate or collect unique strings. Do not use `map[string]bool` as a hand-rolled set.
```
petr-muller:
```suggestion
* Use `k8s.io/apimachinery/pkg/util/sets` (e.g. `sets.New[string]()`) to deduplicate or collect unique strings. Do not use `map[string]bool` as a hand-rolled set.
```

mega-nit, `NewString` is deprecated

.apm/instructions/docs.instructions.md
```diff
@@ -18,5 +18,6 @@ applyTo: "**"
   purpose, data flow, and key code locations; keep them accurate as
   the implementation evolves. If you add a new major feature, create a
   new doc in `docs/features/`.
-* Documentation and code belong in the same PR — never treat a docs update
+* Documentation and code belong in the same PR; never treat a docs update
   as a follow-up task.
+* Do not use em dashes when writing docs. Use commas, parentheses, or periods instead.
```
petr-muller: 😁

.claude/commands/agentic-followup.md
```diff
@@ -0,0 +1,69 @@
+---
+description: 'Address PR review comments for a Jira issue: find PR, fix issues, reply
+  to reviewers, push'
```
petr-muller, line 3: Looks unrelated to the PR, is this generated content from from something that landed earlier and did not regenerate the content? We may need a presubmit that checks that.

pkg/sippyserver/job_run_scan.go
```diff
@@ -170,3 +170,42 @@ func (s *Server) jsonDeleteSymptom(w http.ResponseWriter, req *http.Request) {
 	}
 	w.WriteHeader(http.StatusNoContent)
 }
+
+// Job run symptom re-evaluation handler
+
+func (s *Server) jsonReEvaluateJobRunSymptoms(w http.ResponseWriter, req *http.Request) {
+	user := getUserForRequest(req)
+	log.Infof("symptom re-evaluation POST made by user: %s", user)
```
petr-muller: not new code, the pattern is used elsewhere but I think user should be logged as a field, not inline

pkg/api/jobrunscan/reevaluate.go
```diff
@@ -0,0 +1,608 @@
...
+		if len(s.LabelIDs) == 0 {
+			continue // no value in matching symptoms that don't apply labels
+		}
+		switch s.MatcherType {
+		case jobrunscan.MatcherTypeString, jobrunscan.MatcherTypeRegex, jobrunscan.MatcherTypeFile:
+			filtered = append(filtered, s)
+		case jobrunscan.MatcherTypeCEL:
+			log.Warnf("symptom reEval: skipping symptom %q with unimplemented matcher_type %q (TRT-2466)", s.ID, s.MatcherType)
+		default:
+			log.Warnf("symptom reEval: skipping symptom %q with unknown matcher_type %q", s.ID, s.MatcherType)
```
petr-muller: nit: log as structured values, not inline?

pkg/api/jobrunscan/reevaluate.go
```diff
@@ -0,0 +1,608 @@
...
+	"github.com/openshift/sippy/pkg/db/models/jobrunscan"
+	"github.com/openshift/sippy/pkg/util"
+	log "github.com/sirupsen/logrus"
+	"google.golang.org/api/iterator"
+	"gorm.io/gorm"
+)
+
+const (
+	reEvalSourceTool = "sippy-api-reevaluate"
+	maxJobRunsPerReq = 50
```
petr-muller, line 34: so that's `50 * symptom count * artifact count` scans (right?) in a synchronous POST request. is that realistic to complete in some reasonable time?

pkg/api/jobrunscan/reevaluate.go
```diff
@@ -0,0 +1,608 @@
...
+		labels = append(labels, row.Label)
+	}
+	return labels, nil
+}
+
+// mergeLabels combines manual labels with newly applied symptom labels, deduplicating.
+func mergeLabels(manualLabels []string, bqLabels []models.JobRunLabel) []string {
+	merged := sets.NewString()
+	for _, l := range manualLabels {
+		merged.Insert(l)
```
petr-muller: nit: `Insert` is variadic iirc so we can just `Insert(manualLabels...)`. or actualy just initialize the set with the values right away with `sets.New[string](manualLabels...)`

pkg/api/jobrunscan/reevaluate.go
```diff
@@ -0,0 +1,608 @@
...
+			return labels, err
+		}
+		labels = append(labels, row.Label)
+	}
+	return labels, nil
+}
+
+// mergeLabels combines manual labels with newly applied symptom labels, deduplicating.
+func mergeLabels(manualLabels []string, bqLabels []models.JobRunLabel) []string {
+	merged := sets.NewString()
```
petr-muller: nit: `sets.New[string]()`

`NewString` is deprecated

pkg/api/jobrunscan/reevaluate.go
```diff
@@ -0,0 +1,608 @@
...
+}
+
+// queryNonSymptomLabels queries BQ for labels that were NOT applied by symptom detection.
+func (r *ReEvaluator) queryNonSymptomLabels(ctx context.Context, buildID string, startTime time.Time) ([]string, error) {
+	table := fmt.Sprintf("`%s.job_labels`", r.bqClient.Dataset)
+	queryStr := fmt.Sprintf(
+		`SELECT DISTINCT label FROM %s
+		WHERE prowjob_build_id = @buildID
+		AND (symptom_id IS NULL OR symptom_id = '')
+		AND DATE(prowjob_start) >= @startDate`,
```
petr-muller: I would not notice myself but Claude thinks this could be `DATE(prowjob_start) = @startDate` (like in `clearBQLabels` on L433) for a cheaper query and I think it may be right.

pkg/api/jobrunscan/reevaluate.go
```diff
@@ -0,0 +1,608 @@
...
+	result.PostgresUpdated = true
+
+	result.Status = ReEvalSuccess
+	return result
+}
+
+// evaluateSymptoms runs one artifact query per symptom for the given job run.
+func (r *ReEvaluator) evaluateSymptoms(ctx context.Context, jobRunID int64, symptoms []jobrunscan.Symptom) ([]symptomMatch, error) {
+	var matches []symptomMatch
+	mgr := jobartifacts.NewManager(ctx)
```
petr-muller: Could this be shared between multiple job run calls to have better control over parallelism? The manager looks like something that distributes work between internall worker pools so building it once per jobs seems a bit wasteful.

petr-muller (commented):
Inline comments are mostly nits although it may be still useful to tell an LLM to address them :D 

My main concern is that the operation feels quite heavy & expensive to be performed in a synchronous POST API call handler. The DD mentions that but I don't really buy it:

> For the initial implementation, synchronous processing is sufficient. Individual job run re-evaluation should complete in seconds. If larger batch sizes are needed later, consider async with a task ID + polling pattern.

Even "seconds" * 50 seems like it can hit some client / proxy timeouts enough to get canceled?

The operation also seems to not allow cancellation (at best we could cancel between individual job run refreshes, but once we start deleting/writing new data for a job run, we need to run to completion to avoid inconsistency. But even then we have some runs follow old symptom definitions and some new which is also inconsitent) which does not seem appropriate for an sync API call handler. The new code does not respect cancellation anywhere which is probably intentional, but I *think* it is possible for the operation to "accidentally" respect cancelation if we pass the request ctx to one of the libraries that respect its (like the BQ querying or GCS client calls) and it errors out on the canceled context - and I wonder if that could lead to inconsitent state in that case?

I think that async refresh could actually even lead to a cleaner implementation - just feed some queue with single job refresh requests handled by a worker pool. This is true especially if we do not expect clients to actually need the response data - I'm not sure whether there's intent to consume them for something useful. There's a note about "client retry decisions" but I think that if the server commits to eventually fulfill the submitted refresh request, then client should not really need to make retries?

pkg/api/jobrunscan/reevaluate.go
```diff
@@ -0,0 +1,608 @@
...
+
+	q := r.bqClient.Query(ctx, bqlabel.JobRunLabelsReEvaluateDelete, queryStr)
+	q.Parameters = []bigquery.QueryParameter{
+		{Name: "buildID", Value: buildID},
+		{Name: "startDate", Value: civil.DateOf(startTime.UTC())},
+	}
+	_, err := q.Read(ctx)
+	if err != nil {
+		if strings.Contains(err.Error(), "streaming buffer") {
+			log.Warnf("symptom reEval: BQ delete for %s hit streaming buffer, skipping: %v", buildID, err)
```
mstaeble: What are the implications of failing to delete the labels from BQ and potentially continuing to add more on top of the stale ones?

.claude/commands/agentic-followup.md
sosiouxme: right, i do not even look at these files since they are generated by `make apm`. it must have been missed previously

pkg/api/jobrunscan/reevaluate.go
sosiouxme: right, although the artifact scans are parallelized, and the limit to that parallelization is server-wide (so there's not a lot of point in optimizing anything else).

in my tests locally, a job run eval took a few seconds (likely faster once running in the cloud); this will get longer as more symptoms are added. so it is a very heavy operation. there are various ways to optimize it if we need to, but it does seem unlikely we'll ever want to run it in batches of 50 behind an API call like this; or in fact batches at all. more elsewhere.

pkg/api/jobrunscan/reevaluate.go
sosiouxme: good instruction for the agent. i like fields even though i don't know that we'll ever do much formal log analysis on them

pkg/api/jobrunscan/reevaluate.go
sosiouxme: You're absolutely right!

I can't believe I didn't notice it was doing this. It's supposed to reuse the manager that's instantiated at server startup.

pkg/api/jobrunscan/reevaluate.go
sosiouxme: Almost nothing. There are just duplicate labels. They all get collapsed into a single label on display regardless; it was impossible to implement the cloud function analysis without the potential for creating duplicates.

The only problem if labels don't get deleted is that maybe they should no longer be applied after symptom changes. That's a corner case and nothing too important goes wrong in that case so... I don't think this needs to block the rest of the functionality (it is an unfortunate consequence of trying to misuse a data warehouse as a DB).

pkg/api/jobrunscan/reevaluate.go
sosiouxme: I had that thought and didn't look into it right away since the labels table isn't really that big... but I'll have a look.

sosiouxme (commented):
@petr-muller My thinking on how to approach this evolved during implementation, but didn't really make it into an update to the backend plan.

Yes, evaluating all the symptoms across a job run is heavy. In my testing locally it took a few seconds each, and that will only get longer as there are more symptoms. It's possible to be more clever and not have to read as much, but that would really complicate the next feature, which is compound symptoms, and I don't think it's worth the trouble to be clever. This isn't a common operation and it doesn't need to be fast.

However it's bad UX to give a request and have to wait minutes for a response, especially if that response is "part of the request failed" (or worse "request timed out and you don't know how much succeeded"). So my conclusion was that it probably didn't make sense to actually run these as batch requests; instead, have the frontend run a pool of workers making single-run requests and be able to report progress as they return, and retry failures if appropriate. You could accomplish something similar with a ticket-and-poll approach, but to me that feels like having to implement something complicated twice - once in the backend, and then again to interpret it on the frontend. As a side benefit, running the pool from the frontend means it can hit multiple backend servers to max out the parallel requests.

But maybe there will be a use for batches on the backend; having designed it that way to begin with it doesn't seem necessary to tear it out, but if you think it's pointless and misleading, I guess I wouldn't mind much. Other thoughts welcome.

mstaeble:
> have the frontend run a pool of workers making single-run requests

@sosiouxme What happens if multiple users try to make the same requests concurrently? What happens if the frontend closes the connection or cancels the request? Given that the requests have side-effects in the backend, I suspect it will be harder to ensure consistent behavior if we use the frontend as the orchestrator.

sosiouxme:
@mstaeble these are good points to think about, thanks.

> @sosiouxme What happens if multiple users try to make the same requests concurrently?

They're idempotent-ish. The "-ish" is because with the stream buffer, BQ would refuse to delete what had just been written, so it would write duplicates in BQ - but those are not a problem. So, as long as nothing actually failed, it would end up consistent.

> What happens if the frontend closes the connection or cancels the request? Given that the requests have side-effects in the backend, I suspect it will be harder to ensure consistent behavior if we use the frontend as the orchestrator.

I'd argue the opposite. What happens now if the connection drops is that the backend context is cancelled at some point, and whatever it's doing then halts (returns an error up the stack) the next time it reaches a context checkpoint (I think our backend clients are well-behaved here). And then nothing. It's not going to try to proceed forward or roll back on its own, and with no ability to report results, I don't think it should. Meanwhile the frontend can notice that something is wrong and initiate retries; given idempotency, either these will succeed eventually, or it can report some kind of final failure to the user, who can decide to try again later or just accept that things are broken (they're labels for analysis, not customer orders or something).

Actually with three different storage backends and two server backends all able to fail independently I'd claim it's basically impossible to ensure consistent outcomes :) You're certainly not going to get anything approaching transactionality without a heck of a lot of work that's really not merited by the use case.

I suppose you could implement it with a table enqueueing re-eval requests.  The sippy daemon could act as a single point to read the queue, process the requests, and record a result. That would give you the ability to wait 90 minutes for the stream buffer to clear or handle other long failure modes. I dunno, I just don't think it markedly improves the UX.

mstaeble:
> Actually with three different storage backends and two server backends all able to fail independently I'd claim it's basically impossible to ensure consistent outcomes :)

@sosiouxme I agree with that sentiment. Or at least it is not worth the complexity or effort to ensure consistency. My argument is that observability and logical consistency are easier if known backend components are doing the orchestration rather than myriad frontends running on users' systems. The complexity of managing multiple single-run requests needs to be paid either way--either by the frontend or the backend. In my experience, it is usually a better outcome to put that complexity in the backend and leave the frontend as dumb as possible.

With that said, the main purpose of this PR seems to be to establish the logic for doing the work not the mechanism by which the work is triggered or managed. We can iterate on the latter aspects as part of other work.

petr-muller:
Yeah my motivation for poking into this is that I agree with:

> In my experience, it is usually a better outcome to put that complexity in the backend and leave the frontend as dumb as possible.

I was not entirely sure if the client/frontend really needs the result of the operation or if this could be just client-triggered and the effect would just be propagated through data changes. Seems like the client wants the result so waiting for it makes sense at least in the first iteration over creating some kind of refresh task api that the client would need to poll.

> With that said, the main purpose of this PR seems to be to establish the logic for doing the work not the mechanism by which the work is triggered or managed. We can iterate on the latter aspects as part of other work.

+1

/lgtm
/approve

sosiouxme:
extracted the agent changes to https://github.com/openshift/sippy/pull/3675 and rebased

sosiouxme:
/hold cancel

seems to work. the idea of a backend processor is growing on me but i too feel it can be later if ever.

PR #3663, opened 2026-06-22
UI: show regression/triage ID column in list views

### LINKED ISSUE #3662: [UI] Show regression/triage ID column in list views
## Problem

In Sippy's list views (regressions, triages, etc.), each item shows status, description, and other metadata — but **not the numeric ID**. This makes it difficult to refer to specific regressions when discussing them, especially during AI-assisted triage workflows.

The problem is particularly acute for:
- **Untriaged regressions lists**: installer failures often have identical descriptions, making items visually indistinguishable without an ID
- **Human-AI collaboration**: when the AI agent analyzes 20+ regressions and needs to report back which ones need manual investigation, IDs are the most unambiguous way to reference specific items (e.g. "regression 589 needs deeper investigation" vs repeating a long description that may match multiple entries)

## Requested Change

Add an **ID column** to the data grid in all relevant list views:
- Regressions list (triaged and untriaged)
- Triages list
- Any other views where items have a numeric identifier but don't display it

For example, in the regressions view, a regression with ID `589` should show that ID as a visible column alongside the existing status, description, etc.

## Use Case

During AI-assisted triage, the agent reads regressions via the API (JSON) and can work with IDs internally. However, at the end of analysis it reports results back to the human — e.g. "15 regressions are clear; investigate regressions A, B, C deeper; regression D is unknown; regression E is a composition of multiple issues." Using IDs is the most reliable way for the human to then locate those specific items in the Sippy UI. Without visible IDs, matching by description is error-prone, especially when descriptions are identical or truncated.

## Context

Discussion in Slack ([thread](https://redhat-internal.slack.com/archives/C0A1FQP7YTX/p1782133483146459)) — @sosiouxme confirmed this would be useful.

---

🤖 *This issue was created by AI on behalf of @mkowalski, who has reviewed it.*

mkowalski:
## Problem

In Sippy's list views (regressions, triages, etc.), each item shows status, description, and other metadata — but not the numeric **ID**. This makes it difficult to refer to specific regressions/triages when discussing them, especially during AI-assisted triage workflows.

The problem is particularly acute for:

- **Untriaged regressions lists**: installer failures often have identical descriptions, making items visually indistinguishable without an ID.
- **Human-AI collaboration**: when the AI agent analyzes 20+ regressions and reports back which ones need manual investigation, IDs are the most unambiguous way to reference specific items (e.g. "regression 589 needs deeper investigation" vs. repeating a long description that may match multiple entries).

## Change

The numeric IDs (`Triage.ID` and `TestRegression.ID`) are already present in the API responses, so this is a **frontend-only** change that surfaces them as a column in the relevant DataGrid list views:

| View | File | Column |
| --- | --- | --- |
| Triages list (`/component_readiness/triages`) | `TriagedRegressions.js` | **ID** |
| Unresolved / Untriaged / All regressions tabs | `RegressedTestsPanel.js` | **Regression ID** |
| Triaged regressions list & triage detail page | `TriagedRegressionTestList.js` | **Regression ID** |
| Potential matching triages grid | `RegressionPotentialMatchesTab.js` | **Triage ID** |
| Remove-regressions grid (Update Triage) | `UpdateTriagePanel.js` | **Regression ID** |

(`TriagePotentialMatches.js` already displayed a Regression ID column.)

Columns use the existing `type: 'number'` idiom (as in `JobTable.js`/`FeatureGates.js`) so they sort numerically, align correctly, and expose numeric filter operators.

## Drive-by fix

While wiring up numeric columns I found a gap in the shared client-side filter (`datagrid/filterUtils.js`): the `=` operator that `type: 'number'` columns expose in the filter UI had no `case` and fell through to the `contains` default — so filtering an ID with `=589` would also match `5890`. Added `=` as an exact-match alias of `equals`. No previous `applyFilterModel`-based grid used a numeric column, so this path was never exercised before.

## Tests

- `datagrid/filterUtils.test.js` — covers the `=` exact-match fix and numeric ID filtering (9 tests).
- `component_readiness/TriagedRegressions.test.js` — renders the grid and asserts the new **ID** column header and ID values appear, using two identical-description entries (the exact disambiguation scenario from the issue).

## Verification

- `npx eslint .` (full project) — clean
- `npx prettier --check` on changed files — clean
- Full Jest suite — 23/23 tests pass
- `react-scripts build` — compiles successfully

Closes #3662

---
🤖 *This PR (description and commits) was generated by AI on behalf of @mkowalski, who has reviewed it.*

## Summary by CodeRabbit

* **New Features**
  * Added “Regression ID” and “Triage ID” identifier columns across component readiness data grids to improve tracking and visibility.
* **Bug Fixes**
  * Updated client-side filtering to support the `=` equality operator consistently with existing “equals” behavior.
* **Tests**
  * Added/expanded Jest and React Testing Library coverage to verify new grid ID columns render correctly and filtering works for equality, negation, and numeric comparisons.

### DIFF
--- a/sippy-ng/src/component_readiness/RegressedTestsPanel.js
+++ b/sippy-ng/src/component_readiness/RegressedTestsPanel.js
@@ -144,6 +144,14 @@ export default function RegressedTestsPanel(props) {
         ]
       : []),
     {
+      field: 'regression_id',
+      headerName: 'Regression ID',
+      type: 'number',
+      flex: 6,
+      valueGetter: (params) => params.row.regression?.id,
+      renderCell: (param) => <div>{param.value}</div>,
+    },
+    {
       field: 'component',
       headerName: 'Component',
       flex: 20,
--- a/sippy-ng/src/component_readiness/RegressionPotentialMatchesTab.js
+++ b/sippy-ng/src/component_readiness/RegressionPotentialMatchesTab.js
@@ -159,6 +159,19 @@ export default function RegressionPotentialMatchesTab({
 
   const columns = [
     {
+      field: 'triage_id',
+      headerName: 'Triage ID',
+      type: 'number',
+      flex: 4,
+      align: 'center',
+      valueGetter: (params) => {
+        return params.row.triage.id
+      },
+      renderCell: (param) => (
+        <Typography variant="body2">{param.value}</Typography>
+      ),
+    },
+    {
       field: 'resolution_date',
       valueGetter: (params) => {
         return params.row.triage.resolved?.Valid
--- a/sippy-ng/src/component_readiness/TriagedRegressionTestList.js
+++ b/sippy-ng/src/component_readiness/TriagedRegressionTestList.js
@@ -105,6 +105,16 @@ export default function TriagedRegressionTestList(props) {
 
   const columns = [
     {
+      field: 'id',
+      headerName: 'Regression ID',
+      type: 'number',
+      flex: 8,
+      valueGetter: (params) => {
+        return params.row.id
+      },
+      renderCell: (param) => <div>{param.value}</div>,
+    },
+    {
       field: 'test_name',
       headerName: 'Test Name',
       flex: 50,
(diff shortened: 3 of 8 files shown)


mkowalski:
/close

Superseded by https://github.com/openshift/sippy/pull/3670

PR #3679, opened 2026-06-24
TRT-2734: add release_definitions table and loader

mstaeble:
## Summary

Creates a `release_definitions` table in PostgreSQL and a loader to populate it from BigQuery. This is the first phase of moving release metadata from BigQuery to PostgreSQL ([TRT-2734](https://redhat.atlassian.net/browse/TRT-2734)).

No consumers are switched in this PR. The existing `getReleases()` and `QueryReleases()` paths continue to use BigQuery / hardcoded metadata. A follow-up PR (#3736) will switch consumers to read from the new table once it has been populated by the load cycle.

- Adds `ReleaseDefinition` model with capability constants and `HasCapability` method
- Adds `release-definitions` loader (`--loader release-definitions`) that fetches from BQ and syncs to PG via upsert
- Adds `GetReleasesFromDB` and `DefinitionToRelease` for reading from PG
- Adds `GetReleaseRowsFromBigQuery` for direct BQ row access
- Load command reads release configs from PG with BQ fallback when the table is empty (first run)
- Seeds `release_definitions` for local development
- 7 files changed, +276/-8

## Test plan

- [x] `go build ./...` passes
- [x] `go vet ./...` passes
- [x] `make lint` passes
- [x] `go test ./pkg/... ./cmd/...` passes
- [x] `sippy serve` with seed data works correctly (no behavior change)

### Staging verification

Deployed and ran the loader as a one-off job:

```
sippy load --loader release-definitions --init-database
```

Synced 36 release definitions from BigQuery to staging PostgreSQL (all OCP releases 3.11 through 5.0, OKD, ROSA, ARO, HyperShift, and CAPI entries).

Ref: [TRT-2734](https://redhat.atlassian.net/browse/TRT-2734)

@coderabbitai ignore

### DIFF
--- a/cmd/sippy/load.go
+++ b/cmd/sippy/load.go
@@ -37,6 +37,7 @@ import (
 	"github.com/openshift/sippy/pkg/dataloader/prowloader"
 	"github.com/openshift/sippy/pkg/dataloader/prowloader/gcs"
 	"github.com/openshift/sippy/pkg/dataloader/prowloader/github"
+	releasedefloader "github.com/openshift/sippy/pkg/dataloader/releasedefloader"
 	"github.com/openshift/sippy/pkg/dataloader/releaseloader"
 	"github.com/openshift/sippy/pkg/dataloader/testownershiploader"
 	"github.com/openshift/sippy/pkg/db"
@@ -100,7 +101,7 @@ func (f *LoadFlags) BindFlags(fs *pflag.FlagSet) {
 	f.JiraFlags.BindFlags(fs)
 
 	fs.BoolVar(&f.InitDatabase, "init-database", false, "Migrate the DB before loading")
-	fs.StringArrayVar(&f.Loaders, "loader", []string{"prow", "releases", "jira", "github", "bugs", "test-mapping", "feature-gates"}, "Which data sources to use for data loading")
+	fs.StringArrayVar(&f.Loaders, "loader", []string{"release-definitions", "prow", "releases", "jira", "github", "bugs", "test-mapping", "feature-gates"}, "Which data sources to use for data loading")
 	fs.StringArrayVar(&f.Releases, "release", f.Releases, "Which releases to load (one per arg instance)")
 	fs.StringArrayVar(&f.Architectures, "arch", f.Architectures, "Which architectures to load (one per arg instance)")
 	fs.StringVar(&f.JobVariantsInputFile, "job-variants-input-file", "expected-job-variants.json", "JSON input file for the job-variants loader")
@@ -152,8 +153,6 @@ func NewLoadCommand() *cobra.Command {
 				cacheClient = nil // error hygiene, since we pass this down to quite a few functions
 			}
 
-			releaseConfigs := []sippyv1.Release{}
-
 			// initializing a bigquery client different from the normal one
 			opCtx, ctx := bqcachedclient.OpCtxForCronEnv(ctx, "load")
 			bqc, bigqueryErr := bqcachedclient.New(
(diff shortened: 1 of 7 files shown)


pkg/sippyserver/server.go
```diff
@@ -189,15 +189,19 @@ type Server struct {
...
-// getReleases returns release data, preferring the BigQuery client with caching
-// when available, falling back to the data provider for mock mode.
-func (s *Server) getReleases(ctx context.Context, forceRefresh ...bool) ([]sippyv1.Release, error) {
-	if s.bigQueryClient != nil {
-		refresh := len(forceRefresh) > 0 && forceRefresh[0]
-		return api.GetReleases(ctx, s.bigQueryClient, refresh)
+// getReleases returns release data from PostgreSQL.
+func (s *Server) getReleases(ctx context.Context) ([]sippyv1.Release, error) {
+	if s.db != nil {
+		releases, err := api.GetReleasesFromDB(ctx, s.db)
```
sosiouxme: this removes the cache. reading from a small postgres table is fast, but not nearly as fast as reading from redis, and this page gets called.... a lot. with React calling it several times per UI transition it may even be noticeable at a human timescale. consider whether we don't want caching here?

mstaeble: Good call raising this. I traced all the callers to understand the full impact.

getReleases() is called by 8 HTTP handlers:
  - /api/releases (page load, ~15 calls/hour)
  - /api/component_readiness (every CR request)
  - /api/component_readiness/test_details
  - /api/component_readiness/views
  - 4 triage/regression endpoints

The old Redis cache (8h TTL, key "Releases~") was valuable because it avoided a BigQuery round-trip on every one of these requests. That saved both latency and BQ query cost.

With PostgreSQL, the query is ~7ms for ~20 rows. The CR endpoints themselves take seconds, so 7ms is noise. There's no per-query cost like BQ.

More importantly, these callers are transitional. The CR handlers use release data for two things: resolving relative time strings like "ga-30d" into absolute dates, and generating HATEOAS links with correct time parameters. As we move CR fully to PostgreSQL:

1. The CR test status queries will use pre-aggregated matviews per release (e.g., cr_base_agg_4.21), where the GA-based time window is baked into the matview at build time. No release metadata lookup needed at request time.
2. For any queries that still need release dates, the lookup can fold into the main query as a JOIN against release_definitions rather than a separate round-trip.

That leaves /api/releases as the only endpoint that needs a standalone query, at ~15 calls/hour. Adding a Redis cache layer for that adds complexity with no user-visible benefit. The HTTP round-trip overhead alone (~50-200ms) dwarfs the 7ms PG query.

If we see this become a bottleneck in practice, we can add caching for PQ later, but I don't think it's warranted now.

pkg/api/componentreadiness/dataprovider/postgres/provider.go
```diff
@@ -132,60 +133,8 @@ func (p *PostgresProvider) QueryJobVariants(ctx context.Context) (crtest.JobVari
...
-				rel.GADate = &ga
-			}
-			if meta.product != "" {
-				rel.Product = meta.product
-			}
-		}
-		releases = append(releases, rel)
-	}
-	return releases, nil
+	return api.GetReleasesFromDB(ctx, p.dbc)
```
neisw: Do we have a race condition when this merges prior to load running?  Should this be a two-stage process where we init / load the new table then switch over, or is that happening externally?

mstaeble: Good point. I've split this into two PRs to avoid the race condition:

- #3679 (this PR): Creates the release_definitions table, adds the release-definitions loader, and seeds data for local dev. No consumers are switched — the existing getReleases() and QueryReleases() paths continue to use BigQuery / hardcoded metadata. This is purely additive and safe to merge immediately.
- #3736: Switches getReleases() and the PG provider to read from release_definitions instead of BigQuery / the hardcoded map. This should only merge after #3679 has been deployed and at least one load cycle has populated the table.

cmd/sippy/load.go
```diff
@@ -164,10 +163,14 @@ func NewLoadCommand() *cobra.Command {
...
-					return errors.Wrapf(err, "error querying releases from bq")
-				}
+			}
+
+			// Read release definitions from PG for downstream loader construction.
+			// On the first run the table may be empty; the release-definitions
+			// loader will populate it for subsequent runs.
+			releaseConfigs := []sippyv1.Release{}
+			if dbErr == nil {
+				releaseConfigs, _ = api.GetReleasesFromDB(context.Background(), dbc)
```
neisw: I think we still have an issue here.  The releases are needed for partitioning and the loader.  You could look to have GetReleasesFromDB fallback to GetReleasesFromBigQuery when results are empty or keep the BQ call here for now until we know it is populated.

mstaeble: Good point. I've added a BQ fallback in load.go. It reads from PG first, and falls back to `GetReleasesFromBigQuery` when the table is empty (e.g., first run on a new database). This ensures partitioning and downstream loaders work correctly even before the release-definitions loader has populated the table.

neisw: One last consideration, previously we returned if there was an error loading the releases.  You are now logging the errors in the api calls but drop the errors here.  I don't think we expect the loader to work without releases, at least not the job loader.  

I don't want to overengineer this but should we validate we have at least 1 release and if not then exit?

mstaeble: Updated. Error handling now matches the original behavior: hard exit on query error (`return errors.Wrapf(err, ...)`), silent continue on empty results. It tries PG first, falls back to BQ if PG returns no rows (e.g., first run).

PR #3685, opened 2026-06-24
Batch BigQuery label fetching and skip GCS for cached job runs

mstaeble:
## Summary

- Replace per-job BigQuery label queries with a single bulk prefetch before the worker loop, eliminating thousands of individual BQ round-trips per load cycle
- Move the `prowJobRunCache` check before the GCS `FindAllMatches` listing so already-processed runs skip the expensive object listing entirely
- Keep `createOrUpdateProwJob` before the cache check to ensure job definitions (variants, release) stay current

## Evidence from production logs

  Production fetchdata runs (hourly, June 24 2026) show that the prow loader spends 25-29 minutes processing 16-18K jobs from BigQuery, but only 2-3% actually need GCS processing. The rest are duplicates from the
  12-hour lookback overlap.

  The current code flow in `prowJobToJobRun` fetches GCS artifacts (path resolution, bucket client, JUnit file matching) *before* checking the in-memory cache. This means ~3,500 jobs per run go through full GCS
  I/O only to be discarded as already processed.

  | Metric | 14:00 run | 15:03 run | 16:08 run |
  |---|---|---|---|
  | Jobs from BigQuery | 16,368 | 17,292 | 18,644 |
  | Skipped by in-memory cache (before `prowJobToJobRun`) | ~12,604 (77%) | ~13,444 (78%) | ~14,699 (79%) |
  | GCS fetched then found already processed | 3,297 (20%) | 3,327 (19%) | 3,537 (19%) |
  | GCS fetched and actually needed | 467 (3%) | 521 (3%) | 408 (2%) |
  | **Prow loader time** | **24m 31s** | **25m 18s** | **28m 49s** |

  Moving the cache check before the GCS fetch eliminates ~3,500 unnecessary GCS round-trips per run (~19% of total jobs).

## Evidence from staging

  Deployed the PR image to staging and ran two load cycles against the staging database.

  **Run 1 (cold start, empty staging DB):**
  - 205,482 jobs from BigQuery
  - Bulk label prefetch: 205K build IDs returned 4,636 labels in **12.7 seconds** (single BQ query)
  - Processed 29,705 of 205K jobs in ~43 minutes before being stopped (all jobs required full GCS processing since the DB was empty)
  - 7,291 new job runs inserted

  **Run 2 (warm start, 7K runs already cached):**
  - 21,353 jobs from BigQuery
  - Bulk label prefetch: 21K build IDs returned 363 labels in **2.75 seconds**
  - Of 17,041 jobs processed (at time of observation), only 2,977 (17%) needed GCS processing
  - **14,064 jobs (83%) skipped GCS listing entirely** via the early cache check
  - Cache-hit jobs processed at the rate of thousands per second (no I/O)

  **Label verification:**
  - Staging database shows 55,667 of 544,340 job runs have labels applied
  - Most recent labeled runs (June 24) show correct values (e.g., `ImagePullNeverCompletes`, `TestFailureDuringHighCPUEvents`)

## Test plan

- [x] `go vet ./pkg/dataloader/prowloader/...` passes
- [x] `go test ./pkg/dataloader/prowloader/...` passes
- [x] `go build ./cmd/sippy/...` compiles
- [x] Verify in staging that bulk label prefetch logs show a single BQ query with count and duration
- [x] Verify cached job runs skip GCS listing (83% of jobs skipped in warm run)
- [x] Verify newly imported job runs still have correct labels in the database

### DIFF
--- a/pkg/dataloader/prowloader/prow.go
+++ b/pkg/dataloader/prowloader/prow.go
@@ -269,11 +269,11 @@ func (pl *ProwLoader) Load() {
 	}
 
 	// Pre-fetch labels for all jobs in bulk instead of one BQ query per job.
-	labelsCache, err := pl.prefetchLabels(prowJobs)
-	if err != nil {
+	if lc, err := pl.prefetchLabels(prowJobs); err == nil {
+		pl.labelsCache = lc
+	} else {
 		pl.errors = append(pl.errors, errors.Wrap(err, "error pre-fetching labels from BigQuery"))
 	}
-	pl.labelsCache = labelsCache
 
 	queue := make(chan *prow.ProwJob)
 	errsCh := make(chan error, len(prowJobs))
@@ -312,7 +312,7 @@ func (pl *ProwLoader) Load() {
 
 	// load the test analysis by job data into tables partitioned by day, letting bigquery do the
 	// heavy lifting for us.
-	err = pl.loadDailyTestAnalysisByJob(pl.ctx)
+	err := pl.loadDailyTestAnalysisByJob(pl.ctx)
 	if err != nil {
 		pl.errors = append(pl.errors, errors.Wrap(err, "error updating daily test analysis by job"))
 	}
--- a/pkg/dataloader/prowloader/prow.go
+++ b/pkg/dataloader/prowloader/prow.go
@@ -81,6 +81,7 @@ type ProwLoader struct {
 	gcsClient                    *storage.Client
 	promPusher                   *push.Pusher
 	loadSince                    *time.Time
+	labelsCache                  map[string]pq.StringArray
 }
 
 func New(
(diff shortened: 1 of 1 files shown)


pkg/dataloader/prowloader/prow.go
```diff
@@ -267,6 +268,13 @@ func (pl *ProwLoader) Load() {
 		log.Infof("Partition cleanup complete: detached %d, dropped %d", detached, dropped)
 	}
 
+	// Pre-fetch labels for all jobs in bulk instead of one BQ query per job.
+	labelsCache, err := pl.prefetchLabels(prowJobs)
+	if err != nil {
+		pl.errors = append(pl.errors, errors.Wrap(err, "error pre-fetching labels from BigQuery"))
+	}
+	pl.labelsCache = labelsCache
```
petr-muller: we should only use `labelsCache` if `err` was nil

if we also used `if lc, err := ...; err != nil` syntax we'd avoid polluting the function scope with `err` variable that's easy to reuse later by mistake and we would not need L315 change

petr-muller (approved):
LGTM w/ minor error handling note inline, not blocking, unhold or address at will

/hold

pkg/dataloader/prowloader/prow.go
mstaeble: Done, switched to if-init syntax in b8e53133.

petr-muller (approved):
/hold cancel

PR #3686, opened 2026-06-24
Upgrade golangci-lint to v2.12.2

mstaeble:
## Summary
- Bump the golangci-lint container image in `hack/go-lint.sh` from v2.10.1 to v2.12.2
- Suppress two new gosec false positives introduced by the upgrade:
  - **G117** in `gcs_authentication.go`: intentional OAuth token cache write to a local file
  - **G118** in `daemon_server.go`: `cancel` func is stored in `pendingContexts` and called on shutdown, but gosec cannot trace it through the slice

## Test plan
- [x] Ran `gofmt -w` on both modified Go files (no changes needed)
- [x] Ran `golangci-lint --timeout 10m run ./pkg/dataloader/prowloader/gcs/... ./pkg/sippyserver/...` with 0 issues
- [x] CI lint job passes with the new version

### DIFF
--- a/hack/go-lint.sh
+++ b/hack/go-lint.sh
@@ -25,6 +25,6 @@ else
   $DOCKER run --rm \
     --volume "${PWD}:/go/src/github.com/openshift/sippy${VOLUME_OPTION}" \
     --workdir /go/src/github.com/openshift/sippy \
-    docker.io/golangci/golangci-lint:v2.10.1 \
+    docker.io/golangci/golangci-lint:v2.12.2 \
     golangci-lint --timeout 10m "${@}"
 fi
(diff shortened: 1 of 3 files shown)


hack/go-lint.sh
```diff
@@ -25,6 +25,6 @@ else
   $DOCKER run --rm \
     --volume "${PWD}:/go/src/github.com/openshift/sippy${VOLUME_OPTION}" \
     --workdir /go/src/github.com/openshift/sippy \
-    docker.io/golangci/golangci-lint:v2.10.1 \
+    docker.io/golangci/golangci-lint:v2.12.2 \
```
smg247, line 28: We need to have a matching image added as a supplemental image in openshift/release. Generally, Test Platform does this periodically. The latest one there is still `v2.10.1`.

smg247: You could request they add a new one in #forum-ocp-testplatform on slack

mstaeble: Got it. Not a big deal. I was seeing some lint failures locally do to the version discrepancy. I'll switch to a version that matches what is used in CI.

PR #3695, opened 2026-06-25
TRT-2748: Require minimum failures before marking key tests as pants on fire

smg247:
Key tests like "install should succeed" are highly sensitive and a single post-fix failure is often noise. This adds a threshold (2 failures) before escalating key tests to FailedFixedRegression, reducing false positives. Non-key tests retain the existing single-failure behavior.

The failure counting is injected via a function field on RegressionTracker so tests can verify the threshold logic without a database connection.

## Summary by CodeRabbit

* **New Features**
  * Key tests are now handled with a separate failure threshold, so regressions can be classified more accurately based on how many failures happened after a fix was marked.

* **Bug Fixes**
  * Improved triage results when failures continue after a fix, including safer fallback behavior if failure counting cannot be completed.


dgoodwin (commented):
/lgtm

@cpmeadors this increases the threshold to 2, I'm willing to go to 3 if we need to.

PR #3696, opened 2026-06-25
Make component/capability variants more generic

dgoodwin:
This shouldn't be tied to being a spot check job. Many spot check jobs
will start out running quite often as part of their feature gate
promotion logic.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated periodic job variant handling to use clearer, generic **Component** and **Capability** labels instead of spot-check-specific keys.
  * Strengthened validation for `spotcheck-*` variants to require both component and capability data.
  * Refined variant ownership classification and job-tier overrides to follow the shared job-name pattern rules for consistent categorization.
* **Tests**
  * Updated expected mappings and snapshot data to match the new `Component`/`Capability` labeling.


pkg/variantregistry/ocp.go
```diff
@@ -274,12 +274,12 @@ ORDER BY j.prowjob_job_name;
...
-// validateSpotCheckVariants returns an error if a job has JobTier=spotcheck without both
-// SpotCheckComponent and SpotCheckCapability defined.
-func validateSpotCheckVariants(jobName string, variants map[string]string) error {
+// validateComponentCapabilityVariants returns an error if a job has a spotcheck JobTier without both
+// Component and Capability defined.
+func validateComponentCapabilityVariants(jobName string, variants map[string]string) error {
 	if strings.HasPrefix(variants[VariantJobTier], "spotcheck-") {
-		if variants[VariantSpotCheckComponent] == "" || variants[VariantSpotCheckCapability] == "" {
-			return fmt.Errorf("job %q has JobTier=%s but is missing SpotCheckComponent or SpotCheckCapability", jobName, variants[VariantJobTier])
+		if variants[VariantComponent] == "" || variants[VariantCapability] == "" {
```
dgoodwin, line 281: to @stbenjam point on slack, Capability should probably obsolete Procedure. I don't want to play with that just yet as it's used in views, thus part of the tracked regression data we have triaged. 

I'd like to do a followup replacing Procedure in the sippy postgres whitelist with Component + Capability. I can't think of a reason we need Procedure in sippy postgres/classic right now. This would help limit the combo explosion, though I'm unclear how relevant that is now with Matthew's work.

pkg/variantregistry/ocp.go
```diff
@@ -449,36 +449,36 @@ var (
...
+	VariantRelease          = "Release"
+	VariantReleaseMinor     = "ReleaseMinor"
+	VariantReleaseMajor     = "ReleaseMajor"
+	VariantFromRelease      = "FromRelease"
+	VariantFromReleaseMinor = "FromReleaseMinor"
+	VariantFromReleaseMajor = "FromReleaseMajor"
+	VariantLayeredProduct   = "LayeredProduct"
+	VariantOS               = "OS"
+	VariantComponent        = "Component"  // jobs with an owning component, used to flag regressions if a tailored job is not passing
+	VariantCapability       = "Capability" // jobs with an owning capability, used to flag regressions if a tailored job is not passing
```
mstaeble, line 479: Is this capability the same as the CR capability? Or are we overloading this term?

Is the long-term expectation that many/most jobs will have the component and capability set, even in the absence of the spot-check job tier for the job?

pkg/variantregistry/ocp.go
```diff
@@ -802,14 +808,16 @@ func setSpotCheckClassification(_ logrus.FieldLogger, variants map[string]string
...
-
 	jobNameLower := strings.ToLower(jobName)
 
+	// Check componentCapabilityPatterns first for tier overrides (e.g. spotcheck).
+	for _, p := range componentCapabilityPatterns {
+		if p.jobTier != "" && allSubstringsMatch(jobNameLower, p.substrings) {
+			variants[VariantJobTier] = p.jobTier
+			return
+		}
+	}
```
mstaeble, line 822: nit: Since `setComponentAndCapability` already ran and set VariantComponent/VariantCapability, this could skip the redundant substring matching and use the already-computed result:

```
if variants[VariantComponent] != "" {
  for _, p := range componentCapabilityPatterns {
    if p.jobTier != "" && p.component == variants[VariantComponent] && p.capability == variants[VariantCapability] {
      variants[VariantJobTier] = p.jobTier
      return
    }
  }
}
```
 
This makes the dependency on setter ordering explicit. It's already the case, just not visible in the code.

mstaeble (commented):
Overall the changes look good to me. I just have one clarifying question to help me understand better.

pkg/variantregistry/ocp.go
dgoodwin: Same concept. Most folks refer to this as a feature or feature gate name, but original CR authors went with Capability so I stuck with that. One of the use cases this will be used for, it will be a feature gate name.

dgoodwin: Most of our "main" jobs will not have a specific component/capability, this is for things like the installer where they can't add specific tests for a feature, they can only add a job that uses the configuration for the feature and then run an existing test suite to prove it's ok. 

Could be quite a few jobs having an owner though in the future if this is as successful as I hope it will be.

pkg/variantregistry/ocp.go
dgoodwin: Added, thanks

PR #3726, opened 2026-07-01
Batch BigQuery label pre-fetch to avoid HTTP 413

mstaeble:
## Summary

- The `sippy-backup` CronJob loads 14 days of data from scratch, producing ~332k build IDs. `GatherLabelsFromBQ` sent all of them in a single BigQuery query parameter, exceeding the ~10MB HTTP request body limit (HTTP 413).
- Split the build ID list into batches of 50k before querying, merging results across batches. Incremental fetchdata runs (~15k IDs) fit in a single batch and are unaffected.

## Test plan

- [x] `go vet` and `go test ./pkg/dataloader/prowloader/...` pass
- [x] `make lint` passes (0 issues)
- [x] Verified batching against real BigQuery data: fetched 7,425 real build IDs, split across 3 batches (batch size reduced for testing), confirmed batched results match single-query baseline exactly

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved label loading for large datasets, making results more reliable and less likely to fail on big requests.
  * Large batches are now processed more smoothly, which should reduce query-related issues and improve overall stability.

### DIFF
--- a/pkg/dataloader/prowloader/prow.go
+++ b/pkg/dataloader/prowloader/prow.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"maps"
 	"net/http"
 	"net/url"
 	"os"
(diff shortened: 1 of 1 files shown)


sosiouxme (approved):
/lgtm

PR #3728, opened 2026-07-02
TRT-2768: Import /payload job results from PRs into postgres

dgoodwin:
* Core change here is to begin importing jobs run in a PR via /payload commands as if they were presubmit data. 
* Involves a small hack to normalize the prow job name openshift-origin-31301-ci-5.0-upgrade-from-stable-4.22-e2e-gcp-ovn-rt-upgrade by removing the PR number, so any /payload job run of this job gets the same job name: openshift-origin-ci-5.0-upgrade-from-stable-4.22-e2e-gcp-ovn-rt-upgrade
* Ports the api endpoint for test_results for a PR to now use postgresql instead of BQ.
* This positions the endpoint as cost effective to roll out a risk analysis agent across many repositories, which will examine PR test results for anything suspicious and related to the code change, without worrying about bigquery costs. 
  * In my testing, the response for a 2 week query of presubmit data for one PR in the prod db is now very fast with the sharding.

## Summary by CodeRabbit

- **New Features**
  - Expanded synthetic seeding to include **Presubmits** test data.
  - Added **end-to-end** coverage for `/api/pull_requests/test_results`.
  - Added support for a **`limit`** query parameter.

- **Bug Fixes**
  - Improved **/payload** sub-job handling for more reliable **presubmit-style** job/variant matching.
  - **TestGrid** links are no longer generated for payload presubmit jobs where they don’t apply.

- **API Updates**
  - `/api/pull_requests/test_results` now always uses **PostgreSQL** and returns **job-run–scoped** results, with failures-only-by-default plus optional successes (`include_successes`) and `latest_sha_only`.


cmd/sippy/seed_data.go
```diff
@@ -770,6 +775,198 @@ func syncRegressions(dbc *db.DB) error {
...
+				Output:                  "Expected install to succeed but got timeout after 30m",
+				ProwJobRunTestTimestamp: ri.run.Timestamp,
+				ProwJobRunTestRelease:   "Presubmits",
+			}
+			if err := dbc.DB.Create(&output).Error; err != nil {
+				return fmt.Errorf("failed to create ProwJobRunTestOutput: %w", err)
+			}
+		}
+
+		// Success result for install test (same test, different run aspect)
```
mstaeble: Is this comment accurate? Is says it is for the same install test, but the TestID used is "networkTestID".

cmd/sippy/seed_data.go
```diff
@@ -770,6 +775,198 @@ func syncRegressions(dbc *db.DB) error {
...
+	for _, ri := range runs {
+		// Failure result for install test
+		failResult := models.ProwJobRunTest{
+			ProwJobRunID:        ri.run.ID,
+			ProwJobID:           ri.run.ProwJobID,
+			ProwJobRunRelease:   "Presubmits",
+			ProwJobRunTimestamp: ri.run.Timestamp,
+			TestID:              installTestID,
+			SuiteID:             &suite.ID,
+			Status:              12,
```
mstaeble: Should we use the constants for Status?

https://github.com/dgoodwin/sippy/blob/d5763a01aef34537cb3c36a371780739aee6092d/pkg/apis/sippyprocessing/v1/types.go#L206-L212

pkg/api/prtestresults.go
```diff
@@ -1,324 +1,142 @@
...
+		Select(`pjr.id AS prow_job_run_id,
+			pj.name AS prow_job_name,
+			pjr.url AS prow_job_url,
+			pp.sha AS pr_sha,
+			pjr.timestamp AS prow_job_start,
+			t.name AS test_name,
+			COALESCE(s.name, '') AS test_suite,
+			pjrt.status,
+			COALESCE(pjrto.output, '') AS output`).
+		Joins("JOIN prow_job_run_prow_pull_requests jrpr ON jrpr.prow_pull_request_id = pp.id").
```
mstaeble: Two performance concerns with this join:

**Missing index**: The primary key on prow_job_run_prow_pull_requests is (prow_job_run_id, prow_pull_request_id). Since prow_pull_request_id is the second column, PostgreSQL can't use the PK index for this jrpr.prow_pull_request_id = pp.id lookup and must sequentially scan the join table. This table grows by one row per job-run-to-PR association. A single-column index on prow_pull_request_id would fix this.

**Missing release/timestamp filter on join table**: Every other query in the codebase that uses this table filters on its denormalized prow_job_run_release and prow_job_run_timestamp columns (see pull_request_queries.go:32, repository_queries.go:30, functions.go:89-90). There's a composite index idx_prow_job_run_prow_pull_requests_release_timestamp specifically for this. Adding these conditions to the join would let the planner use it:
```
Joins("JOIN prow_job_run_prow_pull_requests jrpr ON
   jrpr.prow_pull_request_id = pp.id AND
   jrpr.prow_job_run_release = 'Presubmits' AND
   jrpr.prow_job_run_timestamp >= ? AND
   jrpr.prow_job_run_timestamp < ?",
  startDate, endDate)
```

pkg/api/prtestresults.go
```diff
@@ -1,324 +1,142 @@
...
-					AND (`
-		for i := range includeSuccesses {
-			if i > 0 {
-				whereClause += " OR "
-			}
-			whereClause += fmt.Sprintf("deduped.test_name LIKE @IncludeSuccess%d", i)
+	// Start from the PR side and use partition keys (release, timestamp) on
+	// prow_job_run_tests to allow PostgreSQL to prune partitions and avoid
+	// locking every partition in the table.
+	query := dbc.DB.Table("prow_pull_requests pp").
```
mstaeble, line 62: Using dbc.DB.Table(...) with a raw table name bypasses GORM's automatic WHERE deleted_at IS NULL injection. ProwPullRequest, ProwJobRun, and ProwJob all have soft-delete support, so soft-deleted records could leak into results. Other queries in the codebase (e.g., recent_test_failures.go:38-41) add explicit deleted_at IS NULL conditions on each table when using raw table names. Consider adding:
```
Where("pp.deleted_at IS NULL AND pjr.deleted_at IS NULL AND pj.deleted_at IS NULL")
```

pkg/api/prtestresults.go
```diff
@@ -1,324 +1,142 @@
...
+			pjrt.status,
+			COALESCE(pjrto.output, '') AS output`).
+		Joins("JOIN prow_job_run_prow_pull_requests jrpr ON jrpr.prow_pull_request_id = pp.id").
+		Joins("JOIN prow_job_runs pjr ON pjr.id = jrpr.prow_job_run_id").
+		Joins("JOIN prow_jobs pj ON pj.id = pjr.prow_job_id AND pj.release = 'Presubmits'").
+		Joins("JOIN prow_job_run_tests pjrt ON pjrt.prow_job_run_id = pjr.id AND pjrt.prow_job_run_release = 'Presubmits' AND pjrt.prow_job_run_timestamp >= ? AND pjrt.prow_job_run_timestamp < ?", startDate, endDate).
+		Joins("JOIN tests t ON t.id = pjrt.test_id").
+		Joins("LEFT JOIN suites s ON s.id = pjrt.suite_id").
+		Joins("LEFT JOIN prow_job_run_test_outputs pjrto ON pjrto.prow_job_run_test_id = pjrt.id AND pjrto.prow_job_run_test_timestamp = pjrt.prow_job_run_timestamp AND pjrto.prow_job_run_test_release = pjrt.prow_job_run_release").
+		Where("pp.org = ? AND pp.repo = ? AND pp.number = ?", org, repo, prNumber).
```
mstaeble, line 79: There's no index on prow_pull_requests(org, repo, number). The model only has a unique index on (sha, link). This filter is the primary selectivity driver for the entire query (a specific PR typically has 1-5 rows), but without an index PostgreSQL must sequentially scan the table. As the table grows with each unique (PR, SHA) combination, this will get progressively slower.

Consider adding a composite index to the model:
```
Org    string `json:"org" gorm:"index:idx_prow_pull_requests_org_repo_number"`
Repo   string `json:"repo" gorm:"index:idx_prow_pull_requests_org_repo_number"`
Number int    `json:"number" gorm:"index:idx_prow_pull_requests_org_repo_number"`
```

pkg/api/prtestresults.go
```diff
@@ -1,324 +1,142 @@
...
-
-	// Add parameters for includeSuccesses LIKE clauses
-	for i, testName := range includeSuccesses {
-		query.Parameters = append(query.Parameters, bigquery.QueryParameter{
-			Name:  fmt.Sprintf("IncludeSuccess%d", i),
-			Value: "%" + testName + "%", // Wrap in % for SQL LIKE partial matching
+		query = query.Where(conditions)
+	}
+
+	query = query.Order("pjr.timestamp DESC, t.name ASC")
```
mstaeble: There's no result limit or pagination. The default date range is 14 days, and a PR tested many times with many tests could return thousands of rows. The old BigQuery version had a 30-day cap as a safeguard. Consider adding a configurable limit, or at minimum a reasonable ceiling to protect against unexpectedly large result sets.

pkg/api/prtestresults.go
```diff
@@ -1,324 +1,142 @@
...
+	}
+
+	// By default only return failures (no flakes, no successes).
+	// include_successes adds flakes and successes for matching test names.
+	if len(includeSuccesses) == 0 {
+		query = query.Where("pjrt.status = ?", int(sippyprocessingv1.TestStatusFailure))
+	} else {
+		conditions := dbc.DB.Where("pjrt.status = ?", int(sippyprocessingv1.TestStatusFailure))
+		for _, pattern := range includeSuccesses {
+			conditions = conditions.Or("t.name LIKE ?", "%"+pattern+"%")
```
mstaeble: Minor: the OR t.name LIKE '%pattern%' matches regardless of status, so it returns successes, flakes, and failures for matching test names. The old BigQuery version only included successes for matching names (explicitly excluded flakes via adjusted_flake_count = 0). The e2e test confirms this is intentional (statuses["success"]+statuses["flake"] > 0), but it's a behavior change worth noting in the PR description for anyone consuming this API.

mstaeble (commented):
We have an in-flight PR that's relevant here: #3679 (TRT-2734) adds a release_definitions table in PostgreSQL with a Capabilities field and constants like CapPullRequests. It provides the infrastructure to formally define what each release supports.

This PR introduces "Presubmits" as a pseudo-release with the name hardcoded throughout (prow loader routing, PG query filters, seed data, variant overrides). If both PRs land independently, "Presubmits" would exist in prow_jobs.release but not in release_definitions, which could cause issues as the server starts relying on release_definitions to enumerate releases or gate features.

A few things worth considering:

1. "Presubmits" should probably get a ReleaseDefinition row (with CapPullRequests at minimum) so it's a first-class release rather than a magic string that only exists implicitly.
2. The CapPullRequests constant in #3679 is currently unused. This endpoint would be a natural consumer for capability-based gating.
3. The seed data here should eventually seed a "Presubmits" ReleaseDefinition alongside the presubmit jobs/runs.

Not necessarily something to solve in this PR, but wanted to flag the overlap so we can coordinate on how "Presubmits" gets defined as a release. Happy to discuss sequencing.

dgoodwin:
Should be ready for another look.

neisw:
/lgtm
/hold
In case anyone else wants to review / until you are ready to merge

PR #3731, opened 2026-07-03
TRT-2769: WIP: Rework feature gate test filtering to cover component owned jobs

dgoodwin:
- Centralizes the logic for how to find test results for a featuregate
  for re-use in the openshift/api verification job.
- Feature Gate API now returns hateoas links clients can follow.
- Includes a link to the tests tagged for the feature gate name.
- Includes a link to the synthetic "openshift-tests should work" as an
  indicator the job fully passed, for any jobs tied to a Capability.
- Replaces the hack for featuregates containing Install to automatically
  route to all install results and exclude actual tests for the FG.
- Adds feature gate seed data and e2e tests.

## Summary by CodeRabbit

* **New Features**
  * Added a dedicated Feature Gates detail page with tabs for tests by annotation and by capability variant.
  * Feature gate listings now include navigation links to the related test lists and UI detail.
  * Expanded synthetic feature-gate seeding coverage (including NetworkSegmentation and AWSDualStackInstall scenarios).

* **Bug Fixes**
  * Improved Feature Gates navigation so the correct detail page renders for both specific releases and **latest**.

* **Tests**
  * Added end-to-end coverage for the Feature Gates API, including HATEOAS links, filtering, and link follow-through.

* **Chores**
  * Refined end-to-end request helper for absolute URL GETs.

### DIFF
--- a/pkg/sippyserver/server.go
+++ b/pkg/sippyserver/server.go
@@ -10,6 +10,7 @@ import (
 	"net"
 	"net/http"
 	"net/http/httptest"
+	"net/url"
 	"os"
 	"os/signal"
 	"regexp"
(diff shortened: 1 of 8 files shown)


sippy-ng/src/App.js
```diff
@@ -259,16 +259,17 @@ const JobsWrapper = () => {
...
   let { release, feature_gate } = useParams()
   if (release === 'latest') {
     release = findFirstNonGARelease(React.useContext(ReleasesContext))
   }
 
-  return (
-    <Navigate
-      to={pathForTestSubstringByVariant(release, 'FeatureGate:' + feature_gate)}
-      replace
+  return RedirectLatestReleaseWrapper(
```
mstaeble: `RedirectLatestReleaseWrapper` doesn't work for this two-segment route. It calls `useParams()` independently (the manual resolution above doesn't affect the URL), sees `release === 'latest'`, and navigates to `'../' + defaultRelease`. From /feature_gates/latest/NetworkSegmentation that resolves to /feature_gates/4.22, dropping the feature gate segment and landing on the list route. The `findFirstNonGARelease` call above is effectively dead code since the wrapper immediately redirects away.

Fix: replace the wrapper with an explicit navigate that preserves both segments:

```
  const FeatureGateDetailWrapper = () => {
    let { release, feature_gate } = useParams()
    const navigate = useNavigate()

    if (release === 'latest') {
      release = findFirstNonGARelease(React.useContext(ReleasesContext))
      navigate(`/feature_gates/${release}/${feature_gate}`, { replace: true })
    }

    return (
      <FeatureGateDetail
        key={'fg-detail-' + release + '-' + feature_gate}
        release={release}
        featureGate={feature_gate}
      />
    )
  }
```

test/e2e/util/e2erequest.go
```diff
@@ -40,8 +40,12 @@ func BuildE2EURL(apiPath string) string {
 }
 
 func SippyGet(path string, data interface{}) error {
+	return SippyGetAbsolute(BuildE2EURL(path), data)
+}
+
+func SippyGetAbsolute(url string, data interface{}) error {
 	client := &http.Client{Timeout: 30 * time.Second}
-	req, err := client.Get(BuildE2EURL(path)) //nolint:gosec // G704: URL is constructed from test helper's hardcoded localhost base URL
+	req, err := client.Get(url) //nolint:gosec // G704: URL is constructed from test helper's hardcoded localhost base URL
```
mstaeble: nit: This rationale for ignoring the lint complaint is out of date.

pkg/sippyserver/server.go
```diff
@@ -677,10 +678,37 @@ func (s *Server) jsonFeatureGates(w http.ResponseWriter, req *http.Request) {
...
+		}
 		api.RespondWithJSON(http.StatusOK, w, gates)
 	}
 }
 
+func injectFeatureGateHATEOASLinks(fg *apitype.FeatureGate, release, baseAPIURL, baseFrontendURL string) {
+	fg.Links = make(map[string]string, 3)
+
+	annotationFilter := url.QueryEscape(fmt.Sprintf(
+		`{"items":[{"columnField":"name","operatorValue":"contains","value":"FeatureGate:%s]"}]}`,
```
mstaeble: nit: Should use a proper JSON marshaller rather than building the string manually. This would future-proof this against changes to feature gates names that may require JSON escaping.

mstaeble (changes requested):

sippy-ng/src/App.js
dgoodwin: Nice find, thanks.

dgoodwin:
Feedback addressed.

dgoodwin:
/hold

Looks like a refactor problem related to releases in bigquery.

dgoodwin:
make e2e is failing for me locally on a problem with jv_Release in a query, something to do with postgres vs bigquery providers it seems, but I get this failure even on main and presubmits seem to be fine, not sure what it is yet but it's not this PR and I suspect e2e will soon pass here.

mstaeble:
> make e2e is failing for me locally on a problem with jv_Release in a query, something to do with postgres vs bigquery providers it seems, but I get this failure even on main and presubmits seem to be fine, not sure what it is yet but it's not this PR and I suspect e2e will soon pass here.

Very plausibly something that I broke recently.

PR #3740, opened 2026-07-07
replace per-row test ownership upserts with COPY + temp table

mstaeble:
## Summary

- Rewrites the test ownership loader to use COPY + temp table + single upsert SQL instead of ~60,000 individual PostgreSQL round-trips
- COPY raw BigQuery mappings into a temp table, then JOIN against `tests`, `suites`, and `jira_components` to resolve foreign keys server-side in a single `INSERT ... ON CONFLICT` statement
- Stale row deletion uses a single `DELETE ... NOT EXISTS` query instead of collecting IDs and doing `WHERE id NOT IN (huge list)`

## Benchmarks (staging, Go app local, PostgreSQL in AWS RDS)

| Metric | Before | After |
|---|---|---|
| BigQuery mappings | 63,723 | 63,723 |
| Ownership rows upserted | 38,429 | 38,429 |
| PostgreSQL round-trips | ~60,000 | ~5 |
| **Load time** | **3.44 min** | **14 sec** |

## Test plan

- [x] `gofmt`, `go vet`, `golangci-lint` pass
- [x] `make test` passes
- [x] Benchmarked against staging database
- [x] Row count matches prod (difference explained by staging having fewer tests)
- [x] Independent code review found no correctness issues

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved test ownership syncing by switching to a bulk database loading approach for more reliable updates with large datasets.
  * Reduced stale ownership records by ensuring obsolete entries are removed based on the latest mappings.
  * Enhanced error reporting and logging to better surface loading issues and incomplete ownership data.

### DIFF
--- a/pkg/dataloader/testownershiploader/testownershiploader.go
+++ b/pkg/dataloader/testownershiploader/testownershiploader.go
@@ -3,25 +3,24 @@ package testownershiploader
 import (
 	"context"
 	"fmt"
+	"time"
 
+	"github.com/jackc/pgx/v4"
+	"github.com/jackc/pgx/v4/stdlib"
+	v1 "github.com/openshift-eng/ci-test-mapping/pkg/api/types/v1"
 	"github.com/openshift-eng/ci-test-mapping/pkg/bigquery"
-	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
-	"gorm.io/gorm"
-	"gorm.io/gorm/clause"
 
 	"github.com/openshift/sippy/pkg/db"
-	"github.com/openshift/sippy/pkg/db/models"
 )
 
 // TestOwnershipLoader loads test ownership information from BigQuery. This data is generated and
 // pushed to BigQuery from https://github.com/openshift-eng/ci-test-mapping
 type TestOwnershipLoader struct {
-	dbc              *db.DB
-	mappingTableMgr  *bigquery.MappingTableManager
-	errors           []error
-	jiraComponentIDs map[string]uint
-	suiteIDs         map[string]uint
+	ctx             context.Context
+	dbc             *db.DB
+	mappingTableMgr *bigquery.MappingTableManager
+	errors          []error
 }
 
 func New(ctx context.Context, dbc *db.DB, googleServiceAccountCredentialFile, googleOAuthClientCredentialFile string) (*TestOwnershipLoader, error) {
@@ -32,10 +31,9 @@ func New(ctx context.Context, dbc *db.DB, googleServiceAccountCredentialFile, go
 	mappingTableMgr := bigquery.NewMappingTableManager(ctx, client)
 
 	return &TestOwnershipLoader{
-		dbc:              dbc,
-		mappingTableMgr:  mappingTableMgr,
-		jiraComponentIDs: make(map[string]uint),
-		suiteIDs:         make(map[string]uint),
+		ctx:             ctx,
+		dbc:             dbc,
+		mappingTableMgr: mappingTableMgr,
 	}, nil
 }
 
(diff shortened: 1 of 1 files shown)


pkg/dataloader/testownershiploader/testownershiploader.go
```diff
@@ -32,122 +29,149 @@ func New(ctx context.Context, dbc *db.DB, googleServiceAccountCredentialFile, go
...
-				tol.errors = append(tol.errors, errors.WithMessage(res.Error, msg))
-				log.WithError(res.Error).Warning(msg)
-				continue
-			}
-			id := jiraComponent.ID
-			jiraComponentID = &id
-			tol.jiraComponentIDs[m.JIRAComponent] = id
-		}
+	st = time.Now()
+	copied, err := tol.dbc.CopyFrom(tol.ctx, "tmp_test_ownerships", tmpTestOwnershipColumns,
```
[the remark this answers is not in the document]
mstaeble: Good catch. The original code used GORM Exec for CREATE TEMP TABLE and the db.CopyFrom helper (which acquires its own pooled connection) for the COPY, so they could land on different sessions. Fixed by acquiring a single pgx.Conn via stdlib.AcquireConn at the start of loadMappings() and running all operations (CREATE, COPY, upsert, delete, DROP) through it.

pkg/dataloader/testownershiploader/testownershiploader.go
```diff
@@ -32,122 +29,149 @@ func New(ctx context.Context, dbc *db.DB, googleServiceAccountCredentialFile, go
...
+		SELECT
+			tmp.api_version, tmp.unique_id, tmp.name, t.id, tmp.suite, s.id,
+			tmp.product, tmp.priority, tmp.staff_approved_obsolete, tmp.component,
+			tmp.capabilities, tmp.jira_component, jc.id,
+			NOW(), NOW()
+		FROM tmp_test_ownerships tmp
+		INNER JOIN tests t ON t.name = tmp.name AND t.deleted_at IS NULL
+		LEFT JOIN suites s ON s.name = tmp.suite AND s.deleted_at IS NULL
+		LEFT JOIN jira_components jc ON jc.name = tmp.jira_component
+		ON CONFLICT (name, suite) DO UPDATE SET
```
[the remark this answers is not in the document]
mstaeble: Added DISTINCT ON (tmp.name, tmp.suite) ... ORDER BY tmp.name, tmp.suite, tmp.priority DESC to the SELECT so that the highest-priority row wins when BigQuery returns duplicate keys.

pkg/dataloader/testownershiploader/testownershiploader.go
```diff
@@ -32,122 +29,149 @@ func New(ctx context.Context, dbc *db.DB, googleServiceAccountCredentialFile, go
...
+		)
+		AND EXISTS (
+			SELECT 1 FROM tests t
+			WHERE t.id = test_ownerships.test_id AND t.deleted_at IS NULL
+		)
+	`)
+	if deleteResult.Error != nil {
+		log.WithError(deleteResult.Error).Warning("couldn't delete old records")
+		tol.errors = append(tol.errors, deleteResult.Error)
 	}
```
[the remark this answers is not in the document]
mstaeble: The delete path now uses fmt.Errorf("deleting obsolete test_ownerships: %w", err) and returns the error directly (the method signature changed to return error), consistent with the other error paths.

pkg/dataloader/testownershiploader/testownershiploader.go
```diff
@@ -32,122 +31,176 @@ func New(ctx context.Context, dbc *db.DB, googleServiceAccountCredentialFile, go
...
+		)
+		SELECT DISTINCT ON (tmp.name, tmp.suite)
+			tmp.api_version, tmp.unique_id, tmp.name, t.id, tmp.suite, s.id,
+			tmp.product, tmp.priority, tmp.staff_approved_obsolete, tmp.component,
+			tmp.capabilities, tmp.jira_component, jc.id,
+			NOW(), NOW()
+		FROM tmp_test_ownerships tmp
+		INNER JOIN tests t ON t.name = tmp.name AND t.deleted_at IS NULL
+		LEFT JOIN suites s ON s.name = tmp.suite AND s.deleted_at IS NULL
+		LEFT JOIN jira_components jc ON jc.name = tmp.jira_component
```
neisw, line 155: It looks like we would previously skip entries that we failed to find a matching suite / component entry for.  Will this just bring them in with null ids now?  Should we check for these after the fact and log / delete them?

pkg/dataloader/testownershiploader/testownershiploader.go
```diff
@@ -32,122 +31,176 @@ func New(ctx context.Context, dbc *db.DB, googleServiceAccountCredentialFile, go
...
+	if err != nil {
+		return fmt.Errorf("deleting obsolete test_ownerships: %w", err)
 	}
 
 	log.WithFields(log.Fields{
 		"known":    known,
-		"unknown":  unknown,
-		"obsolete": oldRecords.RowsAffected,
-	}).Infof("component loading complete")
+		"unknown":  int64(len(mappings)) - known,
```
neisw: Wondering if we should log a warning when unknown > 0.  It looks like the previous code path would have produced warning entries prior to logging unknown.  Just want to highlight any unexpected gaps where we can.

pkg/dataloader/testownershiploader/testownershiploader.go
mstaeble: The old code skipped the row entirely when a jira_component or suite lookup failed, which meant the ownership record was deleted in the subsequent NOT IN cleanup. The LEFT JOIN approach keeps the ownership row with a null FK instead. This is intentional. t preserves the test-to-component mapping even when the Jira component hasn't been synced yet, and the FK will be filled in on the next load once the component exists. We could add a log event for rows with null jira_component_id if we want more visibility.

pkg/dataloader/testownershiploader/testownershiploader.go
mstaeble: I'll add a warning log when unknown > 0 to maintain the same visibility.

pkg/dataloader/testownershiploader/testownershiploader.go
neisw: That was my main concern, just that previously I think we would have logged a warning for each miss and now we would silently allow it in.  Not sure that anyone is actually reviewing the logs but the entries would be there if someone went looking.

PR #3743, opened 2026-07-07
TRT-2772: Improve agentic solver and review-responder quality

smg247:
## Summary

- **Deduplicate agentic prompts**: Replace standalone `agentic-solve.prompt.md` and `agentic-followup.prompt.md` with small config overlays at `.agentic/{solve,followup}-config.md`. The generic workflow templates now live as heredocs in the step-registry scripts (release repo) and are concatenated with these configs at runtime, removing ~720 lines of duplicated content.
- **Improve review-responder quality**: Rewrite followup prompt with trajectory awareness (check git log before acting), holistic comment synthesis (read all comments together, identify contradictions), follow-existing-patterns guidance, and push-back rules for questions and conflicts.
- **Improve solver quality**: Add simplicity and pattern-reuse guidance to solve prompt.
- **Strengthen testing instructions**: Explicit DI function-type-field pattern and table-driven test preference.
- **Fix MCP server**: Use generic python3 with version check instead of hardcoded python3.12.

## Details

The solve and followup prompts were ~85% identical across repos. The generic workflow (fetch issue, explore codebase, trajectory awareness, holistic comment handling, push-back guidance, security) now lives in the release repo's step-registry heredocs. Only build/test/verify commands and environment info remain in the sippy repo as config overlays.

### New files
- `.agentic/solve-config.md` — build/test commands, MCP tools, environment info, DI pattern guidance
- `.agentic/followup-config.md` — verify/push commands, environment info, DI pattern guidance

### Deleted files
- `.apm/prompts/agentic-{solve,followup}.prompt.md` and all APM-generated copies

Depends on [openshift/release#81565](https://github.com/openshift/release/pull/81565) for the step-registry heredoc templates.

---
Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved active data mode selection to consistently use the startup environment value.
  * Strengthened runtime setup checks to require Python 3.12+ and fail fast when missing/too old.
  * Created virtual environments using the detected `python3` interpreter.

* **Documentation**
  * Expanded testing guidance to prefer table-driven cases and a narrow dependency-injection seam for targeted queries/RPCs.
  * Added standardized local tooling instructions and updated agent workflows for holistic PR review and issue-solving.

* **Chores**
  * Synced prompt/automation metadata and updated associated lockfile hashes.


.agentic/followup-config.md
```diff
@@ -0,0 +1,18 @@
+## Verify and Push
+
+1. Run `make test` and `make lint`.
+2. Run e2e tests using `run_e2e` MCP tool. Must pass before pushing.
```
mstaeble, line 4: Out of curiosity, why do we suggest using the MCP tool for the e2e tests instead of running `make e2e`?

mstaeble (approved):
/lgtm

I see that there is already a hold for the release PR, which needs to land first.

.agentic/followup-config.md
smg247: It is just a thin wrapper for the make target. It tries to resolve some BQ creds (which won't exist here), and then calls `make e2e`. It isn't necessary in this case, but it keeps things consistent for the devcontainer and agentic usage.

mstaeble:
Is there any use case for running the `/agentic-solve` or `/agentic-followup` manually that this change would preclude?

For example, is someone running `/agentic-solve TRT-1234` from their local claude something that we want to support?

smg247:
> Is there any use case for running the `/agentic-solve` or `/agentic-followup` manually that this change would preclude?
> 
> For example, is someone running `/agentic-solve TRT-1234` from their local claude something that we want to support?

I don't see this being effective to use in this manner. It is largely a one-shot process that is inefficient to run locally. If you are working locally, you would want to take advantage of the ability to direct the agent more granularly.

The eventual goal of this is to have any Jira card that is simple enough to be handled by an agent fully implemented without any human interaction outside of the PR review. We aren't quite there yet though.

PR #3768, opened 2026-07-14
Optimize prow loader with bulk writer, COPY protocol, and GCS improvements

mstaeble:
## Summary

Restructures the prow loader from 50 concurrent goroutines each doing their own DB work into a producer/consumer pipeline: 50 GCS fetch workers feed results through a channel to a single bulk writer that batches and commits atomically using PostgreSQL COPY protocol and temp tables.

- **Bulk writer pipeline**: 50 GCS fetch workers -> channel -> single writer batching 100 job runs per transaction. ProwJobRun, annotations, pull requests, and tests all commit atomically per batch.
- **Anti-join pre-filtering**: Identifies new job run IDs via temp table + LEFT JOIN before GCS fetching, eliminating redundant work on steady-state loads.
- **Server-side GCS filtering**: Replaces client-side regex with `MatchGlob` for all `FindAllMatches` callers.
- **Pre-compiled regexes**: Release config regexes compiled once at construction.
- **Batched ProwJob upserts**: 100 per batch instead of per-row.
- **BQ query filter**: Excludes `pending`/`triggered` jobs at the query level.
- **Removed locks and caches**: Per-row mutexes and in-memory caches no longer needed.

## Benchmarks (staging, release 4.19, 1h lookback)

| Scenario | Main | Bulk-Writer | Speedup |
|---|---|---|---|
| Cold load | 108s | 21s | **5.1x** |
| 48h steady-state | ~55s | 43s | 1.3x |

Parity test confirmed: identical run IDs, test counts, and annotations between main and bulk-writer for the same BQ window.

## Test plan

- [x] `make lint` passes
- [x] `make test` passes
- [x] `make e2e` passes
- [x] Parity test against staging DB (identical row counts and per-run test counts)
- [x] Cold load benchmark (48h, 10,895 jobs, 11m25s)
- [x] Steady-state benchmark (48h, 72 new jobs, 43s)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved discovery of job-run artifacts (events, interval files, JUnit reports, and cluster data) across storage paths.
  * More accurate behavior when `events.json` is missing.
  * Excluded non-final job states (pending/triggered) from imported results.
  * Strengthened suite import and test extraction for skipped, flaky, and nested suites.
* **Performance**
  * Faster job-run ingestion via parallel fetching and batched persistence.
* **Tests**
  * Added coverage for test-case extraction scenarios and suite handling.

### DIFF
--- a/pkg/dataloader/prowloader/gcs/gcs_jobrun.go
+++ b/pkg/dataloader/prowloader/gcs/gcs_jobrun.go
@@ -16,43 +16,22 @@ import (
 )
 
 const TestFailureSummaryFilePrefix = "risk-analysis"
-const ClusterDataFilePrefix = "cluster-data_"
-const JunitRegExStr = "\\/.*junit.*xml"
-const intervalFilesRegExStr = "\\/(e2e-events|e2e-timelines).*json"
 
-var (
-	defaultRiskAnalysisSummaryFileRegEx *regexp.Regexp
-	defaultClusterDataFileRegEx         *regexp.Regexp
-	defaultJunitFileRegEx               *regexp.Regexp
-	intervalFilesRegex                  *regexp.Regexp
+const (
+	GlobJunitXML      = "**junit*.xml"
+	GlobEventsJSON    = "**/gather-extra/artifacts/events.json"
+	GlobIntervalsJSON = "**e2e-events*.json"
+	GlobTimelinesJSON = "**e2e-timelines*.json"
+	GlobClusterData   = "**/cluster-data_*.json"
 )
+
+var defaultRiskAnalysisSummaryFileRegEx *regexp.Regexp
 
 func GetDefaultRiskAnalysisSummaryFile() *regexp.Regexp {
 	if defaultRiskAnalysisSummaryFileRegEx == nil {
 		defaultRiskAnalysisSummaryFileRegEx = regexp.MustCompile(fmt.Sprintf("%s.json", TestFailureSummaryFilePrefix))
 	}
 	return defaultRiskAnalysisSummaryFileRegEx
-}
-
-func GetDefaultClusterDataFile() *regexp.Regexp {
-	if defaultClusterDataFileRegEx == nil {
-		defaultClusterDataFileRegEx = regexp.MustCompile(fmt.Sprintf("%s.*json", ClusterDataFilePrefix))
-	}
-	return defaultClusterDataFileRegEx
-}
-
-func GetDefaultJunitFile() *regexp.Regexp {
-	if defaultJunitFileRegEx == nil {
-		defaultJunitFileRegEx = regexp.MustCompile(JunitRegExStr)
-	}
-	return defaultJunitFileRegEx
-}
-
-func GetIntervalFile() *regexp.Regexp {
-	if intervalFilesRegex == nil {
-		intervalFilesRegex = regexp.MustCompile(intervalFilesRegExStr)
-	}
-	return intervalFilesRegex
 }
 
 type GCSJobRun struct {
@@ -76,16 +55,13 @@ func (j *GCSJobRun) SetGCSJunitPaths(paths []string) {
 	j.gcsJunitPaths = paths
 }
 
-func (j *GCSJobRun) GetGCSJunitPaths() ([]string, error) {
+func (j *GCSJobRun) GetGCSJunitPaths(ctx context.Context) ([]string, error) {
 	if len(j.gcsJunitPaths) == 0 {
-		matches, err := j.FindAllMatches([]*regexp.Regexp{GetDefaultJunitFile()})
+		matches, err := j.FindAllMatches(ctx, GlobJunitXML)
 		if err != nil {
 			return nil, err
 		}
-
-		if len(matches) > 0 {
-			j.gcsJunitPaths = matches[0]
-		}
+		j.gcsJunitPaths = matches
 	}
 
 	return j.gcsJunitPaths, nil
@@ -93,7 +69,7 @@ func (j *GCSJobRun) GetGCSJunitPaths() ([]string, error) {
 
 func (j *GCSJobRun) GetCombinedJUnitTestSuites(ctx context.Context) (*junit.TestSuites, error) {
 	testSuites := &junit.TestSuites{}
-	junitPaths, err := j.GetGCSJunitPaths()
+	junitPaths, err := j.GetGCSJunitPaths(ctx)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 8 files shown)


mstaeble:
@coderabbitai full review

pkg/dataloader/prowloader/prow.go
```diff
@@ -1041,36 +1145,265 @@ func (pl *ProwLoader) processGCSBucketJobRun(ctx context.Context, pj *prow.ProwJ
 			OverallResult:  overallResult,
 			TestFailures:   failures,
 			Succeeded:      overallResult == sippyprocessingv1.JobSucceeded,
-			Labels:         labels,
-			Annotations:    annotations,
-		}).Error; err != nil {
-			return err
-		}
+			Labels:         pl.labelsCache[pj.Status.BuildID],
```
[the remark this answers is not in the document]
mstaeble: Good catch. Changed prowJobRunRow.Labels to []string and cast at the assignment site. Fixed.

pkg/dataloader/prowloader/prow.go
```diff
@@ -1041,36 +1145,265 @@ func (pl *ProwLoader) processGCSBucketJobRun(ctx context.Context, pj *prow.ProwJ
...
+		if _, err := tx.Exec(ctx, `
+			INSERT INTO prow_pull_requests (org, repo, link, sha, author, title, number, merged_at, created_at, updated_at)
+			SELECT DISTINCT ON (link, sha) org, repo, link, sha, author, title, number, merged_at, NOW(), NOW()
+			FROM tmp_pull_requests ORDER BY link, sha, merged_at DESC NULLS LAST
+			ON CONFLICT (link, sha) DO UPDATE SET
+				merged_at = COALESCE(EXCLUDED.merged_at, prow_pull_requests.merged_at), updated_at = NOW()
+		`); err != nil {
+			return fmt.Errorf("upserting prow_pull_requests: %w", err)
+		}
+	}
```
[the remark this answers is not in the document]
mstaeble: The old findOrAddPullRequests only set author/title on initial creation (the ErrRecordNotFound branch) and never updated them on subsequent encounters, so this wasn't technically a regression. That said, the suggested improvement is reasonable — added CASE WHEN ... = '' THEN EXCLUDED... ELSE ... for both author and title on the conflict path. Fixed.

pkg/dataloader/prowloader/prow.go
```diff
@@ -1041,36 +1145,265 @@ func (pl *ProwLoader) processGCSBucketJobRun(ctx context.Context, pj *prow.ProwJ
...
+			FROM inserted ins
+			JOIN tests t ON t.id = ins.test_id
+			JOIN tmp_job_run_tests tmp ON tmp.test_name = t.name AND tmp.prow_job_run_id = ins.prow_job_run_id
+			LEFT JOIN suites s2 ON s2.name = tmp.suite_name AND s2.deleted_at IS NULL
+			WHERE tmp.output IS NOT NULL
+				AND ins.suite_id IS NOT DISTINCT FROM s2.id
+		`); err != nil {
+			return fmt.Errorf("inserting prow_job_run_tests: %w", err)
+		}
 	}
```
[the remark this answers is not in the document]
mstaeble: Good find. Changed both the tests and suites inserts from ON CONFLICT (name) DO NOTHING to ON CONFLICT (name) DO UPDATE SET deleted_at = NULL, updated_at = NOW() WHERE ... .deleted_at IS NOT NULL. This ensures re-appearing tests get undeleted and the subsequent INNER JOIN tests ... AND deleted_at IS NULL picks them up. Fixed.

pkg/dataloader/prowloader/prow.go
```diff
@@ -1285,151 +1582,110 @@ func gatherLabelsBatch(ctx context.Context, bqClient *bqcachedclient.Client, bui
...
+			ProwJobRunRelease:   prowJobRelease,
+			TestName:            tc.TestName,
+			SuiteName:           tc.SuiteName,
+			Status:              tc.Status,
+			Duration:            tc.Duration,
+			Output:              tc.Output,
+		})
+		if tc.Status == 12 {
 			failures++
 		}
```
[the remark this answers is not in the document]
mstaeble: Replaced with int(sippyprocessingv1.TestStatusFailure). Fixed.

pkg/dataloader/prowloader/gcs/gcs_jobrun.go
```diff
@@ -16,16 +16,8 @@ import (
...
-const JunitRegExStr = "\\/.*junit.*xml"
-const intervalFilesRegExStr = "\\/(e2e-events|e2e-timelines).*json"
-
-var (
-	defaultRiskAnalysisSummaryFileRegEx *regexp.Regexp
-	defaultClusterDataFileRegEx         *regexp.Regexp
-	defaultJunitFileRegEx               *regexp.Regexp
-	intervalFilesRegex                  *regexp.Regexp
-)
+
```
neisw, line 19: What are your thoughts on keeping the globs defined as constants in a common group?

`**junit*xml` is used in multiple places and keeping the artifact list we pull in grouped together could make it a little easier to maintain / read.

Also noting that the `.` is getting dropped for the junit, not necessarily problematic but was it intentional?

pkg/dataloader/prowloader/prow.go
```diff
@@ -1041,36 +1152,272 @@ func (pl *ProwLoader) processGCSBucketJobRun(ctx context.Context, pj *prow.ProwJ
...
+	for i := range batch {
+		runs = append(runs, batch[i].Run)
+		anns = append(anns, batch[i].Annotations...)
+		prs = append(prs, batch[i].PullRequests...)
+		prAssocs = append(prAssocs, batch[i].PullRequestAssoc...)
+		tests = append(tests, batch[i].Tests...)
+	}
+
+	runCleanup, err := db.CopyToTempTable(ctx, conn, "tmp_prow_job_runs", runs, runCols)
+	defer runCleanup()
```
neisw: code-review:pr prefers

if err != nil { return err} ; defer runCleanup()

multiple places below

pkg/dataloader/prowloader/prow.go
```diff
@@ -598,69 +601,174 @@ ORDER BY
...
+			log.WithError(err).Error("failed to release pgx conn")
+		}
+	}()
+
+	cleanup, err := db.CopyToTempTable(ctx, conn, "tmp_candidate_ids", candidateIDs,
+		[]db.TempColumn[uint]{
+			{Name: "id", Type: "bigint NOT NULL", Value: func(id *uint) any { return *id }},
+		},
+	)
+	defer cleanup()
```
neisw: code-review:pr prefers

if err != nil { return err} ; defer cleanup()

But concern is that [db.CopyToTempTable](https://github.com/mstaeble/sippy/blob/ed02f663f5f0276b01f7a7f86ee3542eff9c8181/pkg/db/pgx.go#L27) might return nil cleanup on error which doesn't appear to be the case.  Feel free to ignore.

pkg/dataloader/prowloader/gcs/gcs_jobrun.go
mstaeble: Good suggestions on both counts. Extracted the globs as named constants (GlobJunitXML, GlobEventsJSON, GlobIntervalsJSON, GlobTimelinesJSON, GlobClusterData) in gcs_jobrun.go and updated all callers. Also added the literal . before extensions in all globs (**junit*.xml, **e2e-events*.json, etc.). The missing dot was unintentional — the original regex (junit.*xml) didn't enforce it either, but it should have.

pkg/dataloader/prowloader/prow.go
mstaeble: Agreed. Restructured all CopyToTempTable call sites to check the error before deferring cleanup:
```
if cleanup, err := db.CopyToTempTable(...); err != nil {
     return err
} else {
    defer cleanup()
}
```

pkg/dataloader/prowloader/prow.go
mstaeble: Same pattern applied here as well.

pkg/dataloader/prowloader/prow.go
mstaeble: Well the linter doesn't like that, so instead it will be the following.

```
cleanup, err := ...
if err != nil {
    return err
}
defer cleanup()
```

PR #3771, opened 2026-07-14
Fix e2e failure due to seed postgres jobs missing release variant

dgoodwin:
fixes 

=== FAIL: test/e2e/componentreadiness TestRegressionCacheLoader (1.27s)
...
    componentreadiness_test.go:86:                                                                                                                                                                                        08:56:29 [83/3805]
                Error Trace:    /Users/dgoodwin/go/src/github.com/openshift/sippy/test/e2e/componentreadiness/componentreadiness_test.go:86
                Error:          Should be empty, but was [component report generation encountered errors: googleapi: Error 400: Unrecognized name: jv_Release; Did you mean release? at [86:844], invalidQuery; googleapi: Error 400: Unrecognized name: jv_Release; Did you mean release? at [86:844], invalidQuery]
                Test:           TestRegressionCacheLoader
                Messages:       regression cache loader had errors: [component report generation encountered errors: googleapi: Error 400: Unrecognized name: jv_Release; Did you mean release? at [86:844], invalidQuery; googleapi: Error
400: Unrecognized name: jv_Release; Did you mean release? at [86:844], invalidQuery]

Failing locally but not in presubmits because GCS_SA_JSON_PATH is not defined in the path the presubmit follows, so the test was skipped. I don't think this was expected.

There's something fishy here, a postgres variants change to fix a bigquery query failure, it appears related to how we cache TestAllVariants, and I'm not totally clear what parts of e2e testing are using postgres vs bigquery providers.

### DIFF
--- a/cmd/sippy/seed_data.go
+++ b/cmd/sippy/seed_data.go
@@ -512,7 +512,12 @@ func seedProwJobs(dbc *db.DB) error {
 	for _, release := range syntheticReleases {
 		for _, job := range syntheticJobs {
 			name := fmt.Sprintf(job.nameTemplate, release)
-			variants := variantMapToArray(job.variants)
+			allVariants := make(map[string]string, len(job.variants)+1)
+			for k, v := range job.variants {
+				allVariants[k] = v
+			}
+			allVariants["Release"] = release
+			variants := variantMapToArray(allVariants)
 			prowJob := models.ProwJob{
 				Kind:     models.ProwKind("periodic"),
 				Name:     name,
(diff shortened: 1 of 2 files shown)


dgoodwin:
@mstaeble tried and failed to fix this here, I'm going to set it aside and we can also just leave things be but just for your awareness there's some info here, it looks like the test in question doesn't even run in CI due to an env var not set.

mstaeble:
> @mstaeble tried and failed to fix this here, I'm going to set it aside and we can also just leave things be but just for your awareness there's some info here, it looks like the test in question doesn't even run in CI due to an env var not set.

See https://github.com/openshift/sippy/pull/3773

PR #3772, opened 2026-07-14
Add NoRegistryClusterInstall capability to variant registry

sadasu:
This change adds Component and Capability tracking for Agent IRI No registry nstaller jobs in the variant registry system.

Jobs matching the pattern '-iso-no-registry` are now classified with:

- Component: Installer
- Capability: NoRegistryClusterInstall
- JobTier: candidate

This enables Sippy to track these jobs separately via the capability-based variant system introduced in PR #3696.

## Summary by CodeRabbit

* **New Features**
  * Added support for identifying no-registry cluster installation jobs and classifying them as installer-related.
  * Updated variant metadata to include the `NoRegistryClusterInstall` capability alongside the installer/agent component.

* **Tests**
  * Added coverage for no-registry, agent-based compact installation jobs and the derived variant attributes, including component/capability matching.


sadasu:
@dgoodwin , @JoelSpeed `verify-feature-promotion` [results](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_api/2859/pull-ci-openshift-api-master-verify-feature-promotion/2077038105440817152) for the [promotion](https://github.com/openshift/api/pull/2859) of  NoRegistryClusterInstall feature is unable to find their [jobs](https://sippy.dptools.openshift.org/sippy-ng/jobs/5.0?filters=%257B%2522items%2522%253A%255B%257B%2522id%2522%253A99%252C%2522columnField%2522%253A%2522name%2522%252C%2522operatorValue%2522%253A%2522contains%2522%252C%2522value%2522%253A%2522iso-no%2522%257D%255D%252C%2522linkOperator%2522%253A%2522and%2522%257D&period=default&sort=asc&sortField=net_improvement).

Proposing this PR to fix that.

/cc @andfasano

dgoodwin:
Did this just happen due to https://github.com/openshift/sippy/pull/3731 merging?

I may have moved too fast there. My change removes the "install should succeed" hack, but it also now requires a full job pass, all tests. That is a major change for the installer. This is consistent with where I want this to go, I'd like jobs like this to map capability as you've done in your PR, but run a reduced and less sensitive set of tests (the main conformance suite is planned to be scaled down to this). These jobs full pass less than 30% of the time, you'll never make it through promotion. Working on what to do now, I think I need to revert my change for the time being. However these jobs can and should map their capability.

pkg/variantregistry/ocp.go
```diff
@@ -767,6 +767,7 @@ var componentCapabilityPatterns = []componentCapabilityEntry{
 	{[]string{"-cpu-partitioning"}, "Node / Kubelet", "CPU Partitioning", "spotcheck-30d"},
 	{[]string{"-etcd-scaling"}, "Etcd", "Scaling", "spotcheck-30d"},
 	{[]string{"-aws-ovn-installer-dualstack"}, "Installer", "AWSDualStackInstall", "candidate"},
+	{[]string{"-iso-no-registry"}, "Installer", "NoRegistryClusterInstall", "candidate"},
```
dgoodwin: The component should match a jira component, which I believe is "Installer / openshift-installer"

andfasano: The exact component type - that we also used to [expose it in CR](https://github.com/openshift-eng/ci-test-mapping/pull/767) - it's `Installer / Agent based installation`

sadasu: Can we have a substring match?

sadasu:
> Did this just happen due to #3731 merging?
> 
> I may have moved too fast there. My change removes the "install should succeed" hack, but it also now requires a full job pass, all tests. That is a major change for the installer. This is consistent with where I want this to go, I'd like jobs like this to map capability as you've done in your PR, but run a reduced and less sensitive set of tests (the main conformance suite is planned to be scaled down to this). These jobs full pass less than 30% of the time, you'll never make it through promotion. Working on what to do now, I think I need to revert my change for the time being. However these jobs can and should map their capability.

Yes, that is a problem. The current problem I am trying to fix is regarding the discoverability of jobs with this featuregate. As part of https://github.com/openshift/api/pull/2859, `veirfy-feature-promotion` gating test is unable to find the jobs in the 1st place.

PR #3778, opened 2026-07-15
TRT-2695: retroactive symptoms (frontend)

sosiouxme:
Use the existing API for re-applying symptoms and make it available wherever there's a JobArtifactQuery dialog; that way when you use the JAQ to modify symptom definitions and see where they would apply, you can then just "make it so".

image
image
image

## Summary by CodeRabbit

- **New Features**
  - Added JAQ “Re-evaluate” action to retroactively match symptoms for selected or all visible job runs.
  - Runs with bounded parallelism and shows execution progress plus success/warning/error snackbars, including optional “reload with fresh data”.
  - Retries recoverable evaluation failures and times out individual requests.
  - Requires write-enabled authentication.

- **Bug Fixes**
  - Improved Prow job identifier consistency across job runs and processing results (now consistently treated as strings).

- **Documentation**
  - Updated the re-evaluation workflow description and completion status.

- **Tests**
  - Added Jest/RTL coverage for UI states, targeting, progress, snackbars, retries, and HTTP failure handling.

### DIFF
--- a/docs/plans/trt-2695-retroactive-symptoms-frontend-plan.md
+++ b/docs/plans/trt-2695-retroactive-symptoms-frontend-plan.md
@@ -31,248 +31,32 @@ component.
 
 Create `sippy-ng/src/jobs/ReEvaluateSymptoms.js`:
 
-```jsx
-import React, { useState } from 'react'
-import {
-  Alert,
-  Button,
-  LinearProgress,
-  Snackbar,
-  Stack,
-  Tooltip,
-  Typography,
-} from '@mui/material'
-import RefreshIcon from '@mui/icons-material/Refresh'
-
-const CONCURRENCY = 10
-const MAX_RETRIES = 1
-
-// Send a single-ID re-evaluate request. Returns the per-run result object.
-async function reEvaluateOne(buildID) {
-  const response = await fetch(
-    process.env.REACT_APP_API_URL + '/api/jobs/runs/reevaluate',
-    {
-      method: 'POST',
-      headers: { 'Content-Type': 'application/json' },
-      body: JSON.stringify({ prow_job_build_ids: [buildID] }),
-    }
-  )
-
-  if (!response.ok) {
-    // The backend returns JSON errors: {"code": N, "message": "..."}
-    let errorMsg = `HTTP ${response.status}`
-    try {
-      const errBody = await response.json()
-      if (errBody.message) errorMsg = errBody.message
-    } catch {
-      // fall back to status text if response isn't JSON
-    }
-
-    if (response.status === 503) {
-      throw new Error(errorMsg)
-    }
-    return {
-      prow_job_build_id: buildID,
-      status: 'eval_error',
-      error: errorMsg,
-    }
-  }
-
-  const data = await response.json()
-  return data.results?.[0] ?? { prow_job_build_id: buildID, status: 'eval_error' }
-}
-
-// Run a pool of workers over a list of IDs, calling onProgress after each.
-async function runPool(ids, onProgress) {
-  const results = []
-  let index = 0
-
-  async function worker() {
-    while (index < ids.length) {
-      const i = index++
-      const result = await reEvaluateOne(ids[i])
-      results.push(result)
-      onProgress([...results])
-    }
-  }
-
-  await Promise.all(
-    Array.from({ length: Math.min(CONCURRENCY, ids.length) }, () => worker())
-  )
-  return results
-}
-
-export default function ReEvaluateButton({
-  prowJobBuildIDs,
-  onComplete,
-  disabled = false,
-}) {
-  const [running, setRunning] = useState(false)
-  const [progress, setProgress] = useState(null) // { total, results: [...] }
-  const [snackbar, setSnackbar] = useState(null) // { severity, message }
-
-  const handleReEvaluate = async () => {
-    if (!prowJobBuildIDs?.length) return
-    setRunning(true)
-    setSnackbar(null)
-
-    const total = prowJobBuildIDs.length
-    setProgress({ total, results: [] })
-
-    try {
-      // Initial pass
-      let results = await runPool(prowJobBuildIDs, (partial) =>
-        setProgress({ total, results: partial })
-      )
-
-      // Retry retryable errors (eval_error, rewrite_error) up to MAX_RETRIES times
-      for (let attempt = 0; attempt < MAX_RETRIES; attempt++) {
-        const retryable = results.filter(
-          (r) => r.status === 'eval_error' || r.status === 'rewrite_error'
-        )
-        if (retryable.length === 0) break
-
-        const kept = results.filter(
-          (r) => r.status !== 'eval_error' && r.status !== 'rewrite_error'
-        )
-        const retryIDs = retryable.map((r) => r.prow_job_build_id)
-        const retryResults = await runPool(retryIDs, (partial) =>
-          setProgress({ total, results: [...kept, ...partial] })
-        )
-        results = [...kept, ...retryResults]
-        setProgress({ total, results })
-      }
-
-      // Summarize
-      const successCount = results.filter((r) => r.status === 'success').length
-      const rewriteErrors = results.filter(
-        (r) => r.status === 'rewrite_error'
-      )
-      const evalErrors = results.filter((r) => r.status === 'eval_error')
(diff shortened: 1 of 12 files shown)


sippy-ng/src/jobs/ReEvaluateSymptoms.js
```diff
@@ -0,0 +1,294 @@
...
+      const i = index++
+      const result = await reEvaluateOne(ids[i])
+      results.push(result)
+      onProgress([...results])
+    }
+  }
+
+  await Promise.all(
+    Array.from({ length: Math.min(CONCURRENCY, ids.length) }, () => worker())
+  )
```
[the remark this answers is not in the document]
sosiouxme: no, i wish for it to continue even if there are errors

sippy-ng/src/jobs/ReEvaluateSymptoms.js
```diff
@@ -0,0 +1,298 @@
...
+        if (retryable.length === 0) break
+
+        const kept = results.filter(
+          (r) => r.status !== 'eval_error' && r.status !== 'rewrite_error'
+        )
+        const retryIDs = retryable.map((r) => r.prow_job_build_id)
+        const retryResults = await runPool(retryIDs, (partial) =>
+          setProgress({ total, results: [...kept, ...partial] })
+        )
+        results = [...kept, ...retryResults]
```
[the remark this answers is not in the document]
sosiouxme: impossible to make the BQ delete-and-write idempotent, but it is fine to have duplicate rows.

sippy-ng/src/component_readiness/JobArtifactQuery.js
```diff
@@ -1775,9 +1783,19 @@ export default function JobArtifactQuery(props) {
...
-      <Stack direction="row" spacing={2}>
+      <Stack direction="row" spacing={2} alignItems="center">
         <JAQOpenJobRunsButton />
         <JAQCopyIdsButton />
+        {capabilitiesContext.includes('write_endpoints') && (
+          <ReEvaluateButton
+            prowJobBuildIDs={(() => {
+              let visible = new Set(filteredRows.map((row) => row.job_run_id))
+              let selected = selectedJobRunIds.intersection(visible)
+              return (selected.size > 0 ? selected : visible).keys().toArray()
```
mstaeble, line 1794: Would it be surprising to the user that everything visible is re-evaluated if nothing is selected? Do we have this behavior so that it is easier for the user to run the re-eval on everything? Could we add a select all button instead, if we don't already have one?

Is there any risk that an accidental click of the re-eval button kicks of a lot of backend work when the user did not intend to since they had not selected anything?

pkg/apis/api/types.go
```diff
@@ -342,7 +342,7 @@ type JobRun struct {
 	Variants              pq.StringArray      `json:"variants" gorm:"type:text[]"`
 	Tags                  pq.StringArray      `json:"tags" gorm:"type:text[]"`
 	TestGridURL           string              `json:"test_grid_url"`
-	ProwID                uint                `json:"prow_id"`
+	ProwID                string              `json:"prow_id"`
```
mstaeble, line 345: Nice change!

sippy-ng/src/jobs/ReEvaluateSymptoms.js
```diff
@@ -0,0 +1,332 @@
...
+        <li key={msg}>
+          {count > 1 ? `(${count}x) ` : ''}
+          {msg}
+        </li>
+      ))}
+    </ul>
+  )
+}
+
+function forceRefreshMessage(successCount, forceRefreshURL) {
```
mstaeble: `successCount` is not used

docs/plans/trt-2695-retroactive-symptoms-frontend-plan.md
```diff
@@ -0,0 +1,256 @@
+# TRT-2695: Frontend Implementation Plan - Re-evaluate Symptoms UI Controls
```
mstaeble, line 1: Do we need to commit implementation plan files?

sippy-ng/src/jobs/ReEvaluateSymptoms.js
```diff
@@ -0,0 +1,332 @@
...
+    const data = await response.json()
+    return (
+      data.results?.[0] ?? { prow_job_build_id: buildID, status: 'eval_error' }
+    )
+  } catch (err) {
+    return {
+      prow_job_build_id: buildID,
+      status: 'eval_error',
+      error:
+        err.name === 'AbortError'
```
mstaeble, line 58: Minor: If the user closes the JAQ dialog while a large batch is in-flight, the pool continues running to completion (or until each request hits the 120s timeout). The requests are harmless, but it's wasted backend activity. Consider accepting an AbortSignal from the parent component (tied to dialog close / component unmount) and checking it before claiming the next ID in the worker loop. Not a blocker.

sippy-ng/src/jobs/ReEvaluateSymptoms.js
```diff
@@ -0,0 +1,332 @@
...
+  } finally {
+    clearTimeout(timeoutId)
+  }
+}
+
+async function runPool(ids, onProgress) {
+  const results = []
+  let index = 0
+
+  async function worker() {
```
mstaeble: Nit: The hand-rolled worker pool is correct, but the shared mutable index takes a moment to reason about. An alternative would be p-limit, which makes the concurrency constraint more declarative:

```
const limit = pLimit(CONCURRENCY)
await Promise.all(ids.map(id => limit(async () => {
  const result = await reEvaluateOne(id)
  results.push(result)
  onProgress([...results])
})))
```

p-limit is already a transitive dependency but not a direct one, so it would need to be added to package.json. Not a blocker: The current approach works fine.

mstaeble (commented):
No blockers. Take a look and see if there is anything that you want to take now or later. Feel free to remove the hold.

/lgtm
/hold

sippy-ng/src/component_readiness/JobArtifactQuery.js
sosiouxme: There's a tooltip on hover that tells them what will happen. There is a select-all button at the top (could be a far scroll away) if they want to be explicit.

The accidental click of the button would kick off stuff, yes... But given that it's idempotent-ish and the buttons are quite large, it didn't seem worthy of "are you sure?"

sippy-ng/src/jobs/ReEvaluateSymptoms.js
sosiouxme: But who would want to close the dialog with such an entertaining progress bar?

There are quite a few more edge cases to consider if this were a common or destructive operation. It's really a question of how airtight you want the experience to be. For me... this is far enough for something that's a rare "true-up" operation.

sippy-ng/src/jobs/ReEvaluateSymptoms.js
sosiouxme: Ah, good point, Claude tends to hand-roll, and I should have thought to ask it to look for an existing package to handle a worker pool. It's almost like it's trained not to add dependencies?

sippy-ng/src/jobs/ReEvaluateSymptoms.js
sosiouxme: and coderabbit didn't complain! you are an uber-coderabbit

docs/plans/trt-2695-retroactive-symptoms-frontend-plan.md
sosiouxme: it seemed weird to me at first when others did. and maybe not? it's worth discussing.

but it's a breadcrumb as to why things are the way they are, which could be useful to humans or agents revising features in the future ("is there some reason this can't work the way i expected it to?"). then again it could just be distracting clutter as the implementation drifts further away. i don't know if anyone has enough experience to say. seems unlikely to be harmful?

sosiouxme:
updated for two of the simplest.

while I don't like leaving even nits in place, i feel like i've been iterating too long already and it's time to get in front of QSE before i'm gone for a while. we can follow up as we feel the need for hygiene or when actual usage gives us the impetus. really the whole UI for this is a bit slapdash and i hope it becomes important enough to completely redesign.

mstaeble (approved):
/lgtm

PR #3800, opened 2026-07-21
NO-JIRA: Fix MCP sippy_ng_start readiness check URL

smg247:
Vite returns 404 for /sippy-ng but 200 for /sippy-ng/ (with trailing slash), causing the readiness check to always time out. Add the trailing slash to all readiness and status URLs.

I discovered this during the migration, but want this separate from the migration to Vite to simplify things.

## Summary by CodeRabbit

- **Bug Fixes**
  - Updated application readiness checks to use the correct trailing-slash URL.
  - Updated start-up status and error messages to consistently display the correct application address.

### DIFF
--- a/mcp/server.py
+++ b/mcp/server.py
@@ -532,16 +532,16 @@ async def sippy_ng_start(
     if existing:
         if not restart:
             pids = ", ".join(str(p) for p in existing)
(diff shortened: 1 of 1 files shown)


mstaeble (approved):
/lgtm

PR #3801, opened 2026-07-21
NO-JIRA: Add CASCADE to migration dropping variant_combination_id

smg247:
Orphaned materialized views (prow_test_report_2d_matview, prow_test_report_7d_matview) depend on the variant_combination_id column, blocking the migration. These matviews are no longer referenced in code. CASCADE drops them automatically.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved database migration reliability by safely removing an obsolete column only when present.
  * Automatically handles related database objects during the migration.

### DIFF
--- a/pkg/db/migrations/000005_drop_daily_summary_variant_combination.up.sql
+++ b/pkg/db/migrations/000005_drop_daily_summary_variant_combination.up.sql
@@ -1,1 +1,1 @@
-ALTER TABLE test_daily_summaries DROP COLUMN IF EXISTS variant_combination_id;
+ALTER TABLE test_daily_summaries DROP COLUMN IF EXISTS variant_combination_id CASCADE;


pkg/db/migrations/000005_drop_daily_summary_variant_combination.up.sql
```diff
@@ -1 +1 @@
-ALTER TABLE test_daily_summaries DROP COLUMN IF EXISTS variant_combination_id;
+ALTER TABLE test_daily_summaries DROP COLUMN IF EXISTS variant_combination_id CASCADE;
```
mstaeble, line 1: I would rather than an explicit drop of the dependencies rather than relying on cascade. But that's me being overly cautious.

Interestingly, I recall having those explicit view drops in this migration at one point. I don't know how it would have ever worked without it.

mstaeble (approved):
/lgtm

PR #3802, opened 2026-07-21
NO-JIRA: Populate regression views when seeding synthetic data

smg247:
The seed data's syncRegressions function was not calling UpsertRegressionView after syncing regressions, so regression_views rows were never created. This caused the "Link additional matching regressions" triage button to fail with "no view provided" because the frontend derives available views from regression view data.

Matches the production code path in regressioncacheloader which calls UpsertRegressionView for each active regression after SyncRegressionsForReport.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved regression tracking by ensuring active regressions are correctly linked to their associated views during synchronization.
  * Added clearer error reporting when regression links cannot be updated.


mstaeble (approved):
/lgtm

PR #3839, opened 2026-07-28
TRT-2856: Extend partition start bound to cover outlier job start times

petr-muller:
## Summary

- `sippy-backup` CronJob pods (fresh-DB `sippy load --init-database --release=4.20`) have been failing repeatedly with:
  ```
  inserting prow_job_run_tests: ERROR: no partition of relation "prow_job_run_tests_p4_20" found for row (SQLSTATE 23514)
  ```
- Root cause: `ensurePartitions()` only creates daily partitions covering `[loadSince-1day, now+2days]`. Jobs that Prow eventually marks `aborted` after getting stuck can report a `StartTime` days before their `completion` time (the field the BigQuery fetch query actually filters on), so such a job can be fetched while its `StartTime` falls outside the partition window. Confirmed one such job directly: `periodic-ci-openshift-release-main-nightly-4.20-e2e-agent-single-node-ipv6-conformance` (build `2076094714775343104`), started 2026-07-12, not marked complete (`aborted`) until 2026-07-15 — an 80 hour gap, well past the existing 1-day grace.
- This isn't a one-off: over the last 30 days, several `aborted` jobs (all releases) had start→completion gaps beyond 24h, with the worst observed around 83h.
- Extend the partition start bound to the earliest `Status.StartTime` actually present in the fetched job set (with the existing `loadSince - 1 day` as a floor), so partitions always cover the real data about to be written, regardless of how anomalous a single job's timestamp is.

This is a follow-up to #3798 (which stops one bad batch from aborting/cancelling the entire load) — this PR addresses the actual partition-range gap that causes the batch to fail in the first place.

## Test plan

- [x] `go build ./pkg/dataloader/prowloader/...`
- [x] `go vet ./pkg/dataloader/prowloader/...`
- [x] `go test ./pkg/dataloader/prowloader/...`
- [x] `golangci-lint run ./pkg/dataloader/prowloader/...` (no new findings)
- [x] Added `TestPartitionStartDate` covering: no jobs, jobs within the grace window, zero-value start times, and an outlier job extending the bound

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved historical job coverage by extending the partition start bound when workflow runs began earlier than the configured load window.
  * Partition selection now considers actual job start times (ignoring zero/unknown values) and falls back to the existing grace-period behavior when none are available.
* **Tests**
  * Added unit coverage for the partition-start calculation, including grace-window, missing data, and earlier-than-window scenarios.

### DIFF
--- a/pkg/dataloader/prowloader/prow.go
+++ b/pkg/dataloader/prowloader/prow.go
@@ -181,12 +181,32 @@ func (pl *ProwLoader) Errors() []error {
 	return pl.errors
 }
 
+// partitionStartDate computes the start of the date range for which partitions must
+// exist to accommodate the given prowJobs. loadSince is used as a floor, with a 1 day
+// grace period since bq imports based on modified time which can include job_run_start_time
+// a day earlier (see https://github.com/openshift/sippy/blob/main/pkg/dataloader/prowloader/prow.go#L473).
+//
+// That grace period covers the common case, but some jobs (e.g. ones Prow eventually marks
+// as aborted after getting stuck) can report a StartTime days before their completion time,
+// which is what the BQ query actually filters on. So the start bound is extended to the
+// earliest job we're actually about to write, to guarantee a partition exists for every row.
+func partitionStartDate(loadSince time.Time, prowJobs []prow.ProwJob) time.Time {
+	startDate := loadSince.AddDate(0, 0, -1)
+	for i := range prowJobs {
+		if st := prowJobs[i].Status.StartTime; !st.IsZero() && st.Before(startDate) {
+			startDate = st
+		}
+	}
+	return startDate
+}
+
 // ensurePartitions creates necessary partitions for partitioned tables.
 // It uses the release list from pl.releases and determines the date range based on:
-//   - pl.loadSince if available, otherwise looks back one week
+//   - pl.loadSince if available, otherwise looks back one week, plus a 1 day grace period
+//   - the earliest prowJobs StartTime, in case it falls outside the above window
 //   - Creates partitions 2 days forward from now
-func (pl *ProwLoader) ensurePartitions() error {
-	startDate := pl.resolveLoadSince()
+func (pl *ProwLoader) ensurePartitions(prowJobs []prow.ProwJob) error {
+	startDate := partitionStartDate(pl.resolveLoadSince(), prowJobs)
 
 	// Create partitions 2 days forward from now
 	endDate := time.Now().AddDate(0, 0, 2)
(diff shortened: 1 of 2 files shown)


pkg/dataloader/prowloader/prow.go
```diff
@@ -181,22 +181,40 @@ func (pl *ProwLoader) Errors() []error {
 	return pl.errors
 }
 
+// partitionStartDate computes the start of the date range for which partitions must
+// exist to accommodate the given prowJobs. loadSince is used as a floor, with a 1 day
```
petr-muller-reviewer: nit: "loadSince is used as a floor" is misleading here. The function uses `loadSince - 1day` as the *default* start date and only moves it *earlier* (toward smaller values), so `loadSince` acts as the upper bound on the default, not a lower bound.

Suggestion: "loadSince (minus a 1-day grace) is the default start date, extended earlier if any job's StartTime precedes it."

pkg/dataloader/prowloader/prow.go
```diff
@@ -181,22 +181,40 @@ func (pl *ProwLoader) Errors() []error {
...
+			startDate = st
+		}
+	}
+	return startDate
+}
+
 // ensurePartitions creates necessary partitions for partitioned tables.
 // It uses the release list from pl.releases and determines the date range based on:
-//   - pl.loadSince if available, otherwise looks back one week
+//   - pl.loadSince if available, otherwise looks back one week, plus a 1 day grace period
```
petr-muller-reviewer: nit: "otherwise looks back one week" — `resolveLoadSince()` falls back to `DefaultLookbackDays = 14`, not 7. Worth correcting while touching these comments.

pkg/dataloader/prowloader/prow.go
```diff
@@ -181,22 +181,40 @@ func (pl *ProwLoader) Errors() []error {
...
+// which is what the BQ query actually filters on. So the start bound is extended to the
+// earliest job we're actually about to write, to guarantee a partition exists for every row.
+func partitionStartDate(loadSince time.Time, prowJobs []prow.ProwJob) time.Time {
+	startDate := loadSince.AddDate(0, 0, -1)
+	for i := range prowJobs {
+		if st := prowJobs[i].Status.StartTime; !st.IsZero() && st.Before(startDate) {
+			startDate = st
+		}
+	}
+	return startDate
```
petr-muller-reviewer, line 201: question: When `partitionStartDate` returns a date earlier than `loadSince - 1day`, it would help operators to log that the bound was extended (e.g. the delta or the earliest outlier time). Currently the only visibility is the log at line 214 showing the final range, but not whether it was extended beyond the default.

petr-muller-reviewer (approved):
Solid, well-targeted fix. Adapting the partition window to actual data instead of guessing a larger grace period is the right design. Code is clean, well-tested, and follows project conventions. Two nits on comment accuracy, one observability question.

petr-muller:
/retitle [TRT-2856](https://redhat.atlassian.net/browse/TRT-2856): Extend partition start bound to cover outlier job start times

mstaeble (approved):
/lgtm

PR #3843, opened 2026-07-29
TRT-2821: Use prefix_max_last_failure instead of LATERAL join for CR timestamps

mstaeble:
## Summary

- Replace the `lastFailureLateral` LATERAL join (which scanned `prow_job_run_tests` partitions per test) with an inline `MAX(e.prefix_max_last_failure)` from `test_cumulative_summaries`, the table already being queried for counts.
- For the GA base path, return `NULL::timestamptz` since only sample `LastFailure` is consumed downstream (by the regression tracker and component report).
- Add a `lastFailureExpr` field to `testStatusSpec` so both prefix-sum and GA paths share the same query builder.
- Add integration tests covering MAX aggregation across multiple jobs, mixed NULL/non-NULL timestamps, and GA base path behavior.

## Test plan

- [x] `go vet` and `go build` pass
- [x] Unit tests pass (`go test ./pkg/api/componentreadiness/...`)
- [x] Integration tests pass (`go test ./test/integration/...`)
- [x] Staging deployment and manual verification against staging-2

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved component readiness results to report the most recent test failure date accurately.
  * Corrected failure-date handling when multiple jobs are grouped together.
  * Preserved available failure information when some grouped jobs have no recorded failure.
  * Ensured cumulative summaries include failures occurring within the relevant reporting window.
  * GA results no longer display an unsupported last-failure timestamp.

### DIFF
--- a/pkg/api/componentreadiness/dataprovider/postgres/cr_queries.go
+++ b/pkg/api/componentreadiness/dataprovider/postgres/cr_queries.go
@@ -15,7 +15,6 @@ import (
 	"github.com/openshift/sippy/pkg/apis/api/componentreport/crstatus"
 	"github.com/openshift/sippy/pkg/apis/api/componentreport/crtest"
 	"github.com/openshift/sippy/pkg/apis/api/componentreport/reqopts"
-	sippyv1 "github.com/openshift/sippy/pkg/apis/sippyprocessing/v1"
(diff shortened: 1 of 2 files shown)


mstaeble:
@coderabbitai full review

mstaeble:
@coderabbitai full review

PR #3863, opened 2026-08-03
Preserve original prow job names in PG provider

mstaeble:
## Summary
- Fixed the PG data provider overwriting the `ProwJob` field on `TestJobRunRows` with the normalized name (`X.X`) instead of preserving the original version number (e.g., `4.22`). The BQ provider already preserved the original name correctly.
- The normalized name is only needed as the map key for cross-release grouping. The `ProwJob` field flows to the test details API response (displayed on the frontend) and into `RegressionJobRun` records, so it should retain the literal version number.
- Added integration test assertions verifying that original job names are preserved in result rows even when grouped under a normalized key.

## Test plan
- [x] `gofmt -w` on modified files
- [x] `go vet` passes
- [x] Unit tests pass (`go test ./pkg/api/componentreadiness/...`)
- [x] All 133 integration tests pass (`make integration`)

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved the original test job names in component readiness results while continuing to combine equivalent versioned jobs correctly.
  * Improved result details so users can distinguish the specific job runs contributing to merged results.

* **Tests**
  * Added coverage validating result merging and preservation of original job names.

### DIFF
--- a/pkg/api/componentreadiness/dataprovider/postgres/provider.go
+++ b/pkg/api/componentreadiness/dataprovider/postgres/provider.go
@@ -458,7 +458,7 @@ WHERE pj.release = ?
 			TestKey:         key,
 			TestKeyStr:      key.Encode(),
 			TestName:        row.TestName,
-			ProwJob:         normalizedName,
+			ProwJob:         row.ProwJobName,
 			ProwJobRunID:    row.ProwJobRunID,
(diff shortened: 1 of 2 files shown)


smg247 (commented):
/lgtm

PR #3871, opened 2026-08-04
Port feature gate promotion logic from o/api to sippy

dgoodwin:
This change unifies sippy's feature gate details page with the logic that was rapidly evolving in the o/api presubmit. Teams will now be able to check their promotion status without opening an o/api PR and waiting for a presubmit to run. The sippy version runs fast enough for live queries. o/api presubmit can use the new API to check validity and just link to this page for results.

New logic includes:
- support for jobs owned by a feature gate (variant Capability=FeatureGateName)
- owned jobs must have all tests passing at least 92% (allowing one but not two failures in the min 14 runs)
- tests owned by other techpreview gates do not count against the above
- api response has hateoas for approximate raw data used for determination (owned job 92% required a custom query for performance reasons)
- ui exposes three tabs for the raw data
- owned jobs are clearly displayed with a link to job analysis
- adds unit, e2e, and seed data coverage

## Summary by CodeRabbit

* **New Features**
  * Added detailed feature-gate pages with promotion readiness, test results, install-gate checks, capability regressions, and owned-job analysis.
  * Added sortable variant results, warnings, errors, thresholds, and filtered analysis links.
  * Added direct API endpoints and navigation links for feature-gate details.
  * Feature-gate lists now include matching jobs and simplified detail links.

* **Documentation**
  * Documented Feature Gates API list and detail endpoints, navigation links, and promotion readiness data.

* **Tests**
  * Added comprehensive coverage for promotion evaluation, regressions, filtering, and detail endpoints.


pkg/api/featuregatepromotion/promotion.go
```diff
@@ -0,0 +1,548 @@
...
+
+	excludedNames := []string{"install should succeed", "openshift-tests should work", "infrastructure should work"}
+
+	type queryResult struct {
+		TestName          string  `gorm:"column:test_name"`
+		WorkingPercentage float64 `gorm:"column:working_percentage"`
+	}
+
+	var rows []queryResult
+	tx := dbc.DB.Raw(`
```
dgoodwin, line 194: My least favourite part of this PR is this forked query. It was too slow using the normal test functions and filters. This is true if you follow the gate_job_tests hateoas link as well.

pkg/api/featuregatepromotion/promotion.go
```diff
@@ -0,0 +1,548 @@
...
+		WHERE e.date = ? AND e.release = ?
+			AND ? = ANY(pj.variants)
+			AND NOT EXISTS (
+				SELECT 1 FROM variant_combinations
+				WHERE ('never-stable' = ANY(variants) OR 'aggregated' = ANY(variants))
+				AND id = pj.variant_combination_id
+			)
+			AND t.name NOT LIKE '%' || ? || '%'
+			AND t.name NOT LIKE '%' || ? || '%'
+			AND t.name NOT LIKE '%' || ? || '%'
```
dgoodwin, line 219: When I grilled claude:

⏺ The three lines look identical in the SQL but each ? binds to a different value from excludedNames:

  excludedNames := []string{"install should succeed", "openshift-tests should work", "infrastructure should work"}

  So it's three distinct filters, one per excluded test name. The SQL just looks repetitive because the pattern t.name NOT LIKE '%' || ? || '%' is the same shape for each. It could be cleaner as a single NOT EXISTS with an unnest or a
  NOT (t.name LIKE ANY(...)), but functionally it's correct as-is.

pkg/api/featuregatepromotion/types.go
```diff
@@ -0,0 +1,55 @@
+package featuregatepromotion
+
+// PromotionStatus represents the overall promotion readiness of a feature gate.
+type PromotionStatus struct {
+	FeatureGate              string                     `json:"feature_gate"`
+	Release                  string                     `json:"release"`
+	Sufficient               bool                       `json:"sufficient"`
+	ResultsByVariant         []VariantResult            `json:"results_by_variant"`
+	CapabilityTestRegessions []CapabilityTestRegression `json:"capability_test_regressions,omitempty"`
```
neisw: CapabilityTestRegessions → CapabilityTestRegressions

pkg/apis/api/types.go
```diff
@@ -1053,15 +1053,57 @@ type SippyViews struct {
...
+	MatchingJobs     []string              `json:"matching_jobs" gorm:"-"`
+	Promotion        *FeatureGatePromotion `json:"promotion,omitempty" gorm:"-"`
+	Links            map[string]string     `json:"links,omitempty" gorm:"-"`
+}
+
+// FeatureGatePromotion represents promotion readiness data included in the detail response.
+type FeatureGatePromotion struct {
+	Sufficient               bool                                  `json:"sufficient"`
+	ResultsByVariant         []FeatureGateVariantResult            `json:"results_by_variant"`
+	CapabilityTestRegessions []FeatureGateCapabilityTestRegression `json:"capability_test_regressions,omitempty"`
```
neisw: CapabilityTestRegessions → CapabilityTestRegressions

dgoodwin: I'm confused how that typo is possible, and how you spotted it, I had to go char by char to see that even with you pointing it out. :) Addressed now.

neisw: It wasn't me...

pkg/api/featuregatepromotion/promotion.go
neisw: Let's take a todo on the performance here and analyze what you need to see if it is a query structure thing vs. schema.  Not necessary for today though.

PR #3872, opened 2026-08-04
TRT-2887: Add `/api/jobs/runs/disruption` endpoint for per-run backend disruption data

smg247:
Queries the BigQuery BackendDisruption table to return disruption seconds per backend for given job run IDs. This enables tooling to identify which CI runs had disruption on specific backends  without downloading full interval artifacts. The current plan is to use this endpoint within the `ci:analyze-disruption` skill within ai-helpers.

Parameters:
- job_run_names (required): comma-separated prow build IDs
- backend_name (optional): filter by backend name pattern (LIKE match)

## Summary by CodeRabbit

* **New Features**
  * Added an API endpoint to retrieve backend disruption details for specified job runs.
  * Supports optional filtering by backend name and time range.
  * Results include disruption duration, job-run metadata, cluster, release, node-update, and status information.
  * Added validation for disruption query parameters, one-hour response caching, and rate limiting.
* **Bug Fixes**
  * Improved handling of optional and empty result fields in API responses.


pkg/util/param/param.go
```diff
@@ -79,6 +79,9 @@ var paramRegexp = map[string]*regexp.Regexp{
 	"includeOutputs": boolRegexp,
 	// pull request test results params
 	"limit": uintRegexp,
+	// disruption params
+	"job_run_names": regexp.MustCompile(`^\d+(,\d+)*$`),
+	"backend_name":  regexp.MustCompile(`^[\w-]+$`),
```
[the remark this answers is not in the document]
smg247: no existing param enforces list length

smg247:
@coderabbitai can you do a review and close everything addressed?

pkg/api/backend_disruption.go
```diff
@@ -0,0 +1,115 @@
...
+    JobName,
+    JobRunName,
+    CAST(JobRunStartTime AS STRING) AS JobRunStartTime,
+    CAST(JobRunEndTime AS STRING) AS JobRunEndTime,
+    Cluster,
+    ReleaseTag,
+    MasterNodesUpdated,
+    JobRunStatus
+FROM ` + "`openshift-ci-data-analysis.ci_data.BackendDisruption`" + `
+WHERE JobRunName IN UNNEST(@JobRunNames)` + filterStr + `
```
neisw: Oddly enough it doesn't look like a massive table but it is 
Partitioned on field JobRunStartTime 

could we find the min/max JobRunStartTime for and add that range to the where clause?

smg247: I am updating it to get the info from postgres and then pass it in the WHERE clause. This should be the most performant way.

PR #3874, opened 2026-08-05
[WIP] TRT-2847: Fix aggregate test details variant filter excluding valid jobs

mstaeble:
@coderabbitai ignore

## Summary

- Fixes a bug where the test_details aggregate fallback returned zero base stats for older release overrides (e.g. 4.21 in the 5.0-main view)
- The SQL variant filter included non-dbGroupBy keys (JobTier, CGroupMode, ContainerRuntime, Owner) from the view's `includeVariants`, which combined to exclude all matching jobs for older releases
- Filters `includeVariants` to only dbGroupBy keys before building the SQL WHERE clause; Go-side `matchRequestedVariants` in `processAggregateRows` handles per-variant matching

## Root cause

For release 4.21, the only jobs with complete variant_combinations (including Suite, CGroupMode, LayeredProduct) had `JobTier:candidate`, which the view's `includeVariants` excludes (only `blocking`, `informing`, `standard`). Jobs with allowed JobTier values lacked the newer variant keys. The AND combination of all filter clauses excluded every job.

## Test plan

- [x] Unit test for `filterVariantsByDBGroupBy` (included)
- [x] Existing unit tests pass (`go test ./pkg/api/componentreadiness/...`)
- [x] Verify against staging DB: fixed SQL returns 2 rows for 4.21 base override (confirmed)
- [x] Deploy to staging and verify the test_details API returns non-zero base stats for the 4.21 override analysis with `dataSource=postgres`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved aggregate readiness filtering by applying only supported dimensions, preventing unsupported filters from affecting database queries.

* **Tests**
  * Added coverage for matching, non-matching, empty, and multiple variant filter scenarios.

### DIFF
--- a/pkg/api/componentreadiness/dataprovider/postgres/provider.go
+++ b/pkg/api/componentreadiness/dataprovider/postgres/provider.go
@@ -529,6 +529,7 @@ func (p *PostgresProvider) queryBaseAggregateTestDetails(ctx context.Context, re
 		includeVariants = map[string][]string{}
 	}
 	includeVariants = mergeRequestedVariants(includeVariants, reqOptions)
+	includeVariants = filterVariantsByDBGroupBy(includeVariants, reqOptions.VariantOption.DBGroupBy)
 
 	testIDs := make([]string, 0, len(reqOptions.TestIDOptions))
 	for _, tid := range reqOptions.TestIDOptions {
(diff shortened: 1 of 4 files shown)


mstaeble:
/close

This is not the correct solution. The older jobs should have their variants updated to include the variants that were recently added to the important variants set.

PR #3852, opened 2026-07-30
TRT-2848: Refresh summary tables incrementally during prow load

mstaeble:
## Summary

- Refreshes `test_daily_totals` and `test_cumulative_summaries` incrementally as part of each prow load batch, replacing the separate full-refresh step that re-scanned the entire `prow_job_run_tests` table
- Extracts batch writing logic from prow.go into a dedicated `pgwriter` package with focused, single-responsibility functions for each SQL step
- Parallelizes carry-forward across releases (4 workers via `errgroup.Group`) to reduce carry-forward time from ~3 minutes to ~50 seconds
- Pins `currentDate` at load start and threads it through all SQL to prevent midnight-crossing corruption
- Adds 32 integration tests covering job runs, annotations, PRs, test results, daily/cumulative summaries, carry-forward, timestamp tracking, and edge cases

## Staging validation

Tested against the staging database with multiple lookback windows:

- **5-minute lookback**: ~69 prow jobs, completed in under 2 minutes including summary upserts
- **1-hour lookback**: 17-23 jobs with 30-43K tests, completed in ~1 minute steady-state
- **12-hour lookback (cold start simulation)**: 364 candidate jobs, completed successfully with correct daily totals and cumulative summaries
- **Carry-forward parallelization**: reduced from ~3 minutes (sequential) to ~50 seconds (4 workers) on staging

## Test plan

- [x] `make lint` passes (0 issues)
- [x] `go test ./pkg/...` passes
- [x] `make integration` passes (150 tests)
- [x] Verified against staging with multiple lookback windows (5m, 1h, 12h)
- [ ] Compare daily totals and cumulative summaries against full-refresh baseline

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## New Features
- Improved persistence of job runs, annotations, pull requests, test results, and test outputs.
- Daily and cumulative summaries now remain consistent across batches, releases, suites, and missed dates.
- Historical summaries automatically carry forward to fill eligible gaps.
- Added explicit date-range backfilling for daily and cumulative summaries.
- Pull-request metadata is preserved when records are reprocessed.

## Bug Fixes
- Improved handling of duplicates, soft-deleted records, empty batches, and invalid future-dated results.
- Refresh operations now focus on materialized views while summary updates occur during data loading.

PR #3880, opened 2026-08-05
TRT-2848: Drop unused idx_test_daily_totals_date index

mstaeble:
## Summary

- Drops `idx_test_daily_totals_date`, a single-column btree index on `test_daily_totals(date)` that existed only to make an unscoped `SELECT MAX(date) FROM test_daily_totals` fast across ~3,700 partitions.
- #3852 ([TRT-2848](https://redhat.atlassian.net/browse/TRT-2848)) removed the last two callers of that query pattern (`dailysummary.MaxSummaryDate` and `cumulativesummary.MaxDailySummaryDate`), which is why this PR was held until that one merged. With those gone, the index is pure write overhead on every insert into `test_daily_totals`, with no remaining reader.
- A full-codebase search confirmed no other query touches `date` on `test_daily_totals`/`test_cumulative_summaries` without a `release` predicate, so nothing else depends on this index.

## Test plan

- [x] `make lint` / `go build ./pkg/db/...`
- [x] `./hack/verify-migrations.sh` passes
- [x] Applied migration up/down against a fresh Postgres 16 instance seeded with the full migration history; confirmed `idx_test_daily_totals_date` is dropped on `up` and exactly recreated on `down` (verified via `\d test_daily_totals`)
- [x] Confirmed against a read-only inspection of staging that the index's current definition (`btree (date)`) matches what this migration expects to drop/recreate

## Summary by CodeRabbit

* **Chores**
  * Updated database migrations to remove an unnecessary index and keep the migration history in sync.
  * Added a rollback path to restore the index if needed.

### DIFF
--- a/pkg/db/migrations/000012_drop_test_daily_totals_date_index.down.sql
+++ b/pkg/db/migrations/000012_drop_test_daily_totals_date_index.down.sql
@@ -1,0 +1,2 @@
+CREATE INDEX IF NOT EXISTS idx_test_daily_totals_date
+    ON test_daily_totals (date);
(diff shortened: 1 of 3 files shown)


smg247 (commented):
/lgtm

PR #3884, opened 2026-08-06
Combined CR query with materialized CTEs

mstaeble:
## Summary
- Folds sample and base `component_readiness` queries into a single SQL statement using two materialized CTEs and UNION ALL branches, eliminating duplicate query planning and table scans.
- Simplifies `getTestStatus` orchestration: the combined query returns both sides in one call, removing goroutine/channel coordination for the common case.
- Adds comprehensive `GenerateReport` integration tests covering prefix-sum aggregation, GA path, mixed lifecycle filtering, disjoint variants between base and sample, capabilities array-overlap filtering, and grid placeholder merging.

## Benchmark results

Tested locally against the staging database (3 runs each, averaged, with warm-up):

| Scenario | Baseline | Combined | Speedup |
|---|---|---|---|
| 5.0-main | 8.57s | 4.19s | 2.04x |
| 4.22-main | 6.91s | 2.91s | 2.37x |
| 5.0-rosa | 0.57s | 0.77s | 0.74x |
| 5.0-hypershift | 1.01s | 1.02s | 0.99x |
| 5.0-ha-vs-single | 5.21s | 2.04s | 2.55x |
| 5.0-x86-vs-arm | 9.14s | 3.55s | 2.57x |
| Etcd-filter | 8.34s | 4.19s | 1.99x |
| Platform-aws | 5.29s | 3.63s | 1.45x |
| ColGroupBy-Topology | 8.27s | 4.18s | 1.97x |
| test_details drilldown | 0.80s | 0.76s | 1.05x |

The combined query provides ~2x speedup on large views and up to 2.6x on cross-variant views. Small views (ROSA, Hypershift) with fewer variant groups show no meaningful change. The test_details drilldown path (which still uses the standalone query) is not regressed.

## Test plan
- [x] Existing integration tests pass (`make integration`)
- [x] New `GenerateReport_*` and `TestCapabilitiesArrayOverlapFilter` integration tests pass
- [x] Benchmark against staging to verify combined query performance matches or improves on standalone
- [x] Verify `test_details` drill-down path still works (standalone query path)
- [x] Verify releasefallback middleware still works (uses `QueryBaseTestStatus`)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved component-readiness report accuracy for release isolation, lifecycle filtering, variant comparisons, missing data, and failure thresholds.
  * Improved handling of overlapping and disjoint capability variants.
  * Prevented incomplete status retrievals from producing misleading results.

* **Performance**
  * Streamlined status collection and report generation.

* **Observability**
  * Added structured completion details, including report duration and result counts.

* **Tests**
  * Expanded coverage for report generation, filtering, aggregation, lifecycle data, and drill-down scenarios.

### DIFF
--- a/pkg/api/componentreadiness/dataprovider/postgres/cr_queries.go
+++ b/pkg/api/componentreadiness/dataprovider/postgres/cr_queries.go
@@ -4,8 +4,11 @@ import (
 	"context"
 	"database/sql"
 	"fmt"
-	"sync"
+	"maps"
+	"strings"
+	"time"
 
+	"cloud.google.com/go/civil"
 	"github.com/lib/pq"
 	"gorm.io/gorm"
 	"k8s.io/apimachinery/pkg/util/sets"
@@ -28,6 +31,8 @@ type variantQuerySetup struct {
 	minimumFailure  int
 }
 
+const queryPlannerHints = "SET LOCAL max_parallel_workers_per_gather = 4; SET LOCAL parallel_setup_cost = 0; SET LOCAL parallel_tuple_cost = 0; SET LOCAL enable_nestloop = off; SET LOCAL enable_sort = off"
+
 func prepareVariantQuery(
 	ctx context.Context,
 	dbc *db.DB,
@@ -35,31 +40,18 @@ func prepareVariantQuery(
 	dbGroupBy sets.Set[string],
 	minimumFailure int,
 ) (*variantQuerySetup, error) {
-	if includeVariants == nil {
-		includeVariants = map[string][]string{}
-	}
-
-	variantLookup, err := lookupVariantValues(ctx, dbc, includeVariants, dbGroupBy)
+	vf, err := resolveVariantFilter(ctx, dbc, includeVariants, dbGroupBy)
 	if err != nil {
 		return nil, err
 	}
-	if len(variantLookup) == 0 {
+	if len(vf.lookup) == 0 {
 		return nil, nil
 	}
 
-	groupMapping := buildVariantGroupMapping(variantLookup)
-
-	filterClause, filterArgs := buildVariantFilterClause(includeVariants)
-
-	variantSubquery := "SELECT vc.id FROM variant_combinations vc"
-	if filterClause != "" {
-		variantSubquery += " WHERE " + filterClause
-	}
-
 	return &variantQuerySetup{
-		groupMapping:    groupMapping,
-		filterArgs:      filterArgs,
-		variantSubquery: variantSubquery,
+		groupMapping:    buildVariantGroupMapping(vf.lookup),
+		filterArgs:      vf.filterArgs,
+		variantSubquery: vf.variantSubquery,
 		minimumFailure:  minimumFailure,
 	}, nil
 }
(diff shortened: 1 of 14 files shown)


pkg/api/componentreadiness/dataprovider/postgres/cr_queries.go
```diff
@@ -404,6 +364,235 @@ func (p *PostgresProvider) scanWithParallelHints(
...
+	if err != nil {
+		return nil, nil, []error{err}
+	}
+	baseLookup, err := lookupVariantValues(ctx, p.dbc, baseIncludeVariants, dbGroupBy)
+	if err != nil {
+		return nil, nil, []error{err}
+	}
+	if len(sampleLookup) == 0 || len(baseLookup) == 0 {
+		return map[string]crstatus.TestStatus{}, map[string]crstatus.TestStatus{}, nil
+	}
```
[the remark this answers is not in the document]
mstaeble: Good catch. Fixed in the latest force-push: changed `||` to `&&` so each side is evaluated independently. When only one side's lookup is empty, the combined query still runs — the empty side's CTE simply returns zero rows, so the report correctly renders MissingBasis or MissingSample. Also added a dedicated integration test (`TestGenerateReport_EmptyBaseLookupStillReturnsSampleResults`) that sets up a cross-compare view where the base-side Topology has no matching variant combinations, and verifies sample results still come through with MissingBasis status.

test/integration/component_readiness_test.go
```diff
@@ -3444,6 +3518,668 @@ func TestTestDetailsReport_FlakeAsFailure(t *testing.T) {
...
+
+	// Test that exists in sample but not base should show MissingBasis
+	newRow := findReportRow(t, report, "NewComponent")
+	newCol := findReportColumn(t, newRow, map[string]string{"Platform": "aws"})
+	assert.Equal(t, crtest.MissingBasis, newCol.Status)
+
+	// Test that exists in both should be assessed normally
+	sharedRow := findReportRow(t, report, "SharedComponent")
+	sharedCol := findReportColumn(t, sharedRow, map[string]string{"Platform": "aws"})
+	assert.GreaterOrEqual(t, int(sharedCol.Status), int(crtest.NotSignificant))
```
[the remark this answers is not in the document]
mstaeble: Fixed. Added explicit `assert.NotEqual(t, crtest.MissingBasis, ...)` and `assert.NotEqual(t, crtest.MissingSample, ...)` checks at both locations (lines 3767-3768 and 3960-3961), alongside the existing `GreaterOrEqual` assertion. A regression that drops base or sample rows will now fail the test instead of silently passing.

pkg/api/componentreadiness/dataprovider/postgres/cr_queries.go
```diff
@@ -404,6 +364,236 @@ func (p *PostgresProvider) scanWithParallelHints(
...
+	baseVarSubquery := "SELECT vc.id FROM variant_combinations vc"
+	if baseFilterClause != "" {
+		baseVarSubquery += " WHERE " + baseFilterClause
+	}
+
+	prowJobJoinTemplate := `JOIN prow_jobs pj ON pj.id = e.prow_job_id
+                AND pj.deleted_at IS NULL
+                AND pj.variant_combination_id IN (%s)
+            JOIN vg ON vg.vcid = pj.variant_combination_id`
```
petr-muller: template is identical to L246 - would it make sense to extract?

pkg/api/componentreadiness/dataprovider/postgres/cr_queries.go
```diff
@@ -404,6 +364,236 @@ func (p *PostgresProvider) scanWithParallelHints(
...
+	sampleProwJobJoin := fmt.Sprintf(prowJobJoinTemplate, sampleVarSubquery)
+	baseProwJobJoin := fmt.Sprintf(prowJobJoinTemplate, baseVarSubquery)
```
petr-muller: I have an LLM that thinks this partially duplicates logic from `prepareVariantQuery`, I had it propose a fix in https://github.com/petr-muller-reviewer/sippy/commit/5ab02afb18e93b1c74ed1638a720d8be530e2b70

pkg/api/componentreadiness/dataprovider/postgres/cr_queries.go
```diff
@@ -107,32 +112,120 @@ func buildDrilldownFilters(reqOptions reqopts.RequestOptions) drilldownFilters {
...
+		lifecycleClause = "\n                AND e.lifecycle = ANY(?)"
+		lifecycleArgs = []any{pq.Array(spec.lifecycles)}
+	}
+
+	innerSQL := fmt.Sprintf(`
+            SELECT e.test_id, e.suite_id, vg.group_id AS variant_group_id,
+                %s
+            %s
+            WHERE %s%s%s
+            GROUP BY e.test_id, e.suite_id, vg.group_id%s`,
```
petr-muller, line 137: While reviewing the PR with an LLM it came up thinking there's a case where it can compute some totals wrong:

> buildInnerAggregation's GROUP BY (test_id, suite_id, variant_group_id — no prow_job_id/lifecycle) is shared by prefixSumSpec, which computes SUM(end) - SUM(start) across the whole group. The prior self-join paired each (test_id, prow_job_id, suite_id, lifecycle) key's own start/end rows before subtracting. Now, a key present only at lookupStart (e.g. because it was reclassified to a new lifecycle key, which pkg/db/cumulativesummary/cumulative_summary.go's delete+insert design explicitly anticipates) still contributes its full start-of-window value to the group sum with nothing at lookupEnd to offset it.

> Worked example: a test's lifecycle reclassifies from blocking to informing mid-window. Old key: prefix_sum_runs=500 at lookupStart, no row at lookupEnd. New key: prefix_sum_runs=50 at lookupEnd only.
> Old (self-join) result: 50 - 0 = 50 — correct.
> New (grouped CASE-WHEN) result: SUM(end)=50, SUM(start)=500 → total_count = 50 - 500 = -450.
Via WHERE agg.total_count > 0 this can also silently drop real data (MissingSample/MissingBasis) or understate counts when other jobs in the same group are positive.

I had it came up with a test to showcase the problem in https://github.com/petr-muller-reviewer/sippy/commit/7b7c2e09a934e9afd7c150f25d28bb6b269c3491 and then propose a fix in https://github.com/petr-muller-reviewer/sippy/commit/29f13ffe2829a6ad035e05d11a2d983d4b04245b . The fix apparently reverts part of the PR and restores the self-join, potentially making performance worse again but I challenged the review LLM about it and it claims it should not be a large harm:

> Per-CTE, the self-join is slightly heavier than the CASE-WHEN, but the index makes it cheap:
>
>  - CASE-WHEN: one scan of the partition over both dates (date IN (start, end)), then hash-aggregate. No join.
>  - Self-join: scans the lookupEnd slice as the driver (release, date index prefix), then joins the lookupStart slice. The join predicate is the full six-column unique key, matched exactly by idx_test_cumulative_summaries_key — a perfect index-backed merge/hash join, no per-row heap thrashing.

Also proposed an alternative:

> If someone wants the single-scan speed and correctness, there is a correct single-scan formulation: keep CASE-WHEN but do the inner GROUP BY on the full key including prow_job_id and lifecycle, compute each key's end − start delta, then sum deltas in an outer aggregate. That gets one pass over both dates without cross-key contamination.

petr-muller (commented):
I have not reviewed this entirely line-by-line, but I read it and it seems fine to me (not that I can entirely grok all the DB/SQL stuff and fit it into head).

While doing a review I had an LLM come up with some claims about logic duplication, and also a potential problematic corner case - I'm listing them inline, they seem plausible to me but I do not have a strong sense about how serious they are and how much we'd want to address them - the PR is fine to me to merge if you don't find them serious.

pkg/api/componentreadiness/dataprovider/postgres/cr_queries.go
mstaeble: The self-join and the grouped CASE-WHEN only diverge when a key has a `lookupStart` row but no `lookupEnd` row. That state can't arise from the cumulative-summary loader. Any key present on a given date is carried forward to every later date. More fundamentally, prefix sums are monotonic by definition, so a key present at start but absent at end is a data-consistency violation, not an input the read path should paper over. Handling it in the query would just mask a loader/data bug.

pkg/api/componentreadiness/dataprovider/postgres/cr_queries.go
mstaeble: Done.

pkg/api/componentreadiness/dataprovider/postgres/cr_queries.go
mstaeble: Done.

pkg/api/componentreadiness/dataprovider/postgres/cr_queries.go
```diff
@@ -28,38 +31,27 @@ type variantQuerySetup struct {
 	minimumFailure  int
 }
 
+const queryPlannerHints = "SET LOCAL max_parallel_workers_per_gather = 4; SET LOCAL parallel_setup_cost = 0; SET LOCAL parallel_tuple_cost = 0; SET LOCAL enable_nestloop = off; SET LOCAL enable_sort = off"
```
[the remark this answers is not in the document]
mstaeble: The `enable_nestloop=off` / `enable_sort=off` hints aren't new to the standalone paths conceptually: they originate from the TRT-2741 PostgreSQL CR provider work, where the standalone prefix-sum and GA queries were benchmarked with them. The standalone path is the implementation that was previously used for the report, so these plans are already exercised. The combined query reuses the same shared hint constant deliberately, so the two paths stay consistent rather than drifting. Keeping them shared.

petr-muller (approved):

PR #3907, opened 2026-08-16
Trt 2709 partitioning phase2 query partitioning

neisw:
Query optimizations based on partitioning migration for

- prow_job_runs
- prow_job_run_annotations
- prow_job_run_prow_pull_requests

Primary modifications 

- Ensure partition filtering is enabled for relevant queries
  - Queries without release &/| timestamp filtering will get defaults
- jobResultFunction required migration to `LANGUAGE plpgsql` to allow the planner access to the query
- postgres_validation_test added for verification across commits / migration phases

## Summary by CodeRabbit

* **New Features**
  * Build-cluster health and analysis support selected releases, with automatic fallback to the active release.
  * Cluster autocomplete respects the selected release and recent activity.
  * Added golden-file validation for PostgreSQL query results.

* **Bug Fixes**
  * Improved release and time-range filtering to prevent data from different releases being combined.
  * Improved accuracy of job-run, pull-request, test, repository, and release health reports.
  * Centralized release update timestamp handling for more consistent reporting.

### DIFF
--- a/pkg/api/autocomplete.go
+++ b/pkg/api/autocomplete.go
@@ -83,6 +83,9 @@ func PrintAutocompleteFromDB(w http.ResponseWriter, req *http.Request, dbc *db.D
 			clusterRelease, err = query.CurrentActiveRelease(dbc)
 			if err != nil {
 				log.WithError(err).Warn("could not determine current development release for cluster autocomplete")
+				// do not return error for autocomplete, default to empty results
+				RespondWithJSON(200, w, result)
+				return
 			}
 		}
 		q = q.Table("prow_job_runs").
--- a/pkg/api/autocomplete.go
+++ b/pkg/api/autocomplete.go
@@ -5,7 +5,10 @@ import (
 	"net/http"
 
 	"github.com/gorilla/mux"
+	log "github.com/sirupsen/logrus"
+
 	"github.com/openshift/sippy/pkg/db"
+	"github.com/openshift/sippy/pkg/db/query"
 )
 
 // LabelOption represents a label with ID and title for autocomplete
@@ -74,9 +77,19 @@ func PrintAutocompleteFromDB(w http.ResponseWriter, req *http.Request, dbc *db.D
 			Select("DISTINCT(author) as name").
 			Order("name")
 	case "cluster":
+		clusterRelease := release
+		if clusterRelease == "" {
+			var err error
+			clusterRelease, err = query.CurrentActiveRelease(dbc)
+			if err != nil {
+				log.WithError(err).Warn("could not determine current development release for cluster autocomplete")
+			}
+		}
 		q = q.Table("prow_job_runs").
 			Select("DISTINCT(cluster) as name").
 			Where("cluster IS NOT NULL").
+			Where("prow_job_release = ?", clusterRelease).
+			Where("timestamp > NOW() - INTERVAL '14 days'").
 			Order("name")
 	case "suites":
 		q = q.Table("suites").
@@ -98,7 +111,7 @@ func PrintAutocompleteFromDB(w http.ResponseWriter, req *http.Request, dbc *db.D
 		RespondWithJSON(404, w, map[string]string{"message": "Autocomplete field not found."})
 	}
 
-	if release != "" {
+	if release != "" && field != "cluster" {
 		q = q.Where("release = ?", release)
 	}
 
--- a/pkg/api/health.go
+++ b/pkg/api/health.go
@@ -84,9 +84,9 @@ func PrintOverallReleaseHealthFromDB(w http.ResponseWriter, dbc *db.DB, release 
 		indicators["tests"] = testsIndicator
 	}
 
-	var lastUpdated time.Time
-	if r := dbc.DB.Raw("SELECT MAX(created_at) FROM prow_job_runs").Scan(&lastUpdated); r.Error != nil {
-		log.WithError(r.Error).Error("error querying last update time")
+	lastUpdated, err := GetLastUpdateTime(dbc)
+	if err != nil {
+		log.WithError(err).Error("error querying last update time")
 		return
 	}
 	log.WithField("lastUpdated", lastUpdated).Info("ran the last update query")
--- a/pkg/api/job_runs.go
+++ b/pkg/api/job_runs.go
@@ -55,11 +55,7 @@ func JobsRunsReportFromDB(dbc *db.DB, filterOpts *filter.FilterOptions, release 
 	}
 
 	if len(release) == 0 {
-		var err error
-		release, err = query.CurrentActiveRelease(dbc)
-		if err != nil {
-			return nil, fmt.Errorf("determining current release: %w", err)
-		}
+		return nil, &ValidationError{Message: "release is required; use useCurrentRelease=true to resolve automatically"}
 	}
 
 	jobsResult := make([]apitype.JobRun, 0)
@@ -453,7 +449,13 @@ func FetchJobRun(dbc *db.DB, jobRunID int64, onlyNewTests bool, preloads []strin
 		q = q.Preload(preload)
 	}
 
-	res := q.Order("timestamp DESC").First(jobRun, jobRunID)
+	partKeys, err := query.LookupProwJobRunPartitionKeys(dbc, jobRunID)
+	if err != nil {
+		return nil, fmt.Errorf("looking up partition keys for job run %d: %w", jobRunID, err)
+	}
+
+	res := q.Where("prow_job_release = ? AND timestamp = ?", partKeys.ProwJobRelease, partKeys.Timestamp).
+		Take(jobRun, jobRunID)
 	if res.Error != nil {
 		return nil, res.Error
 	}
--- a/pkg/api/job_runs.go
+++ b/pkg/api/job_runs.go
@@ -71,6 +71,11 @@ func JobsRunsReportFromDB(dbc *db.DB, filterOpts *filter.FilterOptions, release 
 		Joins("JOIN prow_jobs ON prow_job_runs.prow_job_id = prow_jobs.id")
 
 	addPRJoin := func(q *gorm.DB) *gorm.DB {
+		if len(release) > 0 {
+			return q.
+				Joins(`LEFT JOIN (SELECT DISTINCT ON(prow_job_run_id) prow_job_run_id, prow_pull_request_id FROM prow_job_run_prow_pull_requests WHERE prow_job_run_release = ? ORDER BY prow_job_run_id, prow_pull_request_id DESC) jrpp ON jrpp.prow_job_run_id = prow_job_runs.id`, release).
+				Joins("LEFT JOIN prow_pull_requests pp ON pp.id = jrpp.prow_pull_request_id")
+		}
 		return q.
 			Joins(`LEFT JOIN (SELECT DISTINCT ON(prow_job_run_id) prow_job_run_id, prow_pull_request_id FROM prow_job_run_prow_pull_requests ORDER BY prow_job_run_id, prow_pull_request_id DESC) jrpp ON jrpp.prow_job_run_id = prow_job_runs.id`).
 			Joins("LEFT JOIN prow_pull_requests pp ON pp.id = jrpp.prow_pull_request_id")
@@ -87,6 +92,7 @@ func JobsRunsReportFromDB(dbc *db.DB, filterOpts *filter.FilterOptions, release 
 
 	if len(release) > 0 {
 		q = q.Where("prow_jobs.release = ?", release)
+		q = q.Where("prow_job_runs.prow_job_release = ?", release)
 	}
 	q = q.Where(`prow_job_runs."timestamp" < ?`, reportEnd)
 	q = q.Where(`prow_job_runs."timestamp" >= ?`, lookback)
@@ -132,7 +138,7 @@ func JobsRunsReportFromDB(dbc *db.DB, filterOpts *filter.FilterOptions, release 
 
 		if !needs.needsPRJoin() {
 			g.Go(func() error {
-				return enrichJobRunsWithPRData(dbc, jobsResult, ids)
+				return enrichJobRunsWithPRData(dbc, jobsResult, ids, release)
 			})
 		}
 
@@ -358,7 +364,7 @@ func enrichJobRunsWithTestNames(dbc *db.DB, results []apitype.JobRun, ids []int,
 	return nil
 }
 
-func enrichJobRunsWithPRData(dbc *db.DB, results []apitype.JobRun, ids []int) error {
+func enrichJobRunsWithPRData(dbc *db.DB, results []apitype.JobRun, ids []int, release string) error {
 	type prResult struct {
 		ID                int    `gorm:"column:id"`
 		PullRequestLink   string `gorm:"column:pull_request_link"`
@@ -368,18 +374,21 @@ func enrichJobRunsWithPRData(dbc *db.DB, results []apitype.JobRun, ids []int) er
 		PullRequestAuthor string `gorm:"column:pull_request_author"`
 	}
 	var prResults []prResult
-	if err := dbc.DB.Raw(`
-		SELECT DISTINCT ON(jrpp.prow_job_run_id)
+	q := dbc.DB.Table("prow_job_run_prow_pull_requests jrpp").
+		Select(`DISTINCT ON(jrpp.prow_job_run_id)
 			jrpp.prow_job_run_id AS id,
 			pp.link AS pull_request_link,
 			pp.sha AS pull_request_sha,
 			pp.org AS pull_request_org,
 			pp.author AS pull_request_author,
-			pp.repo AS pull_request_repo
-		FROM prow_job_run_prow_pull_requests jrpp
-			INNER JOIN prow_pull_requests pp ON pp.id = jrpp.prow_pull_request_id
-		WHERE jrpp.prow_job_run_id IN ?
-		ORDER BY jrpp.prow_job_run_id, jrpp.prow_pull_request_id DESC`, ids).Scan(&prResults).Error; err != nil {
+			pp.repo AS pull_request_repo`).
+		Joins("INNER JOIN prow_pull_requests pp ON pp.id = jrpp.prow_pull_request_id").
+		Where("jrpp.prow_job_run_id IN ?", ids).
+		Order("jrpp.prow_job_run_id, jrpp.prow_pull_request_id DESC")
+	if len(release) > 0 {
+		q = q.Where("jrpp.prow_job_run_release = ?", release)
+	}
+	if err := q.Scan(&prResults).Error; err != nil {
 		return err
 	}
 	prMap := make(map[int]*prResult, len(prResults))
@@ -463,7 +472,7 @@ func FetchJobRun(dbc *db.DB, jobRunID int64, onlyNewTests bool, preloads []strin
 			Where("NOT EXISTS (SELECT 1 FROM test_ownerships tow WHERE tow.test_id = prow_job_run_tests.test_id)").
 			Where(`NOT EXISTS (
 				SELECT 1 FROM prow_job_run_tests t2
-				INNER JOIN prow_job_run_prow_pull_requests prmap ON prmap.prow_job_run_id = t2.prow_job_run_id
+				INNER JOIN prow_job_run_prow_pull_requests prmap ON prmap.prow_job_run_id = t2.prow_job_run_id AND prmap.prow_job_run_release = t2.prow_job_run_release
 				INNER JOIN prow_pull_requests prs ON prs.id = prmap.prow_pull_request_id
 				WHERE t2.test_id = prow_job_run_tests.test_id
 				  AND t2.prow_job_run_release = prow_job_run_tests.prow_job_run_release
--- a/pkg/api/job_runs.go
+++ b/pkg/api/job_runs.go
@@ -583,7 +583,7 @@ func findReleaseMatchJobNames(dbc *db.DB, jobRun *models.ProwJobRun, compareRele
 					gosort.Strings(job.Variants)
 					if stringSlicesEqual(variants, job.Variants) {
 
-						runCount, err := query.ProwJobRunCount(dbc, job.ID, compareRelease)
+						runCount, err := query.ProwJobRunCount(dbc, job.ID, compareRelease, time.Now().Add(-14*24*time.Hour))
 						if err != nil {
 							logger.WithError(err).Errorf("Failed to query job run count for %d", job.ID)
 							continue
--- a/pkg/api/job_runs.go
+++ b/pkg/api/job_runs.go
@@ -453,15 +453,7 @@ func FetchJobRun(dbc *db.DB, jobRunID int64, onlyNewTests bool, preloads []strin
 		q = q.Preload(preload)
 	}
 
-	// Two-step load: first fetch partition keys (lightweight cross-partition
-	// scan), then load the full row with partition pruning.
-	partKeys, err := query.LookupProwJobRunPartitionKeys(dbc, jobRunID)
-	if err != nil {
-		return nil, fmt.Errorf("looking up partition keys for job run %d: %w", jobRunID, err)
-	}
-
-	res := q.Where("prow_job_release = ? AND timestamp = ?", partKeys.ProwJobRelease, partKeys.Timestamp).
-		First(jobRun, jobRunID)
+	res := q.Order("timestamp DESC").First(jobRun, jobRunID)
 	if res.Error != nil {
 		return nil, res.Error
 	}
--- a/pkg/api/job_runs.go
+++ b/pkg/api/job_runs.go
@@ -568,14 +568,12 @@ func findReleaseMatchJobNames(dbc *db.DB, jobRun *models.ProwJobRun, compareRele
 					gosort.Strings(job.Variants)
 					if stringSlicesEqual(variants, job.Variants) {
 
-						jobIDs, err := query.ProwJobRunIDs(dbc, job.ID)
-
+						runCount, err := query.ProwJobRunCount(dbc, job.ID, compareRelease)
 						if err != nil {
-							logger.WithError(err).Errorf("Failed to query job run ids for %d", job.ID)
+							logger.WithError(err).Errorf("Failed to query job run count for %d", job.ID)
 							continue
 						}
-
-						totalJobRunsCount += len(jobIDs)
+						totalJobRunsCount += runCount
 						allJobNames = append(allJobNames, job.Name)
 					}
 				}
--- a/pkg/api/job_runs.go
+++ b/pkg/api/job_runs.go
@@ -54,6 +54,14 @@ func JobsRunsReportFromDB(dbc *db.DB, filterOpts *filter.FilterOptions, release 
 		return nil, &ValidationError{Message: fmt.Sprintf("sorting by %s is not supported", filterOpts.SortField)}
 	}
 
+	if len(release) == 0 {
+		var err error
+		release, err = query.CurrentActiveRelease(dbc)
+		if err != nil {
+			return nil, fmt.Errorf("determining current release: %w", err)
+		}
+	}
+
 	jobsResult := make([]apitype.JobRun, 0)
 	lookback := reportEnd.Add(-90 * 24 * time.Hour)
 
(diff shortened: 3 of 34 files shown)


neisw:
@coderabbitai full review

pkg/api/health.go
```diff
@@ -84,9 +84,9 @@ func PrintOverallReleaseHealthFromDB(w http.ResponseWriter, dbc *db.DB, release
 		indicators["tests"] = testsIndicator
 	}
 
-	var lastUpdated time.Time
-	if r := dbc.DB.Raw("SELECT MAX(created_at) FROM prow_job_runs").Scan(&lastUpdated); r.Error != nil {
-		log.WithError(r.Error).Error("error querying last update time")
+	lastUpdated, err := GetLastUpdateTime(dbc)
+	if err != nil {
+		log.WithError(err).Error("error querying last update time")
```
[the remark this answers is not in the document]
neisw: GetLastUpdateTime returns the freshness of the data pipeline, not any specific release's data. Using the active release is intentional here.

pkg/api/job_runs.go
```diff
@@ -72,7 +80,7 @@ func JobsRunsReportFromDB(dbc *db.DB, filterOpts *filter.FilterOptions, release
 
 	addPRJoin := func(q *gorm.DB) *gorm.DB {
 		return q.
-			Joins(`LEFT JOIN (SELECT DISTINCT ON(prow_job_run_id) prow_job_run_id, prow_pull_request_id FROM prow_job_run_prow_pull_requests ORDER BY prow_job_run_id, prow_pull_request_id DESC) jrpp ON jrpp.prow_job_run_id = prow_job_runs.id`).
+			Joins(`LEFT JOIN (SELECT DISTINCT ON(prow_job_run_id) prow_job_run_id, prow_pull_request_id FROM prow_job_run_prow_pull_requests WHERE prow_job_run_release = ? ORDER BY prow_job_run_id, prow_pull_request_id DESC) jrpp ON jrpp.prow_job_run_id = prow_job_runs.id`, release).
```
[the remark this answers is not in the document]
neisw: The PR mapping subquery already has release filter. Adding timestamp to a DISTINCT ON subquery inside a JOIN adds complexity. The outer query already filters prow_job_runs.timestamp, so only matching rows participate.

pkg/db/functions.go
```diff
@@ -75,108 +75,115 @@ $_$;
...
+                ROUND(coalesce(AVG(case when prow_job_runs.timestamp >= p_start AND prow_job_runs.timestamp < p_boundary then prow_job_runs.duration end) / 60000000000.0, 0))::int as previous_average_duration_minutes
         FROM prow_job_runs
         JOIN prow_jobs
                 ON prow_jobs.id = prow_job_runs.prow_job_id
-                                AND prow_jobs.release = $1
-                AND timestamp BETWEEN $2 AND $4
-        WHERE prow_job_runs.prow_job_release = $1
+                AND prow_jobs.release = p_release
+        WHERE prow_job_runs.prow_job_release = p_release
+          AND prow_job_runs.timestamp BETWEEN p_start AND p_endstamp
```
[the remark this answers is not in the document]
neisw: BETWEEN for current period is the existing convention throughout this function. The callers control whether periods overlap. Changing this would alter behavior for all existing consumers.

pkg/mcp/tools/releases.go
```diff
@@ -52,13 +52,8 @@ func (rt *ReleasesTool) GetHandler() func(ctx context.Context, request mcp.CallT
...
 		if rt.deps.DBClient != nil && rt.deps.DBClient.DB != nil {
-			type LastUpdatedQuery struct {
-				Max time.Time
-			}
-			var result LastUpdatedQuery
-			// Assume our last update is the last time we inserted a prow job run.
-			if err := rt.deps.DBClient.DB.Raw("SELECT MAX(created_at) FROM prow_job_runs").Scan(&result).Error; err == nil {
-				lastUpdated = result.Max
+			if t, err := api.GetLastUpdateTime(rt.deps.DBClient); err == nil {
+				lastUpdated = t
```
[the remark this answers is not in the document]
neisw: Silently returning zero LastUpdated is acceptable here; the MCP tool still returns releases successfully. A DB error for staleness shouldn't fail the entire releases response.

neisw:
@coderabbitai resolve

pkg/api/tests.go
```diff
@@ -412,8 +412,13 @@ func GetJobRunTestsCountByLookbackAt(dbc *db.DB, lookbackDays int, today civil.D
 	log.WithField("lookbackDays", lookbackDays).Info("starting lookback count queries")
 
 	// Count job runs from prow_job_runs (one row per run, much smaller than prow_job_run_tests).
+	release, err := query.CurrentActiveRelease(dbc)
```
mstaeble, line 413: If we are limiting the count of job runs to only the current active release, should we also do the same for the count of tests?

pkg/db/query/job_queries.go
```diff
@@ -46,18 +66,16 @@ func ProwJobSimilarName(dbc *db.DB, rootName, release string) ([]models.ProwJob,
...
-		return nil, q.Error
-	}
-	if err := q.Scan(&jobIDs).Error; err != nil {
-		return nil, err
+func ProwJobRunCount(dbc *db.DB, prowJobID uint, release string) (int, error) {
+	var count int64
+	q := dbc.DB.Table("prow_job_runs").
+		Where("prow_job_id = ?", prowJobID).
+		Where("prow_job_release = ?", release).
+		Where("timestamp > NOW() - INTERVAL '14 days'")
```
mstaeble: Should we have callers provide the time window of interest, rather than hard-coding a hidden "previous 14-day" window inside the function logic? This makes it less fragile if the time window is changed in `jobNamesTestResultFunc`. It also makes it easier to pass down the pinned date/time, if it has been configured for the server, instead of always using now.

pkg/db/functions.go
```diff
@@ -75,108 +75,116 @@ $_$;
...
         group by prow_jobs.name, prow_jobs.variants
 ),
-last_pass AS (
-	SELECT prow_job_id, max(timestamp) as last_pass from prow_job_runs where overall_result = 'S' AND prow_job_release = $1 group by prow_job_id
+lp AS (
+    SELECT prow_job_runs.prow_job_id, max(prow_job_runs.timestamp)::timestamp without time zone as last_pass
+    FROM prow_job_runs
+    WHERE overall_result = 'S' AND prow_job_release = p_release
+      AND prow_job_runs.timestamp BETWEEN p_start AND p_endstamp
```
mstaeble: Are we OK with having a null last_pass value if there has not been a successful run in the time window? Previously, we showed the last successful run, if there has ever been a successful run, even if that was a long time ago.

pkg/db/query/pull_request_queries.go
```diff
@@ -29,7 +30,9 @@ func PullRequestReport(dbc *db.DB, filterOpts *filter.FilterOptions, release str
 		Joins("INNER JOIN prow_jobs on prow_job_runs.prow_job_id = prow_jobs.id").
 		Where("prow_jobs.release = ?", release).
 		Where("prow_job_runs.prow_job_release = ?", release).
+		Where("prow_job_runs.timestamp > ?", lookback14d).
```
mstaeble: Are we OK with the implications of limiting this lookback to the previous 14 days? Do we need language in the UI to make it clear that we are only showing PRs that have had job runs in the past 14 days?

pkg/db/query/repository_queries.go
```diff
@@ -27,7 +27,9 @@ func RepositoryReport(dbc *db.DB, filterOpts *filter.FilterOptions, release stri
 		Joins("LEFT JOIN (?) premerge_failures ON premerge_failures.prow_job_ID = prow_jobs.id", averageByJob).
 		Where("prow_jobs.release = ?", release).
 		Where("prow_job_runs.prow_job_release = ?", release).
+		Where("prow_job_runs.timestamp >= ? AND prow_job_runs.timestamp < ?", premergeFailureStart, reportEnd).
 		Where("prow_job_run_prow_pull_requests.prow_job_run_release = ?", release).
+		Where("prow_job_run_prow_pull_requests.prow_job_run_timestamp >= ? AND prow_job_run_prow_pull_requests.prow_job_run_timestamp < ?", premergeFailureStart, reportEnd).
```
mstaeble: With this change, a repo with reverts merged in the last 90 days but no PR job-run activity in the last 14 days won't appear at all, so its revert count is invisible. This contradicts the text in the UI that claims to show the number of reverts the repo has had in the past 90 days. Is there any issue with this inconsistency between the 90-day lookback and the 14-day lookback?

The job counts also goes from "distinct jobs that ever ran PRs for this repo in the release" to "distinct jobs with runs in the last 14 days".

Repos active earlier in the release but idle longer than 14 days disappear. A release that is GA'd or otherwise has no runs in the last 14 days yields an essentially empty repositories table, where previously it showed history.

pkg/db/query/release_queries.go
```diff
@@ -0,0 +1,40 @@
...
+
+	"github.com/openshift/sippy/pkg/db"
+)
+
+// CurrentActiveRelease returns the most relevant active release. It prefers the
+// in-development release with the oldest development_start_date (most mature).
+// If no in-development release exists (all have a GA date), it falls back to
+// the release with the most recent GA date. Used as a fallback for
+// partition-scoped queries when no specific release is available from the caller.
+func CurrentActiveRelease(dbc *db.DB) (string, error) {
```
mstaeble, line 14: **Inconsistent error handling across `CurrentActiveRelease` call sites**

`CurrentActiveRelease` is now called from ~8 places, but each handles a lookup error differently, and a couple of those behaviors are surprising:

- `pkg/api/autocomplete.go:83` — logs a warning, leaves `clusterRelease` empty, then runs `WHERE prow_job_release = ''`, which silently returns an empty cluster list. The user gets "no clusters" with no indication anything went wrong.
- `pkg/sippyserver/server.go:445` (`determineCapabilities`) — logs a warning and skips `BuildClusterCapability`. Since this runs once at startup, a transient DB error here permanently disables the build-cluster feature for the lifetime of the process.
- `pkg/api/job_runs.go:59`, `pkg/api/tests.go:415`, `pkg/sippyserver/metrics/metrics.go:281`, `pkg/sippyserver/server.go:1405` / `:1429`, `pkg/api/releases.go:674` — propagate the error (fail the request / refresh).
- `pkg/dataloader/prowloader/bigqueryjobs.go:28` — warns and falls back to `DefaultLookbackDays`.

So the same helper failing produces three different outcomes: silent empty result, permanent capability loss, or a hard error. At minimum, we should avoid the two footguns above: in `autocomplete` skip the query (or surface an error) rather than filtering on `''`, and in `determineCapabilities` make the build-cluster check resilient to a transient startup error so the capability isn't lost for good.

pkg/api/job_runs.go
```diff
@@ -54,6 +54,14 @@ func JobsRunsReportFromDB(dbc *db.DB, filterOpts *filter.FilterOptions, release
 		return nil, &ValidationError{Message: fmt.Sprintf("sorting by %s is not supported", filterOpts.SortField)}
 	}
 
+	if len(release) == 0 {
```
mstaeble, line 57: Should we throw a 400 error if the release is not provided, instead of surprisingly falling back to the current release?

pkg/db/query/release_queries.go
```diff
@@ -0,0 +1,40 @@
...
+		Where("ga_date IS NOT NULL").
+		Where("product = ?", "OCP").
+		Where("release ~ ?", `^\d+\.\d+$`).
+		Order("ga_date DESC").
+		Limit(1).
+		Pluck("release", &release).Error
+	if err != nil {
+		return "", fmt.Errorf("query latest GA release: %w", err)
+	}
+	return release, nil
```
mstaeble, line 42: Should the function return an error when there is no release? We should not expect the caller to have to check that `release` is not empty in the un-happy scenario where the release_definitions table is empty.

pkg/db/query/job_queries.go
```diff
@@ -12,6 +12,26 @@ import (
...
+// full row with partition pruning.
+type ProwJobRunPartitionKeys struct {
+	ProwJobRelease string    `gorm:"column:prow_job_release"`
+	Timestamp      time.Time `gorm:"column:timestamp"`
+}
+
+// LookupProwJobRunPartitionKeys fetches the partition keys for a prow_job_run
+// by ID. This is intended as the first step of a two-step lookup pattern where
+// the caller then uses these keys to load the full row with partition pruning.
+func LookupProwJobRunPartitionKeys(dbc *db.DB, jobRunID int64) (ProwJobRunPartitionKeys, error) {
```
mstaeble, line 27: Have you benchmarked whether there is generally improved response times when doing this two-step lookup in most scenarios? There should not be a benefit in doing the cross-partition search as a first step unless the partition keys are used to access other tables beyond the prow_job_run tables (such as in `FetchJobRun` which uses the partition keys to select the single partition in the query from the `prow_job_run_tests` table).

The pattern of needing to look up a job run by ID is ubiquitous and fundamental to the access patterns enough that we should consider using a mapping table from ID to release/timestamp so that we can have a fast lookup. The performance of looking up the partition keys from the prow_job_runs table will continue to degrade as there are more releases and more dates.

pkg/db/query/job_queries.go
```diff
@@ -12,6 +12,26 @@ import (
...
+	var keys ProwJobRunPartitionKeys
+	err := dbc.DB.Table("prow_job_runs").
+		Select("prow_job_release, timestamp").
+		Where("id = ?", jobRunID).
```
mstaeble, line 31: With the partition tables, job_run_id is not unique across the table (only within individual partitions of the table). What do we want the behavior to be in the inconsistent scenario where there are multiple rows with the same ID?

pkg/db/query/job_queries.go
```diff
@@ -12,6 +12,26 @@ import (
...
+		Scan(&keys).Error
```
mstaeble: Do we want `LookupProwJobRunPartitionKeys` to throw an error when the job run with the requests ID cannot be found? Should we call `Take` instead of `Scan`?

mstaeble (commented):
There are a number of integration test failures introduced by these changes.

pkg/api/autocomplete.go
```diff
@@ -74,9 +77,19 @@ func PrintAutocompleteFromDB(w http.ResponseWriter, req *http.Request, dbc *db.D
...
+			clusterRelease, err = query.CurrentActiveRelease(dbc)
+			if err != nil {
+				log.WithError(err).Warn("could not determine current development release for cluster autocomplete")
+			}
+		}
 		q = q.Table("prow_job_runs").
 			Select("DISTINCT(cluster) as name").
 			Where("cluster IS NOT NULL").
+			Where("prow_job_release = ?", clusterRelease).
+			Where("timestamp > NOW() - INTERVAL '14 days'").
```
mstaeble, line 95: Is there a risk of omitting clusters from the auto-complete? Are there certain clusters that are only used for bespoke releases of clusters that are used sporadically, where they may not show in the list of clusters used for the current active release in the past 14 days?

pkg/db/query/job_queries.go
```diff
@@ -5,13 +5,40 @@ import (
...
+	res := dbc.DB.Table("prow_job_runs").
+		Select("prow_job_release, timestamp").
+		Where("id = ?", jobRunID).
+		Scan(&keys)
+	if res.Error != nil {
+		return keys, res.Error
+	}
+	if res.RowsAffected == 0 {
+		return keys, gorm.ErrRecordNotFound
+	}
```
mstaeble: I made a [comment](https://github.com/openshift/sippy/pull/3907#discussion_r3800368207), on the previous commit, about returning an error in the not-found case. I think that using `Take` is a cleaner fix than checking for `RowsAffected`.

pkg/db/functions.go
```diff
@@ -75,108 +75,116 @@ $_$;
...
          JOIN prow_jobs ON prow_jobs.name = results.pj_name
-         LEFT JOIN repo_org_jobs ON prow_jobs.id = repo_org_jobs.id
-		 LEFT JOIN retests ON prow_jobs.id = retests.prow_job_id
-		 LEFT JOIN last_pass ON prow_jobs.id = last_pass.prow_job_id
-		 LEFT JOIN job_bugs ON prow_jobs.id = job_bugs.prow_job_id
-    $_$;
+                      AND prow_jobs.release = p_release
+         LEFT JOIN (
+             SELECT prow_pull_requests.org, prow_pull_requests.repo, prow_jobs.id
+             FROM prow_job_runs
```
mstaeble, line 169: Would it be more efficient to bring this JOIN to prow_job_runs into the `retests` CTE, to avoid having to merge the results again? Or are you seeing the planner effectively doing that for us?

pkg/db/functions.go
```diff
@@ -75,108 +75,116 @@ $_$;
...
+                  INNER JOIN prow_job_run_prow_pull_requests on prow_job_run_prow_pull_requests.prow_job_run_id = prow_job_runs.id
+                  INNER JOIN prow_pull_requests on prow_pull_requests.id = prow_job_run_prow_pull_requests.prow_pull_request_id
+                  INNER JOIN prow_jobs ON prow_job_runs.prow_job_id = prow_jobs.id
+             WHERE prow_job_runs.prow_job_release = p_release
+               AND prow_job_runs.timestamp BETWEEN p_start AND p_endstamp
+               AND prow_job_run_prow_pull_requests.prow_job_run_release = p_release
```
mstaeble, line 175: Logically, they should be the same, but I have anecdotally had better outcomes by moving the partition pruning filters into the ON clause of the JOIN rather than in the WHERE clause, particularly for complicated queries with deep nesting.

pkg/db/functions.go
```diff
@@ -75,108 +75,116 @@ $_$;
...
         FROM prow_job_runs
         JOIN prow_jobs
                 ON prow_jobs.id = prow_job_runs.prow_job_id
-                                AND prow_jobs.release = $1
-                AND timestamp BETWEEN $2 AND $4
-        WHERE prow_job_runs.prow_job_release = $1
+                AND prow_jobs.release = p_release
+        WHERE prow_job_runs.prow_job_release = p_release
+          AND prow_job_runs.timestamp BETWEEN p_start AND p_endstamp
         group by prow_jobs.name, prow_jobs.variants
```
mstaeble, line 120: This is pre-existing, but would it be better to group by `prow_jobs.id` instead? Then the outer query can do a JOIN on the ID instead of name. This gives cheaper grouping and cheaper joining.

neisw:
> There are a number of integration test failures introduced by these changes.

They weren't running for me locally and I hadn't dug into why.  Let me know if the changes look reasonable.

pkg/api/job_runs.go
neisw: For now I think we live with this one and refactor later if needed.  Unless you feel we should default to all releases but I feel like that would be would be a worse experience than this.

BuildClusterDetails at /build_clusters/:cluster — no release in the URL, and it doesn't pass release to JobRunsTable. So release defaults to '', and the API gets called without a release parameter.

  That's the one path. Every other usage (Jobs.jsx, VariantStatus.jsx) passes props.release from a release-scoped route. So the fallback is needed to support the build cluster detail view.

  The answer is yes: BuildClusterDetails renders JobRunsTable without a release. The route is /build_clusters/:cluster with no release in the URL or props. Returning a 400 would break that view. The fallback to the active release is the correct behavior here.

pkg/db/query/pull_request_queries.go
neisw: Changed to 90 days which is effectively what we have with current retention.  We can see what performance is like and tune as needed.

pkg/db/query/repository_queries.go
neisw: This and pull_request_queries.go are only available for presubmits.  Using 90 day lookback for now.

pkg/api/autocomplete.go
neisw: Not that I am aware of.  I'm thinking keeping autocomplete responsive would be a priority.

pkg/db/functions.go
neisw: My preference is to update the tool tip to indicate it has not passed within the report window.

pkg/db/query/job_queries.go
neisw: ### Endpoints that load `prow_job_runs` by ID without partition keys

  | Route | Param | Code Path |
  |-------|-------|-----------|
  | `/api/job/run/summary` | `prow_job_run_id` | `FetchJobRun` |
  | `/api/jobs/runs/risk_analysis` | `prow_job_run_id` | `FetchJobRun` |
  | `/api/jobs/runs/intervals` | `prow_job_run_id` | `FetchJobRun` |
  | `/api/jobs/runs/events` | `prow_job_run_id` | `FetchJobRun` |
  | `/api/jobs/artifacts` | `prowJobRuns` | `getJobRun` |
  | `/api/jobs/runs/reevaluate` | `prow_job_build_ids` | `reevaluate` |
  | `/api/jobs/runs/disruption` | `job_run_names` | raw SQL |

  Benchmarking showed the two-step lookup (fetch partition keys, then load with pruning) provides no benefit for `prow_job_runs` since step 1 itself scans all partitions. Both direct load (~414ms) and step 1 alone (~389ms) scan all 4,120 sub-partitions, while a fully pruned query is ~25ms. The two-step pattern only pays off for child tables like `prow_job_run_tests` where the
  partition keys are already available from the loaded parent row.

  Refactored all callers to use direct `First(model, jobRunID)` with `Order("timestamp DESC")` for deterministic selection. Removed `LookupProwJobRunPartitionKeys` and `ProwJobRunPartitionKeys`.

neisw: I suspect many of these at least have the release available to them but we may well want a dedicated mapping table at some point.  That would have to support ~3 years worth of jobruns though.

pkg/api/job_runs.go
mstaeble: What if we were to add an explicit query parameter that indicates that the caller wants to use the current release? The BuildClusterDetails page could use that query parameter.

|release|useCurrentRelease|result
|---|---|---
|X| |Pinned to selected release
| |X|Pinned to current active release
| | |400 error
|X|X|400 error

pkg/db/functions.go
mstaeble: Does the user lose some value information with this approach? Do we have cases where a test has not passed in 2 months, and the user would want to know that this is a long, on-going issue and not just limited to the report window?

pkg/db/query/job_queries.go
mstaeble: Supporting 3 years of job runs should not be a problem. The table would use a B-tree index on the ID, which will scale logartithmically.

pkg/api/autocomplete.go
mstaeble: We could consider in the future having a separate table to store clusters. But I agree that we can postpone that until a later date, if it ever becomes an issue.

pkg/db/query/repository_queries.go
```diff
@@ -13,11 +13,9 @@ import (
 func RepositoryReport(dbc *db.DB, filterOpts *filter.FilterOptions, release string, reportEnd time.Time) ([]api.Repository, error) {
 	end := reportEnd
 
-	premergeFailureStart := reportEnd.Add(-14 * 24 * time.Hour)
-	averageByJob := PullRequestAveragePremergeFailures(dbc, &premergeFailureStart, &end)
-
-	revertCountStart := reportEnd.Add(-90 * 24 * time.Hour)
-	revertCount := RepositoryRevertCount(dbc, &revertCountStart, &end)
+	lookback90d := reportEnd.Add(-90 * 24 * time.Hour)
+	averageByJob := PullRequestAveragePremergeFailures(dbc, release, &lookback90d, &end)
```
mstaeble: Are we intentionally changing this average to the last 90 days? Previously, this was deliberately only considering 14 days, despite the revert count considering 90 days. This change will dilute spikes due to recent issues.

pkg/api/jobartifacts/query.go
```diff
@@ -102,15 +101,9 @@ func (q *JobArtifactQuery) getJobRun(jobRunID int64) (JobRun, error) {
...
-	// Two-step load: first fetch partition keys for pruning, then load full row.
-	partKeys, err := query.LookupProwJobRunPartitionKeys(q.DbClient, jobRunID)
-	if err != nil {
-		return jobRunResponse, fmt.Errorf("looking up partition keys for job run %d: %w", jobRunID, err)
-	}
-
 	jobRunModel := new(models.ProwJobRun)
 	res := q.DbClient.DB.Preload("ProwJob").
-		Where("prow_job_release = ? AND timestamp = ?", partKeys.ProwJobRelease, partKeys.Timestamp).
+		Order("timestamp DESC").
```
mstaeble: I would argue that using `Order(...).First()` is worse than using `Take()`. It can mislead the planner into a worse plan that walks through the partitions newest-first rather than doing a bounded probe of the partitions. Two job runs with the same ID is a consistency error, so choosing the newest row is not any more correct than choosing an arbitrary row.

pkg/db/functions.go
neisw: My take is that this likely isn't used much and not worth a perceived performance hit going back 90 days but we can discuss with the team.

pkg/db/query/job_queries.go
neisw: ### Benchmark: staging (pre-migration) vs migrated (post-migration)
  
  | Query | Staging | Migrated | Change |
  |-------|---------|----------|--------|
  | **FetchJobRunByID** | **106ms** | **29,096ms** | **274x slower** |
  | TestCountsByLookback14ForRelease | 11,158ms | 7,298ms | 1.5x faster |
  | TestCountsByLookback9ForRelease | 7,467ms | 3,584ms | 2.1x faster |
  | TestAnalysisByJob | 2,366ms | 1,970ms | 1.2x faster |
  | TestAnalysisByJobWithVariantFilter | 2,201ms | 1,486ms | 1.5x faster |
  | FindTestsByRelease | 1,253ms | 832ms | 1.5x faster |
  | TestAnalysisOverall | 1,095ms | 710ms | 1.5x faster |
  | TestCountsByLookback14 | 661ms | 190ms | 3.5x faster |
  | TestCountsByLookback9 | 660ms | 115ms | 5.7x faster |
  | JobsRunsReport | 428ms | 476ms | ~same |
  | JobReports | 383ms | 370ms | ~same |
  | JobDetails | 227ms | 264ms | ~same |
  | RecentTestFailures | 202ms | 228ms | ~same |
  | VariantReports | 157ms | 113ms | 1.4x faster |
  | ProwJobHistoricalTestCounts | 148ms | 126ms | ~same |
  | TestDurations | 92ms | 86ms | ~same |
  | IsNewTestQuery | 90ms | 114ms | ~same |
  | TestOutputs | 88ms | 155ms | 1.8x slower |
  | ProwJobRunCount | 73ms | 63ms | ~same |
  | JobRunTestCount | 72ms | 81ms | ~same |
  | BuildClusterHealth | 70ms | 56ms | ~same |
  | QueryTestAnalysis | 68ms | 95ms | ~same |
  | TestAnalysisPassRate | 63ms | 63ms | same |
  | RepositoryReport | 39ms | 139ms | 3.5x slower |
  | PullRequestReport | 34ms | 60ms | 1.8x slower |

  **FetchJobRunByID** (direct `First(model, id)` with no partition keys) is the clear regression at 274x slower, scanning all 4,120 sub-partitions. Partition-pruned queries show 1.5-5.7x improvement. `RepositoryReport` and `PullRequestReport` show moderate regressions from joins through the partitioned `prow_job_run_prow_pull_requests` table.

pkg/db/functions.go
neisw: This function is covered by benchmarking test cases 

  - JobReports  — calls query.JobReports which does dbc.DB.Table("job_results(?, ?, ?, ?)", release, start, boundary, end) directly.
  - JobDetails — calls api.JobDetailsReport which also invokes job_results() under the hood.
  
  Which both report virtually no change between staging and the migrated schema
   JobReports                         │ 383ms       │ 370ms        │ ~same       │
   JobDetails                         │ 227ms       │ 264ms        │ ~same

pkg/api/jobartifacts/query.go
neisw: Hmm Order was added in response to https://github.com/openshift/sippy/pull/3907#discussion_r3800362271.  I consider the duplicate jobRunID to be an edge case so happy to land on the solution you prefer.

pkg/db/query/repository_queries.go
neisw: This was not intentional, reverting it back to the original state.

pkg/api/jobartifacts/query.go
neisw: It seems like we are heading towards ProwJobRunId mapping table for release & timestamp.  If we enforce uniqueness on the release & prowJobRunId that would alleviate the concerns here.  I'm leaning towards reinstating the two phase query and removing the ordering, can align on Take.

pkg/db/functions.go
neisw: Switched it back.  I can't say there is a performance hit or that no one is really using it.  We can leave it with a 90 day lookback until is surfaces as a problem.

pkg/db/query/job_queries.go
neisw: Switched back to the two phase query anticipating we will need to add a mapping table for JobRunID

pkg/api/jobartifacts/query.go
mstaeble: I am aligned. We need the mapping table both for performance and uniqueness. At which point, we've closed the consistency gap and `Take` is the better function to use.

sippy-ng/src/jobs/JobTable.jsx
```diff
@@ -94,7 +94,7 @@ export const getColumns = (config, _openBugzillaDialog) => {
       renderCell: (params) => {
         if (params.value === undefined || params.value === '') {
           return (
-            <Tooltip title="Job has never passed, or pass predates Sippy's history (typically about 90 days)">
+            <Tooltip title="Job has not passed within the report window">
```
mstaeble, line 97: nit: This tooltip change is no longer relevant, since the lookback window is now 90 days.

mstaeble (approved):
/lgtm

PR #3926, opened 2026-08-20
TRT-2883: Fix spurious MissingSample/MissingBasis for tests below MinimumFailure

mstaeble:
## Summary
- The PostgreSQL component-readiness `MinimumFailure` SQL filter dropped below-threshold rows from the combined query (`queryCombinedTestStatus`) in two cases, both of which BigQuery (which has no SQL-level `MinimumFailure` filter, and does all classification in Go) handles correctly:
  1. A test below threshold on one side but at/above it on the other side was dropped from the side it was below threshold on, causing the Go-side merge to report `MissingSample`/`MissingBasis` instead of `NotSignificant`.
  2. A test that only ran during one side's window (no counterpart row at all on the other side) and is below threshold on the side it did run on was dropped from **both** result maps entirely, so it never appeared in the report at all, instead of the `MissingBasis`/`MissingSample` BigQuery would surface.
- `belowThresholdRescueBranchTemplate` (a single UNION ALL branch, replacing the earlier cross-only branch) now handles both cases with one `LEFT JOIN` against a narrow per-side `keys` CTE (`unique_id`, `variant_group_id`, `fail_count`): a below-threshold row is rescued when the other side has a matching row that's `>= threshold`, or when the other side has no matching row at all. Joining the narrow keys projection instead of the full status CTE avoids building a hash table over every column (including a `capabilities text[]` array) just to answer a yes/no threshold question — validated against production data on the real `5.0-main` view to keep the added latency to ~+16% versus ~+27-30% for less targeted approaches (e.g. two separate `EXISTS`/`NOT EXISTS` subqueries, or joining the full-width CTEs directly).
- For the standalone query path (`queryTestStatusCTE`, used only by release-fallback via `QueryBaseTestStatus`), drops the SQL-level `MinimumFailure` filter entirely since there's no "other side" to cross-reference; the existing Go-side `MinimumFailure` check in `component_report.go` already handles it correctly.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `gofmt -l` clean
- [x] `golangci-lint run ./...` (0 issues)
- [x] `go test ./pkg/api/componentreadiness/...`
- [x] `test/integration` full suite (453 tests)
- [x] Updated `TestGenerateReport_MinimumFailureThreshold` to assert `NotSignificant` (not just empty `RegressedTests`) for a test whose failures fall below `MinimumFailure`
- [x] Added `TestQueryTestStatus_OneSidedBelowThreshold` covering both directions (sample-only and base-only below-threshold tests with no counterpart)
- [x] Updated `TestQueryTestStatus_SampleResults` and `TestMinimumFailureWithCapabilityFilter`, which had unintentionally been asserting the old (buggy) silent-drop behavior for one-sided tests

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved component readiness results for tests with low failure counts.
  - Ensured below-threshold tests remain visible and are correctly marked **Not Significant**.
  - Improved handling of tests that ran on only one side, including accurate **Missing Sample** or **Missing Base** statuses.
  - Correctly identifies regressions when one side has qualifying results despite the other being below threshold.
  - Excludes results only when both sides fall below the configured threshold and contain data.

### DIFF
--- a/pkg/api/componentreadiness/dataprovider/postgres/cr_queries.go
+++ b/pkg/api/componentreadiness/dataprovider/postgres/cr_queries.go
@@ -28,7 +28,6 @@ type variantQuerySetup struct {
 	groupMapping    variantGroupMapping
 	filterArgs      []any
 	variantSubquery string
-	minimumFailure  int
 }
 
 const queryPlannerHints = "SET LOCAL max_parallel_workers_per_gather = 4; SET LOCAL parallel_setup_cost = 0; SET LOCAL parallel_tuple_cost = 0; SET LOCAL enable_nestloop = off; SET LOCAL enable_sort = off"
@@ -38,7 +37,6 @@ func prepareVariantQuery(
 	dbc *db.DB,
 	includeVariants map[string][]string,
 	dbGroupBy sets.Set[string],
-	minimumFailure int,
 ) (*variantQuerySetup, error) {
 	vf, err := resolveVariantFilter(ctx, dbc, includeVariants, dbGroupBy)
 	if err != nil {
@@ -52,7 +50,6 @@ func prepareVariantQuery(
 		groupMapping:    buildVariantGroupMapping(vf.lookup),
 		filterArgs:      vf.filterArgs,
 		variantSubquery: vf.variantSubquery,
-		minimumFailure:  minimumFailure,
 	}, nil
 }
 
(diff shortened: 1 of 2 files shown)


mstaeble:
@coderabbitai full review

mstaeble:
@coderabbitai full review

mstaeble:
@coderabbitai full review

petr-muller (approved):

PR #3942, opened 2026-08-24
agent instructions: include coderabbit and other guidance

sosiouxme:
## Summary by CodeRabbit

- **Documentation**
  - Expanded backend, frontend, query, and testing development guidance.
  - Added standards for clearer queries, focused responsibilities, reusable utilities, explicit error handling, and comprehensive automated tests.
  - Clarified asynchronous processing as part of the Sippy system.
  - Added requirements to keep symptoms feature documentation current when related functionality changes.
  - Documented query readability, construction, parameterization, and size guidelines.

- **Chores**
  - Updated generated guidance verification and configuration hashes.


.apm/instructions/docs.instructions.md
```diff
@@ -21,3 +21,7 @@ applyTo: "**"
 * Documentation and code belong in the same PR; never treat a docs update
   as a follow-up task.
 * Do not use em dashes when writing docs. Use commas, parentheses, or periods instead.
+* Files under `pkg/**/jobrunscan**/`, `pkg/**/jobrunannotator**/`, `pkg/api/jobartifacts/**`, and
+  `sippy-ng/src/component_readiness/JobArtifactQuery.js` are part of the symptoms feature documented
+  in `docs/features/job-analysis-symptoms.md`. Update this document when there are changes to data
+  models, API surface, or data flow in the symptoms feature.
```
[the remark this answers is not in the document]
sosiouxme: mmmm, yeah, i disagree that this is worth mentioning. the JAQ dialog is enough.

pkg/db/query/AGENTS.md
```diff
@@ -0,0 +1,18 @@
...
+<!-- Source: local .apm/instructions/query.instructions.md -->
+* BigQuery and SQL query-building code should have inline comments explaining the
+  purpose of each major query section (CTEs, JOINs, window functions, WHERE clauses).
+* Abbreviations like CTE, matview, etc. should be expanded on first use.
+* Functions that construct queries must stay under 200 lines; extract sub-queries or
+  CTEs into helper functions when they grow beyond that.
+
+---
+*This file was generated by APM CLI. Do not edit manually.*
+*To regenerate: `apm compile`*
```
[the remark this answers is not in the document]
sosiouxme: what are you talking about here? you don't add anything directly to these files, it would have to go somewhere else

.apm/instructions/backend.instructions.md
```diff
@@ -3,17 +3,41 @@ description: "Go backend coding standards for Sippy"
...
   - GORM model tags: include `gorm:"type:date"` on date-only columns so GORM and migrations use the correct PostgreSQL type.
+* Check `pkg/util/` for existing helper functions before adding inline utility logic.
+  Avoid calling the same utility function multiple times with identical arguments in the
+  same code path.
+* Never ignore returned errors as `_` without clear justification. Errors should be wrapped
+  with context using `fmt.Errorf` with `%w`. Avoid `panic()` except in `init()` or fatal
+  conditions. Check for nil before dereferencing pointers.
+* Never concatenate or format SQL queries with values directly from user input. Always use
+  placeholders for parameters in queries, preferably named (`@Name`). Avoid string concatenation
+  in SQL query construction unless justified.
```
mstaeble: The directive to avoid string concatenation may be too strong. A lot of the SQL queries are quite complex and involve construction through conditionals that necessitate string concatenation.

sosiouxme: this is from what i told coderabbit; has that been too picky? "avoid unless justified" to me actually feels too weak, but obviously i have to leave room for exactly what you mention where there's no other good way to do it. i was mostly tired of seeing it use concatenation instead of interpolation.

mstaeble: Here is the recommendation from chai-bot.

> Keep SQL structure separate from SQL data. Never interpolate, concatenate, or format data values into SQL text, especially values derived from user input. Bind data values with query parameters/placeholders, preferably named (`@Name`). Concatenating static, code-defined SQL fragments to assemble conditional clauses is acceptable, provided those fragments contain no interpolated runtime values. If a dynamic identifier cannot be parameterized, select it from a strict allowlist.

sosiouxme: that's pretty good in general... i struggle to really capture all the factors that feed into SQL construction for me, between legibility, maintainability, code  complexity / hygiene, and security.

variants are specified at runtime dynamically, but can't be parameterized as they are used in "as foo" table names for joins... to basically every rule we can lay down about this, there is probably some exception where it really just doesn't make sense.

probably much of the value in trying to define these rules is really just in having the discussion about what we want. might be good to use team sync to drag up some of the hairiest examples of query construction we have now and make various cases for how to improve them (sometimes that reaches down into how the entire data model might be refactored for how we end up using it).

sosiouxme: returning to this, i think the existing instruction directed against sql injection is good enough for that purpose, and this is a little out of place here. i'm going to add a little something about sql construction in the query instructions and call it good.

sosiouxme:
/override-sticky ci/prow/e2e
nothing about this can affect e2e

mstaeble (approved):
/lgtm

PR #3956, opened 2026-08-27
TRT-2932: Fix presubmit risk analysis comparing against wrong release

machine424:
Since 3f0ab35ed ("Add OCPMCP mcp-0.5 release to customizations"), JobRunRiskAnalysis for Presubmits was picking mcp-0.5 as the comparison release (first result from GetReleasesFromDB ordered by development_start_date DESC). This caused all presubmit risk results to show "Unknown" since there's no test history overlap.

Filter for the most recent OCP release from the mainline chain (has PreviousRelease set), skipping synthetic releases and non-OCP products. Until non-OCP presubmits are properly supported, this ensures risk analysis works correctly for OCP presubmits.

## Summary by CodeRabbit

- **Bug Fixes**
  - Presubmit job risk analysis now uses the latest eligible OCP release.
  - Excludes presubmit-only releases and releases without a predecessor.
  - Reports an error when no suitable OCP release is available instead of selecting an arbitrary result.

- **Tests**
  - Added coverage for product filtering, release ordering, exclusions, missing predecessors, empty results, and unavailable OCP releases.

### DIFF
--- a/pkg/api/job_runs.go
+++ b/pkg/api/job_runs.go
@@ -25,6 +25,7 @@ import (
 	apitype "github.com/openshift/sippy/pkg/apis/api"
 	"github.com/openshift/sippy/pkg/apis/cache"
 	"github.com/openshift/sippy/pkg/apis/openshift"
+	sippyv1 "github.com/openshift/sippy/pkg/apis/sippy/v1"
 	sippyprocessingv1 "github.com/openshift/sippy/pkg/apis/sippyprocessing/v1"
 	"github.com/openshift/sippy/pkg/bigquery"
 	"github.com/openshift/sippy/pkg/dataloader/prowloader"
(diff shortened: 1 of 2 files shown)


machine424:
/retitle [TRT-2932](https://redhat.atlassian.net/browse/TRT-2932): Fix presubmit risk analysis comparing against wrong release

pkg/api/job_runs.go
```diff
@@ -623,19 +633,19 @@ func JobRunRiskAnalysis(
 	logger = logger.WithField("func", "JobRunRiskAnalysis")
 	// If this job is a Presubmit, compare to test results from master, not presubmits, which may perform
 	// worse due to dev code that hasn't merged. We do not presently track presubmits on branches other than
-	// master, so it should be safe to assume the latest compareRelease in the db.
+	// master, so we use the latest OCP release from the db.
 	compareRelease := jobRun.ProwJob.Release
 	neverStableJob := false
 	if compareRelease == models.ReleasePresubmits {
 		ar, err := GetReleasesFromDB(ctx, dbc)
```
neisw, line 640: Let's try using https://github.com/machine424/openshift-sippy/blob/51b437ff66158ef1cd8150ff6d349bc2658db50b/pkg/db/query/release_queries.go#L14 instead

machine424:
/close
Will be handled in another PR

PR #3965, opened 2026-08-28
Trt 2709 jobrunid mapping csr post backfill

neisw:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved job-run matching across reports, test results, pull requests, repositories, and disruption data.
  * Prevented records from being incorrectly associated across releases or timestamps.
  * Improved handling of missing or duplicate job-run mappings.
  * Ensured infrastructure-failure updates apply to the correct job-run partition.

* **Tests**
  * Added integration coverage for retrieving job-run partition details and handling unknown runs.

### DIFF
--- a/pkg/api/componentreadiness/dataprovider/postgres/provider.go
+++ b/pkg/api/componentreadiness/dataprovider/postgres/provider.go
@@ -395,6 +395,8 @@ FROM target_tests tt
 JOIN prow_job_run_tests pjrt ON pjrt.test_id = tt.test_id
     AND (tt.suite_id = pjrt.suite_id OR (tt.suite_id IS NULL AND pjrt.suite_id IS NULL))
 JOIN prow_job_runs pjr ON pjr.id = pjrt.prow_job_run_id
+    AND pjr.prow_job_release = pjrt.prow_job_run_release
+    AND pjr.timestamp = pjrt.prow_job_run_timestamp
 JOIN prow_jobs pj ON pj.id = pjr.prow_job_id
 JOIN tests t ON t.id = pjrt.test_id
 WHERE pj.release = ?
--- a/pkg/db/infrafailure/infrafailure.go
+++ b/pkg/db/infrafailure/infrafailure.go
@@ -13,6 +13,7 @@ package infrafailure
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
 	log "github.com/sirupsen/logrus"
@@ -81,6 +82,8 @@ const setInfraFailureLabelSQL = `
 UPDATE prow_job_runs
 SET labels = array_append(labels, 'InfraFailure')
 WHERE id = ?
+  AND prow_job_release = ?
+  AND timestamp = ?
   AND (labels IS NULL OR NOT (labels @> ARRAY['InfraFailure']))`
 
 // createDeltasTempTableSQL materializes one job run's test results into a temp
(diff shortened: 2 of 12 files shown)


pkg/api/componentreadiness/dataprovider/postgres/provider.go
```diff
@@ -395,6 +395,8 @@ FROM target_tests tt
 JOIN prow_job_run_tests pjrt ON pjrt.test_id = tt.test_id
     AND (tt.suite_id = pjrt.suite_id OR (tt.suite_id IS NULL AND pjrt.suite_id IS NULL))
 JOIN prow_job_runs pjr ON pjr.id = pjrt.prow_job_run_id
+    AND pjr.prow_job_release = pjrt.prow_job_run_release
+    AND pjr.timestamp = pjrt.prow_job_run_timestamp
```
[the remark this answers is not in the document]
neisw: A JobRunID is specific to a single release and timestamp, if there are multiple releases or timestamps that indicates an error in the system.

test/integration/job_run_id_map_test.go
```diff
@@ -0,0 +1,29 @@
...
+	run := intutil.CreateProwJobRun(t, dbc, job.ID, "4.18", ts, true, v1.JobSucceeded)
+
+	keys, err := query.LookupProwJobRunPartitionKeys(dbc.DB, int64(run.ID))
+	require.NoError(t, err)
+	assert.Equal(t, "4.18", keys.ProwJobRelease)
+	assert.True(t, ts.Equal(keys.Timestamp))
+
+	_, err = query.LookupProwJobRunPartitionKeys(dbc.DB, 999999)
+	require.ErrorIs(t, err, gorm.ErrRecordNotFound)
+}
```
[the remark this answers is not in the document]
neisw: That case should never exist.  If a query goes against prow_job_runs it should still work but would be a performance issue that needed to be corrected.

pkg/db/infrafailure/infrafailure.go
```diff
@@ -204,7 +216,9 @@ func recordInfraFailureInTx(tx *gorm.DB, prowJobRunID int64) (RecordOutcome, err
 		// check (cheap, and the row lock the UPDATE would have taken is moot here)
 		// so callers can distinguish an already-labeled run from a missing one.
 		var exists int
-		check := tx.Raw("SELECT 1 FROM prow_job_runs WHERE id = ? LIMIT 1", prowJobRunID).Scan(&exists)
+		check := tx.Raw(
+			"SELECT 1 FROM prow_job_runs WHERE id = ? AND prow_job_release = ? AND timestamp = ? LIMIT 1",
```
mstaeble, line 220: nit: Thinking out loud here. If we check for the existence of the ID in the `prow_job_run_id_map` table instead, we could make use of the unique index. Otherwise the planner would need to scan all of the rows in the given `prow_job_runs` partition. It is probably insignificant either way, given the number of rows in the partition, though.

mstaeble (approved):
/lgtm

neisw:
/hold
until e2e issues resolve

pkg/db/infrafailure/infrafailure.go
neisw: I can take a TODO and apply it to the followup pr.

mstaeble: I wouldn't even take a TODO. Unless we have evidence of it being a problem (which I doubt), I would leave it as is.

PR #3967, opened 2026-08-31
TRT-2867: move symptom re-evaluation to async daemon process

sosiouxme:
Make the `POST /api/jobs/runs/reevaluate` API asynchronous by moving re-evaluation task execution into `sippy-daemon` using the existing `WorkProcessor` pattern, rather than running heavy BQ/GCS I/O in the main API server pods.

Adds a dependency on River as a postgres-based work pool manager for the asynchronous tasks.

Used agent instruction updates from https://github.com/openshift/sippy/pull/3942 which will merge separately.

## Summary by CodeRabbit

- **New Features**
  - Symptom re-evaluation now runs asynchronously in batches with progress tracking, status polling, retries, deduplication, dry-run support, and cancellation.
  - Requests return a batch identifier and status link for monitoring results.
  - Duplicate build IDs are consolidated, with support for up to 10,000 unique IDs.
  - Completed and stale batches are cleaned up automatically.

- **Bug Fixes**
  - Job artifact details no longer display broken links when URLs are unavailable.

- **Documentation**
  - Updated API and feature documentation for batch workflows, statuses, cancellation, limits, and integration testing.


sosiouxme:
/hold
while I repeat testing with staging env after latest fixes

cmd/sippy-daemon/main.go
```diff
@@ -74,14 +82,39 @@ func NewSippyDaemonCommand() *cobra.Command {
...
+				return errors.WithMessage(err, "couldn't get bigquery client")
+			}
+
+			gcsClient, err := gcs.NewGCSClient(context.TODO(),
+				f.GoogleCloudFlags.ServiceAccountCredentialFile,
+				f.GoogleCloudFlags.OAuthClientCredentialFile,
+			)
+			if err != nil {
+				return errors.WithMessage(err, "couldn't get GCS client")
+			}
```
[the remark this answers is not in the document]
sosiouxme: this is a pre-existing concern, the PR just moves it around. i don't want to address it here.

sippy-ng/src/jobs/ReEvaluateSymptoms.jsx
```diff
@@ -8,92 +8,88 @@ import {
...
+// submitBatch sends all job run IDs in a single POST and returns the batch
+// metadata (batch_id, requested, links).
+async function submitBatch(prowJobBuildIDs) {
+  const response = await fetch(
+    import.meta.env.VITE_API_URL + '/api/jobs/runs/reevaluate',
+    {
+      method: 'POST',
+      headers: { 'Content-Type': 'application/json' },
+      body: JSON.stringify({ prow_job_build_ids: prowJobBuildIDs }),
+      signal: AbortSignal.timeout(REQUEST_TIMEOUT_MS),
```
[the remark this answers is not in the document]
sosiouxme: I don't believe this is worth the trouble; per claude's analysis:
>  AbortSignal.timeout is supported in Node 18+ (our test runtime) and all modern browsers (Chrome 103+, Firefox
100+, Safari 16+). Our tests pass with it. The jsdom concern about v16.7.0 is outdated; the AbortSignal.timeout API
 is a global from Node, not jsdom. Safari 15 is EOL. Not worth adding a polyfill.

cmd/sippy-daemon/main.go
```diff
@@ -146,6 +164,60 @@ func NewSippyDaemonCommand() *cobra.Command {
...
+func setupRiverProcess(ctx context.Context, f *SippyDaemonFlags, dbc *db.DB, bigQueryClient *bigquery.Client, gcsClient *storage.Client, cacheClient cache.Cache) (sippyserver.DaemonProcess, error) {
+	pgxPool, err := workqueue.NewPgxV5Pool(ctx, f.DBFlags.DSN)
+	if err != nil {
+		return nil, fmt.Errorf("creating pgx/v5 pool for River: %w", err)
+	}
+
+	// River schema is also migrated by "sippy migrate" so the API server can
+	// enqueue jobs. The daemon runs it too in case it starts before or without
+	// the migrate step (e.g. local development), otherwise the daemon exits.
+	if err := workqueue.MigrateRiverSchema(ctx, pgxPool); err != nil {
```
mstaeble, line 181: I think we should stick with deliberate migrations rather than surprise migrations within other commands, particularly if this is only to support local development. Is there an issue that makes it difficult to do local development without the migration here?

cmd/sippy/migrate.go
```diff
@@ -31,6 +32,21 @@ func init() {
 				return errors.WithMessage(err, "could not migrate db")
 			}
 
+			// River (work queue) schema is managed separately from Sippy's
+			// golang-migrate migrations. The API server needs the river_job
+			// table to enqueue async batches, so it must exist after migrate.
+			// The daemon also runs this at startup as a fallback for local dev.
+			ctx := cmd.Context()
+			pgxPool, err := workqueue.NewPgxV5Pool(ctx, f.DSN)
```
mstaeble, line 40: Is there any contention having both pgx/v4 and pgx/v5 pools at the same time? Or is it just more open connections?

cmd/sippy/serve.go
```diff
@@ -197,6 +199,24 @@ func NewServeCommand() *cobra.Command {
 				jiraClient,
 			)
 
+			// Wire up async symptom re-evaluation (insert-only River client).
+			pgxPool, err := workqueue.NewPgxV5Pool(context.Background(), f.DBFlags.DSN)
+			if err != nil {
+				log.WithError(err).Warn("unable to create pgx/v5 pool for River, async re-evaluation will be unavailable")
```
mstaeble, line 205: Do we want to have this as a fatal error? Or are we OK with limited functionality if there are issues with the re-evaluator? In the longer term, I suspect we'll use the work queue for more than just re-evalutation, where sippy will not function properly without the work queue.

pkg/db/migrations/000013_workqueue_symptom_re.up.sql
```diff
@@ -0,0 +1,23 @@
...
+CREATE TABLE workqueue_symptom_re_batch_items (
+    id           BIGSERIAL PRIMARY KEY,
+    batch_id     UUID NOT NULL REFERENCES workqueue_symptom_re_batches(id) ON DELETE CASCADE,
+    river_job_id BIGINT,
+    item_key     TEXT NOT NULL,
+    UNIQUE(batch_id, item_key)
+);
+
+CREATE INDEX idx_symptom_re_batch_items_batch
+    ON workqueue_symptom_re_batch_items (batch_id, river_job_id);
```
mstaeble, line 23: Is this meant to be a covering index, used to find `river_job_id` given the `batch_id`? If so, we can use `INCLUDE`.

```
CREATE INDEX idx_symptom_re_batch_items_batch
    ON workqueue_symptom_re_batch_items (batch_id)
    INCLUDE (river_job_id);
```

pkg/sippyserver/workqueue/symptomre/cancel.go
```diff
@@ -0,0 +1,95 @@
...
+		}
+		if item.RiverJobID == nil {
+			continue
+		}
+		if _, err := c.riverClient.JobCancel(ctx, *item.RiverJobID); err != nil {
+			log.WithFields(log.Fields{
+				"batchID":    batchID,
+				"riverJobID": *item.RiverJobID,
+			}).WithError(err).Warn("batch cancel: failed to cancel River job (may already be complete)")
+			continue
```
mstaeble, line 74: Is there a way to distinguish between an error due to the job already having been canceled and other errors, such as transient failures? Do we risk leaving the system in an inconsistent state if some of the jobs in the batch are not canceled but we still mark the batch as canceled?

In other words, is it always appropriate to continue here in the event of an error attempting to cancel a job?

pkg/sippyserver/workqueue/symptomre/cleanup.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+	return result.RowsAffected, nil
+}
+
+// failStaleBatches marks batches stuck in non-terminal statuses (pending,
+// processing, running) as failed if they were created more than the stale
+// timeout ago. This preserves history for the frontend (which would otherwise
+// see a 404) and lets deleteCompletedBatches remove them after the normal
+// retention period.
+func (p *BatchCleanupProcess) failStaleBatches() (int64, error) {
```
mstaeble, line 106: Do we also want to attempt to cancel the jobs associated with the batch that may still be running?

pkg/sippyserver/workqueue/symptomre/status.go
```diff
@@ -0,0 +1,115 @@
...
+			Deduped:   batch.DedupedCount,
+			Completed: counts.Completed,
+			Failed:    counts.Failed,
+			Running:   counts.Running,
+			Pending:   counts.Pending,
+		},
+		Items: items,
+	}
+
+	// Lazy completion: if batch is running/processing and all items are terminal,
```
mstaeble, line 75: Can we separate this lazy completion from the `Query` method? I am reticent to have a method called `Query` do mutations.

pkg/sippyserver/workqueue/symptomre/status.go
```diff
@@ -0,0 +1,115 @@
...
+	if batch.Status == workqueue.BatchStatusRunning || batch.Status == workqueue.BatchStatusProcessing {
+		derived := workqueue.OverallStatus(counts)
+		if derived == workqueue.BatchStatusComplete || derived == workqueue.BatchStatusFailed {
+			now := time.Now()
+			if err := db.Model(&Batch{}).Where("id = ? and status = ?", batchID, batch.Status).
+				Updates(map[string]interface{}{
+					"status":       derived,
+					"completed_at": now,
+				}).Error; err != nil {
+				return nil, fmt.Errorf("updating batch %s completion: %w", batchID, err)
```
mstaeble, line 86: Should a failure of the lazy completion be an terminal error for the query? My impression is that lazy completion is a best-effort side action done because we have the data at this point to do it not because it is fundamental to the query.

pkg/api/jobrunscan/reevaluate.go
```diff
@@ -120,32 +132,93 @@ type symptomMatch struct {
...
+}
+
+// ErrPermanent wraps errors that should not be retried by River.
+var ErrPermanent = errors.New("permanent failure")
+
+// ReEvaluateOneFromCache re-evaluates a single job run using cached symptoms.
+// Returns an error if evaluation fails, which triggers River's retry logic.
+// Permanent failures (e.g. missing job run) are wrapped with ErrPermanent so
+// the River worker can cancel instead of retrying.
+func (r *ReEvaluator) ReEvaluateOneFromCache(ctx context.Context, prowJobBuildID string, dryRun bool) error {
```
mstaeble, line 176: Should we pass in the symptoms hash so that we are not performing a re-eval for a job with an out-of-date symptom hash? Or is the hash only used for dedupe to avoid queue the job again when we can determine that the re-eval will produce the same results as an existing job? It is not meant to lock the job to a certain set of symptoms.

pkg/sippyserver/workqueue/symptomre/submitter.go
```diff
@@ -0,0 +1,82 @@
...
+		gormDB:      gormDB,
+		riverClient: riverClient,
+	}
+}
+
+// Submit creates a batch specification and enqueues a River job for daemon
+// processing. The batch and item rows are written via GORM (pgx/v4); the
+// River job is inserted via the River client (pgx/v5). These are separate
+// transactions: if the GORM write succeeds but the River insert fails, the
+// batch remains in "pending" status until resubmitted or cleaned up.
```
mstaeble, line 44: Can we fail the batch when the River insert fails?

pkg/sippyserver/workqueue/symptomre/workers.go
```diff
@@ -0,0 +1,168 @@
...
+		return fmt.Errorf("loading batch %s: %w", batchID, err)
+	}
+
+	var items []BatchItem
+	if err := w.gormDB.Where("batch_id = ?", batchID).Find(&items).Error; err != nil {
+		return fmt.Errorf("loading batch items for %s: %w", batchID, err)
+	}
+
+	if err := w.gormDB.Model(&Batch{}).Where("id = ?", batchID).
+		Update("status", workqueue.BatchStatusProcessing).Error; err != nil {
```
mstaeble, line 64: Do we need some conditional logic to ensure that we are not processing batches/jobs that have already been canceled?

pkg/sippyserver/workqueue/symptomre/cancel.go
```diff
@@ -0,0 +1,95 @@
...
+
+	cancelled := 0
+	for _, item := range items {
+		if ctx.Err() != nil {
+			return nil, fmt.Errorf("cancellation interrupted: %w", ctx.Err())
+		}
+		if item.RiverJobID == nil {
+			continue
+		}
+		if _, err := c.riverClient.JobCancel(ctx, *item.RiverJobID); err != nil {
```
mstaeble, line 69: This could cancel a job owned by another batch, if the job was a duplicate.

pkg/sippyserver/workqueue/symptomre/workers.go
```diff
@@ -0,0 +1,168 @@
...
+		return fmt.Errorf("updating batch %s to processing: %w", batchID, err)
+	}
+
+	enqueued, deduped, err := w.fanOutItems(ctx, batchID, items, batch.DryRun)
```
mstaeble, line 68: What are the implications to the consistency of the batch if there is a failure while fanning out items, potentially leaving some items inserted without having their river job ID linked in the BatchItem?

pkg/sippyserver/workqueue/river_process.go
```diff
@@ -0,0 +1,62 @@
...
+
+	if err := p.client.Start(ctx); err != nil {
+		log.WithError(err).Error("workqueue: failed to start River client")
+		return
+	}
+	log.Info("workqueue: River client started")
+
+	<-ctx.Done()
+
+	shutdownCtx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
```
mstaeble, line 49: This waits for up to 30 seconds, but the DaemonServer only waits 10 seconds in its shutdown path.

pkg/sippyserver/workqueue/symptomre/functional_test.go
```diff
@@ -0,0 +1,521 @@
...
+	"gorm.io/driver/postgres"
+	"gorm.io/gorm"
+
+	apijobrunscan "github.com/openshift/sippy/pkg/api/jobrunscan"
+	"github.com/openshift/sippy/pkg/db"
+	jobrunscanmodels "github.com/openshift/sippy/pkg/db/models/jobrunscan"
+	"github.com/openshift/sippy/pkg/sippyserver/workqueue"
+)
+
+// These functional tests require a real PostgreSQL instance. They are skipped
```
mstaeble, line 26: Are we able to leverage the postgres container that is created and used by the integration tests at `test/integration`?

mstaeble (commented):
What is the use case for cancellation? I am wondering whether we are adding a fair amount of complexity for a cancellation feature that will not be utilized.

cmd/sippy-daemon/main.go
sosiouxme: it was also because if daemon comes up before the migration, it completely fails. but that's just as temporary as the local development case; run migrations and all is resolved. so i agree it's better to be consistent about where migrations run.

cmd/sippy/migrate.go
sosiouxme: The pools are completely separate, using separate connections. Any contention would be at the level of postgres resources - this does mean more open connections.

cmd/sippy/serve.go
sosiouxme: i think it's fair to make this fatal. there's no reason it should fail if the regular pool worked, so it would be something we should look at.

pkg/api/jobrunscan/reevaluate.go
sosiouxme: Right now the hash is only used to defeat dedupe, in the scenario where you run a re-eval, and then an hour later realize something was wrong, and you change the symptom and run it again - you would not want that to be counted as a dupe of the previous time; you would want the re-eval to proceed.

There's no point in freezing the re-evaluation to a past set of symptoms, and no real way to determine in advance if the changes to the symptoms will in fact make a difference to any given evaluation. We could maybe save some effort by noting at the time this runs that the hash has changed and doing some re-deduping if two queued entries will effectively operate from the same symptom definitions; but all the ways I can think of to do that seem complicated, for what's really a diminishing return.

pkg/db/migrations/000013_workqueue_symptom_re.up.sql
sosiouxme: Thanks for triggering me educating myself about this :) but after looking I don't think so. AFAICS every lookup of batch items is solely by `batch_id` and gets the whole item. There's at least one case (`symptomre/cancel.go`) where we really just need the associated `river_job_id`s and could optimize, but definitely not all. Unless this table starts getting huge (unlikely), I don't know that having another/different index is a worthy optimization.

pkg/sippyserver/workqueue/symptomre/cancel.go
sosiouxme: Two requests create batches including the same build, they both get deduped to the same river job, then one cancels their batch, and the re-eval is canceled for the other batch too even if they still wanted it to run. That's... a bit ugly to prevent, but may be confusing enough when it happens to warrant addressing (of course, they could just re-submit the batch and only the canceled entries would re-run).

pkg/sippyserver/workqueue/symptomre/cancel.go
sosiouxme: Looks like any error we get here is either a bug (e.g. job doesn't exist) or a transient failure.

I think it's fair to consider cancellation "best effort" here. If you can't cancel a job for some reason, it wastes some work by running. With the outcome being idempotent-ish, I don't think the harm merits really pondering how to handle failure to cancel (the batch is marked cancelled anyway, so nothing is left dangling).

pkg/sippyserver/workqueue/symptomre/cleanup.go
sosiouxme: Scenarios where the batch is stale (24 old) but the items are still actually running/queued are pretty unlikely; the main ones being that the daemon went down for a day after work started, or the work is so backed up that it literally has been waiting a day to run.

I think in those scenarios we would likely still want the work to actually occur? Although if that's the case then I suppose failing the batch isn't really right either (if anyone's still paying attention to the result).

Maybe the right thing to do here is to wire in a river client and reuse the cancellation code. If someone's checking the result then eventually they can retry the ones that didn't run.

pkg/sippyserver/workqueue/symptomre/functional_test.go
sosiouxme: for these tests that really just exercise queue mechanics and need no references to real clients/data beyond the DB, yes I think so. i'll look into moving these.

pkg/sippyserver/workqueue/symptomre/status.go
sosiouxme: I hear you and it does feel funny to me too.

Would it suffice just to put the storage in a separate method so the "Query" semantic doesn't apply?

The vast majority of the time, there won't be any active batches. But if there _is_ an active batch, there may be a UI process polling quite frequently (2.5s currently) for updates to show the human watching.

It could omit storing the result when polling (expecting the client to quit anyway once it's complete), and have a process update the DB on a more leisurely schedule... but that also feels awkward.

pkg/sippyserver/workqueue/symptomre/status.go
sosiouxme: True, it would not hurt to leave it and try again if there's another poll request. It doesn't really even matter if it never gets stored.

pkg/sippyserver/workqueue/symptomre/submitter.go
sosiouxme: Effectively, we do. If there's an error on the river insert, we return an error and the response is with no batch ID (maybe I should take it out of the err msg), so they get the failure right away and aren't polling.

The batch records created in the first transaction are removed once they expire. Would you feel better if we proactively mark them failed here? That itself could fail but I suppose it's simple enough...

pkg/sippyserver/workqueue/symptomre/workers.go
sosiouxme: yes, that makes sense. canceled or otherwise no longer in expected status.

pkg/sippyserver/workqueue/symptomre/workers.go
sosiouxme: Well, you only get one shot at batch creation. As written, if anything fails, then no change is made to the batch record. A client polling the batch will see that it never transitions to running, but any batch items that did get river job IDs would be reported in various states of running or completion (the rest show as not enqueued, even if actual river jobs were enqueued). The batch itself would never progress states though until it became stale after a day. Nothing relays the failure of the batch insert job back to the batch's record.

Now, an appropriate response to any such failure is to re-submit the batch. In that case, whatever river jobs got created would de-dupe with the batch contents and if there are no creation errors you'd end up with a consistent batch. The problem is just the client knowing when it needs to do this, preferably well before the 24-hour timeout.

You would think the Query + update process could do something useful with this condition,but technically "batch is processing and only some of the river job ids are filled in" is a valid state on the way to getting them all filled in, unless we wrapped all of the updates in a transaction. If we did that (how big can a transaction be??), and the transaction failed, then we could attempt to mark the batch failed (which might fail).

Instead we could simply log the error and set the batch status to running (which again might fail). Once all items were in a resting state the batch could transition to complete or failed and the client could see a new one was needed. But frankly in that situation I think it would be more useful to just mark it failed so the client can just start over ASAP.

I think I've talked myself into just attempting to mark the batch failed if any errors occur, with a fallback daemon process that marks "processing" batches failed after... a minute or two? The way the code is written there's no good reason for delay in that state, and no reason for the client not to just retry the whole thing.

pkg/sippyserver/workqueue/river_process.go
sosiouxme: not sure anything actually handles shutdown cleanly, but i guess it does not hurt to just make this 9 seconds so at least you get a log message if it's an abrupt cutoff.

sosiouxme (commented):
Thanks for the very illuminating review.

Originally I didn't plan to add cancellation. But it just kind of gnawed at me that we could schedule these huge numbers of jobs with no way to stop them if we changed our minds. I can at least imagine scenarios where someone defines a symptom, kicks off a giant eval for the last month's job runs, and then realizes they made it way too broad or something and it's handing out labels en masse. Rather than waiting for that to finish (and waiting 90 minutes for the BQ write buffer to clear, if they're in the know) they could just stop it now and start over with the right definition (maybe an agent can help schedule just the ones that didn't run and wait to schedule the ones that did after 90 minutes...)

Maybe we never use it. It definitely added more complexity than I anticipated. But having implemented it (and piled coderabbit fixes on top), it may be more work at this point to rip it out than to keep it working. And I bet we'll use it.

pkg/sippyserver/workqueue/symptomre/cancel.go
sosiouxme: I couldn't think of a decent approach to this. The status returned should make it reasonably evident that jobs were canceled... They can re-submit and I'm just going to leave it at that.

pkg/sippyserver/workqueue/symptomre/submitter.go
sosiouxme: I ended up marking the batch canceled in this case... if possible (if that fails too we just get logs in sippy-daemon).

pkg/sippyserver/workqueue/symptomre/status.go
sosiouxme: It ended up being a little more awkward than I thought to extract "query" from "update" but i've refactored and renamed so hopefully the semantics are closer to the functionality.

pkg/sippyserver/workqueue/symptomre/workers.go
sosiouxme: It only attempts to mark the batch failed in this case.

Adding a fallback to the cleanup process that would catch things stuck in `processing` seemed like overkill once I realized there's nothing currently tracking when the last state transition occurred, so I can't tell the difference between a batch that has been in `processing` for an hour vs it sat in `pending` for an hour and is just now finally getting defined.

sosiouxme (commented):
added a bunch of updates after the commit to fix merge conflicts.

@coderabbitai review

sosiouxme:
@coderabbitai resume

pkg/api/jobrunscan/reevaluate.go
```diff
@@ -34,7 +38,7 @@ import (
 
 const (
 	reEvalSourceTool = "sippy-api-reevaluate"
-	maxJobRunsPerReq = 50
+	maxJobRunsPerReq = 10000
```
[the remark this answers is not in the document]
sosiouxme: i don't see how using bounded chunks will help, as they all still need to complete within the time the job runs. however, i can make the insertion more efficient and give river jobs a longer timeout (probably a good idea anyway).

pkg/api/jobrunscan/reevaluate.go
```diff
@@ -255,27 +337,42 @@ func (r *ReEvaluator) reEvaluateOne(ctx context.Context, buildID string, symptom
...
 // evaluateSymptoms runs one artifact query per symptom for the given job run.
-func (r *ReEvaluator) evaluateSymptoms(ctx context.Context, jobRunID int64, symptoms []jobrunscan.Symptom) ([]symptomMatch, error) {
-	var matches []symptomMatch
+func (r *ReEvaluator) evaluateSymptoms(ctx context.Context, jobRun *models.ProwJobRun, symptoms []jobrunscan.Symptom) ([]symptomMatch, error) {
+
+	// handle job runs stored in a bucket different from current usage
+	bucket := util.GcsBucketRoot
+	if jobRun.GCSBucket != "" {
+		bucket = jobRun.GCSBucket
+	}
```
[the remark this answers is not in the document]
sosiouxme: not as quick a win as you might think; but in any case i'm removing this change as it will have to merge anyway with tomorrow's updates to handle the bucket change.

sosiouxme:
@coderabbit review full

pkg/sippyserver/workqueue/symptomre/cancel.go
```diff
@@ -0,0 +1,99 @@
...
+
+	canceled := 0
+	for _, item := range items {
+		if ctx.Err() != nil {
+			return nil, fmt.Errorf("cancellation interrupted: %w", ctx.Err())
+		}
+		if item.RiverJobID == nil {
+			continue
+		}
+		if _, err := c.riverClient.JobCancel(ctx, *item.RiverJobID); err != nil {
```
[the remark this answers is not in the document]
sosiouxme: this is not a "quick" fix. i do not think attempting to handle this edge case merits the extra complexity. if the job gets canceled, so be it - the status is reported correctly.

pkg/sippyserver/workqueue/symptomre/submitter.go
```diff
@@ -0,0 +1,94 @@
...
+		}
+		return nil
+	}); err != nil {
+		return nil, fmt.Errorf("submitting batch: %w", err)
+	}
+
+	if _, err := s.riverClient.Insert(ctx, ProcessBatchArgs{BatchID: batchID}, nil); err != nil {
+		log.WithError(err).Errorf("enqueuing process-batch job for batch %s", batchID)
+		batch.Status = workqueue.BatchStatusCancelled
+		if err2 := db.Save(&batch).Error; err2 != nil {
```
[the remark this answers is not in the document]
sosiouxme: If the context is canceled, I do not really want to extend it for what is really an optional update. it will get cleaned up later.

PR #3991, opened 2026-09-03
trt-2709: update job_runs for partitioning

neisw:
## Summary by CodeRabbit

- **Bug Fixes**
  - Improved job-run reports to associate pull requests, tests, and annotations with the correct release and execution time.
  - Prevented results from unrelated job runs or partitions from appearing in reports.
  - Fixed test-result associations for merged pull requests so they correspond to the appropriate job run.
  - Improved report filtering to respect inclusive start and exclusive end timestamp boundaries.

### DIFF
--- a/pkg/api/job_runs.go
+++ b/pkg/api/job_runs.go
@@ -133,17 +133,17 @@ func JobsRunsReportFromDB(dbc *db.DB, filterOpts *filter.FilterOptions, release 
 		var g errgroup.Group
 
 		g.Go(func() error {
-			return enrichJobRunsWithTestNames(dbc, jobsResult, release)
+			return enrichJobRunsWithTestNames(dbc, jobsResult, release, lookback, reportEnd)
 		})
 
 		if !needs.needsPRJoin() {
 			g.Go(func() error {
-				return enrichJobRunsWithPRData(dbc, jobsResult, release)
+				return enrichJobRunsWithPRData(dbc, jobsResult, release, lookback, reportEnd)
 			})
 		}
 
 		g.Go(func() error {
-			return enrichJobRunsWithAnnotations(dbc, jobsResult, release)
+			return enrichJobRunsWithAnnotations(dbc, jobsResult, release, lookback, reportEnd)
 		})
 
 		if err := g.Wait(); err != nil {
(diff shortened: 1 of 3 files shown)


pkg/api/job_runs.go
```diff
@@ -323,29 +326,36 @@ func testNameFilterSQL(item filter.FilterItem, lookback time.Time) (string, []an
...
+func jobRunPartitionKeyPredicate(alias string, results []apitype.JobRun, release string) (string, []any) {
+	placeholders := make([]string, len(results))
+	args := make([]any, 0, len(results)*3)
+	for i := range results {
+		placeholders[i] = "(?, ?, ?)"
+		args = append(args, results[i].ID, release, results[i].Timestamp)
+	}
+	return fmt.Sprintf("(%s.prow_job_run_id, %s.prow_job_run_release, %s.prow_job_run_timestamp) IN (%s)",
+		alias, alias, alias, strings.Join(placeholders, ", ")), args
+}
```
mstaeble: Have you seen an EXPLAIN for this approach? I suspect that the planner is going to struggle with partition pruning when presented with multi-dimensional IN clauses like this. Given that we already know the specific release and have a time range, can we keep the IN clause limited to just the IDs with explicit bounds for release and date?

Of perhaps we can use this multi-dimensional IN clause **in addition** to explicit release and date range conditions.

neisw: I haven't yet, but will.

From current main:
 FetchJobRunOnlyNewTests      3  7.200292246s  2.400097415s  1.343631606s  3.570613747s
 
 With this fix:
   FetchJobRunOnlyNewTests      3  1.012772084s  337.590694ms  235.590533ms  532.503728ms
   
Indication is this is likely external tooling
``` 
Specifically, onlyNewTests=true adds both NOT EXISTS clauses for ownership and prior merged-PR appearances.

 There are currently no callers passing true; all production/UI paths pass false. Therefore this query likely came from a benchmark, test, older branch, or an external
  caller using FetchJobRun directly. 
  ```
  
  So we can analyze / iterate some more.

Comment (neisw) [reply pkg/api/job_runs.go]: • The current composite IN clauses prune partitions correctly on PostgreSQL 18.4. The reviewer’s safer option, retaining tuple IN and adding explicit release/time bounds, is
  best. Replacing tuples with ID-only IN is substantially worse.

  For a sparse 20-run exact-job page spanning 11 daily partitions:

   Predicate            Tests plan / exec    PR plan / exec    Annotation plan / exec
  ━━━━━━━━━━━━━━━━━━━  ━━━━━━━━━━━━━━━━━━━  ━━━━━━━━━━━━━━━━  ━━━━━━━━━━━━━━━━━━━━━━━━
   Current tuple IN        13.1 / 29.1 ms     5.8 / 0.36 ms             6.8 / 1.93 ms
  ───────────────────  ───────────────────  ────────────────  ────────────────────────
   Tuple IN + bounds        9.7 / 27.9 ms     3.2 / 0.29 ms             5.3 / 1.93 ms
  ───────────────────  ───────────────────  ────────────────  ────────────────────────
   ID IN + bounds          38.6 / 71.6 ms    17.6 / 0.90 ms            22.8 / 3.46 ms

  Key findings:

  - Current tuple IN scanned exactly the 11 relevant daily partitions.
  - Adding explicit release and 90-day bounds retained those 11 partitions while reducing planning time.
  - ID-only IN scanned all 91 daily partitions in the range.
  - For a dense page contained in one day, tuple forms scanned one partition; ID-only still scanned 91.
  - Cold-cache differences were removed by repeating the sparse-page comparison.

  Recommendation: implement the review comment’s second suggestion: keep the composite tuple IN, plus explicit release and report-window timestamp predicates. Do not switch
  to ID-only IN.

Comment (mstaeble) [reply pkg/api/job_runs.go]: > Indication is this is likely external tooling
> 
> Specifically, onlyNewTests=true adds both NOT EXISTS clauses for ownership and prior merged-PR appearances.
> 
>  There are currently no callers passing true; all production/UI paths pass false. Therefore this query likely came from a benchmark, test, older branch, or an external
>   caller using FetchJobRun directly. 
> So we can analyze / iterate some more.

It looks like this comes from the async daemon.

```
cmd/sippy-daemon
└─ NewWorkProcessor(..., 10, 5*time.Minute, ...)
   └─ WorkProcessor.Run
      └─ ticker fires every 5 minutes
         └─ WorkProcessor.work
            └─ fetchPrCommentProspects
               └─ sends PullRequestComment to prospects channel
                  └─ AnalysisWorker.Run
                     └─ determinePrComment
                        ├─ getPrJobsIfFinished
                        ├─ buildProwJobRuns for each PR job
                        └─ NewTestsWorker.analyzeRisks
                           └─ for each job
                              └─ OnlyLatestSha
                                 └─ for each eligible run
                                    └─ getNewTestsForJobRun
                                       └─ fetchJobRun(..., true, ...)
                                          └─ api.FetchJobRun
                                             └─ slow NOT EXISTS query
```

I see log line in the deamon pod like the following.

```
time="2026-09-03T20:41:51.585Z" level=info msg="Finding new tests for job pull-ci-openshift-machine-config-operator-main-unit run 2094504981850755072" func=assessJobRisks org=openshift pull=6472 repo=machine-config-operator runs=1 sha=ee5a2278adf5273c4414af1131aa8b0e96d503dd
```

neisw: Cool, so not an interactive thing impacting people at the moment.

PR #3995, opened 2026-09-04
trt-2709: bump partition retention

neisw:
## Summary by CodeRabbit

* **Bug Fixes**
  * Updated database partition cleanup to use a consistent 365-day retention period.
  * Older partitions are now detached and removed based on their partition date, providing more predictable cleanup behavior.
  * Recent partitions remain preserved during cleanup.
  * Improved cleanup lifecycle validation to confirm that expired partitions are processed while current partitions are retained.


mstaeble (approved):
/lgtm

PR #3996, opened 2026-09-04
NO-JIRA: avoid globals, inject backoff struct instead

petr-muller:
Followup to #3973 , address https://github.com/openshift/sippy/pull/3973#discussion_r3934300929


mstaeble (approved):
/lgtm

PR #4003, opened 2026-09-09
Add quay-3.18 component readiness view

jbpratt:
## Summary by CodeRabbit

- **New Features**
  - Added the `quay-3.18` release view.
  - Added coverage for recent base and sample releases across AWS and GCP environments.
  - Included amd64, IPI, OVN, high-availability, and no-upgrade test variants.
  - Enabled failure analysis, metrics, and regression tracking for this release view.

### DIFF
--- a/config/views.yaml
+++ b/config/views.yaml
@@ -6395,3 +6395,54 @@ component_readiness:
     enabled: false
   prime_cache:
     enabled: false
+- name: quay-3.18
+  base_release:
+    release: "quay-3.18"
+    relative_start: now-30d
+    relative_end: now-7d
+  sample_release:
+    release: "quay-3.18"
+    relative_start: now-7d
+    relative_end: now
+  variant_options:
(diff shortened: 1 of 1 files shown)


neisw:
@jbpratt All component readiness views are based on OCP releases currently.  The data comes from a different backend that wouldn't know anything about the quay-3.18 synthetic release we added here.

We are currently working to bring these queries into our main Sippy DB where you get the `classic` sippy jobs / tests, etc. that you see now for quay-3.18.  But until that work completes, I wouldn't expect this to work for you.

Even then there is still a hard requirement for mapping tests to OCP Jira Components.  I think that could be worked out but it would be another step.  Since you use the regex suite matching, a reference for that would be https://github.com/openshift-eng/ci-test-mapping/pull/808/changes

jbpratt:
Thanks @neisw , just trying to figure this out with chai - closing this for now!

PR #4019, opened 2026-09-14
Install Git for version metadata

smg247:
## Summary
- install Git in the production image builder so the existing Makefile can resolve and embed the source revision

## Summary by CodeRabbit

* **Chores**
  * Updated the build environment to include Git.
  * Adjusted coverage binary generation during builds.


smg247:
/override-sticky ci/prow/e2e
we don't need e2e for this, and it didn't trigger

PR #4043, opened 2026-09-18
feat(config): add quay-io synthetic release

jbpratt:
Add a synthetic release covering the two quay.io stage performance jobs (stage-performance-test and stage-load-smoke). Uses an exact job-name map rather than a regexp since the goal is exactly these two jobs, not a pattern-matched family.

## Summary by CodeRabbit

- **New Features**
  - Added scheduled Quay.io performance and load-smoke test workflows.
  - Enabled Quay.io staging validation jobs for periodic testing.

- **Updates**
  - Classified two Quay.io snapshot jobs as standard release jobs, improving consistency in release tracking.


smg247 (commented):
/lgtm

PR #4046, opened 2026-09-21
NO-JIRA: validate gcs bucket

neisw:
## Summary by CodeRabbit

- **Bug Fixes**
  - Improved retrieval of job run events and intervals from Google Cloud Storage by deriving the bucket and object path from job run URLs.
  - Added fallback handling when stored bucket information is outdated, while recording a warning.
  - Invalid, incomplete, or unsupported job run URLs are now rejected instead of being processed incorrectly.
  - Prevented unrelated URLs containing matching bucket text from being misidentified as valid storage locations.
  - Added test coverage for matching buckets, stale bucket data, and unsupported URLs.


smg247 (commented):
/lgtm

